PR #130, opened 2016-10-04
Feature/internal user check

MitchellSternke:
- Added checks to see if the user is Microsoft internal. This is primarily done by checking if the user is using a known Microsoft domain server (the registry is also checked for a special key on Windows).
- If the user is Microsoft internal, we will capture the computer name and username when sending telemetry

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,13 +68,15 @@
     "http-proxy-agent": "^1.0.0",
     "https-proxy-agent": "^1.0.0",
     "pretty-data": "^0.40.0",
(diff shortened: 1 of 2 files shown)


kevcunnane:
Mitch - can this review be closed out? It's been listed for 6 weeks now, I'm guessing it's no longer relevant...

MitchellSternke:
@kevcunnane  Sure - I'll close it for now. From the privacy review it seems like we wouldn't be able to collect this information anyway without risks.

PR #357, opened 2016-11-16
Update platform.ts for elementary OS 0.4

cxh004405:
Update platform script to correctly detect elementary OS 0.4 ID value. ID is 'elementary' rather than 'elementary OS'. Added fall-through case statement value to handle both values. Currently it is defaulting to Windows and generating a "Failed to load SQL Tools Service" message when attempting to install.


endeavorchan:
cool

PR #369, opened 2016-11-17
Web app side tests

### LINKED ISSUE #324: Test for Angular data grid wrapper

anthonydresser:
Flushes out most of the missing tests on the web app side of the extension. Also cleans up some of the karma configuration files.

resolves #324


src/views/htmlcontent/src/js/components/app.component.spec.ts
```diff
@@ -66,54 +236,470 @@ describe('AppComponent', function (): void {
...
-            // let slickgrid = results.getElementsByTagName('slick-grid')[0];
-            // let viewport = slickgrid.firstElementChild.getElementsByClassName('slick-viewport')[0]
+            expect(results.className.indexOf('hidden')).toEqual(-1);
+            let resultspane = <HTMLElement> ele.querySelector('#resultspane');
+            resultspane.click();
+            fixture.detectChanges();
+            expect(results.className.indexOf('hidden')).not.toEqual(-1);
+        });
+
+        it('should render all grids when there are alot but only subset of data', () => {
```
kevcunnane: >alot but only subset of data [](start = 51, length = 28)

What does this mean? Is it "when there is some but not all data"? I don't get the wording

src/views/htmlcontent/src/js/components/app.component.spec.ts
```diff
@@ -66,54 +236,470 @@ describe('AppComponent', function (): void {
...
+            spyOn(shortcutService, 'buildEventString').and.returnValue('');
+            spyOn(shortcutService, 'getEvent').and.returnValue(Promise.resolve('event.copySelection'));
+            spyOn(dataService, 'copyResults');
+            dataService.sendWSEvent(batch1);
+            dataService.sendWSEvent(completeEvent);
+            fixture.detectChanges();
+            let keyboardEvent = document.createEvent('KeyboardEvent');
+            let initMethod = typeof keyboardEvent.initKeyboardEvent !== 'undefined' ? 'initKeyboardEvent' : 'initKeyEvent';
+
+            keyboardEvent[initMethod](
```
kevcunnane: Can we simplify this since a bunch of tests are calling it? E.g. if we only have 1-3 params to actually configure, put this in its own function that does the right thing and shortens this test code / uses function & param names to clarify intent

kevcunnane:
:shipit:

src/views/htmlcontent/src/js/components/app.component.spec.ts
anthonydresser: I meant for this to mean testing lots of data sets...I'll update the name...I also realized I didn't actually test what I meant to test. Basically when a certain number of grids are loaded, it will load them all, but only load the data for what is showing on the screen.

PR #371, opened 2016-11-17
Time string parse issue

### LINKED ISSUE #332: Query execution time - format milliseconds as "002" instead of "2" to avoid confusion
- MSSQL Extension Version: 1.5
- VSCode Version:
- OS Version: Any

Steps to Reproduce:

1. Run a query `WAITFOR DELAY '00:00:05'
2. Observe that a millisecond-level result such as "00:00:05.2" is shown

Expected: Format should be "00:00:05.002" if we are showing this level of precision. It's hard to tell if .2 means 2/10ths of a second or 200 milliseconds.

anthonydresser:
Rebased off #369 fixes #332 . Fixes problem with inserting missing characters when milliseconds is less than 100. Added tests to cover this.


src/views/htmlcontent/src/js/utils.ts
```diff
@@ -58,8 +58,9 @@ export function parseNumAsTimeString(value: number): string {
     let hs = h < 10 ? '0' + h : '' + h;
     let ms = m < 10 ? '0' + m : '' + m;
     let ss = s < 10 ? '0' + s : '' + s;
+    let mss = tempVal < 10 ? '00' + tempVal : tempVal < 100 ? '0' + tempVal : '' + tempVal;
```
kevcunnane: Suggestion: Couldn't you just check the length of the mss value and if < 3 left pad with 0s? Similar behavior for hrs and seconds too - have a function that takes an int, required #digits (2 or 3) and pads as needed

kevcunnane:
:shipit:

src/views/htmlcontent/src/js/utils.ts
anthonydresser: I guess I don't see the fundamental difference between how this is doing it and what you are suggesting. Only difference I see is if you are checking the length as a string or as a number.

Issue #364, opened 2016-11-17
Failed to load sql tools service


ellwoods:
- MSSQL Extension Version: 0.1.5
- VSCode Version: 1.7.1
- OS Version: Linux Mint 18

Steps to Reproduce:

1. Install MSSQL
2. Reload
Installing SQL tools service message appears at the bottom of the application
then ERROR mssql: Filed to load Sql Tools Service

The only logs I can find say - 

[Extension Host] vscode-icons is active!
mssql: Failed to load Sql Tools Service
[Extension Host] rejected promise not handled within 1 second
spawn EACCES: Error: spawn EACCES
    at exports._errnoException (util.js:1026:11)
    at ChildProcess.spawn (internal/child_process.js:313:11)
    at Object.exports.spawn (child_process.js:392:9)
    at LanguageClient.createConnection (/home/steve/.vscode/extensions/ms-mssql.mssql-0.1.5/node_modules/vscode-languageclient/lib/main.js:824:32)
    at LanguageClient.resolveConnection (/home/steve/.vscode/extensions/ms-mssql.mssql-0.1.5/node_modules/vscode-languageclient/lib/main.js:533:37)
    at LanguageClient.start (/home/steve/.vscode/extensions/ms-mssql.mssql-0.1.5/node_modules/vscode-languageclient/lib/main.js:465:14)
    at _server.getServerPath.then.serverPath (/home/steve/.vscode/extensions/ms-mssql.mssql-0.1.5/out/src/languageservice/serviceclient.js:146:46)

kevcunnane:
Thanks for raising this. This platform isn't handled in our platform list right now - we handle Linux Mint v14 and 16 but for v18 we need to add an additional check. We will work to fix in the next release and improve handling in this area.

**Bug fix info:**
Looking at https://github.com/OmniSharp/omnisharp-vscode/blob/5c6ccf55e8671f08c480bf5991ebd6a8580510ce/src/platform.ts, it seems that C# extension has added support for this since we last updated. To fix, we should:

- Add matching support for all platforms as done by C# extension

- Change the error handling to not default to Windows. Instead we should return unknown_distribution just like their extension and handle in a similar manner.

ellwoods:
Many thanks for getting back to me, I was unsure which versions of Mint were supported. Is there some official documentation somewhere that I clearly missed?

thanks again

kevcunnane:
[...]

PR #400, opened 2016-11-22
Added support for more platforms

llali:
- Added support for linuxmint and galliumos Linux distros ans win86
- Refactored the code in platform.ts 
- Trying /etc/os-release and fallback to /usr/lib/os-release to get linux os release info
- setting a display name for each runtime that is used for the folder name when the service is installed. The reason the display name is different than the runtime is that if the service is already installed and user gets the update comes with this PR, the extension won't try to install the service again just because the folder name is different

### DIFF
--- a/src/configurations/dev.config.json
+++ b/src/configurations/dev.config.json
@@ -3,15 +3,16 @@
         "downloadUrl": "https://github.com/Microsoft/sqltoolsservice/releases/download/v{#version#}/microsoft.sqltools.servicelayer-{#fileName#}",
         "version": "0.1.4",
         "downloadFileNames": {
-            "Windows": "win-x64-netcoreapp1.0.zip",
-            "OSX": "osx-x64-netcoreapp1.0.tar.gz",
-            "CentOS": "centos-x64-netcoreapp1.0.tar.gz",
-            "Debian": "debian-x64-netcoreapp1.0.tar.gz",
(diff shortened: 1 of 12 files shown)


llali:
#364

llali:
Fix for #364

src/configurations/dev.config.json
```diff
@@ -3,15 +3,16 @@
...
-            "OSX": "osx-x64-netcoreapp1.0.tar.gz",
-            "CentOS": "centos-x64-netcoreapp1.0.tar.gz",
-            "Debian": "debian-x64-netcoreapp1.0.tar.gz",
-            "Fedora": "fedora-x64-netcoreapp1.0.tar.gz",
-            "OpenSUSE": "opensuse-x64-netcoreapp1.0.tar.gz",
-            "RHEL": "rhel-x64-netcoreapp1.0.tar.gz",
-            "Ubuntu14": "ubuntu14-x64-netcoreapp1.0.tar.gz",
-            "Ubuntu16": "ubuntu16-x64-netcoreapp1.0.tar.gz"
+            "Windows_7_64": "win-x64-netcoreapp1.0.zip",
+            "Windows_7_86": "win-x86-netcoreapp1.0.zip",
```
kevcunnane: We may not want to support this. C# doesn't and we're moving away from this in general. Given the very small number of users, I think we should take this out.

kevcunnane:
:shipit:

Issue #352, opened 2016-11-16
mssql activates on all startup contexts


kburtram:
- MSSQL Extension Version: 0.1.5
- VSCode Version: 1.7+
- OS Version: All

Steps to Reproduce:

1. Install mssql and start VS Code

Expected: mssql doesn't activate until a SQL file is opened

Actual: mssql always activates on start-up

PR #407, opened 2016-11-23
Extension no longer activates on all startup contexts

MitchellSternke:
Fixes https://github.com/Microsoft/vscode-mssql/issues/352

Changed it so that we only activate once a SQL file is opened or the user tries to run one of our commands. It adds a bit of a delay to the first use (less than a second), but makes sure our extension doesn't load if the user doesn't need it.

Used this as a guide: https://code.visualstudio.com/docs/extensionAPI/activation-events

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,13 @@
     "Azure SQL Data Warehouse"
   ],
   "activationEvents": [
-    "*"
+    "onLanguage:sql",
+    "onCommand:extension.connect",
+    "onCommand:extension.runQuery",
+    "onCommand:extension.disconnect",
+    "onCommand:extension.manageProfiles",
+    "onCommand:extension.chooseDatabase",
+    "onCommand:extension.cancelQuery"
   ],
   "main": "./out/src/extension",
   "extensionDependencies": [


kevcunnane:
My one question is what's the behavior when the tools service isn't installed - in that case do things freeze for 5+ seconds? If so, can/should we make this experience better?

Overall happy to :shipit: as this is a nice forcing function to improve in this aspect anyhow. If we notice a poor experience let's create a separate task to add feedback in the command palette (e.g. show a progress bar and then launch the action, instead of failing with a message as we do now).

MitchellSternke:
@kevcunnane I tested it without the service installed and it does appear to freeze after executing the first command, although there is a visual indicator that the service is being installed in the status bar. Once the service is installed, the command finally starts executing. The overall delay was a little less than 5 seconds, which would probably be longer on a non-local connection.

We could certainly have a message popup appear letting the user know the service is being installed and that the extension won't be usable until that finishes. That way they would be aware of the wait time.

kevcunnane:
That sounds like a good plan - please sync with Leila who is doing similar work to help clarify when the service is being installed. So long as it's clearer what is going on, it's fine since it's a 1-time thing. Thanks for looking into this.

Issue #341, opened 2016-11-13
Executing no-op query never completes 


kburtram:
- MSSQL Extension Version: 0.1.5
- VSCode Version: 1.7+
- OS Version: CentOS

Steps to Reproduce:

1. Execute a no-op query such as "GO" or "-- no-op"

Expected: Completes with message such as no results

Actual: Runs in "Executing query..." state indefinitely

PR #422, opened 2016-11-28
Fixed issue where no-op queries would never complete

MitchellSternke:
Fixes https://github.com/Microsoft/vscode-mssql/issues/341

For service layer changes, also see https://github.com/Microsoft/sqltoolsservice/pull/157. I put the description for the bug solution in a comment there.

Also added unit test

### DIFF
--- a/test/queryRunner.test.ts
+++ b/test/queryRunner.test.ts
@@ -75,33 +75,6 @@ suite('Query Runner tests', () => {
 
     });
 
-    test('Handles Query Error Correctly', () => {
-        let testuri = 'uri';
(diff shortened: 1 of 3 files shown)


kevcunnane:
:shipit:

test/queryRunner.test.ts
```diff
@@ -102,6 +102,36 @@ suite('Query Runner tests', () => {
         });
     });
 
+    test('Handles Info Message Correctly', () => {
```
MitchellSternke: Correct!

Issue #277, opened 2016-11-05
Adding quote at end of word incorrect adds 2 quote marks


kevcunnane:
For other file types (e.g. JSON, JS, C#) if you have a word like test, and you add a quote at the start (e.g. "test or 'test), then typing in a quote at the end of the word "test" or 'test' will only add the single " or ' character.
For .sql files, if you type his it adds 2 characters, e.g. "test"" or 'test''. 

- What makes this more annoying is hitting delete removes both characters - to succeed in adding the end quote you have to hit the right key then delete the 2nd quote

- This is an inconsistency with other editor types and quite annoying. We are basically failing to correctly detect we're at the end of a word and handle appropriately

kburtram:
Fix with commit https://github.com/Microsoft/sqltoolsservice/commit/84eb164d8733521f688a6f1b7a613bd0a32beb21

PR #426, opened 2016-11-29
Update SQL language configuration to handle auto-closing characters

kburtram:
This is a fix for this issue https://github.com/Microsoft/vscode-mssql/issues/277.

### DIFF
--- a/src/languageservice/serviceclient.ts
+++ b/src/languageservice/serviceclient.ts
@@ -4,7 +4,7 @@
  * ------------------------------------------------------------------------------------------ */
 'use strict';
 
-import { ExtensionContext, workspace } from 'vscode';
+import { ExtensionContext, workspace, languages } from 'vscode';
 import { LanguageClient, LanguageClientOptions, ServerOptions,
     TransportKind, RequestType, NotificationType, NotificationHandler,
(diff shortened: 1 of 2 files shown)


MitchellSternke:
Does this mean that https://github.com/Microsoft/vscode-mssql/blob/dev/syntaxes/sql.configuration.json is no longer needed if we are configuring it using code?

src/languageservice/serviceclient.ts
```diff
@@ -146,9 +146,43 @@ export default class SqlToolsServiceClient {
...
+     * @memberOf SqlToolsServiceClient
+     */
+    private initializeLanguageConfiguration(): void {
+        languages.setLanguageConfiguration('sql', {
+            comments: {
+                lineComment: '--',
+                blockComment: ['/*', '*/']
+            },
+
+            brackets: [
```
llali: I don't see any change in sql.configuration.json. shouldn't we remove the configurations for brackets there?

kburtram:
I left the sql.configuration.json file there since it's referenced in the package.json to define the extension contribution points in the vscode-extension manager.  We could probably remove the contents of the file, but and have it contain an empty JSON object, but I didn't test that.

llali:
:shipit:

PR #430, opened 2016-11-29
Default font

### LINKED ISSUE #428: Result Grid font should be customizable
The results grid font should be customizable through the user settings.  Also, the dev branch has changed the font to be monospaced.  We should probably keep the previous font, unless there was a specific UX review that concluded monospace is better in this case.  Personally I like the previous font better.

image

anthonydresser:
Adds extension setting for default font for results grid and adds old default.

resolves #428

### DIFF
--- a/package.json
+++ b/package.json
@@ -396,8 +396,13 @@
         },
         "mssql.messagesDefaultOpen": {
           "type": "boolean",
-          "description": "true for the messages pane to be open by default; false for closed",
+          "description": "True for the messages pane to be open by default; false for closed",
           "default": true
+        },
+        "mssql.resultsFontFamily": {
+          "type": "string",
+          "description": "Set the font family for the results grid; set to blank to use the editor set font",
+          "default": "serif"
         },
         "mssql.saveAsCsv": {
           "type": "object",
--- a/package.json
+++ b/package.json
@@ -401,8 +401,13 @@
         },
         "mssql.resultsFontFamily": {
           "type": "string",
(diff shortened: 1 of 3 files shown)


src/models/SqlOutputContentProvider.ts
```diff
@@ -78,8 +78,11 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
             } catch (e) {
                 prod = false;
             }
+            let mssqlConfig = self._vscodeWrapper.getConfiguration('mssql');
```
kevcunnane: It's cleaner to have a single line that gets the fontFamily object, then a 2nd line that does the split and join operations rather than duplicating these. Ideally we'd avoid double-reading from the mssqlConfig also

kevcunnane: Shouldn't the 'mssql' and 'resultsFondFamily' constants be in a constants file?

package.json
```diff
@@ -396,9 +396,14 @@
         },
         "mssql.messagesDefaultOpen": {
           "type": "boolean",
-          "description": "true for the messages pane to be open by default; false for closed",
+          "description": "True for the messages pane to be open by default; false for closed",
           "default": true
         },
+        "mssql.resultsFontFamily": {
```
kburtram: should we also have a value to determine the font size?  I think now it's controlled by the editor size, which may be good enough, but an explicit option seems better in my opinion.

PR #432, opened 2016-11-30
Update slickgrid

### LINKED ISSUE #431: Update slickgrid lib

anthonydresser:
Updates slickgrid to work with the new refactor of the other repo. Waiting on release fork as to not destabilize then next release.

Resolves #431


kevcunnane:
I just want to confirm the refactor that's happened:

- We now have 2 dependencies that are cleanly separated from each other: pure slickgrid library (which is agnostic to angular and can be used in any HTML app) and angular2-slickgrid (which is an angular wrapper over slickgrid)

- The plugins that are specific to our application are now back inside it. Specifically, the drag control behavior is now build in our repo and any changes can be managed just inside this repository

Is this correct? And the purpose is that it should be much easier to update with any bug fixes from the maintained SlickGrid fork that's managed by the community? If so, looks good and :shipit: as it seems cleanly factored to me.

anthonydresser:
Exactly @kevcunnane

PR #433, opened 2016-11-30
Instrumentation for Linux distribution telemetry

MitchellSternke:
Added Linux distribution as a common property to all telemetry events. It will be empty on other platforms.


kburtram:
Please remember to update any relevant telemetry documentation that we've now capturing this datapoint.

MitchellSternke:
@kburtram Will do.

PR #436, opened 2016-11-30
Updating status in the output when installing the service

llali:
- Opening the output window if the service is not installed to show the progress
- Adding logs to the output windows when the service is installing 
- Showing the download percentage to the status bar
- Fixed the issue with getting 500 from http client
- Added tests for platforms and service client
- Added a telemetry attribute to "ExtensionActivated" event to show if the service already installed
- Showing an error message in UI when the platform is not supported  
- Showing an info message to user when initializing the service that the commands won't work until the service is initialized


package.json
```diff
@@ -73,7 +73,8 @@
     "typescript": "^1.8.9",
     "uglify-js": "mishoo/UglifyJS2#harmony",
     "vscode": "^0.11.0",
-    "yargs": "https://registry.npmjs.org/yargs/-/yargs-3.32.0.tgz"
+    "yargs": "https://registry.npmjs.org/yargs/-/yargs-3.32.0.tgz",
+    "chai": "^3.5.0"
```
MitchellSternke: nit: these are in alphabetical order

test/platform.test.ts
```diff
@@ -20,4 +22,223 @@ suite('Platform Tests', () => {
             done();
         });
     });
+
```
anthonydresser: Why are you using chai expect as opposed to assert which is already a part of the vscode testing framework?

llali: no specific reason, I saw it somewhere else and I wanted to try it out. the test can be more readable this way

---
In reply to: [90500837](https://github.com/Microsoft/vscode-mssql/pull/436#discussion_r90500837) [](ancestors = 90500837)

anthonydresser: I mean I have no problem with it, but it means we are mixing assertion libraries.

llali: if you don't think it will cause any issue then I'm going to keep it just to play with it a little bit more 

---
In reply to: [90504000](https://github.com/Microsoft/vscode-mssql/pull/436#discussion_r90504000) [](ancestors = 90504000)

PR #437, opened 2016-11-30
Feature: Progressive Results Part 2

benrr101:
Adding support for rendering result sets as they are returned from the service layer.

In more detail, this change contains:
* New class for result set completion notifications to mirror the object from the service layer
* New event handler for when a result completion notification comes through
* New web socket behavior for when a result set is completed. It basically splits the existing batch complete notification into two separate handlers: one for rendering the result sets, one for rendering the batch's messages.
* New logic for re-sending web socket messages to preview pane once preview pane has connected. In essence, rather than re-sending all the resultsets, batch, and completion objects, a queue of messages to send is built up and replayed once the web socket has a connection. This method was chosen because it was simpler logic than attempting to rebuild the order of the messages from the object trees.
* Unit tests for the new result set logic have been added as well.


src/controllers/localWebService.ts
```diff
@@ -33,10 +36,26 @@ export default class LocalWebService {
         this.app.set('view engine', 'ejs');
         Utils.logDebug(Constants.msgLocalWebserviceStaticContent + LocalWebService.staticContentPath);
         this.server.on('request', this.app);
+
+        // Handle new connections to the web socket server
         this.wss.on('connection', (ws) => {
             let parse = querystring.parse(url.parse(ws.upgradeReq.url).query);
-            self.clientMap.set(parse.uri, ws);
-            self.newConnection.emit('connection', parse.uri);
+            if (self.wsMap.has(parse.uri)) {
```
anthonydresser: Have you tested this with multiple result windows and switching between them? I suspect this won't work.

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -310,72 +310,45 @@ export class AppComponent implements OnInit, AfterViewChecked {
...
-                            columnDefinitions.push({
-                                id: i.toString(),
-                                name: columnName,
-                                type: self.stringToFieldType('string'),
-                                formatter: self.textFormatter
-                            });
-                        }
+            switch (event.type) {
+                case 'complete':
+                    self.complete = true;
```
anthonydresser: I was going to fix this, but might as well since you are editing this area; in this case you should also set self.messagesAdded since one of the side effects of setting complete to true is another message is added to the message pane.

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -410,9 +406,10 @@ export class AppComponent implements OnInit, AfterViewChecked {
                     self.placeHolderDataSets.push(undefinedDataSet);
                     self.messagesAdded = true;
                     self.onScroll(0);
-                }
+                break;
+                default:
```
anthonydresser: Might be useful for debugging to added a console.error message here saying there was an unexpect event emitted or something.

src/controllers/localWebService.ts
```diff
@@ -57,16 +76,29 @@ export default class LocalWebService {
...
+                mapping.pendingMessages.push(message);
+            }
+        } else {
+            // There isn't a mapping, so create it
+            let mapping = new WebSocketMapping();
+            this.wsMap.set(uri, mapping);
 
-        if (this.clientMap.has(uri) && this.clientMap.get(uri).readyState === ws.OPEN) {
-            this.clientMap.get(uri).send(JSON.stringify(temp));
+            // Append the message to the queue
```
anthonydresser: See my previous comment, I was doing this the way I was for a reason 😄

src/controllers/localWebService.ts
benrr101: Ah, yep. You are totally right. Ok, I'll think on this a little more.

benrr101:
@anthonydresser, I added changes as per your comments. To address the multiple windows/queries issue, I've set it up so that a history of the messages passed via web socket are kept. On any web socket connection, the messages are replayed. When a query is executed, the web socket mapping is reset, clearing the history and taking care of re-executing the query issue. To cut down on memory footprint, instead of storing the json stringified message, the un-stringified message is kept, which just has a reference to the resultset/batch summary objects.

src/controllers/localWebService.ts
```diff
@@ -57,16 +82,38 @@ export default class LocalWebService {
...
+        mapping.pendingMessages.push(message);
+    }
+
+    /**
+     * Purges the queue of messages to send on the web socket server for the given uri
+     * @param   uri URI of the web socket server to reset
+     */
+    resetSocket(uri: string): void {
+        if (this.wsMap.has(uri)) {
+            this.wsMap.delete(uri);
```
anthonydresser: Theres a better user experience that could happen here, but it's not really necessary atm.

anthonydresser:
Also I'd double check to make sure the html tests pass for this and the progress bar one

benrr101:
Tests are working, will merge now 😃

PR #438, opened 2016-11-30
Fixed broken unit test

MitchellSternke:
This test broke sometime after automatic extension activation was removed. To fix, I forced the extension to activate by opening an untitled SQL file.


benrr101:
Works as expected, :shipit:

PR #453, opened 2016-12-04
Added additional check for errors

anthonydresser:
Was getting a error on my local machine related to the map existing but no the websocket, which was causing events to get missed. Added an additional check to prevent this.


kevcunnane:
@benrr101  has a duplicate PR out, please pick one to check in.

benrr101:
This one was first, and the code is identical, so I closed the other one.

PR #454, opened 2016-12-04
Fixing issue where very fast web socket calls might fail

benrr101:
This is a high-pri bug that needs to go in ASAP. In a slight oversight on my part, I missed the possibility of the mapping for url to web socket server existing, but the socket not existing yet. The result is that very fast queries would not send anything more than their first update.


anthonydresser:
I opened a pull request that does this exact thing yesterday? Did you not see it?

benrr101:
Ah, no, I didn't see it. I was scanning my email in more of a panic looking for "you broke it, fix it fix it fix it now!" and not "I fixed it already" emails. Since you got yours out first, I'll let you have the credit 😃

PR #462, opened 2016-12-05
Adding telemetry for intellisense activation, completion and service initializing failures

llali:
- Adding telemetry for intellisense activation, completion and service initializing failures
- had to fix the service download code to send the error when promise fails


src/languageservice/serviceclient.ts
```diff
@@ -177,6 +181,7 @@ export default class SqlToolsServiceClient {
                 }).catch(err => {
                     Utils.logDebug(Constants.serviceLoadingFailed + ' ' + err );
                     Utils.showErrorMsg(Constants.serviceLoadingFailed);
+                    Telemetry.sendTelemetryEvent('ServiceInitializingFailed');
```
kburtram: maybe "Telemetry.sendTelemetryEventForException" is better here?

src/languageservice/serviceclient.ts
```diff
@@ -249,11 +254,19 @@ export default class SqlToolsServiceClient {
...
+
         });
+        client.onNotification(LanguageServiceContracts.DefinitionSentNotification.type, this.handleLanguageServiceDefinitionSentNotification());
 
         return client;
     }
 
+    private handleLanguageServiceDefinitionSentNotification(): NotificationHandler<LanguageServiceContracts.DefinitionSentParams> {
+        return (event: LanguageServiceContracts.DefinitionSentParams): void => {
+            Telemetry.sendTelemetryEvent('DefinitionRequested');
```
kburtram: the volume of this event will likely be pretty high.  is there a reason we need to capture a separate event for each definition request vs. an aggregated metric?

src/languageservice/download.ts
```diff
@@ -199,6 +199,8 @@ export default class ServiceDownloadProvider {
                 .then(inStream => {
                     this.install(inStream, installDirectory).then ( installed => {
                         resolve(installed);
+                    }).catch(installError => {
+                        reject(installError);
```
kburtram: any reason we don't send telemetry event here?  is the error propagated and logged elsewhere?

llali: the telemetry event for this will be sent in serviceClient 

---
In reply to: [91185859](https://github.com/Microsoft/vscode-mssql/pull/462#discussion_r91185859) [](ancestors = 91185859)

src/languageservice/serviceclient.ts
llali: Exception telemetry in already send in mainController for all type of exception happens while initializing, I thought we may want to also have the number of times the service initializing fails 

---
In reply to: [91185018](https://github.com/Microsoft/vscode-mssql/pull/462#discussion_r91185018) [](ancestors = 91185018)

PR #465, opened 2016-12-06
Filter error paths to only include source files.

kevcunnane:
- minimize path sent via telemetry as we only want the path after the /out/ section of the path


llali:
:shipit:

PR #471, opened 2016-12-06
Fix password saving for the Credential Store when savePassword is enabled 

### LINKED ISSUE #353: Password is not saved when connection created from user settings file with savePasswortrue
- MSSQL Extension Version: 1.5
- VSCode Version: 1.7.1
- OS Version: macOS

Steps to Reproduce:

1. Hit F1 -> Manage Connection Profiles -> Edit to open the connection list
2. Copy & Paste and existing connection with password, or add one (see below)
3. Set password as blank but have `"savePassword": true`
4. Save, open a .sql file and connect to the database. Enter your password when prompted
5. Disconnect, then reconnect to that DB

Expected:
- Password is saved and you are not prompted to re-enter every time

Actual:
- Password is not saved. Every time you reconnect you must enter it

```
"mssql.connections": [
        {
            "authenticationType": "SqlLogin",
            "server": "myserver",
            "database": "",
            "user": "test",
            "password": "",
            "savePassword": true
        }
    ]
```

Raymondd:
Fixes #353 Modified the removeProfile function to accept an optional boolean for keeping Credential Store profiles. Also added a test to for this option.

### DIFF
--- a/src/models/connectionCredentials.ts
+++ b/src/models/connectionCredentials.ts
@@ -119,7 +119,7 @@ export class ConnectionCredentials implements IConnectionCredentials {
                     // then transfer the password to the credential store
                     if (profile.savePassword && !wasPasswordEmptyInConfigFile) {
                         // Remove profile, but keep credential store if savePassword was enabled
-                        connectionStore.removeProfile(profile, profile.savePassword).then(() => {
+                        connectionStore.removeProfile(profile, true).then(() => {
                             connectionStore.saveProfile(profile);
                         });
                     // Or, if the user answered any additional questions for the profile, be sure to save it
--- a/src/models/connectionCredentials.ts
+++ b/src/models/connectionCredentials.ts
@@ -118,7 +118,8 @@ export class ConnectionCredentials implements IConnectionCredentials {
                     // If this is a profile, and the user has set save password to true and stored the password in the config file,
                     // then transfer the password to the credential store
                     if (profile.savePassword && !wasPasswordEmptyInConfigFile) {
-                        connectionStore.removeProfile(profile).then(() => {
+                        // Remove profile, but keep credential store if savePassword was enabled
+                        connectionStore.removeProfile(profile, profile.savePassword).then(() => {
                             connectionStore.saveProfile(profile);
                         });
                     // Or, if the user answered any additional questions for the profile, be sure to save it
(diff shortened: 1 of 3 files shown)


anthonydresser:
@Raymondd I'd recommend doing a merge from dev for this branch, looks like the builds are failing because they can't find the config which was pushed into dev yesterday. Once you upload the commit to merge from dev, it should trigger another build and it should fix the broken build.

src/models/connectionStore.ts
```diff
@@ -308,9 +308,10 @@ export class ConnectionStore {
      * from the credential store
      *
      * @param {IConnectionProfile} profile the profile to be removed
+     * @param {Boolean} keepCredentialStore optional value to keep the credential store after a profile removal
      * @returns {Promise<boolean>} true if successful
      */
-    public removeProfile(profile: IConnectionProfile): Promise<boolean> {
+    public removeProfile(profile: IConnectionProfile, keepCredentialStore: Boolean = false): Promise<boolean> {
```
kevcunnane: please use lowercase `boolean`. We use that consistently across the extension

src/models/connectionCredentials.ts
```diff
@@ -118,7 +118,8 @@ export class ConnectionCredentials implements IConnectionCredentials {
                     // If this is a profile, and the user has set save password to true and stored the password in the config file,
                     // then transfer the password to the credential store
                     if (profile.savePassword && !wasPasswordEmptyInConfigFile) {
-                        connectionStore.removeProfile(profile).then(() => {
+                        // Remove profile, but keep credential store if savePassword was enabled
+                        connectionStore.removeProfile(profile, profile.savePassword).then(() => {
```
kevcunnane: Minor: profile.SavePassword will always be true in this scenario, so you might as well always pass true.

test/connectionStore.test.ts
```diff
@@ -216,6 +216,32 @@ suite('ConnectionStore tests', () => {
             }).catch(err => done(new Error(err)));
     });
 
+    test('RemoveProfile should not remove password from CredentialStore if keepCredentialStore is enabled', (done) => {
```
kevcunnane: It might be worthwhile adding another test for the keepCredentialStore = false case. This is implicitly covered by the removeProfile tests but it would be good to handle this here too.

PR #480, opened 2016-12-07
Batch start notification

benrr101:
This is the change to the extension that corresponds to Microsoft/sqltoolsservice#169, which sends a notification to indicate that a batch has completed execution. This change makes it so that the batch start message is visible as soon as the batch starts executing, and allows for more accurate time of the batch start to be visible.

### DIFF
--- a/src/controllers/QueryRunner.ts
+++ b/src/controllers/QueryRunner.ts
@@ -7,7 +7,7 @@ import {QueryNotificationHandler} from './QueryNotificationHandler';
 import VscodeWrapper from './vscodeWrapper';
 import { BatchSummary, QueryExecuteParams, QueryExecuteRequest,
     QueryExecuteCompleteNotificationResult, QueryExecuteSubsetResult,
-    QueryExecuteResultSetCompleteNotificationParams, ResultSetSummary,
+    QueryExecuteResultSetCompleteNotificationParams,
     QueryExecuteSubsetParams, QueryDisposeParams, QueryExecuteSubsetRequest,
     QueryDisposeRequest, QueryExecuteBatchNotificationParams } from '../models/contracts/queryExecute';
 import { QueryCancelParams, QueryCancelResult, QueryCancelRequest } from '../models/contracts/QueryCancel';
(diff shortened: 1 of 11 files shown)


src/controllers/QueryRunner.ts
```diff
@@ -181,31 +181,45 @@ export default class QueryRunner {
...
+        this._batchSets[batch.id] = batch;
+        this.eventEmitter.emit('batchStart', batch);
+    }
+
+    public handleBatchComplete(result: QueryExecuteBatchNotificationParams): void {
+        let batchGiven = result.batchSummary;
+        let batchLocal = this._batchSets[batchGiven.id];
+
+        // Store the info we don't get from the initial batch notification
+        batchLocal.hasError = batchGiven.hasError;
```
anthonydresser: Is there a reason you can't just assign it? Worst case scenario you overwrite old data with new data, which would probably be preferable anyways. Individually updating each field means in the future if you want to update the batch you have to add a line here.

If the object definitions are not the same, you could at least do a for of loop since they seem to have the same names.

src/controllers/QueryRunner.ts
```diff
@@ -181,31 +181,45 @@ export default class QueryRunner {
...
+        let batchSet = this._batchSets[resultSet.batchId];
 
-        // Create the batch summary if we haven't seen this one yet
-        if (this._batchSets[resultSet.batchId] === undefined) {
-            let batchSummary = new BatchSummary();
-            batchSummary.id = resultSet.batchId;
-            batchSummary.resultSetSummaries = new Array<ResultSetSummary>();
-            this._batchSets[resultSet.batchId] = batchSummary;
+        // Store the result set in the batch
+        if (!batchSet.resultSetSummaries) {
```
anthonydresser: Why not just do this when you create the batch?

src/models/SqlOutputContentProvider.ts
```diff
@@ -294,8 +294,11 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
             queryRunner.eventEmitter.on('resultSet', (resultSet) => {
                 this._service.broadcast(resultsUri, 'resultSet', resultSet);
             });
-            queryRunner.eventEmitter.on('batch', (batch) => {
-                this._service.broadcast(resultsUri, 'batch', batch);
+            queryRunner.eventEmitter.on('batchStart', (batch) => {
```
anthonydresser: Since there are starting to be a few of these events (of which the event names have to persist correctly through the pipeline, might be worth making these enums.

src/views/htmlcontent/test/app.component.spec.ts
```diff
@@ -270,14 +277,101 @@ describe('AppComponent', function (): void {
...
+            dataService.sendWSEvent(batchStart);
+            fixture.detectChanges();
+
+            let results = ele.querySelector('#results');
+            expect(results).toBeNull('results pane is visible');
+
+            // Messages should be visible
+            let messages = ele.querySelector('#messages');
+            expect(messages).not.toBeNull('messages pane is not visible');
+            expect(messages.className.indexOf('hidden')).toEqual(-1);
```
anthonydresser: I didn't learn this till later, but this probably isn't the best way to determine if something is visible. It can have the class, but the class can be overwritten by a different class. A better way would be to get the [computed style](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) of the element and check the display property.

test/queryRunner.test.ts
```diff
@@ -139,14 +174,34 @@ suite('Query Runner tests', () => {
             testQueryNotificationHandler.object,
             testVscodeWrapper.object
         );
+        queryRunner.batchSets[0] = {
+            executionElapsed: null,         // tslint:disable-line:no-null-keyword
```
anthonydresser: any reason why you are using null here (and other places in this PR) rather than undefined? We shouldn't make a habit of ignoring tslint rules (the exception being in the angular ignoring unused fields).

src/views/htmlcontent/test/app.component.spec.ts
benrr101: Now that I think about it, yeah, we should be checking the style attributes instead of class names. I think in this case it works to check the class name, but that's only the case because we tend to explicitly hide elements.

test/queryRunner.test.ts
benrr101: Yes, I am using nulls because null is what comes back from the service layer. Regardless of the "truthyness" of undefined vs null, I think for our testing scenarios it makes more sense to test with the same data that the service will send back than to use different data just because the tslinting rules say so.

src/controllers/QueryRunner.ts
benrr101: The main reason for this is to ensure that we don't override the result set summaries for the batch. It's not really a huge issue, so I'll go ahead and make this change.

src/controllers/QueryRunner.ts
benrr101: It seemed to logically make more sense to do it here instead of modifying the object that came back from the service. But since we already do that for the selection lines, I don't see why we can't continue the trend.

src/models/SqlOutputContentProvider.ts
benrr101: I agree, but since this is a larger change, and the hope was to have this all checked in soon for the Dec release, I'll hold off on this until I get going on part 3.

kevcunnane: Perhaps make a task / issue to track this?

---
In reply to: [91421282](https://github.com/Microsoft/vscode-mssql/pull/480#discussion_r91421282) [](ancestors = 91421282)

kevcunnane:
:shipit:

PR #481, opened 2016-12-08
Fix issue where deselect is not supported in results view

### LINKED ISSUE #328: Deselect is not supported in results view
- MSSQL Extension Version: latest dev
- VSCode Version: 1.7
- OS Version: Windows/OSX

Steps to Reproduce:

1. Shift-select an area in the results view, moving right and down. Notice the cursor is in the lower right.
image

2. Hit shift+left

Expected:
The selection box should be bound by the starting location of the shift-select and the location of the cursor. This is the behavior of SSMS, as well as the click+drag selection
image

Actual:
The cursor moves left, but the selection is unchanged
image

coquagli:
Fixes #328. Adds logic to handleKeyDown(), navigateRight(), and navigateLeft() to check if the cell that was previously active should be included or discluded from the selection.

### DIFF
--- a/src/views/htmlcontent/src/js/slick.dragrowselector.js
+++ b/src/views/htmlcontent/src/js/slick.dragrowselector.js
@@ -116,10 +116,16 @@
 
                 if (e.shiftKey) {
                     var last = _ranges.pop();
-                    var fromRow = Math.min(activeCell.row, last.fromRow);
-                    var fromCell = Math.min(newRangeColumn, last.fromCell);
-                    var toRow = Math.max(activeCell.row, last.toRow);
+
                     var toCell = Math.max(newRangeColumn, last.toCell);
+                    if(last.fromCell <= newRangeColumn) {
+                        toCell = Math.max(newRangeColumn, last.toCell - 1);
+                    }
+
+                    var fromRow = Math.min(activeCell.row, last.fromRow);
+                    var toRow = Math.max(activeCell.row, last.toRow);
+                    var fromCell = Math.min(newRangeColumn, last.fromCell);
+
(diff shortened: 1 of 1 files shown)


src/views/htmlcontent/src/js/slick.dragrowselector.js
```diff
@@ -116,10 +116,16 @@
 
                 if (e.shiftKey) {
                     var last = _ranges.pop();
+
+                    var toCell = Math.max(newRangeColumn, last.toCell);
+                    if(last.fromCell <= newRangeColumn) {
```
kburtram: could you add a ' ' between the 'f' and '(' in conditional statements to match existing coding format of the file.

anthonydresser: It is possible to add eslint to our repo as well if desired to lint the js files, but i think there are only 2 currently.

coquagli: Will do.

src/views/htmlcontent/src/js/slick.dragrowselector.js
```diff
@@ -235,10 +259,10 @@
 
         function handleHeaderClick(e, args) {
             var columnIndex = _grid.getColumnIndex(args.column.id);
-            if(e.ctrlKey || e.metaKey){
+            if (e.ctrlKey || e.metaKey){
```
benrr101: 👍 on fixing the linting issues

benrr101:
If it's working, I say :shipit: but also take a look at the suggestion I proposed

coquagli:
Just made the changes Ben suggested. Also added comment explaining why the changes are needed.

PR #489, opened 2016-12-08
Changed resultSet.totalRows to resultSet.rowCount

### LINKED ISSUE #472: Preview Release 2 candidate: When multiple grids are rendered with larger than 8 rows, 8 is the minimum rows to display per grid.
Each grid should display 8 or lower number of rows per grid.
image

coquagli:
Fixes #472. Looks like the variable name was changed, causing maxHeight and minHeight to always be 'inherit'.

### DIFF
--- a/src/views/htmlcontent/src/js/components/app.component.ts
+++ b/src/views/htmlcontent/src/js/components/app.component.ts
@@ -363,10 +363,10 @@ export class AppComponent implements OnInit, AfterViewChecked {
                     };
 
                     // Precalculate the max height and min height
-                    let maxHeight = resultSet.totalRows < self._defaultNumShowingRows
-                        ? Math.max((resultSet.totalRows + 1) * self._rowHeight, self.dataIcons.length * 30) + 10
+                    let maxHeight = resultSet.rowCount < self._defaultNumShowingRows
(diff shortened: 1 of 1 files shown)


src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -363,10 +363,10 @@ export class AppComponent implements OnInit, AfterViewChecked {
                     };
 
                     // Precalculate the max height and min height
-                    let maxHeight = resultSet.totalRows < self._defaultNumShowingRows
-                        ? Math.max((resultSet.totalRows + 1) * self._rowHeight, self.dataIcons.length * 30) + 10
+                    let maxHeight = resultSet.rowCount < self._defaultNumShowingRows
```
kevcunnane: Are there any unit tests for this path? Can we add some to ensure this doesn't break again?

coquagli: The lines themselves are covered in the tests (probably due to the fact that these lines are within `ngOnInit()` ), but I don't think any tests specifically verify maxHeight  or anything like that.

benrr101: I'm not entirely certain how to do it, but I think we could test the computed height of the children of #results when given different number of rows. We can likely address this after the current release

benrr101:
:shipit:

Issue #484, opened 2016-12-08
More graceful crash recovery when creating natively compiled functions  


abist:
- MSSQL Extension Version:
- VSCode Version:
- OS Version:

Steps to Reproduce:

1. Execute the following query to create a function:

```
CREATE FUNCTION [dbo].[ufnGetAccountingEndDate_native]()
RETURNS [datetime] 
WITH NATIVE_COMPILATION, SCHEMABINDING
AS 
BEGIN ATOMIC WITH (TRANSACTION ISOLATION LEVEL=SNAPSHOT, LANGUAGE=N'us_english')

    RETURN DATEADD(millisecond, -2, CONVERT(datetime, '20160701', 112));

END
```

2. Execute any other query for example:

` SELECT * FROM SYS.OBJECTS  `

Expected: Error for the first query, and successful execution for the second query.
Actual: An error message will show up and "Executing query..." stays in the second query result from the previous query unless run another time.

If we execute the same query twice,  it gets fixed the second time. So the problem is not acknowledging the execution error caused by this [bug.](https://github.com/Microsoft/vscode-mssql/issues/342)

Issue #345, opened 2016-11-14
An Exception in SQL Parser doesn't show any error in result panel


llali:
MSSQL Extension Version: 0.1.5
VSCode Version: 1.7+
OS Version: Windows 10

Steps to Reproduce:

1. Run a script that can cause an exception in SQL Parser. (  creating natively compiled function can cause an exception now)

Expected: The result panel should show an error message that the query execution has failed.
Actual: The result panel shows" Executing" and it won't go away

PR #491, opened 2016-12-09
fixed results pane for various queries

abist:
This fixes problems #484 and #345 where compiling a native function gives a spinning wheel forever with the message "Executing query...". It also fixes another bug, where if you execute 2 or more different queries, then an unsuccessful query retained the result from a previous successful query or a successful query retained the error message from a previous unsuccessful query.


benrr101:
At some point we should try to add unit tests for these scenarios -- they're pretty essential scenarios. But for now, this fixes the issues so we should :shipit:

PR #499, opened 2016-12-10
Removed workaround for showTextDocument

### LINKED ISSUE #482: Update show text document
The vscode team has updated their showtextdocument api (Microsoft/vscode#12280), which we are working around in at least one place in the code https://github.com/Microsoft/vscode-mssql/blob/dev/src/controllers/QueryRunner.ts#L344.

Should consider cleaning up the code accordingly.

coquagli:
Fixes #482. Removed the workaround for `showTextDocument`. This change relies on updates that are present in VS Code - Insiders, but are not yet present in standard VS Code. This change was tested with the most recent VS Code - Insiders build (1.8.0-insider, commit 40243d1e01f426f54670f3dbaa6b77d18469310d). 

I looked through the code and didn't find anywhere else where an obvious workaround is present. It would be best if someone else double-check that, though.

### DIFF
--- a/src/controllers/QueryRunner.ts
+++ b/src/controllers/QueryRunner.ts
@@ -339,22 +339,12 @@ export default class QueryRunner {
         const self = this;
         return new Promise<void>((resolve, reject) => {
             self._vscodeWrapper.openTextDocument(self._vscodeWrapper.parseUri(self.uri)).then((doc) => {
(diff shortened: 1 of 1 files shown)


anthonydresser:
The side effect in vscode versions that do not have the update is that highlighting a selection takes time to resolve.

kevcunnane:
A quick note that this will not be merged until the fix is release in a full (non-insiders) VSCode build.

PR #507, opened 2016-12-12
setting the telemtry proeprties to empty list if it's null

llali:
When the server sends the telemetry notification, the properties can be null when is not set in server. Client was only checking for  "unidentified" so the null properties was causing an error.


kburtram:
I merged my version of the fix with commit https://github.com/Microsoft/vscode-mssql/commit/6cda1b48be40e9dbf90ab7f342c1f38a3ab4e8ea.

PR #512, opened 2016-12-13
updated test to check bad query execution

### LINKED ISSUE #495: Unit tests for query execution in all scenarios
We need to make sure in all scenarios we have
== 1 query start WS event
&gt;= 1 batch start WS event
&gt;= 1 batch complete WS event
== 1 query complete WS event

This will be subject to change in the future, but we need these tests at this point

abist:
Fixes #495. Added tests for new emits and behaviors such as running failing queries.


kevcunnane:
:shipit:

PR #526, opened 2016-12-14
SlickGrid plugin tests

anthonydresser:
This PR will most likely reduce code coverage because I am adding more files to be covered under code coverage. 

1. Specifically I added the 2 JS slickgrid plugins under code coverage. 

2. I've also added a basic test to show how tests can be done on this area.

3. I've changed the karma tests to only run coverage based on the json similarly to the extension code coverage.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,6 +1,12 @@
 // #docregion
 const path = require('path');
-const coverconfig = require('./out/coverconfig.json')
+const fd = require('fs');
+const coverconfig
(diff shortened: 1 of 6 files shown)


kburtram:
It looks like this change is lowering code coverage.  Is there anything we do about that?

kevcunnane:
We need to write additional tests per Anthony's note. It's capturing coverage for .js files so more accurately represents our coverage numbers. We can put an issue to add tests in this area on the backlog and plan to do them

PR #527, opened 2016-12-14
Fix multi-line display formatting and copy formatting 

### LINKED ISSUE #511: Multi-line UDF formatting Issue
- VSCode Version: 1.7.2
- OS Version: Win10

Steps to Reproduce:

1. Add a multi-line user defined function with white space at the top
```


-- Example of DB function
CREATE FUNCTION avgSalary () 
  RETURNS MONEY 
  AS 
  BEGIN 
    DECLARE @salary MONEY 
      SELECT @salary = AVG(SALARY) 
      FROM CUSTOMERS
      RETURN(@salary) 
  END 
```
2. Execute Query to find function definitions: 
```
SELECT name, definition, type_desc 
  FROM sys.sql_modules m 
INNER JOIN sys.objects o 
  ON m.object_id=o.object_id
```

Output:
newlines

SSMS seems to remove all newlines and keep comments:
ssms

Raymondd:
Fixes #511 by adding a property to the settings file 'copyRemoveNewLine' that determines if newlines are removed when data is copied from results grid (default is true). Also always replaces newlines with spaces for data displayed in the results grid (same as SSMS).

### DIFF
--- a/src/controllers/QueryRunner.ts
+++ b/src/controllers/QueryRunner.ts
@@ -354,8 +354,8 @@ export default class QueryRunner {
     private removeNewLines(inputString: string): string {
         // This regex removes all newlines in all OS types
         // Windows(CRLF): \r\n
-        // Linux(LF): \n
-        // MacOS(CR): \r
+        // Linux(LF)/Modern MacOS: \n
+        // Old MacOs: \r
         let outputString: string = inputString.replace(/(\r\n|\n|\r)/gm, '');
         return outputString;
     }
--- a/src/controllers/QueryRunner.ts
+++ b/src/controllers/QueryRunner.ts
@@ -352,7 +352,7 @@ export default class QueryRunner {
     }
 
     private removeNewLines(inputString: string): string {
-        // This regex removes all newlines in all OStypes
+        // This regex removes all newlines in all OS types
         // Windows(CRLF): \n\r
         // Linux(LF): \n
         // MacOS(CR): \r
--- a/src/controllers/QueryRunner.ts
+++ b/src/controllers/QueryRunner.ts
@@ -353,7 +353,7 @@ export default class QueryRunner {
 
     private removeNewLines(inputString: string): string {
         // This regex removes all newlines in all OS types
-        // Windows(CRLF): \n\r
+        // Windows(CRLF): \r\n
         // Linux(LF): \n
         // MacOS(CR): \r
         let outputString: string = inputString.replace(/(\r\n|\n|\r)/gm, '');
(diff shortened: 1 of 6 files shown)


src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -363,9 +363,15 @@ export class AppComponent implements OnInit, AfterViewChecked {
                         return new Promise<IGridDataRow[]>((resolve, reject) => {
                             self.dataService.getRows(offset, count, resultSet.batchId, resultSet.id).subscribe(rows => {
                                 let gridData: IGridDataRow[] = [];
-                                for (let i = 0; i < rows.rows.length; i++) {
+                                for (let row = 0; row < rows.rows.length; row++) {
```
anthonydresser: This is already being done at https://github.com/Microsoft/vscode-mssql/blob/ce0181c16464352c5804eb68a8d73e824bf131b3/src/views/htmlcontent/src/js/components/app.component.ts#L544 I'm confused, was something not working?

src/controllers/QueryRunner.ts
```diff
@@ -339,6 +344,22 @@ export default class QueryRunner {
...
+        let config = this._vscodeWrapper.getConfiguration(Constants.extensionConfigSectionName);
+        let removeNewLines: boolean = config[Constants.configCopyRemoveNewLine];
+        return removeNewLines;
+    }
+
+    private removeNewLines(inputString: string): string {
+        // This regex removes all newlines in all OS types
+        // Windows(CRLF): \r\n
+        // Linux(LF): \n
+        // MacOS(CR): \r
```
anthonydresser: This comment isn't exactly right, old mac os using /r (pre osx) but osx now follows the same new lines as linux https://en.wikipedia.org/wiki/Newline

src/views/htmlcontent/src/js/components/app.component.ts
Raymondd: Checkout bug #511. Seems like it wasn't being done for these cases.

src/controllers/QueryRunner.ts
Raymondd: Yup, you're right. I'll fix that comment.

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -539,13 +534,13 @@ export class AppComponent implements OnInit, AfterViewChecked {
...
     /**
-     * Format all text to replace /n with spaces
+     * Format all text to replace all new lines with spaces
      */
     textFormatter(row: number, cell: any, value: string, columnDef: any, dataContext: any): string {
         let valueToDisplay = value;
         let cellClasses = 'grid-cell-value-container';
         if (value) {
-            valueToDisplay = value.replace(/\n/g, ' ');
+            valueToDisplay = value.replace(/(\r\n|\n|\r)/g, ' ');
```
Raymondd: Yeah, it fixed the bug still. I was slightly confused also. UDF definitions seem to be formatted differently from other output.

PR #531, opened 2016-12-14
Updated ReadMe with a What's New section

kevcunnane:
- Added What's New to match C# extension and make it easier to decide whether to install the update
- Added Supported Operating Systems directly in the Readme since this is important to Linux users and helps avoid confusion
- Added comment on how to install offline. We had issues raised on this and would like to cover the most common acquisition scenarios in our readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,8 @@ Welcome to **mssql** <sup>preview</sup> for Visual Studio Code! An extension for
 
 * Connect to Microsoft SQL Server, Azure SQL Database and SQL Data Warehouses.
 * Create and manage connection profiles and most recently used connections.
-* Write T-SQL script with IntelliSense, T-SQL snippets, syntax colorizations, T-SQL error validations and ```GO``` batch separator.
-* Execute the script.
-* View the result in a slick grid.
+* Write T-SQL script with IntelliSense, Go to Definition, T-SQL snippets, syntax colorizations, T-SQL error validations and ```GO``` batch separator.
+* Execute your scripts and view results in a simple to use grid.
 * Save the result to json or csv file format and view in the editor.
 * Customizable extension options including command shortcuts and more.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,7 +89,27 @@ See [customize options] and [manage connection profiles] for more details.
...
+
+Currently this extension supports the following operatings systems:
+
+* Windows (64-bit only)
+* macOS
+* Ubuntu 14.04 / Linux Mint 17 / Linux Mint 18 / Elementary OS 0.3
+* Ubuntu 16.04 / Elementary OS 0.4
+* Debian 8.2
+* CentOS 7.1 / Oracle Linux 7
+* Red Hat Enterprise Linux (RHEL)
```
benrr101: Version?

README.md
```diff
@@ -83,7 +89,27 @@ See [customize options] and [manage connection profiles] for more details.
...
 ## Change Log
-The current version is ```0.2.0```. See the [change log] for more detail.
+The current version is ```0.2.0```. See the [change log] for a detailed list of changes in each version.
+
+## Supported Operating Systems
+
+Currently this extension supports the following operatings systems:
+
+* Windows (64-bit only)
+* macOS
```
benrr101: Version?

README.md
```diff
@@ -83,7 +89,27 @@ See [customize options] and [manage connection profiles] for more details.
...
 
 ## Change Log
-The current version is ```0.2.0```. See the [change log] for more detail.
+The current version is ```0.2.0```. See the [change log] for a detailed list of changes in each version.
+
+## Supported Operating Systems
+
+Currently this extension supports the following operatings systems:
+
+* Windows (64-bit only)
```
benrr101: Version?

README.md
```diff
@@ -83,7 +89,27 @@ See [customize options] and [manage connection profiles] for more details.
 ```
 
 ## Change Log
-The current version is ```0.2.0```. See the [change log] for more detail.
+The current version is ```0.2.0```. See the [change log] for a detailed list of changes in each version.
+
+## Supported Operating Systems
+
+Currently this extension supports the following operatings systems:
```
benrr101: You should mention that this is "official" support and maybe add a note that it may work on other OS that can run VSCode and dotnet core

README.md
llali: 10.11 and by the way we might have a bug that we don't say 10.12 is not supported. we have to verify

---
In reply to: [92522143](https://github.com/Microsoft/vscode-mssql/pull/531#discussion_r92522143) [](ancestors = 92522143)

llali:
:shipit:

README.md
kevcunnane: Added to the wiki page, let's keep this clean and add the advanced info there.

README.md
kevcunnane: I'm matching the c# extension's doc. We don't discriminate (and at least try to install & run). I'd like to wait until our first Issue of "this fails on RHEL 4" is raised).

README.md
kevcunnane: We build for 10.11 (El Capitan) but run on 10.10 and 12 (Sierra). Effectively should work in all cases

PR #533, opened 2016-12-15
fixed null values highlighting in dark theme

### LINKED ISSUE #510: Foreground text for NULL cell should be white in Dark Theme
- MSSQL Extension Version: 0.2
- VSCode Version: 1.7
- OS Version: macOS

Steps to Reproduce:

1. Set VSCode to dark theme 
2. Connect to a database and execute query `select * from sys.databases`
3. Select any cell in the source_database_id column

Expected:
- Foreground text color is white, as it is for other non-NULL cells
Actual:
- Forground text color is black, making it hard to see.

image

abist:
Fixes #510.

### DIFF
--- a/src/views/htmlcontent/src/css/color-theme-dark.css
+++ b/src/views/htmlcontent/src/css/color-theme-dark.css
@@ -28,6 +28,10 @@ slick-grid.active .grid .slick-cell.selected {
   background-color: var(--color-cell-bg-grid-selected);
 }
 
+.slick-cell .grid-cell-value-container.missing-value {
+  color: var(--color-content) !important;
+}
+
 .boxRow.content.horzBox.slickgrid {
   border: solid 1px #2D2D30;
 }
--- a/src/views/htmlcontent/src/css/color-theme-dark.css
+++ b/src/views/htmlcontent/src/css/color-theme-dark.css
@@ -28,8 +28,8 @@ slick-grid.active .grid .slick-cell.selected {
   background-color: var(--color-cell-bg-grid-selected);
 }
 
-.slick-cell .grid-cell-value-container.missing-value {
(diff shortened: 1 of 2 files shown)


src/views/htmlcontent/src/css/color-theme-dark.css
```diff
@@ -28,6 +28,10 @@ slick-grid.active .grid .slick-cell.selected {
   background-color: var(--color-cell-bg-grid-selected);
 }
 
+.slick-cell .grid-cell-value-container.missing-value {
+  color: var(--color-content) !important;
```
kevcunnane: To help me understan how this works:
- I see this color used for the .header css? Is that css used for all cells?
- What does `!important` do?

abist: - It's the default color for all other entries other than null, I believe, I can check again on chrome debugger.
- !important prioritizes the style given and takes precedence over any other tags which might be affecting that component. I had to put it in because otherwise it was retrieving style from SlickGrid.css, which is a generated .css file

src/views/htmlcontent/src/html/sqlOutput.ejs
```diff
@@ -19,14 +19,14 @@
         font-size: var(--font-size);
       }
     </style>
+    <link rel="stylesheet" type="text/css" href="lib/css/slick.grid.css"/>
+    <link rel="stylesheet" type="text/css" href="lib/css/SlickGrid.css"/>
```
kburtram: have you verified there are not other style changes that result from reordering these CSS files?

abist: I tried fiddling after the changes, and I didn't see any other style changes.

PR #540, opened 2016-12-15
Fix persisting connection even when file is renamed 

### LINKED ISSUE #502: Renaming a file causes connection to be lost
- VSCode Version: 1.8
- OS Version: OpenSuse

Steps to Reproduce:

1. Open a SQL file, e.g., test.sql
2. Connect to a database
3. Rename SQL file, e.g., test2.sql

Expected: Connection is not lost

Actual: Connect is lost

Raymondd:
Fixes #502  I added a rename file detection function based off of the untitled file detection function. I also reworked the onDidCloseTextDocument function to take care of a small bug where is wasn't properly closing some files normally. I will be adding unit tests in another PR for this file which expands beyond these changes, since some other functions (onDidOpenTextDocument, onDidCloseTextDocument, and onDidSaveTextDocument) weren't being covered fully.

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -597,33 +597,17 @@ export default class ConnectionManager {
         }
     }
 
-    public onDidRenameTextDocument(oldFileUri: string, newFileUri: string): void {
-        // Check if the old file is connected and the new one is not connected
-        if (!this.isConnected(newFileUri) && this.isConnected(oldFileUri)) {
-
-            // Connect the saved uri and disconnect the untitled uri on successful connection
-            let creds: Interfaces.IConnectionCredentials = this._connections[oldFileUri].credentials;
-
-            // Connect new Uri using same credential then disconnect old Uri
-            this.connect(newFileUri, creds).then(result => {
-                if (result) {
-                    this.disconnect(oldFileUri);
-                }
-            });
-        }
-    }
-
-    public onUntitledFileSaved(untitledUri: string, savedUri: string): void {
+    public transferFileConnection(oldFileUri: string, newFileUri: string): void {
         // Is the new file connected or the old file not connected?
-        if (!this.isConnected(untitledUri) || this.isConnected(savedUri)) {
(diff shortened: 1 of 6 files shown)


src/controllers/mainController.ts
```diff
@@ -378,33 +380,47 @@ export default class MainController implements vscode.Disposable {
...
+
+        // Determine which event caused this close event
+        if (this._lastSavedUri &&
+                closedDocumentUriScheme === Constants.untitledScheme &&
+                this._lastSavedTimer.getDuration() < Constants.untitledSaveTimeThreshold) {
+            // Untitled file was saved
+            this._connectionMgr.onUntitledFileSaved(closedDocumentUri, this._lastSavedUri);
+
+        } else if (this._lastOpenedUri &&
+                this._lastOpenedTimer.getDuration() < Constants.renamedOpenTimeThreshold) {
```
llali: can you please add comment here explaining why based the condition the file is renamed? it's not clear to me

llali:
I know you want to add test in another PR but the coverage decreasing in this PR. Is there any way you add tests here?

Raymondd:
Sure! I can just add all tests here.

src/controllers/mainController.ts
Raymondd: A file rename event consists of an openTextDocument event immediately followed by a closeTextDocument event. So this condition is basically saying if the last event which occurred was an openTextDocument event and it happened just before the this closeTextDocument event then the file must have been renamed.

src/controllers/connectionManager.ts
```diff
@@ -597,6 +597,22 @@ export default class ConnectionManager {
         }
     }
 
+    public onDidRenameTextDocument(oldFileUri: string, newFileUri: string): void {
```
kburtram: >public onDidRenameTextDocument(oldFileUri: string, newFileUri: string): void [](start = 4, length = 76)

this looks basically the same as onUntitledFileSaved.  Can we add a new method and make both the event handlers call a single implementation of this logic?

PR #542, opened 2016-12-15
Fix results pane column behavior

### LINKED ISSUE #523: Results for the same file appear in different panes when three editor panes are open
VS Code state before `select * from sys.databases` is run in the middle pane:
image

**Actual** VS Code state after `select * from sys.databases` is run in the middle pane:
image

**Expected** VS Code state after `select * from sys.databases` is run in the middle pane:
image

(Note that here I had the same file open twice, but this is not needed to reproduce the issue)

Steps to Reproduce:
1. Open 3 editor panes
2. Open a SQL file in the left or center pane, then run a query
3. Move the results pane to the right pane
4. Run the query again (or any query in the same SQL file)

Raymondd:
Fixes #523 Behavior for results window set to reuse the existing window in its pane. If a new results window has no pane, then it is defaulted to the 2nd pane.


Raymondd:
Results pane duplication fixed. Will open a new issue for default ViewColumn.

PR #544, opened 2016-12-15
Fix/changelog html

kevcunnane:


kevcunnane:
Accidentally based this off Dev. Will fix to base off master instead

PR #551, opened 2016-12-19
Fix infinite loop copying multiple columns

kburtram:
Prevent infinite loop copying multiple columns.


kburtram:
I'm going to merge this into master for Preview 2 release.  This is a one character change to fix an infinite loop bug breaking the "copy multiple columns" scenario.

PR #553, opened 2016-12-19
Update publisher back to ms-mssql.

kburtram:


kburtram:
Publisher must be ms-mssql to publish to our marketplace account.

PR #555, opened 2016-12-19
Add back Getting Started Guide command

kburtram:
The Getting Started Guide command was removed from the package.json file, apparently as the result of a mismerge.  Unfortunately 0.2.0 went out without this command, but we should add it back unless it was an explicit decision to remove this command.


llali:
:shipit:

PR #556, opened 2016-12-19
Switch publisher in telemetry test to ms-mssql

kburtram:


kburtram:
These are "test fix" follow-ups to previous PR to get CI builds.  Please review the commit if there are any follow-ups needed.

PR #557, opened 2016-12-19
Fixing download progress status

llali:
- Fixed the bug with downloading the service from the download center and calculating the download percentage to show in UI.

- refactoring the download code to be able to add more tests.
- Fixed the problem with the temp file was not deleted after installing was done


kburtram:
:shipit:

src/languageservice/httpClient.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+   /*
+    * Calculate the download percentage and stores in the progress object
+    */
+    public handleDataReceivedEvent(progress: IDownloadProgress, data: any, logger: ILogger, statusView: IStatusView): void {
+        progress.downloadedBytes += data.length;
+
+        // Update status bar item with percentage
+        let newPercentage = Math.ceil(100 * (progress.downloadedBytes / progress.packageSize));
```
llali: I can add a check  for 0 just in case

---
In reply to: [93141773](https://github.com/Microsoft/vscode-mssql/pull/557#discussion_r93141773) [](ancestors = 93141773)

src/languageservice/serviceclient.ts
```diff
@@ -132,7 +134,9 @@ export default class SqlToolsServiceClient {
             _channel = window.createOutputChannel(Constants.serviceInitializingOutputChannelName);
             let logger = new Logger(text => _channel.append(text));
             let serverStatusView = new ServerStatusView();
-            let downloadProvider = new ServiceDownloadProvider(config, logger, serverStatusView);
+            let httpClient = new HttpClient();
+            let decompressProvider = new DecompressProvider();
```
llali: I think it's fine. it's lightweight and I need to pass everything in the constructor for unittests

---
In reply to: [93142367](https://github.com/Microsoft/vscode-mssql/pull/557#discussion_r93142367) [](ancestors = 93142367)

PR #565, opened 2016-12-22
Switch publisher in telemetry test to ms-mssql (#556)

zMrAzzz:
* Switch publisher in telemetry test to ms-mssql

* Update a couple more hard-coded publisher references.


kburtram:
@zMrAzzz thanks for preparing this PR.  I think the change to the publisher should already be in master.  Please let me know if there's still something else that needs to happen for this test fix.

kburtram:
I think this change was picked up already.  Please reopen if it's still needed.  Thanks!

PR #569, opened 2016-12-29
added new command to create new SQL query document

llali:
- Added new command "mssql: New query" TODO: not sure what to use for the short but
- The new command creates new SQL query untitled document in the temp folder but because it's untitled, the file doesn't get created so user can decide to save it or not

- the new query command activates the extension is not already activated. 
- The name of the document will be SQLQuery[number].sql 
- after a new SQL query is created the new connection command will be called for user to open a connection

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -16,6 +16,7 @@ import { IPrompter } from '../prompts/question';
 import CodeAdapter from '../prompts/adapter';
 import Telemetry from '../models/telemetry';
 import VscodeWrapper from './vscodeWrapper';
+import UntitledSqlDocumentService from './untitledSqlDocumentService';
 import { ISelectionData } from './../models/interfaces';
 import * as path from 'path';
 import fs = require('fs');
@@ -38,6 +39,7 @@ export default class MainController implements vscode.Disposable {
     private _lastSavedTimer: Utils.Timer;
     private _lastOpenedUri: string;
     private _lastOpenedTimer: Utils.Timer;
+    private _untitledSqlDocumentService: UntitledSqlDocumentService;
 
     /**
      * The main controller constructor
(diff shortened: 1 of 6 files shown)


src/controllers/mainController.ts
```diff
@@ -52,7 +54,10 @@ export default class MainController implements vscode.Disposable {
         }
         if (vscodeWrapper) {
             this._vscodeWrapper = vscodeWrapper;
```
benrr101: We can simplify this with the or operator into `this._vscodeWrapper = vscodeWrapper || new VscodeWrapper()`
Or we could use a ternary operator. But I think we should avoid using if/else for this kind of logic

src/controllers/untitledSqlDocumentService.ts
```diff
@@ -0,0 +1,62 @@
...
+import VscodeWrapper from './vscodeWrapper';
+import vscode = require('vscode');
+import path = require('path');
+import os = require('os');
+const fs = require('fs');
+
+/**
+ * Service for creating untitled documents for SQL query
+ */
+export default class UntitledSqlDocumentService {
```
benrr101: I'm sure you explored this already, but I feel like there has to be some command built into VS Code to just create an untitled document. I'm thinking by opening an in-memory untitled doc, the header for the file would be "untitled 1" or some such. Can we verify that there isn't any way to programmatically create an untitled document before we go ahead with this change?

test/untitledSqlDocumentService.test.ts
```diff
@@ -0,0 +1,128 @@
+import * as TypeMoq from 'typemoq';
```
benrr101: I see there's a good number of tests added here, but Coveralls says this change drops code coverage by 32%. That's a huge change. Can we make sure we understand what's going on with coveralls's assessment before we move forward?

src/controllers/untitledSqlDocumentService.ts
llali: that one doesn't let me change the language to sql

---
In reply to: [94176030](https://github.com/Microsoft/vscode-mssql/pull/569#discussion_r94176030) [](ancestors = 94176030)

test/untitledSqlDocumentService.test.ts
llali: yes not sure why. I see a permission error in travis job, might be related 

---
In reply to: [94176306](https://github.com/Microsoft/vscode-mssql/pull/569#discussion_r94176306) [](ancestors = 94176306)

src/controllers/untitledSqlDocumentService.ts
benrr101: Is it possible to change the language of the document after it's created? I took a look at the API but it claims that the languageId of a TextDocument is readonly. I can't really understand why that would be the case, and I'd want further clarification from VS Code folks.

llali: I tried that also and it didn't work. I created a feature request in VsCode: https://github.com/Microsoft/vscode/issues/17917#issuecomment-269953344

benrr101: Fair enough, seems silly they don't have a programmatic way of doing that, but it's a project that's still growing.

PR #581, opened 2017-01-05
Change explicit keycode values to jquery constants

anthonydresser:
- Removed explicit keycodes in favor of jquery constants.

- Fixed a couple bugs I found during making this change
    - Fixed problem that mousedown event was not properly being applied
    - Fixed selection problem that file was not being loaded


kevcunnane:
:shipit:

PR #582, opened 2017-01-05
Fix new result pane column behavior

### LINKED ISSUE #549: Default column for a new results pane
MSSQL Extension Version: 0.2.0
VSCode Version: 1.8
OS Version: Win10

Steps to Reproduce:
1. Open three columns in VSCode. 
2. Run a query that opens a results pane.

Actual: 
Results pane opens in 2nd column (current default for new panes). 

Expected:
Not sure. Where should the results pane default to?

Raymondd:
Fixes #549.
splitPaneSelection configuration property added. Corresponding behavior added to vscodeWrapper class.

### DIFF
--- a/package.json
+++ b/package.json
@@ -441,8 +441,12 @@
           "type": "boolean",
           "description": "[Optional] Configuration options for copying multi-line results from the Results View",
           "default": true
+        },
+        "mssql.splitPaneSelection": {
+          "type": "string",
+          "description": "[Optional] Configuration options for which column new result panes should open in (Options: {'next'|'same'|'last'})",
+          "default": "next"
         }
-
       }
     }
   },
--- a/package.json
+++ b/package.json
@@ -444,12 +444,12 @@
         },
         "mssql.splitPaneSelection": {
           "type": "string",
-          "description": "[Optional] Configuration options for which column new result panes should open in (Options: {'next'|'same'|'last'})",
+          "description": "[Optional] Configuration options for which column new result panes should open in",
           "default": "next",
           "enum": [
             "next",
-            "same",
-            "last"
+            "current",
+            "end"
           ]
         }
       }
--- a/package.json
+++ b/package.json
@@ -445,7 +445,12 @@
         "mssql.splitPaneSelection": {
           "type": "string",
           "description": "[Optional] Configuration options for which column new result panes should open in (Options: {'next'|'same'|'last'})",
-          "default": "next"
+          "default": "next",
+          "enum": [
+            "next",
+            "same",
+            "last"
+          ]
         }
       }
     }
--- a/src/controllers/vscodeWrapper.ts
+++ b/src/controllers/vscodeWrapper.ts
@@ -269,5 +269,4 @@ export default class VscodeWrapper {
     public get workspaceRootPath(): string {
         return vscode.workspace.rootPath;
     }
-
 }
--- a/src/controllers/vscodeWrapper.ts
+++ b/src/controllers/vscodeWrapper.ts
@@ -269,4 +269,20 @@ export default class VscodeWrapper {
     public get workspaceRootPath(): string {
         return vscode.workspace.rootPath;
     }
+
+    /**
+     * The folder that is open in VS Code. `undefined` when no folder
+     * has been opened.
+     *
+     * @readonly
+     * @see vscode.workspace.rootPath
+     */
+    public doesResultPaneExist(resultsUri: string): boolean {
+        let resultPaneURIMatch = vscode.workspace.textDocuments.find(tDoc => tDoc.uri.toString() === resultsUri);
+        return (resultPaneURIMatch !== undefined);
+    }
+
+    public newResultPaneViewColumn(): vscode.ViewColumn {
+        return vscode.ViewColumn.Two;
+    }
 }
--- a/src/controllers/vscodeWrapper.ts
+++ b/src/controllers/vscodeWrapper.ts
@@ -270,43 +270,4 @@ export default class VscodeWrapper {
         return vscode.workspace.rootPath;
     }
 
-    /**
-     * Returns wether or not a result pane with the same URI exists
-     * @param The string value of a Uri.
-     * @return Existence boolean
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -441,8 +441,17 @@
           "type": "boolean",
           "description": "[Optional] Configuration options for copying multi-line results from the Results View",
           "default": true
+        },
+        "mssql.splitPaneSelection": {
+          "type": "string",
+          "description": "[Optional] Configuration options for which column new result panes should open in (Options: {'next'|'same'|'last'})",
```
kevcunnane: I think you can remove the `(Options: {'next'|'same'|'last'})` part - the enum values are listed as options by VSCode when altering them

package.json
```diff
@@ -441,8 +441,17 @@
...
+          "default": "next",
+          "enum": [
+            "next",
+            "same",
```
kevcunnane: `current` instead of `same`, `end` instead of `last` unless you mean `previous` ?

src/controllers/vscodeWrapper.ts
```diff
@@ -269,4 +269,5 @@ export default class VscodeWrapper {
     public get workspaceRootPath(): string {
         return vscode.workspace.rootPath;
     }
+
```
kevcunnane: Seems like you can undo the change in this file

src/models/SqlOutputContentProvider.ts
```diff
@@ -313,16 +313,15 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
         let paneTitle = Utils.formatString(Constants.titleResultsPane, queryRunner.title);
         // Always run this command even if just updating to avoid a bug - tfs 8686842
 
-        // Find any URIs which match the one we are about to display
-        let resultPaneURIMatch = vscode.workspace.textDocuments.find(tDoc => tDoc.uri.toString() === resultsUri);
 
         // Check if the results window already exists
-        if (resultPaneURIMatch !== undefined) {
+        if (this.doesResultPaneExist(resultsUri)) {
             // Implicity Use existsing results window by not providing an pane
```
kevcunnane: typo `existing` instead of `existsing` (since you're changing this area anyhow)

src/models/SqlOutputContentProvider.ts
```diff
@@ -535,4 +534,51 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
         }
         return path.join(os.tmpdir(), columnName + '_' + String(Math.floor( Date.now() / 1000)) + String(process.pid) + '.' + linkType);
     }
+
+    /**
+     * Returns wether or not a result pane with the same URI exists
```
kevcunnane: typo `whether`

src/models/SqlOutputContentProvider.ts
```diff
@@ -535,4 +534,51 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
...
+     * @param The string value of a Uri.
+     * @return Existence boolean
```
kevcunnane: `@return boolean true if pane exists`

src/models/SqlOutputContentProvider.ts
```diff
@@ -535,4 +534,51 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
...
+     * public for testing purposes
+     */
+    public doesResultPaneExist(resultsUri: string): boolean {
+        let resultPaneURIMatch = vscode.workspace.textDocuments.find(tDoc => tDoc.uri.toString() === resultsUri);
```
kevcunnane: please use `this._vscodeWrapper.textDocuments()` instead of direct references so it's easier to mock.

src/models/SqlOutputContentProvider.ts
```diff
@@ -535,4 +534,51 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
...
+        let splitPaneSelection = config[Constants.configSplitPaneSelection];
+        let viewColumn: vscode.ViewColumn;
+
+
+        switch (splitPaneSelection) {
+        case 'same' :
+            viewColumn = this._vscodeWrapper.activeTextEditor.viewColumn;
+            break;
+        case 'last' :
+            viewColumn = vscode.ViewColumn.Three;
```
kevcunnane: Is three the max value? Could this ever change?

test/sqlOutputContentProvider.test.ts
```diff
@@ -1,105 +1,160 @@
...
+            done();
+        } catch (err) {
+            done(new Error(err));
+        }
+    });
+
+});
+
+
+// TODO: rewrite all the outputprodiver handle tests (old ones kept for reference)
```
kevcunnane: is there an issue tracking it? if so what's the issue #? also typo `outputprovider`

test/sqlOutputContentProvider.test.ts
```diff
@@ -1,105 +1,160 @@
...
+            });
+        }
+
+        class Case {
+            position: number;
+            config: string;
+            expectedColumn: number;
+        }
+
+        let cases: Case[] = [
```
kevcunnane: It would be nice if each of these was a test case. See https://mochajs.org/#dynamically-generating-tests for how to easily convert this (basically define cases at the `suite` level and in the foreach define a test.

package.json
Raymondd: It does mean 'end'. Will change.

src/models/SqlOutputContentProvider.ts
Raymondd: Three is the maximum number of panes one can have open in VSCode.

test/sqlOutputContentProvider.test.ts
Raymondd: I could dynamically generate each test, but would it be worth it to have 9 test cases and an entire suite for this one function?

kevcunnane: Sounds good

PR #585, opened 2017-01-06
Maintain focus in text editor after executing query

### LINKED ISSUE #261: Edit window loses focus after executing query
The editor window loses focus when executing a query.  The experience seems better if the focus isn't lost so that the query can easily be edited and re-executed if needed.  This is consistent with SSMS behavior.

kburtram:
This is to fix #261.  Unfortunately the vscode editor management extensibility points are a bit clunky making the code more complicated than this should be.

### DIFF
--- a/src/models/SqlOutputContentProvider.ts
+++ b/src/models/SqlOutputContentProvider.ts
@@ -330,7 +330,7 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
         }
 
         // reset focus back to the text document after showing query results window
-        previewCommandPromise.then((e, f, g) => {
+        previewCommandPromise.then(() => {
             // get the result pane text editor to determine which column it was shown in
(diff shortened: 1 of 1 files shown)


src/models/SqlOutputContentProvider.ts
```diff
@@ -314,16 +314,38 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
         // Always run this command even if just updating to avoid a bug - tfs 8686842
 
         // Find any URIs which match the one we are about to display
-        let resultPaneURIMatch = vscode.workspace.textDocuments.find(tDoc => tDoc.uri.toString() === resultsUri);
+        let resultPaneTextDocument = vscode.workspace.textDocuments.find(tDoc => tDoc.uri.toString() === resultsUri);
```
kevcunnane: Raymond is checking in a change to this same method - I'd recommend merging his changes in and then applying your fix.

PR #591, opened 2017-01-10
Feature: Progressive Messages

### LINKED ISSUE #576: Progressive Messages, Extension Support
Support for progressive display of messages in the extension

benrr101:
This change enables progressive rendering of messages as they are returned from the service. It also changes the state changes for a query such that the a query will always get a QueryComplete notification if the query is started successfully, and will always get an Error if the query fails to start. Batch start notifications are changed to be sent to the preview pane as a message with a link property. Correspondingly, the handler for changing the text selection in the query now uses a GET handler instead of a POST handler, with the URI being generated in the SQL output provider.

This will fix #576


src/controllers/QueryRunner.ts
```diff
@@ -100,70 +104,39 @@ export default class QueryRunner {
...
-                        executionStart: undefined
-                    }];
-                this.eventEmitter.emit('batchStart', self._batchSets[0]);
-                this.eventEmitter.emit('batchComplete', self._batchSets[0]);
-            } else {
-                // register with the Notification Handler
-                self._notificationHandler.registerRunner(self, queryDetails.ownerUri);
-            }
-            self.eventEmitter.emit('complete');
+            self._notificationHandler.registerRunner(self, queryDetails.ownerUri);
```
Raymondd: Shouldn't there be a handler for result.messages since this is how the service returns errors?

benrr101: Nope, we've removed that. If you get a Result back (ie, the promise is resolved), you're guaranteed to get a QueryCompleted event back and any number of batch start, batch end, resultset, and message events. If you get an error back (ie, the promise is rejected), you're not going to get anything else. The message to show in the extension will be provided in the error

src/controllers/QueryNotificationHandler.ts
```diff
@@ -43,7 +45,7 @@ export class QueryNotificationHandler {
     private handleCompleteNotification(): NotificationHandler<any> {
```
Raymondd: May want to rename this function to handleQueryCompleteNotification to match the one in query runner (handleQueryComplete). Also it will remove ambiguity as to what is complete. Just a suggestion.

src/controllers/QueryRunner.ts
Raymondd: Ohh nice. Okay that looks good then.

src/controllers/QueryRunner.ts
```diff
@@ -198,6 +173,7 @@ export default class QueryRunner {
 
         // Store the batch again to get the rest of the data
         this._batchSets[batch.id] = batch;
+        this._totalElapsedMilliseconds += new Date(batch.executionEnd).getTime() - new Date(batch.executionStart).getTime();
```
Raymondd: Doesn't a batch come with ExecutionElapsedTime property already? Why would be calculate it in here?

benrr101: An unfortunate oversight. The elapsed time is a string and a JS Date can't be instantiated from a TimeSpan string (eg, 00:01:00.00). So, I'm stuck recalculating it, for now.

src/controllers/QueryNotificationHandler.ts
benrr101: Good idea, it does make sense. I'll make that change.

src/controllers/QueryRunner.ts
benrr101: Ideally, I'd like to return the total execution time for the query with the query complete event.

Raymondd:
I think this are good changes. They seems to really streamline the service API calls. I can't really comment on the rendering parts, but all the controller updates look good to me.

src/controllers/QueryRunner.ts
anthonydresser: https://github.com/Microsoft/vscode-mssql/blob/dev/src/views/htmlcontent/src/js/utils.ts#L21

src/controllers/QueryRunner.ts
```diff
@@ -58,6 +60,8 @@ export default class QueryRunner {
         this._uri = _ownerUri;
         this._title = _editorTitle;
         this._isExecuting = false;
+        this.batchSets = [];
```
kburtram: >batchSets [](start = 13, length = 9)

should this _batchSets to be consistent with all the other fields be initialized, or is there a reason to use the setter?

src/models/SqlOutputContentProvider.ts
```diff
@@ -150,10 +150,15 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
...
         });
 
-        // add http post handler for setting the selection in the editor
-        this._service.addPostHandler(Interfaces.ContentType.EditorSelection, (req, res): void => {
+        // add http get handler for setting the selection in the editor
+        this._service.addHandler(Interfaces.ContentType.EditorSelection, (req, res): void => {
             let uri = req.query.uri;
-            let selection: ISelectionData = req.body;
+            let selection: ISelectionData = {
+                startLine: parseInt(req.query.startLine, 10),
```
kburtram: >10 [](start = 57, length = 2)

I assume this is the number base..if 10 is the default value I'd recommend leaving off the parameter to simplify the method call.

kburtram:
:shipit:

src/models/SqlOutputContentProvider.ts
anthonydresser: it is the base but it is not the default. There is no default, the radix for the conversion is a required input for some reason 🙄

src/controllers/QueryRunner.ts
benrr101: Ah, good code. I didn't see that there.

src/controllers/QueryRunner.ts
benrr101: I agree with you. I think this was leftover from before, but I'll change it to make it match.

PR #592, opened 2017-01-10
Removing DataSet/BatchSet Promise

benrr101:
Removing dead code that is no longer being used now that we have the websocket implementation. As far as I can tell, these web server handlers are no longer being called from the preview pane and since they were the only things that hinged on the dataset/batchset promise, we should be able to safely remove these.


benrr101:
Checks are passing locally, will check in.

Issue #587, opened 2017-01-09
Results Tab Renders HTML instead of plain text.


joeygreen:
- MSSQL Extension Version: 0.2.0
- VSCode Version: 1.8.1
- OS Version: Windows 10

Steps to Reproduce:

1. Query a table that has a column containing HTML
2. Check it out in the results tab.

benrr101:
Oh yes, this is a very good catch! Thanks for reporting this. I'll be working on this right away.

benrr101:
Alright, I've got a code change for this issue merged into our dev branch. It will be in our next preview release (which I believe is slated for end of January). If you want to take it for a spin, feel free to take our next nightly build and follow the instructions to side-load it via the VSIX file.

joeygreen:
Thanks for the hard work!

PR #604, opened 2017-01-12
Fix/html encode display values

benrr101:
Addresses #587 

Adding a fairly robust utility function to encode results from databases with HTML entities -- since JS doesn't have a built in one. This fixes issues where displaying result sets with HTML would fail. It also fixes an undiscovered issue where tooltips for cells would be cut off if they contained a `'` or `"`. Since the change is in the presentation layer, copy/paste and save as functionality still works as expected.


anthonydresser:
Would probably be beneficial to add some basic tests since it is not a stand alone function.

benrr101:
@anthonydresser Not a bad idea. I'll add selected tests for this (obviously don't want to enumerate the entire Unicode space)

PR #609, opened 2017-01-12
Fix copy bug by adding null check

### LINKED ISSUE #596: Cannot copy from results grid if selection contains null
Steps to Reproduce:
1. Run a query that will have null values in the resultset 
2. Make a selection in the grid that includes null values

Expected Behavior:
All values are copied to clipboard, including the null values

Actual Behavior:
Nothing is copied to the clipboard. 

Copyable:
image

Uncopyable:
image

coquagli:
Fixes #596. Note that the value in `row` comes back as `null` and not `undefined` for `null` data.

### DIFF
--- a/src/controllers/QueryRunner.ts
+++ b/src/controllers/QueryRunner.ts
@@ -295,7 +295,7 @@ export default class QueryRunner {
                         for (let row of result.resultSubset.rows) {
                             // iterate over the cells we want from that row
                             for (let cell = range.fromCell; cell <= range.toCell; cell++) {
-                                if (row[cell] && self.shouldRemoveNewLines()) {
+                                if (self.shouldRemoveNewLines()) {
                                     // This regex removes all new lines in all forms of new line
                                     copyString += self.removeNewLines(row[cell]) + '\t';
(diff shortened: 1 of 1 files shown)


src/controllers/QueryRunner.ts
```diff
@@ -295,7 +295,7 @@ export default class QueryRunner {
                         for (let row of result.resultSubset.rows) {
                             // iterate over the cells we want from that row
                             for (let cell = range.fromCell; cell <= range.toCell; cell++) {
-                                if (self.shouldRemoveNewLines() && row[cell]) {
+                                if (row[cell] && self.shouldRemoveNewLines()) {
```
abist: shouldn't we check for null explicitly?

coquagli: I am not sure. This works currently, but JavaScript as some subtleties to it that might matter for things like this.

abist: you're right
http://stackoverflow.com/questions/5515310/is-there-a-standard-function-to-check-for-null-undefined-or-blank-variables-in

PR #615, opened 2017-01-14
Add QueryRunner queue to ensure correct notification ordering

### LINKED ISSUE #613: Fix race condition in QueryRunner event handling
The [changes to connection behavior in the tools service](https://github.com/Microsoft/sqltoolsservice/pull/176) has exposed a race condition with how the QueryRunner events are handled in `QueryNotificationHandler`. Messages like the following indicate that it is possible for`registerRunner` to be called after `handleBatchStartNotification`:

`Notification handler 'query/batchStart' failed with message: Cannot read property 'handleBatchStart' of undefined`

We should fix this by ensuring that `handleQueryCompleteNotification` and similar functions can only be called after `registerRunner`. The current proposed solution is to queue up events if they arrive before `registerRunner` is called.

coquagli:
Fixes #613. If `registerRunner` is not called first, notifications are put in a queue and are run once `registerRunner` is called.

I was thinking of writing tests for this, but was unsure how. Is it possible to fire fake notifications using a real or mock `SqlToolsServiceClient`?

### DIFF
--- a/src/controllers/QueryNotificationHandler.ts
+++ b/src/controllers/QueryNotificationHandler.ts
@@ -16,6 +16,7 @@ import { NotificationHandler } from 'vscode-languageclient';
 export class QueryNotificationHandler {
     private static _instance: QueryNotificationHandler;
     private _queryRunners = new Map<string, QueryRunner>();
+    private _handlerCallbackQueue: ((run: QueryRunner) => void)[];
 
     static get instance(): QueryNotificationHandler {
         if (QueryNotificationHandler._instance) {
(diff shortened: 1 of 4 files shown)


coquagli:
Seeing as there is a coverage drop, I will be writing tests for this. The following scenarios seem important to test for notifications and cleanup being handled correctly:
- `registerRunner` comes in first (queue not used)
- `registerRunner` comes in somewhere in the middle (queue partially used)
- `registerRunner` comes in somewhere in the last (queue fully used)
- `QueryRunner` is used then reused by a subsequent query


After setting up the notification handler and the callback (e.g. using [onNotification](https://github.com/Microsoft/vscode-mssql/blob/967108cfdf554f922d8faa50c424938b5abee42f/src/languageservice/serviceclient.ts#L328) like [this](https://github.com/Microsoft/vscode-mssql/blob/4d2e6910eb7247fc7b23bcca4d6cdb5d26fee910/src/controllers/QueryNotificationHandler.ts#L33)), how do I trigger the callback? Through [sendRequest](https://github.com/Microsoft/vscode-mssql/blob/967108cfdf554f922d8faa50c424938b5abee42f/src/languageservice/serviceclient.ts#L317)?

coquagli:
Tests are in, so ping on this.

src/controllers/QueryNotificationHandler.ts
```diff
@@ -36,50 +41,102 @@ export class QueryNotificationHandler {
...
+        // _handlerCallbackQueue will be non-empty. Run all handlers in the queue first
+        // so that notifications are handled in order they arrived
+        while (this._handlerCallbackQueue.length > 0) {
+            let handler: NotificationHandler<any> = this._handlerCallbackQueue.shift();
+            handler(runner);
+        }
+
+        // Set the runner for any other handlers if the runner is in use by the
+        // current query or a subsequent query
+        if (runner.hasCompleted === false) {
```
benrr101: `!runner.hasCompleted`?

coquagli: Oops. #truthiness

Issue #625, opened 2017-01-18
Tests: Update the service requests handler tests 


Raymondd:
Update and refactor the commented out tests in the webServiceRequestsHandler.test.ts file and create a new test suite for them. It is currently difficult to test them the way they are set up, may need some refactoring.

Issue #584, opened 2017-01-06
Tests: Update 'SqlOutputProvider Tests' Suite


Raymondd:
Started in #582.
Update and refactor the commented out tests the SqlOutputContentProvider.test.ts file and re-add them to the 'SqlOutputProvider Tests' Suite.

PR #626, opened 2017-01-18
Tests: Output Content Provider query related functionality 

Raymondd:
Addresses #584.
The request handler tests seem like they may need to be refactored to be properly testable. I created Issue #625 to track these changed since they are going to be significantly different from these query related tests.


test/webServiceRequestHandlers.test.ts
```diff
@@ -0,0 +1,76 @@
+// TODO: rewrite all the outputprovider handle tests (old ones kept for reference)
```
kevcunnane: I'd remove this test class if it's no longer relevant. People can go back in Git history if they need to retrieve at all. Unless there's a reason to preserve?

src/models/SqlOutputContentProvider.ts
```diff
@@ -275,8 +275,12 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
         queryRunner.runQuery(selection);
         let paneTitle = Utils.formatString(Constants.titleResultsPane, queryRunner.title);
         // Always run this command even if just updating to avoid a bug - tfs 8686842
+        this.displayResultPane(resultsUri, paneTitle);
+    }
 
-        // Check if the results window already exists
+    // Intentionally created anonymously for testing purposes
+    public displayResultPane = function(resultsUri: string, paneTitle: string): void {
```
kevcunnane: This is kinda goofy. I think using TypeMoq you can just mock a real instance of this class and override this one function, avoiding the need to add in a setDisplayResultPane and declare this anonymously. However it's not critical to fix.

test/webServiceRequestHandlers.test.ts
Raymondd: I implement this TODO in another PR. Should have probably created this file in that one, but I will send the other one in right after this one, so this shouldn't be a problem.

src/models/SqlOutputContentProvider.ts
Raymondd: I couldn't do a partial Moq, but I did fix this goofy syntax in the class.

Issue #641, opened 2017-01-23
Localization framework setup for the extension


Raymondd:
Setup a working framework for localization on the extension side. 
Resources: 
[NLS Library](https://github.com/Microsoft/vscode-nls)
[NLS Setup Scripts](https://github.com/Microsoft/vscode-nls-dev)

PR #647, opened 2017-01-25
Task: localization framework setup

Raymondd:
Addresses #641 

- modified gulp build command to also run localization processes
- added localization directory with all needed sub-folders
- added an example of machine translated localization 
- setup localization activation based on vscode's environment 

It worked for me in Spanish, but that is about all I can test personally. 
May add another task for machine translations.


Raymondd:
going to continue work on this branch before I pull.

PR #655, opened 2017-01-31
Fix header scroll/selection issues

### LINKED ISSUE #622: Results grid resets scroll location and selection when resizing a column
- MSSQL Extension Version: 0.2.0 alpha 18 (with fix #604)
- VSCode Version: 1.8.1
- OS Version: Windows 10 x64

Steps to Reproduce:

1. Open an SQL file or create a new one.
2. Connect to a DB and run any query that would have enough results to scroll.
3. Scroll down and select anything.
4. Resize a column.

Expected:
The column is resized and current scroll position and cell selection remains intact.

Actual:
The resized column is now selected entirely and the scroll position resets to the top.

coquagli:
Fixes #622. Specifically, this fixes some of the annoying behavior regarding what happens when columns are resized or when the column headers are clicked. 

Currently, if a `click` event occurs in the column header, all cells for that column are selected. Additionally, when resizing a column, if the `click` event occurs in the header, all cells in the modified column would become selected and the column would be resized. The ideal behavior would be to disable the "select all cells in this column" behavior for the next click if a column is resized. This can be accomplished by ignoring the next click all together after a column resize. 

A `click` event is only fired by the DOM if `mousedown` and `mouseup` occur in the same HTML element. Also, SlickGrid only exposes `click` (not `mousedown` and `mouseup`). So it seems that the events exposed by SlickGrid are insufficient to correctly check for all cases that can occur. The missing case is when a `mousedown` for a column resize occurs in the headers, and the respective `mouseup` occurs outside the headers. For this case I added the `setTimeout` to act as somewhat of a "mock `mouseup`". This is fairly janky, so let me know if this is not acceptable and if you have any other ideas to solve the issue.

### DIFF
--- a/src/views/htmlcontent/src/js/slick.dragrowselector.js
+++ b/src/views/htmlcontent/src/js/slick.dragrowselector.js
@@ -273,7 +273,7 @@
                 _ranges.push(new Slick.Range(0, last, _grid.getDataLength()-1, last));
             } else {
                 _ranges = [new Slick.Range(0, columnIndex, _grid.getDataLength()-1, columnIndex)];
-                _grid.setActiveCell(0, columnIndex + 1);
+                _grid.resetActiveCell();
             }
             setSelectedRanges(_ranges);
             e.stopImmediatePropagation();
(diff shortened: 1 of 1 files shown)


kburtram:
Do you know what happened to the code coverage run with this PR?

PR #656, opened 2017-02-01
Support Copy&Paste of Messages

### LINKED ISSUE #385: Copy & Paste from output message doesn't work as expected
- MSSQL Extension Version: 0.1.5
- VSCode Version: 1.7.1
- OS Version: Ubuntu 16.04 LTS

After running the query, it is not possible to copy the results from the output messages.

Steps to Reproduce:

1. Run any query (eg, select * from sys.dm_exec_query_stats)
2. Go to the Result pane, and try to copy the message (eg, "Total execution time")
3. It will copy the selected cell from Grid Results instead of the text message

kevcunnane:
Fixes #385. Added keyboard shortcut `Ctrl+C` and context menu support for copying from the Messages section. To do this:
- Messages have a context menu handler with a single option of `Copy`.
- Added use of the Rangy library which simplifies selection and text range manipulation, including pretty print support for copying as text. We need to copy as text since the HTML includes our anchor links to the current page and in dark mode the text is copied as black text on a black background


src/views/htmlcontent/src/html/app.html
```diff
@@ -16,6 +16,8 @@
                         showHeader="true"
                         [resized]="dataSet.resized"
                         (mousedown)="navigateToGrid(i)"
+                        [selectionModel]="selectionModel"
```
coquagli: I'm fairly certain that this file does absolutely nothing and can probably be removed. The hardcoded string in app.component.ts is where the actual HTML is located.

kevcunnane: No it's there for reference only. Talking with Anthony he found it useful to have it for HTML intellisense, but it was getting quite out of date (as you can see from my update). In the future we could remove it, but I'm trying to be consistent for now.

src/views/htmlcontent/src/html/messagescontextmenu.component.html
```diff
@@ -0,0 +1,5 @@
+<ul class="contextMenu" style="position:absolute" [class.hidden]="!visible" [style.top]="position.y" [style.left]="position.x">
```
kevcunnane: This file is also for reference only, in order to be consistent with our other modules.

src/views/htmlcontent/src/js/components/messagescontextmenu.component.ts
```diff
@@ -0,0 +1,86 @@
...
+        const self = this;
+        $(document).on('click', () => {
+            self.hide();
+        });
+    }
+
+    show(x: number, y: number, selectedRange: IRange): void {
+        this.selectedRange = selectedRange;
+        let selectedText = (selectedRange && selectedRange.toString) ? selectedRange.toString() : '';
+        this.isDisabled = selectedText.length === 0;
```
coquagli: This context menu poses an interesting scenario if text isn't selected because `Copy` is the only option. It looks like different applications handle context menu copying differently - some don't list `Copy` if text is not selected, and some do what you did and grey it out. I think the current behavior is probably best, but I just want to make sure you thought of the possibility of disabling the context menu when there is no text selection.

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -163,9 +173,15 @@ export class AppComponent implements OnInit, AfterViewChecked {
...
         },
         'event.copySelection': () => {
-            let activeGrid = this.activeGrid;
-            let selection = this.slickgrids.toArray()[activeGrid].getSelectedRanges();
-            this.dataService.copyResults(selection, this.renderedDataSets[activeGrid].batchId, this.renderedDataSets[activeGrid].resultId);
+            let range: IRange = this.getSelectedRangeUnderMessages();
+            let messageText = range ? range.text() : '';
+            if (messageText.length > 0) {
+                this.executeCopy(messageText);
+            } else {
```
coquagli: Considering that we don't have a strong differentiation of focus between the messages container and the grid, when the last click occurred in the messages container should a Ctrl-C copy the selected grid cells? This is the current behavior.

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -442,6 +461,58 @@ export class AppComponent implements OnInit, AfterViewChecked {
...
+            selectedRange = sel.getRangeAt(0).intersection(elRange);
+        }
+        elRange.detach();
+        return selectedRange;
+    }
+
+    // Copy text as text
+    executeCopy(text: string): void {
+        let input = document.createElement('textarea');
+        document.body.appendChild(input);
```
coquagli: Might also be a good idea to ensure that this is never visible. Something like `input.style.position = 'absolute'; input.style.bottom = '100%'` might work.

src/views/htmlcontent/src/js/components/messagescontextmenu.component.ts
kevcunnane: I did think about this. My worry is that disabling it completely impacts discoverability - how will the user know it's available? I think having a context menu but greying out works reasonably well. It's definitely a tradeoff though.

src/views/htmlcontent/src/js/components/app.component.ts
kevcunnane: Yeah that's how it works with this code, and I'd consider it a limitation / non ideal behavior. We could add a click event and explicitly deselect all active cells in the grid if this happens, but it's kinda hacky and not great. I've gone for "if message text is selected, use it, otherwise always copy cells". Not exactly perfect though. Let me know if you want a change and I can work to get click-based deselect working

src/views/htmlcontent/src/js/components/app.component.ts
kevcunnane: Will implement this and add a new commit

kevcunnane:
Yes, Ctrl-A is an option  we could definitely look at. For now I'm going to leave it out of scope. I think we may need tab order so that users can move to the Messages section and use keyboard selection in general for improved accessibility. We can look at those as a separate issue.

src/views/htmlcontent/src/js/components/messagescontextmenu.component.ts
coquagli: Those were my exact thoughts.

src/views/htmlcontent/src/js/components/app.component.ts
coquagli: I think the current behavior is fine but not ideal. It's hard to make a comparison with SSDT and SSMS because they put messages and resultsets in different tabs.

PR #665, opened 2017-02-02
Fixing query/dispose API contract

benrr101:
As per Microsoft/sqltoolsservice#223, the query/dispose API will send a result only upon successful disposal of the query. An error will be send if disposal failed for some reason.

I will update the version of the tools service as soon as that pull request is completed.


kevcunnane:
>     public dispose(): Promise<void> {

Do we have unit tests for this? Seems like a good chance to get some in :)

---
Refers to: src/controllers/QueryRunner.ts:230 in 51bf540. [](commit_id = 51bf540e542ba903f49ef73b1e9010dc308d180b, deletion_comment = False)

kevcunnane:
:shipit:

PR #670, opened 2017-02-03
Fix Results Panel not Refreshing Automatically

### LINKED ISSUE #669: Results Panel not Refreshing Automatically
- VSCode Version: 1.9.0
- OS Version: Windows 10

Just upgraded to stable version 1.9.0 and noticed that the results panel does not refresh automatically every time I run the query. It used to work fine with the previous version of VSCode.

Steps to Reproduce:

1. Execute a query
2. Modify the query and run again, the results pane will not automatically refresh with the results for the modified query.

kevcunnane:
- Fixes #669
- Ensures that TextDocumentContentProvider.provideTextDocumentContent always provides an updated string when being refreshed. In VSCode 1.9.0, it's vital to do this as there is a diff on the provided content and the document is only refreshed if this is different

### DIFF
--- a/src/models/SqlOutputContentProvider.ts
+++ b/src/models/SqlOutputContentProvider.ts
@@ -411,13 +411,13 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
     public provideTextDocumentContent(uri: vscode.Uri): string {
         // URI needs to be encoded as a component for proper inclusion in a url
         let encodedUri = encodeURIComponent(uri.toString());
-
-        // return dummy html content that redirects to 'http://localhost:<port>' after the page loads
+        let timeNow = new Date().getTime();
         return `
         <html>
         <head>
             <script type="text/javascript">
                 window.onload = function(event) {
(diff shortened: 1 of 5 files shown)


src/models/SqlOutputContentProvider.ts
```diff
@@ -411,13 +411,13 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
     public provideTextDocumentContent(uri: vscode.Uri): string {
         // URI needs to be encoded as a component for proper inclusion in a url
         let encodedUri = encodeURIComponent(uri.toString());
-
-        // return dummy html content that redirects to 'http://localhost:<port>' after the page loads
+        let timeNow = new Date().getTime();
```
anthonydresser: Might be good to add a comment referencing a github issue to know why this was done in the future when everyone has forgotten about this problem.

kevcunnane: Good point - updated in latest commit

Issue #653, opened 2017-01-31
Localization framework setup for sqltoolsservice


Raymondd:
Setting up localization framework as a .Net Core application. It will be quite a bit different from the extension setup. 
[.Net Core Localization Guide](https://microsoft.sharepoint.com/teams/celoc4dev/SitePages/DotNetCoreGlobalization.aspx)

PR #675, opened 2017-02-03
Task: localization for the extension

Raymondd:
Addresses task #653 

-Moved constants into a folder and split them into to categories (localizable and non-localizable) 
- Updated all imports for localizable constants
- Added localization workflow to gulp build task
- Add new gulp tasks for XLIFF file conversions
- Added example XLIFF file translations 
- Added switch to have localization off by default 
- Added CLI to send locale to sqltoolsextension  (behind localization switch)


src/constants/localizedConstants.ts
```diff
@@ -0,0 +1,275 @@
+import * as nls from 'vscode-nls';
+
+// Localizable Strings (localized at runtime)
```
kevcunnane: Should we document how to add new strings to this? E.g. do we need to update the enu xliff file at the same time? What process will we have for keeping the strings in a localizable state?

test/localization.test.ts
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+import * as LocalizedConstants from '../src/constants/localizedConstants';
+import assert = require('assert');
+
+suite('Localization Tests', () => {
+
+    test('Default Localiztion Test' , done => {
```
kevcunnane: typo - `Localization` not `Localiztion`. Please fix in all 3 tests

kevcunnane:
:shipit:

src/constants/localizedConstants.ts
Raymondd: If the string is to be localized then yes, you would want to add a string to this file and then to the xliff file. I was thinking of adding another automation step (like the tools service has) where you just add your new string to a json file and  run a gulp task to put it into the XLIFF and into the proper constants file.

Raymondd:
I want https://github.com/Microsoft/sqltoolsservice/pull/226 to go in first, so I can bump the sqltools service version in this PR to keep the localization branches in sync with each other.

PR #677, opened 2017-02-04
Fix: cleaning previewHtml api calling logic

### LINKED ISSUE #676: PreviewHtml API usage throws an error
Our usage of the previewHtml command to vscode throws and error due to incorrect usage. 
Pointed out in this issue with vscode: https://github.com/Microsoft/vscode/issues/19654

Raymondd:
Fixes #676 . 

- Only calling this.update when a results pane is already open 
- Bug where an open results pane that is hidden cannot be found by `this.doesResultPaneExist(resultsUri)` is still present 
(This issue is going to be raised to the vscode team, because it looks like it is a bug in the `vscode.workspace.textDocuments` api call which is not able to find hidden previewHtml documents)

### DIFF
--- a/src/models/SqlOutputContentProvider.ts
+++ b/src/models/SqlOutputContentProvider.ts
@@ -289,6 +289,9 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
 
     // Function to render resultspane content
     public displayResultPane(resultsUri: string, paneTitle: string): void {
+        // Get the active text editor
+        let activeTextEditor = this._vscodeWrapper.activeTextEditor;
+
         // Check if the results window already exists
         if (this.doesResultPaneExist(resultsUri)) {
             // Update the existing results pane with the given reusltsUri
@@ -307,7 +310,6 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
                     resultPaneColumn = resultPaneTextEditor.viewColumn;
                 }
 
-                let activeTextEditor = this._vscodeWrapper.activeTextEditor;
                 // only reset focus to the text editor if it's in a different column then the results window
                 if (resultPaneColumn !== undefined
                     && resultPaneColumn !== activeTextEditor.viewColumn) {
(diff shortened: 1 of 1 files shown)


src/models/SqlOutputContentProvider.ts
```diff
@@ -289,36 +289,34 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
...
+        // Get the active text editor
         let activeTextEditor = this._vscodeWrapper.activeTextEditor;
-        let previewCommandPromise;
-        let resultPaneColumn;
+
+        // Check if the results window already exists
         if (this.doesResultPaneExist(resultsUri)) {
-            // Implicity Use existing results window by not providing a pane
-            previewCommandPromise = vscode.commands.executeCommand('vscode.previewHtml', resultsUri, paneTitle);
+            // Update the existing results pane with the given reusltsUri
```
kevcunnane: typo: `resultsUri`

src/models/SqlOutputContentProvider.ts
```diff
@@ -289,36 +289,34 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
...
+            this.update(vscode.Uri.parse(resultsUri));
```
kevcunnane: Aren't we already calling this in another file? Won't this double-reload? I'm basically wondering if we need to do anything in this case

src/models/SqlOutputContentProvider.ts
```diff
@@ -289,36 +289,34 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
...
-            let resultPaneTextEditor = this._vscodeWrapper.visibleEditors.find(
-                editor => editor.document.uri.toString() === resultsUri);
-
-            // get the result pane column from the text editor
-            if (resultPaneTextEditor !== undefined) {
-                resultPaneColumn = resultPaneTextEditor.viewColumn;
-            }
+            let resultPaneColumn = this.newResultPaneViewColumn();
+            // Open new window then reset focus back to the editor
+            vscode.commands.executeCommand('vscode.previewHtml', resultsUri, resultPaneColumn, paneTitle).then(() => {
```
kevcunnane: How about we add in some error handling this time? At least log to the output channel if something unexpected happens...

src/models/SqlOutputContentProvider.ts
Raymondd: This is also being called directly after a query is run, so yeah it is being double reloaded. I will remove this one.

src/models/SqlOutputContentProvider.ts
Raymondd: Adding error catching and logging.

PR #685, opened 2017-02-07
Fix: Removed trailing tabs from copy/paste

### LINKED ISSUE #668: Pasted columns into CSV in Excel from Results panel add extra blank column
- VSCode Version: 1.8.1
- OS Version: Windows 7 Enterprise - 64-bit  - SP1

Steps to Reproduce:

1.  open new file
2.  switch language to sql
3.  connect to db
4.  run select query
5.  select column from results with cursor
6.  copy
7.  paste to CSV already open in excel (2013)
8.  column pastes correctly but extra blank column overwrites column to the immediate right

If you copy 2 columns only one blank column will end up to the right.  It's always n+1 columns pasted.

This is a minor thing because I can export to a CSV just fine and copy over to my other CSV.  It just adds a couple steps.

Thanks!

Raymondd:
Fixes #668.
This was code I recently touched, so I figured I would take a look and the fix was simple. 

- Prevents tabs from being placed at the end of a line

### DIFF
--- a/src/controllers/QueryRunner.ts
+++ b/src/controllers/QueryRunner.ts
@@ -15,6 +15,7 @@ import { QueryCancelParams, QueryCancelResult, QueryCancelRequest } from '../mod
 import { ISlickRange, ISelectionData } from '../models/interfaces';
 import Constants = require('../models/constants');
 import * as Utils from './../models/utils';
+import * as os from 'os';
 
 const ncp = require('copy-paste');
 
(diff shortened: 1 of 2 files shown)


src/controllers/QueryRunner.ts
```diff
@@ -275,25 +275,18 @@ export default class QueryRunner {
                         if (self.shouldIncludeHeaders(includeHeaders)) {
                             let columnHeaders = self.getColumnHeaders(batchId, resultId, range);
                             if (columnHeaders !== undefined) {
-                                for (let header of columnHeaders) {
-                                    copyString += header + '\t';
-                                }
-                                copyString += '\r\n';
+                                copyString += columnHeaders.join('\t') + '\r\n';
```
kevcunnane: >\r\n [](start = 74, length = 4)

Minor: we should use os.EOL for the newline character instead or \r\n. On Unix (Mac and Linux) it's just \n. See [http://stackoverflow.com/questions/10864486/node-js-constant-for-platform-specific-new-line](http://stackoverflow.com/questions/10864486/node-js-constant-for-platform-specific-new-line)

kevcunnane:
:shipit:

PR #698, opened 2017-02-15
TSql Formatter support

### LINKED ISSUE #617: Feature. SQL code formatting
- VSCode Version: v 1.8.1
- OS Version: Windows 10. x64

Hi guys! First of all,  Thank you for that great extension. 
Suggestion is will be better in the future to have SQL code formatting. Some extension for SQL Management Studio like ApexSQL refactor has this feature... 
For the initial realization simple formatting is enough(very common), without customization... 

So why I like use this extension with vs code? Because is working smoothly and fast. This is most important 👍

kevcunnane:
Fixes #617. Adds support for formatter settings and includes an updated SqlToolsService reference with the formatter implemented


package.json
```diff
@@ -456,9 +456,45 @@
...
+        },
+        "mssql.format.placeSelectStatementReferencesOnNewLine": {
+          "type": "boolean",
+          "description": "Should references to objects in a select statements be split into separate lines? E.g. for 'SELECT C1, C2 FROM T1' both C1 and C2 will be on separate lines",
+          "default": false
+        },
         "mssql.applyLocalization": {
           "type": "boolean",
           "description": "[Optional] Configuration options for localizing into VSCode's configured locale (must restart VSCode for settings to take effect)",
+
```
Raymondd: extra line, probably added by mistake?

package.json
```diff
@@ -456,9 +456,45 @@
             "end"
           ]
         },
+        "mssql.format.alignColumnDefinitionsInColumns": {
+          "type": "boolean",
+          "description": "Should column definitions be aligned?",
```
Raymondd: are these settings optional? I think the convention we were using before is to prepend the description with [Optional] if they are.

kevcunnane: I'm trying to match other formatter settings. None of these use the [Optional] tag and it's more concise. I'd prefer to keep it this way since, by definition, formatter settings are optional.

image

package.json
kevcunnane: Yes, happened during merge. Will push a fix

PR #699, opened 2017-02-15
Fix for gulp-cli install failure

kevcunnane:
- The workaround for gulp install stopped working as the necessary gulp cli version no longer worked. This was breaking Jenkins builds
- Updated per recommendation in https://github.com/gulpjs/gulp-cli/issues/84


anthonydresser:
What exactly is this supposed to be fixing? I don't think gulp cli should be a part of that command, but you definitely need gulp 4.0 to be part of it.

PR #719, opened 2017-02-23
Fix release notes and 1 config option

kevcunnane:
- Config option was not setup to support discovery of the "saveAsCsv.IncludeHeaders" option - it only gave an empty object instead of the true/false option that's standard for other settings. Updated & verified that the option is still correctly passed through to the SqlToolsService
- Updated release notes, readme & changelog to reflect new release
- Updated Readme to include all config options

### DIFF
--- a/src/views/htmlcontent/src/docs/index.html
+++ b/src/views/htmlcontent/src/docs/index.html
@@ -3,7 +3,7 @@
 
 <h2 id="7">Version 0.3.0</h2>
 <ul>
-<li>Release date: February, 2016</li>
+<li>Release date: February 24, 2016</li>
 <li>Release status: Public Preview</li>
 </ul>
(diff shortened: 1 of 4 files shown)


src/views/htmlcontent/src/docs/index.html
```diff
@@ -3,7 +3,7 @@
 
 <h2 id="7">Version 0.3.0</h2>
 <ul>
-<li>Release date: February, 2016</li>
+<li>Release date: February 24, 2016</li>
```
kburtram: >li>Release date: February 24, 2016< [](start = 1, length = 35)

still haven't removed this file yet?  we should switch to popping up a web page for the next release.

kburtram:
:shipit:

PR #736, opened 2017-02-27
Fixing regression where null cells are copied as blank strings

benrr101:


kburtram:
Do you know if anyone is looking at why the CI builds are failing?

benrr101:
No idea. But from the history, it looks like it isn't just my changes.

kburtram:
Please make sure to get clean CI builds before merging.

PR #743, opened 2017-02-28
Fix CI test runs broken by gulp-remote-src

kevcunnane:
- gulp-remote-src v0.4.2 breaks the vscode test runner due to https://github.com/ddliu/gulp-remote-src/issues/11. This fixes the issue for us by explicitly binding to 0.4.1 which still works


benrr101:
Thank you Kevin.... 😃
Apologies I didn't solve it before you got to it

benrr101:
Question -- should we be merging this straight into master?

kevcunnane:
I chose master since we're focused on releasing. Will merge back to Dev after. It's up to you - I can rebase if needed.

anthonydresser:
Are you sure this is explicitly binding to 0.4.1? That's not my understanding of the syntax you used. Might want to double check what version a npm install pulls down.

benrr101:
Understood, I can see the reason. Buuuut, we have master marked as a protected branch. I don't think you're even allowed to do a PR to master.

kevcunnane:
@anthonydresser - I used the `npm install --save-dev gulp-remote-src@0.4.1` and verified that this pulls down that version into my local machine, and fixed the issue. Hoping AppVeyor build confirms this.

@benrr101 that's not what protected means. We have both dev and master that can be targeted for PRs. Protected means certain actions like rewriting history and deleting are blocked.

anthonydresser:
@kevcunnane directly installing it might install the correct version, but will running just `npm install` install the correct version? since that is what the CI machines will be doing. The CI machines will soon let us know.

benrr101:
Ah. The stipulation is that the PR has to pass its pre-checkin validations and have the assigned reviewers sign off.

Proceed 🤜

anthonydresser:
This is what I was afraid of
`npm install
.......
+-- gulp-remote-src@0.4.2 `

https://ci.appveyor.com/project/kburtram/vscode-mssql/build/1.0.420#L763

kevcunnane:
Well I stand corrected :) So it looks like removing the ^ should fix this? I'll try that and go again.

kevcunnane:
And... that now works. Thanks Anthony! Please review based on the latest commit.

anthonydresser:
Is this the reason Travis is no longer running?

kevcunnane:
@anthonydresser yes, this is what broke. I need to create an issue against VSCode since any other extension running tests has this. not sure why they don't hit it in their own extension CI runs.

PR #758, opened 2017-03-02
Updater extension to 0.4.0

kevcunnane:
- This will make detecting dev users easier since we're already on a newer version than the publicly released version.


kburtram:
@kevcunnane do we still need this PR or should we close?

kevcunnane:
I'll close it out. Given recent discussions, we should bump this to 1.0 in the next week or so and can plan for future alpha releases to be at that version number instead.

PR #766, opened 2017-03-08
using the new VS code API to create untitled document for SQL language

llali:
- had to update the vs code engine to 1.9.0 to use the new API

- had to fix some codes in test classes because of using new engine version

- removed some of the tests for untitledSqlDocumentService because the code in the service related to tests were removed.

### DIFF
--- a/test/mainController.test.ts
+++ b/test/mainController.test.ts
@@ -112,14 +112,17 @@ suite('MainController Tests', () => {
     // Saved Untitled file event test
     test('onDidCloseTextDocument should call untitledDoc function when an untitled file is saved' , done => {
         // Scheme of older doc must be untitled
-        document.uri.scheme = LocalizedConstants.untitledScheme;
+        let document2 = <vscode.TextDocument> {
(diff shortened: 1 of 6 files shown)


test/mainController.test.ts
```diff
@@ -112,14 +112,17 @@ suite('MainController Tests', () => {
...
     test('onDidCloseTextDocument should call untitledDoc function when an untitled file is saved' , done => {
         // Scheme of older doc must be untitled
-        document.uri.scheme = LocalizedConstants.untitledScheme;
+        let document2 = <vscode.TextDocument> {
+            uri : vscode.Uri.parse(`${LocalizedConstants.untitledScheme}:${docUri}`),
+            languageId : 'sql'
+        };
 
-        // A save untitled doc constitutes an saveDoc event directly followed by a closeDoc event
+        // // A save untitled doc constitutes an saveDoc event directly followed by a closeDoc event
```
kburtram: extra //

PR #784, opened 2017-03-19
Fix #782 UI for Export as Excel

### LINKED ISSUE #782: UI for Export As Excel
Now that we have support in the sqltoolsservice for exporting to Excel files (thanks to @wjzhou!), we need to expose it via the VS Code extension and get it into our next release.

kevcunnane:
- Added support for the saveAsExcel operation that has been added to the SqlToolsService
- Updated tests to cover new scenario
- Open Excel files in default program on the machine since VSCode won't / can't display this

Note: Icon is requested, for now using CSV icon.


src/models/resultsSerializer.ts
```diff
@@ -128,6 +133,22 @@ export default class ResultsSerializer {
         return saveResultsParams;
     }
 
+    private getConfigForExcel(): Contracts.SaveResultsAsExcelRequestParams {
+        // get save results config from vscode config
+        let config = vscode.workspace.getConfiguration(Constants.extensionConfigSectionName);
+        let saveConfig = config[Constants.configSaveAsCsv];
```
kburtram: >configSaveAsCsv [](start = 42, length = 15)

should this be configSaveAsExcel?

src/models/resultsSerializer.ts
```diff
@@ -256,16 +280,26 @@ export default class ResultsSerializer {
...
+    public openSavedFile(filePath: string, format: string): void {
         const self = this;
-        let uri = vscode.Uri.file(filePath);
-        self._vscodeWrapper.openTextDocument(uri).then((doc: vscode.TextDocument) => {
-            // Show open document and set focus
-            self._vscodeWrapper.showTextDocument(doc, 1, false).then(undefined, (error: any) => {
+        if (format === 'excel') {
+            // This will not open in VSCode as it's treated as binary. Use the native file opener instead
+            // Note: must use filePath here, URI does not open correctly
+            opener(filePath, undefined, (error, stdout, stderr) => {
```
kburtram: >opener(filePath [](start = 12, length = 15)

did you have a chance to see what this does on mac, linux, and windows w/o excel installed?  As long as it doesn't blow-up the behavior should be fine.

src/models/resultsSerializer.ts
kevcunnane: Well here's the thing - the option was already set with "csv" in it, but I don't think we should duplicate the "copy with headers" setting. So I'm proposing we leave it alone and use for both.

src/models/resultsSerializer.ts
kevcunnane: I didn't have a chance to test - will test on Linux but likely post checkin. This simply calls the default file open action on the machine, so it should prompt for a sane default. I handle any error callbacks from it so it should report errors if something does go wrong.

PR #787, opened 2017-03-20
Sending the connection parameter as a list of options

llali:
- The new protocol in the service accept a list of options and connection parameters so change the extension to do that. Have to change the service version after I have a new service build


src/models/connectionCredentials.ts
```diff
@@ -43,36 +43,36 @@ export class ConnectionCredentials implements IConnectionCredentials {
      */
     public static createConnectionDetails(credentials: IConnectionCredentials): ConnectionDetails {
         let details: ConnectionDetails = new ConnectionDetails();
-        details.serverName = credentials.server;
-        if (credentials.port && details.serverName.indexOf(',') === -1) {
+        details['serverName'] = credentials.server;
```
kevcunnane: This should be `details.options['serverName']`. You changed the class to add a sub-property, but since you can assign fields using `['xyz']` Typescript isn't complaining.

kevcunnane:
:shipit:

PR #810, opened 2017-04-08
Utilize Error Objects

benrr101:
This change ended up being a bit larger than expected.
* The vscode-languageclient node module has been updgraded to v3.2.0 which includes
  * Updated language server protocol support (ResponseError object returned on error)
  * Enforced wait for ready before working with LanguageClient (required rewrite of serviceClient initialization logic, rewrite of unit tests for serviceClient)
  * Support for dynamic registration (which requires obnoxious extra type parameters for all send requests and notification handlers)
* Updates to code paths to accept a ResponseError object in promise rejection. I can't verify 100% that this is working in all cases due to difficulty in generating error conditions during manual testing and few unit tests that handle error conditions. However, it should not degrade the experience. Worst case, an error message or log may appear with `[object Object]` or `undefined` indicating we how we can track down and quickly fix the issue.


src/languageservice/httpClient.ts
```diff
@@ -134,7 +134,7 @@ export default class HttpClient implements IHttpClient {
             });
 
             response.on('error', err => {
-                reject(new PackageError(`Response error: ${err.code || 'NONE'}`, pkg, err));
+                reject(new PackageError(`Response error: ${err.stack}`, pkg, err));
```
kburtram: >stack [](start = 63, length = 5)

we need to verify that the stack isn't logged via telemetry anywhere since this can violate PII requirements.

src/models/SqlOutputContentProvider.ts
```diff
@@ -345,7 +345,7 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
         // Cancel the query
         queryRunner.cancel().then(success => undefined, error => {
             // On error, show error message
-            self._vscodeWrapper.showErrorMessage(Utils.formatString(LocalizedConstants.msgCancelQueryFailed, error));
+            self._vscodeWrapper.showErrorMessage(Utils.formatString(LocalizedConstants.msgCancelQueryFailed, error.message));
```
kburtram: >message [](start = 115, length = 7)

let's also be careful this doesn't log to telemetry any PII

benrr101:
Recalling this PR b/c it goes a lot further than it needs to in order to get errors working with the updated service.

PR #834, opened 2017-04-22
Fix issue where resize handle is invisible in light theme

### LINKED ISSUE #832: Resize handle is invisible in light theme
Steps to Reproduce:

1. F1 -> "Preferences: Color Theme" -> "Light (Visual Studio)"
2. Open a SQL file and run a query
3. When the results window opens, try to drag the "Messages" region using the top of the grey handle

**Expected:** The message resize handle is not white so it can be seen
image

**Actual:** The message resize handle white - the same color as the background
image

coquagli:
Fixes #832.


coquagli:
Not sure if I trust the coveralls drop. I only changed CSS files.

abist:
@coquagli I've been seeing similar coverage in my PRs, so it's not you.

PR #841, opened 2017-04-26
Fix issue where clicking on JSON/XML cell values opened a blank editor window

### LINKED ISSUE #814: Open XML/JSON cells as linked files is broken in Dev
This was regressed by changes to support displayValue in the cells. We need to correctly parse this when dealing with the openLink event by using this method:

```
    private getCellValueString(dataContext: JSON, colDef: any): string {
        let returnVal = '';
        let value = dataContext[colDef.field];
        if (Services.DBCellValue.isDBCellValue(value)) {
            returnVal = value.displayValue;
        } else if (typeof value === 'string') {
            returnVal = value;
        }
        return returnVal;
    }
```

- MSSQL Extension Version: Dev
- VSCode Version:
- OS Version:

Steps to Reproduce:

1. Create a table and query it with `FOR JSON PATH` syntax, e.g.
```
create table t1 (ID int, c2 nvarchar(50))
INSERT into t1 VALUES (1, 'myvalue)
select * from dbo.t1 FOR JSON PATH, ROOT('t1')
```
2. Click on the JSON columns value in the results
Expected:
- JSON data is shown in a new window
Actual:
- editor window is opened but no data is shown. This is due to an exception thrown when converting this

MattIrv:
Fixes #814 

- Adds code to handle cells' displayValue property when opening links
- Updates link opening code to open the links as new untitled files instead of generating temporary files for them


src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -538,23 +538,36 @@ export class AppComponent implements OnInit, AfterViewChecked {
...
+        let value = self.getCellValueString(dataContext, colDef);
         $(cellRef).children('.xmlLink').click(function(): void {
-            self.dataService.openLink(value, colDef.name, 'json');
+            self.dataService.openLink(value, colDef.name, linkType);
         });
     }
 
+    private getCellValueString(dataContext: JSON, colDef: any): string {
+        let returnVal = '';
+        let value = dataContext[colDef.field];
```
coquagli: This will throw if `dataContext` is undefined. This would occur if a cell was `null`. I'm not sure if it's even possible for cells to be `null` entries, however. Still might be good to check.

src/models/SqlOutputContentProvider.ts
```diff
@@ -475,7 +471,7 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
             }
         }
 
-        vscode.workspace.openTextDocument(uri).then((doc: vscode.TextDocument) => {
+        vscode.workspace.openTextDocument({ language: linkType }) .then((doc: vscode.TextDocument) => {
```
coquagli: It looks like having a bad language type won't cause a crash, but will instead default to 'Plain Text'. This isn't terrible or anything, but since we are already checking above if we have 'xml' or 'json' it might be a good idea to not launch this unless we have 'xml' or 'json'.

coquagli:
I'm getting some strange behavior that the link turns into `[object Object]` when clicked on, ma king it unclickable until tinkered with for a bit:

image

It looks like you didn't introduce this, but it might be good to fix because it's related.

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -538,23 +538,36 @@ export class AppComponent implements OnInit, AfterViewChecked {
      */
     xmlLinkHandler = (cellRef: string, row: number, dataContext: JSON, colDef: any) => {
         const self = this;
-        let value = dataContext[colDef.field];
-        $(cellRef).children('.xmlLink').click(function(): void {
-            self.dataService.openLink(value, colDef.name, 'xml');
-        });
+        self.handleLink(cellRef, row, dataContext, colDef, 'xml');
```
kevcunnane: Minor: you can use `this` instead of `self` here since we are no longer using it inside an embedded function. Same for the jsonLinkHandler method

kevcunnane:
@coquagli I'd suggest creating a separate issue to track that. Do you have a reliable repro given you say it happens even without this PR?

src/models/SqlOutputContentProvider.ts
MattIrv: In the current code I don't think we'll get here without a linkType of xml or json, since we only add the link for those types. But even if we somehow did have a different linkType, we would want to open the link with the default behavior, so I'll keep it as-is.

coquagli:
@kevcunnane Looks like cloning a new repo fixed my `[object Object]` issue.

PR #852, opened 2017-04-29
Changes to vscode-mssql extension allowing empty passwords for SqlLogin

huypha:
Allowing empty passwords for SqlLogin. Adding in new emptyPasswordInput property for profiles. This is paired with a PR to sqltoolsservice to allow empty passwords

### DIFF
--- a/src/models/connectionCredentials.ts
+++ b/src/models/connectionCredentials.ts
@@ -115,9 +115,10 @@ export class ConnectionCredentials implements IConnectionCredentials {
                 if (isProfile) {
                     let profile: IConnectionProfile = <IConnectionProfile>credentials;
 
-                    // If this is a profile, and the user has set save password to true and stored the password in the config file,
+                    // If this is a profile, and the user has set save password to true and either
+                    // stored the password in the config file or purposefully set an empty password
                     // then transfer the password to the credential store
-                    if (profile.savePassword && !wasPasswordEmptyInConfigFile) {
+                    if (profile.savePassword && (!wasPasswordEmptyInConfigFile || profile.emptyPasswordInput)) {
                         // Remove profile, then save profile without plain text password
                         connectionStore.removeProfile(profile).then(() => {
                             connectionStore.saveProfile(profile);
@@ -180,7 +181,7 @@ export class ConnectionCredentials implements IConnectionCredentials {
                     credentials.authenticationType = value;
                 }
             },
-            // Username must be pressent
+            // Username must be present
             {
                 type: QuestionTypes.input,
                 name: LocalizedConstants.usernamePrompt,
@@ -204,7 +205,10 @@ export class ConnectionCredentials implements IConnectionCredentials {
                     }
                     return undefined;
                 },
-                onAnswered: (value) => credentials.password = value
+                onAnswered: (value) => {
+                    credentials.password = value;
+                    (<IConnectionProfile>credentials).emptyPasswordInput = utils.isEmpty(credentials.password);
+                }
             }
         ];
         return questions;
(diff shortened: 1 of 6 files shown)


src/models/connectionCredentials.ts
```diff
@@ -180,7 +181,7 @@ export class ConnectionCredentials implements IConnectionCredentials {
                     credentials.authenticationType = value;
                 }
             },
-            // Username must be pressent
+            // Username must be present
```
benrr101: Good catch 👍

src/models/connectionStore.ts
```diff
@@ -131,8 +131,8 @@ export class ConnectionStore {
                 // Don't try to lookup a saved password if savePassword is set to false for the credential
                 resolve(credentialsItem);
             } else if (ConnectionCredentials.isPasswordBasedCredential(credentialsItem.connectionCreds)
-                    && Utils.isEmpty(credentialsItem.connectionCreds.password)) {
-
+                    && Utils.isEmpty(credentialsItem.connectionCreds.password)
+                    && credentialsItem.connectionCreds['emptyPasswordInput'] === false) {
```
MattIrv: This causes it to only save if emptyPasswordInput is explicitly false. Since old configs won't have a value for emptyPasswordInput, I think it would be safer to instead check if emptyPasswordInput is not explicitly true, that way undefined is treated as false.

Since this is typescript and not javascript I'd also suggest looking into a more type safe way to do this, or at least explaining in a comment that emptyPasswordInput will be present only when connectionCreds is actually a ConnectionProfile object.

src/models/connectionCredentials.ts
```diff
@@ -214,8 +218,13 @@ export class ConnectionCredentials implements IConnectionCredentials {
         return utils.isEmpty(credentials.user) && ConnectionCredentials.isPasswordBasedCredential(credentials);
     }
 
+    // Prompt for password if this is a password based credential and the password for the profile was empty
+    // and not explicitly set as empty. If it was explicitly set as empty, only prompt if pw not saved
     private static shouldPromptForPassword(credentials: IConnectionCredentials): boolean {
-        return utils.isEmpty(credentials.password) && ConnectionCredentials.isPasswordBasedCredential(credentials);
+        return utils.isEmpty(credentials.password)
+            && ConnectionCredentials.isPasswordBasedCredential(credentials)
+            && (!((<IConnectionProfile>credentials).emptyPasswordInput) ||
```
benrr101: I think to clean this up a wee bit, we should move the || to the next line

benrr101: And potentially store off some of the lengthier conditions as variables

src/models/connectionCredentials.ts
```diff
@@ -214,8 +218,13 @@ export class ConnectionCredentials implements IConnectionCredentials {
...
+             ((<IConnectionProfile>credentials).emptyPasswordInput && !(<IConnectionProfile>credentials).savePassword));
```
MattIrv: `<IConnectionProfile>credentials).emptyPasswordInput` is redundant here since the first part of the or clause checks it

src/models/connectionCredentials.ts
```diff
@@ -204,7 +205,10 @@ export class ConnectionCredentials implements IConnectionCredentials {
                     }
                     return undefined;
                 },
-                onAnswered: (value) => credentials.password = value
+                onAnswered: (value) => {
+                    credentials.password = value;
+                    (<IConnectionProfile>credentials).emptyPasswordInput = utils.isEmpty(credentials.password);
```
MattIrv: Is this type assertion guaranteed to be true? If so, I think we should just update the method to take a IConnectionProfile object. If not, this code should probably check if credentials is actually an IConnectionProfile object.

MattIrv: Don't see this addressed

src/models/connectionStore.ts
```diff
@@ -126,13 +126,16 @@ export class ConnectionStore {
     public addSavedPassword(credentialsItem: IConnectionCredentialsQuickPickItem): Promise<IConnectionCredentialsQuickPickItem> {
         let self = this;
         return new Promise<IConnectionCredentialsQuickPickItem>( (resolve, reject) => {
+            let isEmptyPasswordInput: boolean = typeof(credentialsItem.connectionCreds['emptyPasswordInput'] !== 'undefined')
```
MattIrv: I think your parentheses are in the wrong place here. Also you can simplify checking for undefined by just using `credentialsItem.connectionCreds['emptyPasswordInput']` as if it were a boolean in the if statement

test/connectionCredentials.test.ts
```diff
@@ -78,6 +78,51 @@ suite('ConnectionCredentials Tests', () => {
             connectionStore.object);
     }
 
+    function ensureRequestAndSavePassword(emptyPassword: boolean): (done: MochaDone) => void {
+
```
MattIrv: (nit) extra line break

test/connectionCredentials.test.ts
```diff
@@ -78,6 +78,51 @@ suite('ConnectionCredentials Tests', () => {
...
+        return (done: MochaDone) => {
+            // Setup Profile Information to have savePassword off and blank
```
MattIrv: This comment says savePassword is off, but it's true below. Which is right?

huypha: It looks like the original intention of the test was to have save password as on. The comment probably wasn't updated after a copy-paste. Will change

PR #861, opened 2017-05-03
Support connecting with connection strings

MattIrv:
Adds connection string support, including
- Allows connection credentials to have connection strings
- Ignores other parameters when a connection string is given in credentials
- Allows user to create profiles with connection strings by giving one in place of the server prompt
- Tests connection string behavior

### DIFF
--- a/src/connectionconfig/connectionconfig.ts
+++ b/src/connectionconfig/connectionconfig.ts
@@ -69,8 +69,8 @@ export class ConnectionConfig implements IConnectionConfig {
         let profiles: IConnectionProfile[] = [];
         let compareProfileFunc = (a, b) => {
             // Sort by profile name if available, otherwise fall back to server name or connection string
-            let nameA = a.profileName ? a.profileName : a.server ? a.server : a.connectionString;
-            let nameB = b.profileName ? b.profileName : b.server ? b.server : b.connectionString;
+            let nameA = a.profileName ? a.profileName : (a.server ? a.server : a.connectionString);
+            let nameB = b.profileName ? b.profileName : (b.server ? b.server : b.connectionString);
             return nameA.localeCompare(nameB);
         };
 
--- a/src/connectionconfig/connectionconfig.ts
+++ b/src/connectionconfig/connectionconfig.ts
@@ -68,30 +68,15 @@ export class ConnectionConfig implements IConnectionConfig {
     public getConnections(getWorkspaceConnections: boolean): IConnectionProfile[] {
         let profiles: IConnectionProfile[] = [];
         let compareProfileFunc = (a, b) => {
-            // Sort by profile name if available, otherwise fall back to server name
-            let nameA = a.profileName ? a.profileName : a.server;
-            let nameB = b.profileName ? b.profileName : b.server;
+            // Sort by profile name if available, otherwise fall back to server name or connection string
+            let nameA = a.profileName ? a.profileName : a.server ? a.server : a.connectionString;
+            let nameB = b.profileName ? b.profileName : b.server ? b.server : b.connectionString;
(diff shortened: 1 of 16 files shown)


src/models/utils.ts
```diff
@@ -234,15 +237,16 @@ export function isSameProfile(currentProfile: interfaces.IConnectionProfile, exp
...
+ * match on all key properties (connectionString or server, db, auth type, user) being identical.
  * Other properties are ignored for this purpose
  *
  * @param {IConnectionCredentials} conn the connection to check
  * @param {IConnectionCredentials} expectedConn the connection to try to match
  * @returns boolean that is true if the connections match
  */
 export function isSameConnection(conn: interfaces.IConnectionCredentials, expectedConn: interfaces.IConnectionCredentials): boolean {
-    return expectedConn.server === conn.server
+    return (conn.connectionString || expectedConn.connectionString) ? conn.connectionString === expectedConn.connectionString :
```
abist: I think this should be ``(conn.connectionString && expectedConn.connectionString)`` in case ``conn`` or ``expectedConn`` is null.

MattIrv: The point of using or is so that we return false if one has a connection string but not the other.

src/connectionconfig/connectionconfig.ts
```diff
@@ -66,17 +66,18 @@ export class ConnectionConfig implements IConnectionConfig {
...
     public getConnections(getWorkspaceConnections: boolean): IConnectionProfile[] {
-        let profiles = [];
+        let profiles: IConnectionProfile[] = [];
         let compareProfileFunc = (a, b) => {
-            // Sort by profile name if available, otherwise fall back to server name
-            let nameA = a.profileName ? a.profileName : a.server;
-            let nameB = b.profileName ? b.profileName : b.server;
+            // Sort by profile name if available, otherwise fall back to server name or connection string
+            let nameA = a.profileName ? a.profileName : a.server ? a.server : a.connectionString;
+            let nameB = b.profileName ? b.profileName : b.server ? b.server : b.connectionString;
```
huypha: Potentially either add in braces or have the result of (a.server ? a.server : a.connectionString) stored in a separate variable to make this more easily readable

test/utils.test.ts
```diff
@@ -32,3 +34,51 @@ suite('Utility Tests - parseNumAsTimeString', () => {
...
+    let connection4 = Object.assign(new ConnectionCredentials(), {
+        connectionString: connectionString
+    });
+
+    test('should return true for matching non-connectionstring connections', () => {
+        expect(Utils.isSameConnection(connection1, connection2)).to.equal(true);
+    });
+
+    test('should return true for non-matching non-connectionstring connections', () => {
+        connection2.server = 'some-other-server';
```
huypha: 'should return true' -> 'should return false'

src/models/connectionInfo.ts
```diff
@@ -167,10 +179,11 @@ export function getUserNameOrDomainLogin(creds: Interfaces.IConnectionCredential
  */
 export function getTooltip(connCreds: Interfaces.IConnectionCredentials, serverInfo?: ConnectionContracts.ServerInfo): string {
     let tooltip: string =
-           'Server name: ' + connCreds.server + '\r\n' +
+           connCreds.connectionString ? 'Connection string: ' + connCreds.connectionString + '\r\n' :
```
MattIrv: Looks fine on my Mac. Maybe Windows requires it and it doesn't make a difference on other platforms?

abist: Yeah was just wondering. I'm aware that this code existed earlier, so don't worry about it for now.

Issue #417, opened 2016-11-25
[suggestion] Please support Arch Linux


tinchou:
Hi,

I haven't had time to investigate why the code for this extension has [so many special cases for specific distributions](https://github.com/Microsoft/vscode-mssql/blob/dev/src/models/platform.ts#L245), but I will assume you have a reason for that and I'll thank you for giving us a great experience.

This issue is to request support for Arch Linux. I've seen you have support for distros not as popular (based on [this list](http://distrowatch.com/dwres.php?resource=major)) such as Elementary OS, so I hope you'll take it into account.

Thanks again for your Linux efforts!

kevcunnane:
Hi @tinchou, thanks for raising this issue. Our backing service is built on .Net Core. This lets us reuse a lot of existing code and add new features more rapidly.  The downside is that there are some distributions that .Net Core does not support. Arch Linux is one of these - the issue https://github.com/dotnet/coreclr/issues/4409 looks like the most relevant assessment of state, and this is just to compile on the platform. We'd need support for building all-in-one installers to match the other distributions. 

At present, this request is blocked by lack of support in .Net Core. We will keep it open as a tracking item, and if this support level changes we can take a look at whether it's possible to add support to our extension.

benrr101:
However, if you're feeling particularly gutsy, it looks like there's some unofficial (at least from dotnet's point of view) means of installing the dotnet-cli via the AUR https://aur.archlinux.org/packages/dotnet-cli/

As an unpopular distro user myself (Scientific Linux), I'd like to see as many Linux distros supported as possible, even if they're not "officially supported". I'll try and keep an eye out to see if we can reduce the number of special cases for distros and enable overrides for those to allow unpopular or outdated distros to run our extension.

kevcunnane:
It looks like Omnisharp has figured out a workaround where Ubuntu 16.04 package works - see [issue](https://github.com/OmniSharp/omnisharp-vscode/issues/564) and [pull request](https://github.com/OmniSharp/omnisharp-vscode/pull/1017).

We should look at enabling the same for our code, run testing & verification on an Arch linux platform.

@tinchou if you are feeling adventurous, you could try doing this by changing the switch statement in platform.js to be as follows and seeing if the extension works. We'll get to this after we complete the current release as we're busy focusing on that right now.

```
case 'arch':
   // NOTE: currently Arch Linux seems to be compatible enough with Ubuntu 16 that this works,
[...]

PR #882, opened 2017-05-25
Adding support for Arch Linux

benrr101:
The change has been taken directly from #417 

I'll try to run a quick test of this, but if anyone wants to confirm that the change works, I'll welcome that as well.

Additionally, we will need to make sure that we have a quick note added to the wiki to describe the short requirements for running on Arch.


kburtram:
@benrr101 I've changed the default branch to master..so please keep in mind when merging we'll need to port to master.

bdruth:
👍 Can we get this merged?

PR #886, opened 2017-05-31
Update to angular4

anthonydresser:
Updates to angular4 and the newest version of the slickgrid angular wrapper. Had to circumvent the tests because angular4 added generics to their injector, and Typescript wouldn't allow me to cast from MockDataService to DataService


kburtram:
@anthonydresser are we waiting on anything with this PR?  Otherwise, could we either merge or close?

anthonydresser:
I need to revisit this, since we find a lot of change detection problems when doing this elsewhere. I'll close for now and come back to it and go through and update all the change detection problems.

PR #888, opened 2017-06-01
Added the sqlListColumns snippet

eashi:
This snippet lists all the columns and their types of all tables whose name contain the supplied string.

### DIFF
--- a/snippets/mssql.json
+++ b/snippets/mssql.json
@@ -240,5 +240,24 @@
 			"GO"
 		],
 		"description": "List databases"
+	},
+	
+	"List columns": {
+		"prefix": "sqlListColumns",
+		"body": [
+			"-- List columns and their types of all tables whose name like '${TableName}'",
+			"select ",
+    			"TableName = tbl.table_schema + '.' + tbl.table_name, ",
+    			"ColumnName = col.column_name, ",
+    			"ColumnDataType = col.data_type",
+			"FROM information_schema.tables tbl",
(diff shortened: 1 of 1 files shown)


eashi:
Ops, I pushed the button too quickly, I will check the tests.

kburtram:
@eashi the test failure doesn't appear related to this change.  It could be related to recent branch configuration changes we made.

This change looks good to me.  I'd like @erickangMSFT to take a quick look at the snippet before merging.

kburtram:
@eashi could you please complete the contributor license agreement at https://cla.microsoft.com.?  This is a requirement for us to accept contributions from outside Microsoft.  I think the process is very quick.  Thanks!

erickangMSFT:
@eashi Can you please keep the consistency of lower or upper casing for keywords? Other snippets use upper case for keyword so the output would be:

-- List columns and their types of all tables whose name like 'TableName'
SELECT
TableName = tbl.table_schema + '.' + tbl.table_name, 
ColumnName = col.column_name, 
ColumnDataType = col.data_type
FROM information_schema.tables tbl
INNER JOIN information_schema.columns col 
ON col.table_name = tbl.table_name
AND col.table_schema = tbl.table_schema
WHERE tbl.table_type = 'base table' AND tbl.table_name LIKE '%TableName%'
GO

eashi:
Thanks @kburtram and @erickangMSFT :)

PR #914, opened 2017-07-10
Update SqlToolsService to netcore2.0 and fix test failures

kevcunnane:
- Upgraded to .net core 2.0 version of the SqlToolsService
- Fixed a number of issues breaking lab builds. Notably, set the typescript version to 2.3.4 since the latest version 2.4.1 has intentional breaking changes that will take time to resolve & fix propery
- Fixed test failures caused by upgrading TypeMoq, which was again necessary due to upstream dependency changes


kburtram:
:shipit:

kevcunnane:
I'm going to merge this in as it fixes AppVeyor and Travis builds are taking forever (and previously broken). Will work to see if it gets fixed in the build and do separate PR if we still have issues on that build

PR #927, opened 2017-07-12
Add disconnect option when changing databases

### LINKED ISSUE #904: Suggestion: Add Disconnect to status bar server connection shortcut
- MSSQL Extension Version: 1.0.0
- VSCode Version: 1.13.1
- OS Version: any

If you are connected to a sql server and click on the server info text in the status bar, that opens the "choose database" dialog.

Is it possible to change that behavior so you have the option to either choose a database or just disconnect?

MattIrv:
Closes #904 

This change adds an option to disconnect when clicking on the connection status in the bottom-right, as an alternative to changing the connected database. If you choose to disconnect, it will prompt for a confirmation before disconnecting.

### DIFF
--- a/src/views/connectionUI.ts
+++ b/src/views/connectionUI.ts
@@ -210,8 +210,7 @@ export class ConnectionUI {
             self.vscodeWrapper.showQuickPick<vscode.QuickPickItem>(pickListItems, pickListOptions).then( selection => {
                 if (selection === disconnectItem) {
                     self.handleDisconnectChoice().then(() => resolve(undefined), err => reject(err));
-                }
-                if (typeof selection !== 'undefined') {
+                } else if (typeof selection !== 'undefined') {
                     resolve((selection as Interfaces.IConnectionCredentialsQuickPickItem).connectionCreds);
                 } else {
                     resolve(undefined);
--- a/src/views/connectionUI.ts
+++ b/src/views/connectionUI.ts
@@ -178,7 +178,7 @@ export class ConnectionUI {
                                         databaseNames: Array<string>): Promise<Interfaces.IConnectionCredentials> {
         const self = this;
         return new Promise<Interfaces.IConnectionCredentials>((resolve, reject) => {
-            const pickListItems = databaseNames.map(name => {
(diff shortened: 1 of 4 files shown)


src/views/connectionUI.ts
```diff
@@ -195,21 +195,49 @@ export class ConnectionUI {
...
                 placeHolder: LocalizedConstants.msgChooseDatabasePlaceholder
             };
 
             // show database picklist, and modify the current connection to switch the active database
-            self.vscodeWrapper.showQuickPick<Interfaces.IConnectionCredentialsQuickPickItem>(pickListItems, pickListOptions).then( selection => {
+            self.vscodeWrapper.showQuickPick<vscode.QuickPickItem>(pickListItems, pickListOptions).then( selection => {
+                if (selection === disconnectItem) {
+                    self.handleDisconnectChoice().then(() => resolve(undefined), err => reject(err));
+                }
                 if (typeof selection !== 'undefined') {
```
kevcunnane: Shouldn't this be an else if? While resolve will eventually return from the promise it doesn't actually block the subsequent code from being run I think?

src/views/connectionUI.ts
```diff
@@ -195,21 +195,49 @@ export class ConnectionUI {
...
+
             const pickListOptions: vscode.QuickPickOptions = {
                 placeHolder: LocalizedConstants.msgChooseDatabasePlaceholder
             };
 
             // show database picklist, and modify the current connection to switch the active database
-            self.vscodeWrapper.showQuickPick<Interfaces.IConnectionCredentialsQuickPickItem>(pickListItems, pickListOptions).then( selection => {
+            self.vscodeWrapper.showQuickPick<vscode.QuickPickItem>(pickListItems, pickListOptions).then( selection => {
+                if (selection === disconnectItem) {
+                    self.handleDisconnectChoice().then(() => resolve(undefined), err => reject(err));
```
kevcunnane: At this point if you reject what happens? I think it's possible you should handle it by showing the error to the user.

MattIrv: Turns out that rejecting here does show the error to the user

MattIrv:
Received in person sign-off from Kevin

PR #934, opened 2017-07-13
Execute SQL statement at cursor location

kburtram:


src/controllers/mainController.ts
```diff
@@ -307,6 +316,15 @@ export default class MainController implements vscode.Disposable {
 
                 Telemetry.sendTelemetryEvent('RunQuery');
 
+                if (execStatment) {
```
benrr101: I'm not certain on how this is supposed to work. Isn't the contract to send a line and column number? If so, how does the start/end position play into this?
I think it might be a bit cleaner to factor out the prompting/retry logic and have two onRunQuery methods, one for a single statement, one for a selection/entire doc.

kburtram: @benrr101 I've refactored that code that to make it more clear what's going on.  When there is no selection range the editor selection property gives the cursor position, otherwise we use the start position in the selection as the cursor position.

PR #938, opened 2017-07-14
Fix retry and empty file bugs for exec current statement

kburtram:
This fixes two bugs that got missed on the previous "Exec current stmt" check-in.

1. Executing an empty file gives an error
2. Executing disconnected or non-SQL fails during retry since 'this' context gets lost.


kburtram:
Fixing regressions from prior checkin.  I'm merging to get this into the release candidate.  Please send me any feedback directly and I'll do a follow-up PR.

PR #940, opened 2017-07-14
Support for SLES 12.2

kburtram:
.Net 2 now supports the latest SLES distros.  We shouldn't have the downloader block this platform anymore.


src/configurations/dev.config.json
```diff
@@ -10,6 +10,7 @@
             "Debian_8": "debian-x64-netcoreapp2.0.tar.gz",
             "Fedora_23": "fedora-x64-netcoreapp2.0.tar.gz",
             "OpenSUSE_13_2": "opensuse-x64-netcoreapp2.0.tar.gz",
+            "SLES_12_2": "opensuse-x64-netcoreapp2.0.tar.gz",
```
kburtram: we just run the opensuse binaries on sles and this works fine...actually I think all the Linux archives contain the exact same bits at this point so they all could probably refer to the same archive.

PR #941, opened 2017-07-14
Fix #933 support Mac versions lower than Sierra

### LINKED ISSUE #933: MSSQL 1.1 SQL Tools Service component could not start. After installing v1.1
- MSSQL Extension Version:1.1
- VSCode Version: 1.14.0
- OS Version: OSX 10.11.6

Greetings all!

I recently followed the steps outlined here: https://github.com/Microsoft/vscode-mssql/blob/master/KERBEROS_HELP.md and am currently getting an error after installing mssql-1.1.0.vsix.

The error I am seeing in the SQLToolsService output channel is the following:

```
dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
  Referenced from: /Users/aUserNAmeThatItotallyJustChanged/.vscode/extensions/ms-mssql.mssql-1.1.0/sqltoolsservice/1.0.0-alpha.45/OSX/libcoreclr.dylib (which was built for Mac OS X 10.12)
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _clock_gettime
  Referenced from: /Users/aUserNAmeThatItotallyJustChanged/.vscode/extensions/ms-mssql.mssql-1.1.0/sqltoolsservice/1.0.0-alpha.45/OSX/libcoreclr.dylib (which was built for Mac OS X 10.12)
  Expected in: /usr/lib/libSystem.B.dylib

[Error - 11:50:55 AM] Connection to server got closed. Server will not be restarted.
```

Steps to Reproduce:

1. Install mssql-1.1.0.vsix using Extensions: Install from VSIX

2. Choose mssql-1.1.0.vsix downloaded from https://github.com/Microsoft/vscode-mssql/releases/tag/v1.1.0-alpha.1

3.  Reload

4. SqlToolService Initialization

5. Receive error here

(obligatory optional steps)
6. ???

7. Profit

It looks like a similar experience was expressed in https://github.com/Microsoft/vscode-mssql/issues/356 by user xs9627 though he/she didn't report the above mentioned error.

Edit: I'm a little sad that mine was not the 1000th issue opened. :(

Edit 2: Thanks for the swift action kevcunnane!

kevcunnane:
- Added version check to detect macOS versions lower than Sierra (10.12.0). If this is discovered, use the v1.0 version of the tools service and disable new features
- Updated downloader to fall back to v1 version and added tests to verify this
- Added handling in new commands (use of Integrated Auth and execute current statement) to return an error to the user.


src/configurations/config.ts
```diff
@@ -14,6 +14,8 @@ import {IConfig} from '../languageservice/interfaces';
 */
 export default class Config implements IConfig {
      private static _configJsonContent = undefined;
+     private sqlToolsServiceConfigKey: string;
```
kburtram: >sqlToolsServiceConfigKey [](start = 13, length = 24)

should private fields start with an '_' prefix?

src/controllers/mainController.ts
```diff
@@ -410,6 +416,14 @@ export default class MainController implements vscode.Disposable {
     }
 
     /**
+     * Verifies the tools service version is high enough to support certain commands
+     */
+    private CanRunV2Command(): boolean {
```
kburtram: >CanRunV2Command [](start = 12, length = 15)

this method should be camelCased...we should probably change CanRunCommand at the same time

src/languageservice/serviceclient.ts
```diff
@@ -24,9 +24,12 @@ import {PlatformInformation} from '../models/platform';
 import {ServerInitializationResult, ServerStatusView} from './serverStatus';
 import StatusView from '../views/statusView';
 import * as LanguageServiceContracts from '../models/contracts/languageService';
+import { IConfig } from '../languageservice/interfaces';
 let vscode = require('vscode');
 
 let opener = require('opener');
+
+
```
kburtram: nit: random new lines?

src/configurations/config.ts
kevcunnane: They should!

src/controllers/mainController.ts
kevcunnane: Agreed

src/languageservice/serviceclient.ts
kevcunnane: Will remove

src/models/platform.ts
```diff
@@ -87,6 +89,16 @@ export class PlatformInformation {
         return getRuntimeDisplayName(this.runtimeId);
     }
 
+    public isMacVersionLessThan(version: string): boolean {
+        if (this.isMacOS) {
+            let versionInfo = plist.parse(fs.readFileSync('/System/Library/CoreServices/SystemVersion.plist', 'utf-8'));
```
kburtram: >plist.parse(fs.readFileSync('/System/Library/CoreServices/SystemVersion.plist', 'utf-8')) [](start = 30, length = 89)

not sure, but this looks like the type of code that could throw...it way be good to try/catch this method to avoid any surprises

src/controllers/mainController.ts
kburtram: I just noticed @MattIrv  renamed canRunCommand in his active PR, so you don't need to rename that one

PR #942, opened 2017-07-15
Show an error message for certain commands if no editor has focus

### LINKED ISSUE #937: Trying to use the 'Refresh Intellisense Cache' command gives an error if results pane has focus
- MSSQL Extension Version: Master 4f0ba2e
- VSCode Version: 1.13.1
- OS Version: OS X 10.12

Steps to Reproduce:
1. Connect to a server and execute a query, then focus in the results pane
2. Try to execute the 'Refresh Intellisense Cache' command, and see that it gives an error

Expected behavior:
The command runs successfully as it would have if the query window had focus

Other commands, like Execute Query, have similar issues.

MattIrv:
For commands that require a SQL document, we now show an error if the user tries to execute those commands without a text document having focus.

Prior to this change, we did not display such an error message but none of those commands worked without a text document having focus, so they would fail with error messages that did not help the user figure out what was wrong.

Closes #937


localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -284,6 +284,9 @@
         <trans-unit id="disconnectConfirmationMsg">
             <source xml:lang="en">Are you sure you want to disconnect?</source>
         </trans-unit>
+        <trans-unit id="noActiveEditorMsg">
+            <source xml:lang="en">A SQL query editor window must have focus before executing this command</source>
```
kburtram: >query editor window [](start = 40, length = 19)

you probably could shorten "SQL query editor window" to "SQL editor" without losing much in this error message

src/controllers/mainController.ts
```diff
@@ -408,7 +411,7 @@ export default class MainController implements vscode.Disposable {
     /**
      * Verifies the extension is initilized and if not shows an error message
      */
-    private CanRunCommand(): boolean {
+    private canRunCommand(): boolean {
```
kburtram: >canRunCommand [](start = 12, length = 13)

nice, I just asked Kevin to rename this method in another code review.  I'll leave a comment for him to not do that to avoid conflicts

kburtram:
@MattIrv please take a look at the CI builds before merging.  I see validateTextDocumentHasFocus caused one of the failures

src/controllers/mainController.ts
kevcunnane: Mine is checked in at this point so you'll need to merge and fix the conflicts unfortunately.

PR #943, opened 2017-07-15
Fix  #939 Allow messages to display run time for each batch when there are multiple batches

### LINKED ISSUE #939: Allow messages to display run time for each batch when there are multiple batches
- MSSQL Extension Version: 1.1.0
- VSCode Version: 1.14.0
- OS Version: Windows 10

Steps to Reproduce:

1. enter two separate queries, separated by GO in teh editor window
2. execute query for the full windows
3. Messages window will display start time, starting line # and # of affected rows for each message. Then it will display the total running time. 

Is it possible to display the running time for each batch in addition to the total?
THanks

kevcunnane:
- Added this feature behind a settings flag
- Some time parsing fixes to support this


src/controllers/queryRunner.ts
```diff
@@ -213,7 +213,12 @@ export default class QueryRunner {
 
         // Store the batch again to get the rest of the data
         this._batchSets[batch.id] = batch;
-        this._totalElapsedMilliseconds += <number>(Utils.parseTimeString(batch.executionElapsed) || 0);
+        let executionTime =  <number>(Utils.parseTimeString(batch.executionElapsed) || 0);
```
MattIrv: nit: extra space here before `<number>`

src/controllers/queryRunner.ts
```diff
@@ -364,6 +369,22 @@ export default class QueryRunner {
         return outputString;
     }
 
+    private sendBatchTimeMessage(batchId: number, executionTime: string): void {
+        // get config copyRemoveNewLine option from vscode config
+        let config = this._vscodeWrapper.getConfiguration(Constants.extensionConfigSectionName);
+        let showBatchTime: boolean = config[Constants.configShowBatchTime];
+        if (!!showBatchTime) {
```
MattIrv: Is !! necessary here?

kevcunnane: Probably not. We used it elsewhere in these cases but given the config should give back a sensible value for the default I can skip it.

PR #944, opened 2017-07-16
Update to latest tools service build

kburtram:


kburtram:
This is only buimping the tools service version in a config file so I'll merge now so that the latest bits are picked up in Monday's build for "sign-off" testing.

PR #951, opened 2017-07-17
Update to use 1.1.0 GA SqlToolsService release from download center

kevcunnane:


kevcunnane:
Auto merging, manual testing all passed too

PR #952, opened 2017-07-17
Fix #948 save password is not working

### LINKED ISSUE #948: Save password is not working in RHEL
Created a new profile with saved password but when try to connect, it ask for password every time

kevcunnane:


kburtram:
@kevcunnane thanks for fixing this.  I've been trying to get my machine environment setup all day and haven't had a chance to debug yet 🙁

test/connectionStore.test.ts
```diff
@@ -576,5 +576,69 @@ suite('ConnectionStore tests', () => {
...
 
+    test('addSavedPassword not lookup password if password already set', (done) => {
+        // Given a profile with emptyPasswordInput set
+        let passwordAlreadySetProfile = Object.assign(new ConnectionProfile(), {
+            profileName: 'defaultNamedProfile',
+            server: 'namedServer',
+            database: 'bcd',
+            authenticationType: utils.authTypeToString(interfaces.AuthenticationTypes.SqlLogin),
+            user: 'cde',
+            password: 'asdf!@#$'
```
kburtram: FYI, everywhere we add the text 'password" we'll get a TSA credential scan bug.  if we do need to include a password I would recommend using the sting "password" which is more obviously not a real password

kburtram: hmm, I wonder if all the strings '...password..' in this file are going to open bugs

PR #967, opened 2017-08-01
Fix #891 Create Connection Profile Aborts Without Warning

### LINKED ISSUE #891: Create Connection Profile Aborts Without Warning
- MSSQL Extension Version:
- VSCode Version:
- OS Version:

Steps to Reproduce:

1. Start creating a new connection profile
2. enter some data
3. change focus to another window (for example to copy a long, complex password)
4. All your work is lost without warning!

kevcunnane:
- VSCode added an option that supports ignoring focusOut events so that prompts stay visible even on tabbing to other apps! Fixed and verified manually. Tests pass

### DIFF
--- a/src/models/connectionProfile.ts
+++ b/src/models/connectionProfile.ts
@@ -52,7 +52,7 @@ export class ConnectionProfile extends ConnectionCredentials implements IConnect
                 }
         });
 
-        return prompter.prompt(questions).then(answers => {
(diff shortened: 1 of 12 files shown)


kevcunnane:
Default should be false - this is a general prompter framework and in other cases you want things to disappear on click away (example: switch database). I think I'll leave as-is here.

PR #980, opened 2017-09-07
Fix travis build for linux

llali:
According to https://github.com/Microsoft/vscode/issues/33998
the fix was adding libsecret-1-dev as dependency

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,11 @@ os:
   - linux
   - osx
 
+addons:
+  apt:
+    packages:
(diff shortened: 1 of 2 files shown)


kburtram:
@llali could you check why the CI builds are failing on the PR before merging?

llali:
@karlb@microsoft.com 
I'm not going to merge this. Well at least until I figure out if I have to change the version of vs code. The  travis build for master if failing and I was trying changing the version of vs code in this PR to see if I can fix it

PR #988, opened 2017-09-19
Add language flavor to fix #978

### LINKED ISSUE #978: Add Language Flavor notification support to allow disabling MSSQL processing on a per-file basis
We implemented Language Flavor notifications in the backend so that we can filter out SQL files we don't want to be treated as MSSQL. This is useful for dynamically skipping files if you have a mix of MSSQL and non-MSSQL files with a .sql extension. We should implement the front-end to:
- Show a button with `MSSQL` or `Other SQL` as the option
- On click, allow user to apply the filter and send the notification

This will mean users can easily opt out of the processing for each file and save some pain. We have a number of issues raised related to this.

kevcunnane:
Fixes #978: adds a new status indicator for Language Flavor so that users can choose `MSSQL` or `None` for provider type. If `None` then MSSQL intellisense and syntax checking are skipped.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -3,6 +3,15 @@
     "version": "0.1.0",
     "configurations": [
         {
+            "type": "node",
+            "request": "launch",
+            "name": "Gulp task",
+            "program": "${workspaceRoot}/node_modules/gulp/bin/gulp.js",
+            "args": [
+                "ext:localization:xliff-to-package.nls"
+            ]
+        },
+        {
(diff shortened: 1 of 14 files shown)


Yensan:
I am wandering this plugin will support for MySQL etc. , But it is like MS just want this plugin be an T-SQL for Microsoft SQL Server, Azure SQL Database and SQL Data Warehouses. 
If VScode can ship a plugin that can connect all SQL database...

PR #991, opened 2017-09-20
using vscode API to store and load connection profiles

llali:
using workspaceConfiguration  to store connections to user/workspace settings and also loading from the settings.
Fixed the tests based on latest changes. 
Also fixed a bug which connections using Windows auth were added twice to recent history.


src/connectionconfig/connectionconfig.ts
```diff
@@ -77,16 +63,14 @@ export class ConnectionConfig implements IConnectionConfig {
         };
 
         // Read from user settings
-        let parsedSettingsFile = this.readAndParseSettingsFile(ConnectionConfig.configFilePath);
-        let userProfiles = this.getProfilesFromParsedSettingsFile(parsedSettingsFile);
+        let userProfiles = this.getProfilesFromSettingsFile();
```
kevcunnane: Minor: remove `File` from the name since it's no longer a real file read, it's to/from an API that could store anywhere

PR #993, opened 2017-09-20
prompt user for new credential if connection failed with login failed error

llali:
- If connection failed and it's "login failed" error, prompt user to get new credential, try connection using the new credential 

TODO: have to add tests

### DIFF
--- a/src/views/connectionUI.ts
+++ b/src/views/connectionUI.ts
@@ -276,10 +276,10 @@ export class ConnectionUI {
                 if (result) {
                     let connectionWithoutCredentials = Object.assign({}, connection, { user: '', password: '', emptyPasswordInput: false });
                     ConnectionCredentials.ensureRequiredPropertiesSet(
-                        connectionWithoutCredentials,
-                        true,
-                        false,
-                        true,
+                        connectionWithoutCredentials, // connection profile
+                        true,                         // isProfile
+                        false,                        // isPasswordRequired
+                        true,                         // wasPasswordEmptyInConfigFile
                         this._prompter,
                         this._connectionStore, connection).then(connectionResult => {
                             resolve(connectionResult);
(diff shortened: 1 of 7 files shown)


src/views/connectionUI.ts
```diff
@@ -441,6 +466,18 @@ export class ConnectionUI {
         });
     }
 
+    private promptForRetryConnectWithDifferentCredentials(): PromiseLike<boolean> {
+        // Ask if the user would like to fix the profile
+        return this._vscodeWrapper.showErrorMessage(LocalizedConstants.msgPromptRetryConnectionDifferentCredentials
+            , LocalizedConstants.retryLabel).then(result => {
```
kevcunnane: Minor: I'd just put this on 1 line even if it's a bit long

src/views/connectionUI.ts
```diff
@@ -269,6 +269,31 @@ export class ConnectionUI {
...
 
+    public createProfileWithDifferentCredentials(connection: IConnectionCredentials): Promise<IConnectionCredentials> {
+
+        return new Promise<IConnectionCredentials>((resolve, reject) => {
+            this.promptForRetryConnectWithDifferentCredentials().then(result => {
+                if (result) {
+                    let connectionWithoutCredentials = Object.assign({}, connection, { user: '', password: '', emptyPasswordInput: false });
+                    ConnectionCredentials.ensureRequiredPropertiesSet(
+                        connectionWithoutCredentials,
+                        true,
```
kevcunnane: Minor: can you add comments for what these values are? A bunch of true / false / true makes it hard to read.

src/views/connectionUI.ts
llali: the long is too long so I get formatting error if I don't put them in to lines

---
In reply to: [140110249](https://github.com/Microsoft/vscode-mssql/pull/993#discussion_r140110249) [](ancestors = 140110249)

PR #996, opened 2017-09-21
Fix error where language flavor showed on new untitled docs

kevcunnane:
- Incorrectly checked active editor instead of the doc passed into the onDidOpenTextDocument method

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -602,7 +602,7 @@ export default class MainController implements vscode.Disposable {
         }
         this._connectionMgr.onDidOpenTextDocument(doc);
 
-        if (this._vscodeWrapper.isEditingSqlFile) {
+        if (doc && doc.languageId === Constants.languageId) {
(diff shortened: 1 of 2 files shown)


src/controllers/mainController.ts
```diff
@@ -602,7 +602,7 @@ export default class MainController implements vscode.Disposable {
         }
         this._connectionMgr.onDidOpenTextDocument(doc);
 
-        if (this._vscodeWrapper.isEditingSqlFile) {
+        if (doc && doc.languageId === Constants.languageId) {
```
kburtram: do we need to fix the isEditingSqlFile property, or this is really checking a different condition?

kevcunnane: This is a different thing. These are essentially 2 different events (`onDidOpenTextDocument` vs `onDidChangeActiveEditor`. Since the file open is processed before the editor changes, the `activeEditor` value still refers to the old editor in the vscode API. So basically I was checking the last file, not the current one here.

PR #997, opened 2017-09-21
Added contributions to CHANGELOG.md as this is what users view when o…

kevcunnane:
…penening to see release notes.


kevcunnane:
Will auto-merge this as it's already in the README. Adding here since on opening and clicking "View Release Notes" we'd like to have people see the request to help with localization.

PR #1020, opened 2017-11-08
Bump SQL Tools Service for release

MattIrv:


src/configurations/dev.config.json
```diff
@@ -1,7 +1,7 @@
 {
     "service": {
-        "downloadUrl": "https://download.microsoft.com/download/0/C/9/0C9EBA2F-0D0C-458E-9935-078B64FB5BDE/microsoft.sqltools.servicelayer-{#fileName#}",
-        "version": "1.2.0",
+        "downloadUrl": "https://download.microsoft.com/download/4/D/9/4D983AEF-AF2B-48AF-AB03-78C6DE276BA1/microsoft.sqltools.servicelayer-{#fileName#}",
+        "version": "1.3.0",
```
kevcunnane: Note: can you take the bits you put up on download studio and release on the SqlToolsService Github as version 1.3.0? And also update our automated scripts so that all new versions will be 1.3.0 or higher?

Alternatively, please use the version `1.2.1` instead.

MattIrv: 1.2.1 makes sense, as these are minor targeted changes. I'll update it.

PR #1032, opened 2017-11-17
Improve Integrated Auth troubleshooting

saurabh500:
Adding details about the required Sql Server setup to allow Integrated Authentication using Kerberos.


saurabh500:
@kevcunnane Sounds good. I will work with the folks mentioned above. 
Looks like a lot of people are facing issues because they don't have Kerberos enabled on Sql Server.

Issue #1025, opened 2017-11-13
Error. mssql: Execution failed: Length cannot be less than zero.


4174:
- MSSQL Extension Version: 1.2.1
- VSCode Version: 1.18.0
- OS Version: Windows 7 x64

Steps to Reproduce:

1. Start new workspace.
2. Add two folders.
3. Open SQL file (with SQL commands).
4. Press Ctrl + Shift + E. No reaction.
5. Press F7 (bonded early). See error.

Run from folder (not from workspace) works as usual.

kevcunnane:
@MattIrv could you take a look at this issue? Does it reproduce for you?

MattIrv:
@4174  Thanks for reporting this. I can't seem to reproduce it, so if you're still getting this error message, could you please let me know the answers to the following questions so I can continue trying to track it down:
- Does this error happen for all SQL files or just for a particular one? Are you testing both from the folder and workspace with the same query? If it happens for some queries and not others, is there any chance you could post a simple query that causes it to happen?
- When you say "Open SQL file" in step 3, are you creating a new untitled file from scratch (ctrl+n), or are you opening a saved file somewhere on disk? If it's a saved file, am I understanding correctly that the file is saved outside of either of the two folders you opened?
- In step 5 what command do you have F7 mapped to?

Thanks!

4174:
@MattIrv Thank You for answer

It's seems that problem is folder I run script from. If it's local folder then no problem, but problem arises when i open file from network share.

4174:
Installed update 1.18.**1** doesn't resolve error.

MattIrv:
@4174 thanks for the additional details. I can reproduce this issue, but only when a backslash character `\` is somewhere in the path to the file (either in the filename itself or in the name of one of the parent directories). I'm not sure what's causing this bug yet, but I'll keep looking into it.

Are there any backslashes or other characters that Windows might not expect (potentially any of `:`, `*`, `?`, `<`, `>`, or `|`) somewhere in the full path of the file that gives you this error?

MattIrv:
Stack trace
```

    Unknown error System.ArgumentOutOfRangeException: Length cannot be less than zero.
    Parameter name: length
       at System.String.Substring(Int32 startIndex, Int32 length)
       at System.Uri.CreateHostStringHelper(String str, UInt16 idx, UInt16 end, Flags& flags, String& scopeId)
       at System.Uri.CreateHostString()
[...]

PR #1037, opened 2017-11-30
Bump SQL Tools Service version to support special characters in paths

MattIrv:
This bumps the SQL Tools Service version to include the fix for #1025


MattIrv:
When we released 1.2.0 this got updated to use the download center since that was the newest release, alpha or otherwise. Now that there's an alpha release we want to use we can change it back

PR #1038, opened 2017-12-01
Localization check-in for December update

Monasr:
check in the latest localization XLIFF files for December update


kevcunnane:
I'll close this as your new PR is the right one to accept

PR #1080, opened 2018-04-03
Update docs for 1.3.1

MattIrv:


MattIrv:
@abist I looked through the commit logs for this repo and SQL Tools Service and I think those are all of the meaningful commits that have come in since the last release. I know that they close a bunch of issues but I'm not sure we need to call out all the issues here, especially for the syntax highlighting changes. But if there's anything I missed or anything else you think needs to be included just let me know

abist:
@MattIrv Sounds good. Also look out for the App Veyor build fail.

PR #1086, opened 2018-04-11
[Discussion] [Live Share] Restrict language services to local files

lostintangent:
In preparation for [Visual Studio Live Share](http://aka.ms/vsls) adding support for "guests" to receive remote language services for SQL files, this PR simply updates the current `DocumentSelector` to be limited to `file` and `untitled` (unsaved) files. This way, when someone has this extension installed, and joins a Live Share session (where files use the `vsls:` scheme), their language services will be entirely derived from the remote/host side, which provides a more accurate and project-wide experience (guests in Live Share don't have local file access to the project they're collaborating with).

*Note: As an example, the TypeScript/JavaScript language services that come in-box with VS Code [already have this scheme restriction](https://github.com/Microsoft/vscode/blob/master/extensions/typescript-language-features/src/utils/fileSchemes.ts#L12), and so this PR replicates that behavior.*

### DIFF
--- a/src/languageservice/serviceclient.ts
+++ b/src/languageservice/serviceclient.ts
@@ -276,7 +276,10 @@ export default class SqlToolsServiceClient {
     private createLanguageClient(serverOptions: ServerOptions): LanguageClient {
         // Options to control the language client
         let clientOptions: LanguageClientOptions = {
-            documentSelector: ['sql'],
+            documentSelector: [
+                { language: 'sql', scheme: 'file' },
+                { language: 'sql', scheme: 'untitled' }
+            ],
             synchronize: {
                 configurationSection: 'mssql'
             },


kburtram:
The change looks good conceptually to me, but I see this related issue in the CI build.  Does this mean we need to update to a newer vscode language protocol client?

```
src\languageservice\serviceclient.ts(278,13): error TS2322: Type '{ documentSelector: { language: string; scheme: string; }[]; synchronize: { configurationSection:...' is not assignable to type 'LanguageClientOptions'.
3278
  Types of property 'documentSelector' are incompatible.
3279
    Type '{ language: string; scheme: string; }[]' is not assignable to type 'string | string[]'.
3280
      Type '{ language: string; scheme: string; }[]' is not assignable to type 'string[]'.
3281
        Type '{ language: string; scheme: string; }' is not assignable to type 'string'.
```

kburtram:
@MattIrv can you please help move this PR along?  I think we may need to upgrade our vscode language protocol from the error, though it could be something else.

lostintangent:
@kburtram Apologies! Yes, we need to update the `vscode-languageclient` library. I made that change, and it requires a bigger change set, since there's a handful of type definitions that have changed. I'll take a stab at this and will update the PR.

PR #1097, opened 2018-05-23
New Feature: support to all versions of Linux Deepin, based on Debian distro.

ChristianGrimberg:
### Hi lovers of Visual Studio Code and Microsoft SQL Server!

I add a new feature to support the extention service of __SQLToolService__ to __Linux Deepin__ to all versions. __Linux Deepin__ is based on __Linux Debian__, and the new lines of code is added to support this OS system and can install __SQLToolService__ on __Linux Deepin__. I test this in two Computers and works everything with __mssql__ extention.

Let me know if you agree or not.

Please sorry for my bad english.

### DIFF
--- a/src/models/platform.ts
+++ b/src/models/platform.ts
@@ -26,8 +26,7 @@ export enum Runtime {
     SLES_12_2 = <any>'SLES_12_2',
     RHEL_7 = <any>'RHEL_7',
     Ubuntu_14 = <any>'Ubuntu_14',
-    Ubuntu_16 = <any>'Ubuntu_16',
-    deepin = <any>'deepin'
+    Ubuntu_16 = <any>'Ubuntu_16'
 }
 
 export function getRuntimeDisplayName(runtime: Runtime): string {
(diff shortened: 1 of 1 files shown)


grass0916:
It's work for me! Thanks for your submit :)
Wish the request can be accepted soon.

kburtram:
@MattIrv this looks good to me, but could you please take a quick look and merge if this is ready?

src/models/platform.ts
```diff
@@ -304,6 +306,7 @@ export class PlatformInformation {
             case 'rhel':
                 return Runtime.RHEL_7;
             case 'debian':
+            case 'deepin':
                 return Runtime.Debian_8;
```
MattIrv: Since this returns `Runtime.Debian_8` for `deepin` here, does the runtimes enum need to include `Runtime.deepin` above on line 30?

ChristianGrimberg: Yes, it's true, it would not be necessary.

ChristianGrimberg: I re-run the extension installer with this commit and everything works:
deepinscreenshot_code_20180601200610

MattIrv: Great, thank you!

ChristianGrimberg:
Thank You :thumbsup:

Issue #1101, opened 2018-05-31
RMB menu (Copy etc) unavailable in results when multiple columns are selected


nivek7:
- MSSQL Extension Version:  1.3.1
- VSCode Version:  1.23.1
- OS Version:  MAC OS  High Sierra 10.13.4

Steps to Reproduce:

1.  Return multi-column results from any query
2.  Shift click (or cmd click) to highlight 2 or more columns, RMB is grayed out (unavailable)

If you click the column header for one column, you can use RMB to Copy etc

MattIrv:
In order to get this working for non-contiguous column/row selections we would have to rewrite much of our logic about how we copy and export results, so I'll leave this issue open to track requests related to that. I also have a pull request out now to at least make it work for columns/rows that are next to each other.

abist:
This is now fixed as the copy-paste logic is now rewritten.

PR #1106, opened 2018-06-08
Combine grid selections when copying/exporting results

MattIrv:
This updates the logic when opening the right click context menu for grids so that all selections on the grid get combined into a single large selection if possible.

This partially addresses #1101 where it wasn't possible to select multiple columns or rows that were next to each other and then copy or export the values.

In order to fully address it we would also have to update our serialization code to handle non-contiguous columns and rows, which will be more complicated and can be done in a separate PR.

### DIFF
--- a/src/views/htmlcontent/src/js/components/app.component.ts
+++ b/src/views/htmlcontent/src/js/components/app.component.ts
@@ -470,13 +470,15 @@ export class AppComponent implements OnInit, AfterViewChecked {
 
     openContextMenu(event: {x: number, y: number}, batchId, resultId, index): void {
         let selection = this.slickgrids.toArray()[index].getSelectedRanges();
-        if (selection.length > 1) {
-            selection = this.tryCombineSelections(selection);
-        }
+        selection = this.tryCombineSelections(selection);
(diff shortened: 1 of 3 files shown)


src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -468,14 +470,47 @@ export class AppComponent implements OnInit, AfterViewChecked {
 
     openContextMenu(event: {x: number, y: number}, batchId, resultId, index): void {
         let selection = this.slickgrids.toArray()[index].getSelectedRanges();
+        selection = this.tryCombineSelections(selection);
         this.contextMenu.show(event.x, event.y, batchId, resultId, index, selection);
     }
 
+    private tryCombineSelections(selections: ISlickRange[]): ISlickRange[] {
```
kevcunnane: How expensive is this? E.g. any chance this'll take more than a few milliseconds? Just want to ensure we don't hold up the UI responsiveness with pre-processing logic.

MattIrv: I tested it out on some queries with pretty big result sets and it performed well, but I'll double check

PR #1111, opened 2018-06-22
Update query results to use VS Code's new web view API

### LINKED ISSUE #1109: Update results display to use VS Code web view API
VS Code now exposes a web view API that we can use for the results grid, including an option to persist the web view when it's not active, which will fix issues like #916 

I have some code for this ready to go in the feature/useWebviewApi branch, but it needs a bit more testing. Right now it opens results in a new web view every time you execute the query, even if a result tab for that query is already open. Also the "Started executing query at Line _" link doesn't work correctly if the results and query are in different editor columns.

MattIrv:
Updates the query result display to use VS Code's new web view API. The major benefit of this is that users' selections and scroll positions will no longer be lost when switching between tabs.

Closes #1109

### DIFF
--- a/src/models/sqlOutputContentProvider.ts
+++ b/src/models/sqlOutputContentProvider.ts
@@ -30,7 +30,7 @@ class ResultsConfig implements Interfaces.IResultsConfig {
     messagesDefaultOpen: boolean;
 }
 
-export class SqlOutputContentProvider implements vscode.TextDocumentContentProvider {
+export class SqlOutputContentProvider {
     // CONSTANTS ///////////////////////////////////////////////////////////
     public static providerName = 'tsqloutput';
     public static providerUri = vscode.Uri.parse('tsqloutput://');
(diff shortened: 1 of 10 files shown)


src/models/sqlOutputContentProvider.ts
```diff
@@ -339,33 +340,28 @@ export class SqlOutputContentProvider implements vscode.TextDocumentContentProvi
...
-                    this._vscodeWrapper.showTextDocument(activeTextEditor.document, activeTextEditor.viewColumn);
-                }
-            }, err => {
-                // Output to console if an error occurs
-                Utils.logToOutputChannel(err);
+        // Check if the results window already exists
+        let panel = this._resultsPanes.get(resultsUri);
+        if (!panel) {
+            panel = vscode.window.createWebviewPanel(resultsUri, paneTitle, resultPaneColumn, {
+                retainContextWhenHidden: true,
```
anthonydresser: We should probably think if we actually want this. This has a stark impact on performance as noted by vscode. As it currently stands we built the application to not need this (replaying events).

MattIrv: That's a good point. I didn't see any performance impact in testing but I'm not familiar enough with it to know when you start seeing those issues.

I think maybe it's best to add a setting to let the user control it. Do you think we should set it to `false` by default and let users enable it if they want and that way we can test it for a release or two before enabling it for good?

PR #1125, opened 2018-07-20
Dev snippets

hungrybear88:
Have two snippets:

The first helps to list procedures by SCHEMA;
The second helps to create documentation for SQL Stored Procedure

updated from edsonlourenco
@MattIrv @kburtram


kburtram:
@MattIrv could you please work with @hungrybear88 on this PR?  The snippets look fine from a quick glance but the PR validations are failing.  Let's also run the snippets by @vickyharp to get her opinion.

The tradeoff here is that we want to make sure to only include snippets that are likely to be useful to a significant percentage of the user base, otherwise, we'd encourage users to manage their own personal snippet libraries.

anthonydresser:
@kburtram this PR is pretty old, but still valid, should we take it? the coverage decrease doesn't seem relevant to the PR.

abist:
@kburtram We could take this PR as well if we're taking snippets.

kburtram:
Thanks for opening this PR!  I'm closing out due to age of the PR to clean-up older PRs.  In general, I don't think we're expanding the set of snippets at the moment and recommend users add the snippets to their local config for use if desired.

PR #1154, opened 2018-10-28
Use correct tag for gulp package

eashi:
For new machine setup, running `npm install` errors because the tag used for the gulp package in package.json does not exist (4.0). The error was like this:

```
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but npm-shrinkwrap.json was generated for lockfileVersion@0. I'll try to do my best with it!
npm WARN deprecated postinstall-build@2.1.3: postinstall-build's behavior is now built into npm! You should migrate off of postinstall-build and use the new `prepare` lifecycle script with npm 5.0.0 or greater.
npm ERR! code 1
npm ERR! Command failed: C:\Program Files\Git\cmd\git.EXE checkout 4.0
npm ERR! error: pathspec '4.0' did not match any file(s) known to git.
npm ERR!

```

**The change uses the tag "v4.0.0" which exists in the gulp repo here:** [https://github.com/gulpjs/gulp/releases/tag/v4.0.0](https://github.com/gulpjs/gulp/releases/tag/v4.0.0). 

I have no idea, though, how this worked previously, maybe there was another tag "4.0" that was  removed? or NPM versions handles discovering the tags differently?

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "coveralls": "^2.11.15",
     "decache": "^4.1.0",
     "del": "^2.2.1",
-    "gulp": "github:gulpjs/gulp#4.0",
+    "gulp": "github:gulpjs/gulp#v4.0.0",
     "gulp-clean-css": "^2.0.13",
     "gulp-concat": "^2.6.0",
     "gulp-filter": "^5.0.0",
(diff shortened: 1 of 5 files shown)


eashi:
After the first commit, I received errors of the compiler complaining typings used in different places. It appeared that the gulp 4.0.0 uses a typescript compiler that by default includes the parent `@typings` folder. So in this commit I excluded all the files in the parent level so the `views` folder can compile separately.

MattIrv:
Thank you for fixing this!

PR #1197, opened 2019-03-21
Update vscode language client module

kburtram:
The current vscode language client module is no longer compatible with current builds of VS Code.  This updates to the same version we're using in ADS, which fixes the compat issue with VS Code 1.32.3.  We'll want to upgrade both the VS Code and ADS extensions to the latest language client in the near future.

This PR fixes broken IntelliSense described in https://github.com/Microsoft/vscode-mssql/issues/1194.

Screen Shot 2019-03-21 at 12 07 28 AM


test/serviceClient.test.ts
```diff
@@ -40,39 +40,39 @@ suite('Service Client tests', () => {
...
+    //     let fixture: IFixture = {
+    //         installedServerPath: 'already installed service',
+    //         downloadedServerPath: undefined,
+    //         platformInfo: new PlatformInformation('win32', 'x86_64', undefined)
+    //     };
+
+    //     setupMocks(fixture);
+    //     let serviceClient = new SqlToolsServiceClient(testConfig.object, testServiceProvider.object, logger, testStatusView.object, vscodeWrapper.object);
+
+    //     return serviceClient.initializeForPlatform(fixture.platformInfo, undefined).then( result => {
```
kburtram: The initialization sequence has changed with the new language client and this callback doesn't occur at the same time as before, causing these tests to fail.  I'll try to rewrite the tests, or reenable as a separate PR.

anthonydresser:
At some point it might be worth looking into upgrading to the latest language client, since I'm assuming the same problem might arise? I think they are on v5 now.

kburtram:
@anthonydresser yes, we will want to move to v5 at some point in the near future.  I'm moving to v3 now since that is what ADS uses and has been extensively tested against the current SQL Tools Service implementation.  I want to upgrade the ADS and VS Code extensions to v5 at the same time, along with any associated tools service changes.

src/languageservice/serviceclient.ts
```diff
@@ -191,7 +191,10 @@ export default class SqlToolsServiceClient {
                         });
                     } else {
                         this.initializeLanguageClient(serverPath, context);
-                        resolve(new ServerInitializationResult(false, true, serverPath));
+
+                        this._client.onReady().then(() => {
```
kevcunnane: Can we catch the failure case here? Or else return this .then from so that the outer .catch will be called?
- Recommend changing to async for clarity on this. Inner thenables have this problem, e.g. what are we returning?

kevcunnane: Basically because we don't return `this._client.onReady().then` we risk having unhandled promise instead of the outer .catch on the `this._server.getServerPath(platformInfo.runtimeId).then` being called.

src/languageservice/serviceclient.ts
```diff
@@ -288,9 +291,9 @@ export default class SqlToolsServiceClient {
         client.onReady().then( () => {
             this.checkServiceCompatibility();
 
+            client.onNotification(LanguageServiceContracts.TelemetryNotification.type, this.handleLanguageServiceTelemetryNotification());
```
kevcunnane: This might risk missing some early telemetry notification, but I don't think that's a major risk.

kburtram: I think we've turned-off telemetry for the extension at this point, so no risk for the current release.  When moving to the v5 language client I'll look into refactoring how we register notification handlers.

src/languageservice/serviceclient.ts
kburtram: I switched to await, so I think the exception would go to the existing .catch block below.  Please let me know if this doesn't address the concern.

README.md
```diff
@@ -220,7 +220,7 @@ See [customize options] and [manage connection profiles] for more details.
     "mssql.logDebugInfo": false,
     "mssql.saveAsCsv.includeHeaders": true,
     "mssql.saveAsCsv.delimiter": ",",
-    "mssql.saveAsCsv.lineSeperator": null,
+    "mssql.saveAsCsv.lineSeparator": null,
```
kburtram: Fix the user-facing portion of https://github.com/Microsoft/vscode-mssql/issues/1198.

jsalterjim:
Problems with spelling of `lineSeparator` still exist.

src/models/contracts.ts:
```
    lineSeperator: string = undefined;
```
src/models/resultsSerializer.ts:
```
                saveResultsParams.lineSeperator = saveConfig.lineSeparator;
```

src/models/resultsSerializer.ts
```diff
@@ -80,8 +80,8 @@ export default class ResultsSerializer {
             if (saveConfig.delimiter !== undefined) {
                 saveResultsParams.delimiter = saveConfig.delimiter;
             }
-            if (saveConfig.lineSeperator !== undefined) {
-                saveResultsParams.lineSeperator = saveConfig.lineSeperator;
+            if (saveConfig.lineSeparator !== undefined) {
+                saveResultsParams.lineSeperator = saveConfig.lineSeparator;
```
jsalterjim: `saveResultsParams.lineSeperator` is misspelled (lineSeparator)

kburtram:
@jsalterjim thanks for locating these spelling errors!  I'll look into get them updated in a future PR.

This PR was related to tactical changes to prepare for the v1.5 "hotfix" release.  So I only fixed the spelling errors that would be displayed to the user in the product UI.  I left the spelling of internal variables for a future PR.  The `saveResultsParam` spelling error will involve fixing the same spelling error in an external component (SQL Tools Service) which expects the value with the incorrect spelling, so that can't be updated in isolation.

Issue #1201, opened 2019-03-26
1.5.0  Install/Connect problems


nickbunyan:
Issue Type: <b>Bug</b>

Having run 1.4 successfully for months, I attempted to update to 1.5. After an extremely long processing time it claimed to have installed correctly.
On restarting VSCode, everything looked OK but the content of the sqltoolsservice sub-directory was missing.  Further investigation showed an error message during the post-install processing which failed on the download of the following file:

Microsoft.SqlTools.ServiceLayer-win-x86-netcoreapp2.2.zip 

The error occurred because of an invalid Certificate File in the download path.

I downloaded the file manually and expanded the contents into the corresponding sub-directory based on the 1.4. layout. This allowed the extension to initialise and load.

When trying to use the Connect option on the taskbar at the bottom of screen however, the message ' You have No Profiles defined' (or similar??)  was displayed AND the create new connection icon at the top of the screen was greyed out.

Interestingly - if I use the keyboard shortcut CTRL_SHIFT_E  to try and execute the SQL code, my stored profiles appear in a dropdown at the top of the screen, and everything actually works as far as running a query goes.

This is all running in a Windows 7 Service Pack1  32-bit environment.

Extension version: 1.5.0
VS Code version: Code 1.32.3 (a3db5be9b5c6ba46bb7555ec5d60178ecc2eaae4, 2019-03-14T22:49:40.955Z)
OS version: Windows_NT ia32 6.1.7601

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-4600U CPU @ 2.10GHz (4 x 2693)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: enabled<br>surface_synchronization: enabled_on<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Memory (System)|2.92GB (0.80GB free)|
|Process Argv|C:\Users\BunyNic\.vscode\extensions\ms-mssql.mssql-1.5.0\README.md|
|Screen Reader|no|
|VM|0%|

</details>

kimera76:
[...]

PR #1204, opened 2019-03-27
Read system certificate authorities for HTTPS request

kburtram:
I think https://github.com/Microsoft/vscode-mssql/issues/1201 is related to issues with Node.JS not reading system Root Certificate Authorities for HTTPS requests.  This causes "self-signed cert" issues in environments with an HTTPS decrypting proxy.

### DIFF
--- a/src/languageservice/httpClient.ts
+++ b/src/languageservice/httpClient.ts
@@ -10,6 +10,7 @@ import {parse as parseUrl, Url} from 'url';
 import * as https from 'https';
 import * as http from 'http';
 import {getProxyAgent, isBoolean} from './proxy';
+import { getTrustedCertificates } from '../utils/getTrustedCertificates';
 
 let fs = require('fs');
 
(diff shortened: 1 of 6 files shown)


src/utils/getTrustedCertificates.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export function getTrustedCertificates(importSetting: boolean = true): any {
+    try {
+        return importSetting ? getCertificatesFromSystem() : [];
+    } catch (e) {
+        return [];
+    }
+}
+
+function getCertificatesFromSystem(): (string | Buffer)[] {
```
kburtram: I copied this method from the `vscode-docker` extension.

src/languageservice/httpClient.ts
```diff
@@ -77,11 +78,13 @@ export default class HttpClient implements IHttpClient {
         };
 
         if (url.protocol === 'https:') {
+            let caList = getTrustedCertificates();
```
kburtram: we should put this behind a user setting.  I'm not sure if the default should be on or off?

kburtram:
This didn't seem to fix the problem as intended so I'm going to abandon this PR for now.

PR #1215, opened 2019-04-11
Added create and drop snippets for views

testingcan:
I was wondering why there were no dedicated snippets for creating and dropping views, only for stored procedures, tables and databases. Since I could not find any discussion about this or a reason I added two snippets `sqlCreateView` and `sqlDropView`, adapted from the other drop and create-commands


abist:
We just added the scripting feature that allows scripting for creating, dropping, altering and executing database objects, so this will be unnecessary now.

kburtram:
@abist snippets are a somewhat different use-case than the scripting menu item commands.  Snippets are often used while editing a SQL script to speed up typing common operation and are keyboard focused.  The scripting menu commands are for generating a targeted SQL code block for a single operation and are mouse focused.  

To replace a snippet workflow the user would (1) be editing a SQL script (2) switch from keyboard to mouse (3) locate View in OE tree and run a scripting command (4) copy the generated SQL code from new editor then close the new editor (5) switch back to the document from step 1 and paste the content into the document (6) switch back to keyboard and resume editing the script.

Hope that helps understand the difference in use-cases with these two features.  That said, it looks like there is a merge conflict blocking taking this PR.

abist:
@kburtram That explains a lot! I can take this in for 1.8 if we want after getting it up to date.

PR #1222, opened 2019-04-22
Clean up linting errors and add in SDL-required rules

Charles-Gagnon:
This adds in the SDL-required rules for compliance with TSA scans and fixes a number of linting errors that were flagged. The version of tslint was updated to the latest current as well as a dependency requirement by the tslint-microsoft-contrib package. Because of this some deprecated rules were removed (they are handled by compiler settings instead)


tslint.mssdl.required.json
```diff
@@ -0,0 +1,24 @@
...
+    "no-banned-terms": true,
+    "no-delete-expression": true,
+    "no-disable-auto-sanitization": true,
+    "no-document-domain": true,
+    "no-duplicate-parameter-names": true,
+    "no-eval": true,
+    "no-exec-script": true,
+    "no-function-constructor-with-string-args": true,
+    "no-octal-literal": true,
+    // The actual SDL list includes no-reserved-keywords currently but this has been deprecated
```
abist: Sometimes the json files error out when they have comments in it
Let's make sure it doesn't happen with this one

Charles-Gagnon: What do you mean "errors out"? Yes, technically comments in JSON files aren't permitted but TSLint specifically allows that :

From https://palantir.github.io/tslint/usage/configuration/

>  tslint.json configuration files may have JavaScript-style // single-line and /* multi-line */ comments in them (even though this is technically invalid JSON). If this confuses your syntax highlighter, you may want to switch it to JavaScript format.

abist: TIL!

src/models/platform.ts
```diff
@@ -58,6 +58,80 @@ export function getRuntimeDisplayName(runtime: Runtime): string {
...
+ * seen adoption in "recent" versions of all major distributions.
+ * https://www.freedesktop.org/software/systemd/man/os-release.html
+ */
+export class LinuxDistribution {
+    public constructor(
+        public name: string,
+        public version: string,
+        public idLike?: string[]) { }
+
+    public static GetCurrent(): Promise<LinuxDistribution> {
```
kburtram: we should use `camelCasing` naming convention in TypeScript code, so for exampl, `getCurrent` here.  Ideally tslint would flag these.

kburtram: hmm, looks like you just moved this code around, so it's optional if you'd like to update.

src/views/htmlcontent/tslint.json
```diff
@@ -60,7 +54,6 @@
     "no-string-literal": false,
     "no-switch-case-fall-through": false,
     "no-trailing-whitespace": true,
-    "no-unreachable": true,
```
kburtram: why do we want to remove this rule?

test/istanbultestrunner.ts
```diff
@@ -226,3 +164,65 @@ class CoverageRunner {
     }
 }
 
+function _readCoverOptions(testsRoot: string): ITestRunnerOptions {
```
kburtram: probably should remove the `_` prefix since that isn't our normal coding convention in this project

src/views/htmlcontent/tslint.json
Charles-Gagnon: This rule no longer exists in tslint - it's replaced by the compiler flag noImplicitReturns

test/istanbultestrunner.ts
Charles-Gagnon: There's a ton of leading underscores and the tslint was set up to specifically allow that so I'm going to leave that one in for now.

src/models/platform.ts
Charles-Gagnon: Correct, this was just moved (it failed tslint use-before-declared rule, since it's used down below on line 208)

PR #1225, opened 2019-04-23
Move rulesDirectory from JSON to tasks

Charles-Gagnon:
This is so TSA scans can use the same JSON file (they use a different node_modules location for the additional rules package)


Charles-Gagnon:
Going to merge this in now since the changes are so minimal - let me know if you have any comments you'd like me to address.

PR #1252, opened 2019-06-21
Correct typo in package.nls.json

gcamp0730:
Resubmitting PR with only this one commit


kburtram:
@gcamp0730 thanks for separating this commit out.  We should make the other change to the code, but since those messages are used across a few products that would need to be a coordinated change.

PR #1259, opened 2019-07-15
Commiter Work - Object Explorer

abist:
Changes made -
- [x] Cleaned up copyright header for all files
- [x] Added new extension viewlet for queries and Object Explorer
- [x] Added existing `New Query` command as context menu action for object explorer
- [x] Added Object Explorer as MRU

Screenshots - 

EDIT: Updated Screenshot with Icons
Untitled


src/constants/constants.ts
```diff
@@ -1,3 +1,8 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
```
Charles-Gagnon: It'd be great to add a gulp hygiene task like we have for ADS that checks for these as a precommit hook

kevcunnane: We do - checking in changes to a file will flag missing copyright (and just a space before it, as I've discovered a bunch).

kevcunnane: Oh sorry - wrong repo. Ignore my comment.

src/controllers/localWebService.ts
```diff
@@ -43,8 +48,10 @@ export default class LocalWebService {
         this.server.on('request', this.app);
 
         // Handle new connections to the web socket server
-        this.wss.on('connection', (ws) => {
-            let parse: any = querystring.parse(<string>url.parse(ws.upgradeReq.url).query);
+        this.wss.on('connection', (ws, req) => {
+            // upgradeReq was removed from the API
+            // See - https://github.com/websockets/ws/pull/1099
+            let parse: any = querystring.parse(<string>url.parse(req.url).query);
```
Charles-Gagnon: Was this actually broken? Or did you just happen to see that it was removed and did this now so that if we updated the version later on it wouldn't break? Because right now it seems like this isn't actually changing/fixing anything

abist: No it was broken in master when I started working on this. And I saw the last commit was bumping the version for this library.

package.nls.json
```diff
@@ -2,6 +2,10 @@
 "extension.runQuery":"Execute Query",
 "extension.runCurrentStatement":"Execute Current Statement",
 "extension.cancelQuery":"Cancel Query",
+"extension.addObjectExplorer":"Add Object Explorer",
+"extension.removeObjectExplorerNode":"Remove",
```
Charles-Gagnon: Wouldn't something like "Show Object Explorer" be a better title? Or am I misunderstanding what this command is doing?

abist: Well, there's a `View: Object Explorer` already that we get for free since I added a new `Viewlet`/`View` for Object Explorer. This command basically adds a node to the tree for a new server.

Charles-Gagnon: Then shouldn't it be Add Server or Add Connection or something? Right now it's implying that it's adding "Object Explorer" (whatever that means) so it seems like this could be better clarified

---
In reply to: [303570661](https://github.com/microsoft/vscode-mssql/pull/1259#discussion_r303570661) [](ancestors = 303570661)

package.json
```diff
@@ -123,6 +126,23 @@
     "ws": ">=3.3.1"
   },
   "contributes": {
+    "viewsContainers": {
+      "activitybar": [
+        {
+          "id": "objectExplorer",
+          "title": "Object Explorer",
+          "icon": "./images/server_page_inverse.svg"
+        }
```
Charles-Gagnon: Does this not support light/dark images?

Charles-Gagnon: [https://code.visualstudio.com/api/references/contribution-points#contributes.viewsContainers](https://code.visualstudio.com/api/references/contribution-points#contributes.viewsContainers) Looks like not - just double check this to be sure though if you wouldn't mind

---
In reply to: [303623862](https://github.com/microsoft/vscode-mssql/pull/1259#discussion_r303623862) [](ancestors = 303623862)

package.nls.json
abist: @Charles-Gagnon Sure, I might need to consult Alan once he's back for certain strings.

src/controllers/localWebService.ts
Charles-Gagnon: Alright. I'd just remove the comment though. It has no context here unless you know the history (that it used to be ws.upgradeReq). The logic just using req.url seems fine by itself - not sure why it was using upgradeReq in the first place.

---
In reply to: [303567328](https://github.com/microsoft/vscode-mssql/pull/1259#discussion_r303567328) [](ancestors = 303567328)

src/objectExplorer/objectExplorerProvider.ts
```diff
@@ -0,0 +1,161 @@
...
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import * as vscode from 'vscode';
+import ConnectionManager from '../controllers/connectionManager';
+import { NodeInfo } from '../models/contracts/objectExplorer/nodeInfo';
+import { ObjectExplorerService } from './objectExplorerService';
+import { ConnectionCredentials } from '../models/connectionCredentials';
+
+export class TreeNodeInfo extends vscode.TreeItem {
```
Charles-Gagnon: >TreeNodeInfo [](start = 13, length = 12)

I'd suggest putting this into its own file - no reason to overload this file with additional stuff

src/objectExplorer/objectExplorerService.ts
```diff
@@ -0,0 +1,150 @@
...
+                return this._treeNodeToChildrenMap.get(this._currentNode);
+            } else {
+                // expansion
+                await this.expandNode(element, element.sessionId);
+                return;
+            }
+        } else {
+            if (this._rootTreeNodeArray.length === 0) {
+                this.createSession();
+                return;
```
Charles-Gagnon: >return [](start = 16, length = 6)

You should return an empty array here and above

src/controllers/mainController.ts
```diff
@@ -113,10 +115,38 @@ export default class MainController implements vscode.Disposable {
...
+                    vscode.window.registerTreeDataProvider('objectExplorer', this._objectExplorerProvider)
+                );
+                this.registerCommand(Constants.cmdAddObjectExplorer);
+                this._event.on(Constants.cmdAddObjectExplorer, async () => {
+                    if (self._objectExplorerProvider.objectExplorerExists) {
+                        return self._objectExplorerProvider.createSession();
+                    } else {
+                        self._objectExplorerProvider.objectExplorerExists = true;
+                        return vscode.commands.executeCommand('workbench.view.extension.objectExplorer');
+                    }
```
Charles-Gagnon: This command ID should go with the rest of the ones in constants.ts I would think. Also I'm not seeing this registered anywhere?

src/controllers/mainController.ts
```diff
@@ -113,10 +115,38 @@ export default class MainController implements vscode.Disposable {
...
+                this._context.subscriptions.push(
+                    vscode.window.registerTreeDataProvider('objectExplorer', this._objectExplorerProvider)
+                );
+                this.registerCommand(Constants.cmdAddObjectExplorer);
+                this._event.on(Constants.cmdAddObjectExplorer, async () => {
+                    if (self._objectExplorerProvider.objectExplorerExists) {
+                        return self._objectExplorerProvider.createSession();
+                    } else {
+                        self._objectExplorerProvider.objectExplorerExists = true;
+                        return vscode.commands.executeCommand('workbench.view.extension.objectExplorer');
```
Charles-Gagnon: This seems odd to have it set here. Shouldn't the objectExplorerProvider be the one setting this whenever it's ready?

src/controllers/mainController.ts
abist: Every `view` that is registered with VSCode automatically gets a `workbench.view.extension.view_name` command that opens it up.

src/controllers/mainController.ts
Charles-Gagnon: What does this even control anyways? I'm not sure exists is the right term here - it seems like this is more saying whether the provider is ready or not?

---
In reply to: [303634495](https://github.com/microsoft/vscode-mssql/pull/1259#discussion_r303634495) [](ancestors = 303634495)

abist: This is just for the case when a user opens the `Object Explorer` for the first time. The provider doesn't know that yet.

src/controllers/mainController.ts
```diff
@@ -499,11 +529,22 @@ export default class MainController implements vscode.Disposable {
...
+                this._untitledSqlDocumentService.newQuery().then((uri) => {
+                    // connect to the node if the command came from the context
+                    if (!this.connectionManager.isConnected(sessionId)) {
+                        const connectionCreds = this._objectExplorerProvider.getConnectionCredentials(sessionId);
+                        this._statusview.languageFlavorChanged(uri.toString(), Constants.mssqlProviderName);
+                        return this.connectionManager.connect(uri.toString(), connectionCreds);
+                    }
+                });
+            } else {
+                return this._untitledSqlDocumentService.newQuery().then(x => {
```
Charles-Gagnon: >x [](start = 72, length = 1)

suggest updating this to uri as well to be clear what it actually is

Charles-Gagnon:
So is this only going to display active connections then? That is if I make some connections and then close VS Code the next time I open it up those connections won't show up in OE until you connect to them again?

package.json
```diff
@@ -154,6 +174,35 @@
...
+        {
+          "command": "extension.objectExplorerNewQuery",
+          "when": "viewItem == Server"
+        },
+        {
+          "command": "extension.objectExplorerNewQuery",
+          "when": "viewItem == Database"
+        },
+        {
+          "command": "extension.objectExplorerNewQuery",
```
kburtram: if we could simply add "SELECT TOP 1000" here that would be awesome.  That's one of the top use-cases users want to complete when viewing a table in OE.

package.nls.json
kburtram: yeah, let's use the same text we use in ADS for this purpose: "Add Connection".

test/mainController.test.ts
```diff
@@ -199,21 +204,21 @@ suite('MainController Tests', () => {
 
     test('onNewQuery should call the new query and new connection' , () => {
 
-        untitledSqlDocumentService.setup(x => x.newQuery()).returns(() => Promise.resolve(true));
+        untitledSqlDocumentService.setup(x => x.newQuery()).returns(() => Promise.resolve(undefined));
```
kburtram: can we add some tests for these changes?

src/objectExplorer/objectExplorerService.ts
```diff
@@ -0,0 +1,153 @@
...
+                return this._treeNodeToChildrenMap.get(this._currentNode);
+            } else {
+                // expansion
+                await this.expandNode(element, element.sessionId);
+                return [];
+            }
+        } else {
+            if (this._rootTreeNodeArray.length === 0) {
+                this.createSession();
+                return;
```
Charles-Gagnon: >return [](start = 16, length = 6)

Still should be returning an empty array here

src/controllers/mainController.ts
```diff
@@ -113,10 +113,51 @@ export default class MainController implements vscode.Disposable {
...
+                this._event.on(Constants.cmdRemoveObjectExplorerNode, () => {
+                    return this._objectExplorerProvider.removeObjectExplorerNode(this._objectExplorerProvider.currentNode);
+                });
+                this.registerCommand(Constants.cmdRefreshObjectExplorerNode);
+                this._event.on(Constants.cmdRefreshObjectExplorerNode, () => {
+                    return this._objectExplorerProvider.refreshNode(this._objectExplorerProvider.currentNode);
+                });
+                this._context.subscriptions.push(vscode.commands.registerCommand(Constants.cmdScriptSelect, async (treeNodeInfo) => {
+                    const objectNames = treeNodeInfo.label.split('.');
+                    const tableName = `[${objectNames[0]}].[${objectNames[1]}]`;
```
Charles-Gagnon: >onst tableName = `[${objectNames[0]}].[${objectNames[1]}]` [](start = 21, length = 58)

These names need to be escaped

src/controllers/vscodeWrapper.ts
```diff
@@ -169,8 +174,8 @@ export default class VscodeWrapper {
      * @return A promise that resolves to a [document](#TextDocument).
      * @see vscode.workspace.openTextDocument
      */
-    public openMsSqlTextDocument(): Thenable<vscode.TextDocument> {
-        return vscode.workspace.openTextDocument({ language: 'sql'});
+    public openMsSqlTextDocument(content?: string): Thenable<vscode.TextDocument> {
```
Charles-Gagnon: >content [](start = 33, length = 7)

Update doc comments

test/mainController.test.ts
Charles-Gagnon: Yes, please add some tests

---
In reply to: [305093888](https://github.com/microsoft/vscode-mssql/pull/1259#discussion_r305093888) [](ancestors = 305093888)

src/models/contracts/objectExplorer/refreshSessionRequest.ts
```diff
@@ -0,0 +1,22 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import { ExpandParams } from './expandNodeRequest';
+import { RequestType } from 'vscode-jsonrpc';
+
+/**
+ * Parameters to the ExpandRequest.
```
Charles-Gagnon: >ExpandRequest [](start = 21, length = 13)

RefreshRequest

src/models/contracts/objectExplorer/refreshSessionRequest.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export class RefreshParams extends ExpandParams {}
```
Charles-Gagnon: >ExpandParams [](start = 35, length = 12)

I don't think this should be extending ExpandParams. They have nothing to do with each other so shouldn't be directly related. It's fine in this case to just copy the params if you want to use the same ones

src/models/contracts/objectExplorer/nodeInfo.ts
```diff
@@ -0,0 +1,51 @@
...
+     * The path functions as a unique ID and is used to disambiguate the node when sending requests for expansion.
+     * A common ID is needed since processes do not share address space and need a unique identifier
+     */
+    public nodePath: string;
+
+    /**
+     * The type of node - for example Server, Database, Folder, Table
+     */
+    public nodeType: string;
+
```
Charles-Gagnon: How is this set? It would be greatly preferential to have this be a string enum since these should all be known types

src/models/contracts/objectExplorer/createSessionRequest.ts
```diff
@@ -0,0 +1,67 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import {NotificationType, RequestType} from 'vscode-languageclient';
+// tslint:disable-next-line:no-unused-variable
+import { ConnectionDetails } from '../connection';
```
Charles-Gagnon: Why is this necessary?

src/models/contracts/objectExplorer/closeSessionRequest.ts
```diff
@@ -0,0 +1,79 @@
...
+    public errorMessage: string;
+}
+
+
+ // ------------------------------- < Close Session Request > ----------------------------------------------
+
+/**
+ * Establishes an Object Explorer tree session for a specific connection.
+ * This will create a connection to a specific server or database, register
+ * it for use
```
Charles-Gagnon: This comment seems wrong

src/controllers/untitledSqlDocumentService.ts
```diff
@@ -16,16 +17,16 @@ export default class UntitledSqlDocumentService {
     /**
      * Creates new untitled document for SQL query and opens in new editor tab
      */
-    public newQuery(): Promise<boolean> {
+    public newQuery(content?: string): Promise<Uri> {
```
Charles-Gagnon: >content [](start = 20, length = 7)

Update doc comments

src/controllers/mainController.ts
```diff
@@ -499,11 +540,21 @@ export default class MainController implements vscode.Disposable {
...
+            if (sessionId) {
+                const uri = await this._untitledSqlDocumentService.newQuery(content);
+                // connect to the node if the command came from the context
+                if (!this.connectionManager.isConnected(sessionId)) {
+                    const connectionCreds = this._objectExplorerProvider.getConnectionCredentials(sessionId);
+                    this._statusview.languageFlavorChanged(uri.toString(), Constants.mssqlProviderName);
+                    return this.connectionManager.connect(uri.toString(), connectionCreds);
+                }
+            } else {
+                return this._untitledSqlDocumentService.newQuery().then(uri => {
```
Charles-Gagnon: >then [](start = 67, length = 4)

Just use await here

src/controllers/mainController.ts
```diff
@@ -499,11 +540,21 @@ export default class MainController implements vscode.Disposable {
     /**
      * Opens a new query and creates new connection
      */
-    public onNewQuery(): Promise<boolean> {
+    public async onNewQuery(sessionId?: string, content?: string): Promise<boolean> {
```
Charles-Gagnon: >boolean [](start = 75, length = 7)

Do you want to update this to also return a URI since you did for newQuery()?

package.json
```diff
@@ -154,6 +174,39 @@
...
+          "command": "extension.addObjectExplorer",
+          "title": "%extension.addObjectExplorer%",
+          "group": "navigation"
+        }
+      ],
+      "view/item/context": [
+        {
+          "command": "extension.objectExplorerNewQuery",
+          "when": "viewItem == Server"
+        },
```
Charles-Gagnon: Apparently context keys support OR now [https://github.com/microsoft/vscode/commit/893635042ccd4dbdf919a4493a062330aec2cbc7](https://github.com/microsoft/vscode/commit/893635042ccd4dbdf919a4493a062330aec2cbc7)

test/mainController.test.ts
abist: I'm writing tests now, I haven't completed the PR yet. The next iteration will have them :)

src/models/contracts/objectExplorer/refreshSessionRequest.ts
abist: I'm just following what we have in SQL Tools Service, and ADS does the same thing. They both have the same class members so it kind of makes sense.

src/models/contracts/objectExplorer/nodeInfo.ts
abist: This comes from the tools service

abist:
@Charles-Gagnon Still not done with the PR. I'll re-request a review once I'm finished with it.

src/models/contracts/objectExplorer/createSessionRequest.ts
abist: Linter otherwise complains with the following error - 
```
ERROR: (no-unused-variable) vscode-mssql/src/models/contracts/objectExplorer/createSessionRequest.ts[8, 1]: All imports on this line are unused.
```

Charles-Gagnon: Which makes sense since it looks like you aren't actually using ConnectionDetails anywhere. So you should just remove it if that's the case.

Charles-Gagnon: Oh wait nevermind it's in the generic definition. And actually I recall seeing that issue before...there was a way to fix it though IIRC. Let me look it up - otherwise this is fine

src/controllers/mainController.ts
abist: Think it's fine since it doesn't need it

abist:
@Charles-Gagnon merging soon

src/controllers/mainController.ts
```diff
@@ -113,10 +114,52 @@ export default class MainController implements vscode.Disposable {
...
+                    return this._objectExplorerProvider.removeObjectExplorerNode(this._objectExplorerProvider.currentNode);
+                });
+                this.registerCommand(Constants.cmdRefreshObjectExplorerNode);
+                this._event.on(Constants.cmdRefreshObjectExplorerNode, () => {
+                    return this._objectExplorerProvider.refreshNode(this._objectExplorerProvider.currentNode);
+                });
+                this._context.subscriptions.push(vscode.commands.registerCommand(Constants.cmdScriptSelect, async (treeNodeInfo) => {
+                    const objectNames = treeNodeInfo.label.split('.');
+                    const tableName = `[${escapeCharacters(objectNames[0])}].[${escapeCharacters(objectNames[1])}]`;
+                    const databaseName =  treeNodeInfo.parentNode.parentNode.nodeType === Constants.databaseString ?
```
Charles-Gagnon: It might be a good idea to 

1. Check that the current treeNodeInfo.nodeType === Constants.table
2. Instead of assuming that the parent DB is exactly two nodes above, just walk the parents until you find one with nodeType === Constants.database

And then probably display an error if either of those fail (mostly so users are more likely to report the issue and we can investigate)

It's probably not strictly necessary but I really prefer we be safer on things like this since if this did change and start failing for some reason it'd be more annoying to track down without a clear error message.

src/controllers/untitledSqlDocumentService.ts
```diff
@@ -15,19 +16,17 @@ export default class UntitledSqlDocumentService {
...
-                        resolve(true);
-                    });
-                });
+                let doc = await this.vscodeWrapper.openMsSqlTextDocument(content);
+                // Show the new untitled document in the editor's first tab and change the focus to it.
+                await this.vscodeWrapper.showTextDocument(doc, 1, false);
+                resolve(doc.uri);
             } catch (error) {
                 reject(error);
             }
```
Charles-Gagnon: With the async you can get rid of the resolve/reject and the whole try/catch

package.json
Charles-Gagnon: Not planning on using the new or capabilities? It would be good to try them out to make sure they work. 

---
In reply to: [309266302](https://github.com/microsoft/vscode-mssql/pull/1259#discussion_r309266302) [](ancestors = 309266302)

src/constants/constants.ts
```diff
@@ -55,6 +66,7 @@ export const gettingStartedGuideLink = 'https://aka.ms/mssql-getting-started';
 export const changelogLink = 'https://aka.ms/vscode-mssql-changelog';
 export const integratedAuthHelpLink = 'https://aka.ms/vscode-mssql-integratedauth';
 export const sqlToolsServiceCrashLink = 'https://github.com/Microsoft/vscode-mssql/wiki/SqlToolsService-Known-Issues';
+export const databaseString = 'Database';
```
Charles-Gagnon: I'd rather see this as a string enum here at least so we know what it's intended use is and makes it easier to discover the nodeTypes that are supported.

anthonydresser:
Late to the party here, but why don't we have icons?

abist:
@anthonydresser Wasn't sure if we wanted to increase the size of the extension just for icons. The extension already has minimal UI, so I was just following that.

anthonydresser:
All the other trees in vscode (at least the onces written by microsoft) have icons, even if its just a subset of icons (maybe server, database, object) its nice from a visual perspective to understand where in the tree you are. @yualan Thoughts?

abist:
@kburtram I added the icons and the MRU functionality. I'll go ahead and merge it tonight if I don't hear about any concerns with the new additions.

PR #1267, opened 2019-08-01
Adding command for loading completion extension

shengyfu:


package.nls.json
```diff
@@ -10,6 +10,7 @@
 "extension.showGettingStarted":"Getting Started Guide",
 "extension.newQuery":"New Query",
 "extension.rebuildIntelliSenseCache":"Refresh IntelliSense Cache",
+"mssql.loadCompletionExtension":"Load completion extension into sql language service",
```
kburtram: if this text will be show to user we should capitalize `SQL`.

package.json
```diff
@@ -211,6 +212,11 @@
         "command": "extension.rebuildIntelliSenseCache",
         "title": "%extension.rebuildIntelliSenseCache%",
         "category": "MS SQL"
+      },
+      {
+        "command": "mssql.loadCompletionExtension",
+        "title": "%mssql.loadCompletionExtension%",
+        "category": "MS SQL"
```
kburtram: does this show up in the Command Palette?

src/models/contracts/languageService.ts
```diff
@@ -111,3 +111,27 @@ export class DidChangeLanguageFlavorParams {
...
+    /// </summary>
+    public assemblyPath: string;
+    /// <summary>
+    /// The type name for the completion extension
+    /// </summary>
+    public typeName: string;
+    /// <summary>
+    /// Property bag for initializing the completion extension
+    /// </summary>
+    public properties: {};
```
kburtram: we don't have any better type here?

package.json
shengyfu: This is just for testing, it's removed now.

package.nls.json
shengyfu: No need to expose the command in Command Palette, removed now.

src/models/contracts/languageService.ts
shengyfu: This is to match the properties dictionary needed for initializing a CompletionExtension.

Issue #1234, opened 2019-04-24
A11y_mssql for VSCode_Results Tab: In high contrast mode, Color contrast ratio for "CTRL+ALT+R" is 4.4:1 which is less than  7:1


InduPriya1805:
GitHubTags: #A11y_mssql for VSCode; #A11yMAS; #A11yTCS; #C+E-H; #DesktopApp; #Win10; #Color Contrast; #MAS1.4.3; #mssql for VSCode

**Environment:**
Visual Studio Code – Insider (install from https://na01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fcode.visualstudio.com%2Finsiders.&data=02%7C01%7C%7C00d9f7de52714b323f9908d607afe0c5%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636704849297354687&sdata=cmMqckjoNPGtCS3SZ00bCqCnFprKgiux%2BIjyvzgU0hE%3D&reserved=0)
Version: 1.34.0-insider (user Setup)
Mssql version: 1.6.0
OS Version: Microsoft Windows 10 Enterprise (10.0.17763 Build 17763)

**MAS ID Violated:** MAS 1.4.3

**Note:**
Similarly Color contrast ratio also fails for "CTRL+ALT+Y" which is present at Messages dropdown

**Repro Steps:**
1. Open Visual Studio Code in your system
2. Navigate to "Extensions" and install "SQl Server (mssql)" extension
3. Press F1 and select MS SqL : Connect and connect to any sql server
4. "Write  code (SELECT *
FROM sys.objects;) in the new document and press F1 and select ""MS SQL : Execute Query""  "
5. Change the color theme to high contrast(Files-->Preferences-->Color Theme-->High Contrast)
6. Verify color contrast ratio for "CTRL+ALT+R" is >= 7:1 or not

**Actual Result:**
 Color contrast ratio for "CTRL+ALT+R" is 4.4:1 which is less than  7:1

**Expected Result:**
Color contrast ratio for "CTRL+ALT+R" should be greater than or equal to  7:1

**User Impact:**
If color contrast ratio is less than 7:1 in high contrast theme then low vision users finds difficult in seeing the content present on the screen

**MAS Reference:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={a73546c7-765f-489c-b18f-afb659fe99e6}

**Attachments:**
[...]

PR #1268, opened 2019-08-01
Fixed bug: In high contrast mode, Color contrast ratio for CTRL+ALT+R is 4.4:1 which is less than 7:1

geleems:
This resolve https://github.com/microsoft/vscode-mssql/issues/1234
In the dark theme, the font color for the shortcut strings should be white like other label string so that it fulfills >= 7:1 contrast ratio requirement.

Before fix:
image

After fix: 
image

### DIFF
--- a/src/views/htmlcontent/src/css/color-theme-dark.css
+++ b/src/views/htmlcontent/src/css/color-theme-dark.css
@@ -103,5 +103,5 @@ slick-grid.active .grid .slick-cell.selected {
 }
 
 .shortCut {
-  color: grey;
+  color: rgb(255, 255, 255);
 }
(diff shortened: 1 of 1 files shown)


src/views/htmlcontent/src/css/color-theme-dark.css
```diff
@@ -103,5 +103,5 @@ slick-grid.active .grid .slick-cell.selected {
 }
 
 .shortCut {
-  color: grey;
+  color: rgb(255, 255, 255);
```
alanrenmsft: I think you should use var(--color-content) instead. can you give it a try?

geleems: sure.

PR #1269, opened 2019-08-04
Update vscode-languageclient to support preselect option

Charles-Gagnon:
This required updating the typescript version as well - which then required doing some other fixes so everything built cleanly.

Did some basic manual testing and verified all tests passed successfully.

### DIFF
--- a/src/views/htmlcontent/tsconfig.json
+++ b/src/views/htmlcontent/tsconfig.json
@@ -7,6 +7,7 @@
         "sourceMap": true,
         "rootDir": ".",
         "target": "ES6",
-        "types": []
+        "types": [],
+        "noUnusedLocals": true
     }
 }
--- a/src/views/htmlcontent/tsconfig.json
+++ b/src/views/htmlcontent/tsconfig.json
@@ -7,7 +7,6 @@
         "sourceMap": true,
         "rootDir": ".",
         "target": "ES6",
-        "types": [],
-        "noUnusedLocals": true
+        "types": []
     }
 }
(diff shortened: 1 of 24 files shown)


src/views/htmlcontent/tsconfig.json
```diff
@@ -7,6 +7,7 @@
         "sourceMap": true,
         "rootDir": ".",
         "target": "ES6",
-        "types": []
+        "types": [],
+        "noUnusedLocals": true
```
anthonydresser: I'm not really a fan of this approach, since it means making everything protected in angular code, I would rather move to using ESLint over TSLint and keep it has a linting rule that can be disabled with comments. https://github.com/palantir/tslint/issues/4534

anthonydresser:
How are we integrating into the preselect option? I don't believe we have anything after preselect in the sqltoolsserivce.

Charles-Gagnon:
What do you mean "integrating into the preselect option"? @shengyfu is working on Intellicode support for SQL that uses this flag to select the items which should be put first in the list - the current version of the vscode-languageclient the extension is using doesn't have that so he needs this update in order for his changes to work (most of the changes are in the SqlToolsService side)

src/views/htmlcontent/tsconfig.json
Charles-Gagnon: Hm, that definitely seems like the right long-term approach but I don't think I can prioritize that work over the other stuff I'm assigned to right now. How about as a temporary fix I remove this option and revert the private -> protected changes then create a follow-up issue to move over to ESLint that me or someone else can do once we have time? That way @shengyfu can get unblocked with his Intellicode work.

PR #1270, opened 2019-08-05
Change result view web protocol

abist:
Updated the web view protocol


anthonydresser:
This will solve the security issues but won't solve the remote issues. 

Solving the remote issues will by definition solve the security issues. Is this the goal? @kburtram

abist:
@anthonydresser I'm just trying to close the security bugs, not focusing on remote in particular

src/controllers/mainController.ts
```diff
@@ -143,7 +143,7 @@ export default class MainController implements vscode.Disposable {
         return new Promise<boolean>( (resolve, reject) => {
             // Ensure telemetry is disabled
             Telemetry.disable();
-            SqlToolsServerClient.instance.initialize(self._context).then(serverResult => {
+            SqlToolsServerClient.instance.initialize(self._context).then(async serverResult => {
```
Charles-Gagnon: Why is this being made async? I'm not seeing any changes that would require that

PR #1276, opened 2019-08-15
Extension rewrite and features

abist:
- [x] Remove websockets and use webviews for queries
- [ ] Add tests using webviews
- [x] Handle opening and closing of tabs and saving state for multiple queries
- [x] Fix link to query batches
- [x] Add SQLCMD support
- [ ] Continuous execution time display
- [ ] Merge OE
- [ ] Fix build


src/controllers/sqlOutput.ejs
```diff
@@ -1,36 +1,17 @@
...
-        font-weight: var(--font-weight);
-        font-size: var(--font-size);
-      }
-    </style>
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' vscode-resource: https:; script-src vscode-resource: 'self' 'unsafe-inline' 'unsafe-eval' https:; style-src vscode-resource: 'self' 'unsafe-inline';"/>
+    <base href="<%=basehref%>">
     <link rel="stylesheet" type="text/css" href="lib/css/slick.grid.css"/>
     <link rel="stylesheet" type="text/css" href="lib/css/SlickGrid.css"/>
     <link rel="stylesheet" type="text/css" href="dist/css/styles.min.css"/>
-    <%if (theme === 'vscode-light') { %>
```
anthonydresser: I think we are still going to want this, I just didn't implement it when I was doing this. You should be able to get the theme style from vscode's apis

tasks/htmltasks.js
```diff
@@ -32,6 +32,18 @@ gulp.task('html:lint', () => {
     .pipe(gulpTsLint.report())
 });
 
+// Compile protocol for server and client communication
```
anthonydresser: We should be able to get rid of this entire tasks file. Ideally we will compile the entire project at once. we will still need a separate task to webpack the webview application separately but that is a separate issue.

src/controllers/sqlOutput.ejs
abist: I added the tags to the css file instead to handle the specific styling. Will find a better way to imitate the conditional styling.

src/views/htmlcontent/src/js/components/contextmenu.component.ts
```diff
@@ -83,7 +80,8 @@ export class ContextMenu implements OnInit {
         this.index = index;
         this.selection = selection;
         this.isDisabled = (selection.length > 1);
-        this.position = { x: x, y: y};
+        jQuery('.contextMenu').get(0).style.top = `${y.toString()}px`;
```
anthonydresser: Why did you remove the angular binding? its preferable over useing discord

abist: angular binding doesnt refresh the view because the main html is static

abist:
I'm going to replace this PR with a new one since I branched off this as I was changing the project structure.

PR #1282, opened 2019-08-30
Web View API and project restructure

abist:
Continued from https://github.com/microsoft/vscode-mssql/pull/1276


abist:
@kburtram @anthonydresser @Charles-Gagnon if I could get a review I can reply to the SLA emails for an ETA

.travis.yml
```diff
@@ -39,11 +39,6 @@ install:
 script:
   - gulp build
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then gulp cover:enableconfig; fi
-  - if [ $TRAVIS_OS_NAME == "osx" ]; then
```
anthonydresser: At some point we should readd the karma tests, they test the front end code.

gulpfile.js
```diff
@@ -66,9 +89,110 @@ gulp.task('ext:compile-src', (done) => {
...
+gulp.task('ext:copy-images', (done) => {
+    return gulp.src([
+            config.paths.project.root + '/src/views/htmlcontent/src/images/**/*'])
+        .pipe(gulp.dest('out/src/views/htmlcontent/src/images'));
+});
+
+// Clean angular slickgrid library
+gulp.task('ext:clean-angular2-slickgrid', function() {
+    return del(config.paths.project.root + '/node_modules/angular2-slickgrid/**/*.ts');
+});
```
Charles-Gagnon: Is there a reason you wouldn't just have this delete the entire directory completely?

Charles-Gagnon:
> 

Why comment this out?

---
Refers to: gulpfile.js:282 in 1eec446. [](commit_id = 1eec446814729f724de65bf54ff3dcb81a49503f, deletion_comment = False)

src/controllers/queryRunner.ts
```diff
@@ -222,6 +223,29 @@ export default class QueryRunner {
...
+            for (let resultSetId = 0; resultSetId < batchSet.resultSetSummaries.length; resultSetId++) {
+                let resultSet = batchSet.resultSetSummaries[resultSetId];
+                this.eventEmitter.emit('resultSet', resultSet, true);
+            }
+        }
+        // We're done with this query so shut down any waiting mechanisms
+        this._statusView.executedQuery(this.uri);
+        this.eventEmitter.emit('complete', Utils.parseNumAsTimeString(this._totalElapsedMilliseconds), true);
+        return Promise.resolve(true);
+    }
```
Charles-Gagnon: This can just be return true;

src/controllers/queryRunner.ts
```diff
@@ -222,6 +223,29 @@ export default class QueryRunner {
...
+            const batchSet = this.batchSets[batchId];
+            this.eventEmitter.emit('batchStart', batchSet);
+            let executionTime = <number>(Utils.parseTimeString(batchSet.executionElapsed) || 0);
+            this._totalElapsedMilliseconds += executionTime;
+            if (executionTime > 0) {
+                // send a time message in the format used for query complete
+                this.sendBatchTimeMessage(batchSet.id, Utils.parseNumAsTimeString(executionTime));
+            }
+            this.eventEmitter.emit('batchComplete', batchSet);
+            for (let resultSetId = 0; resultSetId < batchSet.resultSetSummaries.length; resultSetId++) {
```
Charles-Gagnon: So for every batch you send out a batchStart followed immediately by a batchComplete? Is that just to have the new tab "catch up" with the completed results? What happens if a batch is still executing when it's switched?

src/controllers/queryRunner.ts
```diff
@@ -222,6 +223,29 @@ export default class QueryRunner {
         this.eventEmitter.emit('batchComplete', batch);
     }
 
+    // Refreshes the webview panel with the query results when tabs are changed
+    public async refreshQueryTab(): Promise<boolean> {
```
Charles-Gagnon: Use js doc comments (/** **/) for header comments

src/controllers/sqlOutput.ejs
```diff
@@ -0,0 +1,45 @@
...
+  </head>
+  <!-- 3. Display the application -->
+  <body>
+    <my-app></my-app>
+  </body>
+    <% if (prod) { %>
+    <script src="dist/js/app.bundle.js"></script>
+    <% } else { %>
+    <script src="views/htmlcontent/src/js/lib/jquery-1.8.3.js"></script>
+    <script src="views/htmlcontent/src/js/lib/jquery.event.drag-2.2.js"></script>
```
Charles-Gagnon: Isn't this content gone? Is this file even still needed?

src/controllers/webviewController.ts
```diff
@@ -0,0 +1,113 @@
...
+        sendMessage: message => webview.postMessage(message),
+        onMessage: message => webview.onDidReceiveMessage(message)
+    };
+}
+
+export class WebviewPanelController implements vscode.Disposable {
+    public readonly proxy: IWebviewProxy;
+    private _panel: vscode.WebviewPanel;
+    private _disposables: any[] = [];
+    private _isDisposed: boolean = false;
```
Charles-Gagnon: vscode.Disposable[]?

src/languageservice/interfaces.ts
```diff
@@ -28,7 +28,7 @@ export interface IConfig {
 export interface IPackage {
     url: string;
     installPath?: string;
-    tmpFile: tmp.SynchronousResult;
+    tmpFile: tmp.SynchrounousResult;
```
Charles-Gagnon: >SynchrounousResult [](start = 17, length = 18)

The original was the correct spelling

src/languageservice/serviceDownloadProvider.ts
```diff
@@ -10,7 +10,7 @@ import * as path from 'path';
 import {IConfig, IStatusView, IPackage, PackageError, IHttpClient, IDecompressProvider} from './interfaces';
 import  {ILogger} from '../models/interfaces';
 import Constants = require('../constants/constants');
-import * as tmp from 'tmp';
+import tmp = require('tmp');
```
Charles-Gagnon: Why did this change?

gulpfile.js
abist: yes, we need the compiled js files

src/models/connectionCredentials.ts
```diff
@@ -1,7 +1,7 @@
-/* --------------------------------------------------------------------------------------------
- * Copyright (c) Microsoft Corporation. All rights reserved.
```
Charles-Gagnon: It would be great in the future to try and keep unrelated changes like these in another PR...sending the header fixes out before hand and getting that in beforehand would help reduce the randomization when reviewing this

abist: I had this stuff initially but like always the PR kept growing since I don't hear anything. I would commit in parts if I got reviews

src/models/sqlOutputContentProvider.ts
```diff
@@ -237,53 +94,77 @@ export class SqlOutputContentProvider {
...
+            queryCallback: any): Promise<void> {
+        let queryRunner = this.createQueryRunner(statusView ? statusView : this._statusView, uri, title);
+        if (this._panels.has(uri)) {
+            let panelController = this._panels.get(uri);
+            if (panelController.isDisposed) {
+                this._panels.delete(uri);
+                await this.createWebviewController(uri, title, queryRunner);
+            } else {
+                return queryCallback(queryRunner);
+            }
```
Charles-Gagnon: This could result in something being returned but the signature says it's a void return - do you expect the callback to have a return type? Probably best just to have it also be void and then call it directly - then have the return called separately.

```
queryCallback(queryRunner);
return;
```

src/controllers/queryRunner.ts
abist: https://github.com/microsoft/vscode-mssql/pull/1282/files#diff-460f7bbdef98ad458411a73a20676d65R57

.travis.yml
abist: From the tests it looks like we just checked whether it loaded on chrome. Now that we use webviews, do we still need those tests?

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -267,7 +268,7 @@ export class AppComponent implements OnInit, AfterViewChecked {
     // Datasets currently being rendered on the DOM
     private renderedDataSets: IGridDataSet[] = this.placeHolderDataSets;
     private messages: IMessage[] = [];
-    private scrollTimeOut: number;
+    private scrollTimeOut: any;
```
Charles-Gagnon: >any [](start = 27, length = 3)

Why is this any now?

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -320,10 +323,25 @@ export class AppComponent implements OnInit, AfterViewChecked {
...
             switch (event.type) {
+                case 'start':
+                    self.uri = event.data;
+                    // Empty the data set if the query is run
+                    // again on the same panel
+                    if (self.hasRunQuery) {
+                        self.dataSets = [];
+                        self.placeHolderDataSets = [];
+                        self.renderedDataSets = self.placeHolderDataSets;
+                        self.messages = [];
```
Charles-Gagnon: This is going to have these both pointing at the same object - is that intended? So changes made to one will also affect the other.

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -355,7 +371,7 @@ export class AppComponent implements OnInit, AfterViewChecked {
                         ? (self._defaultNumShowingRows + 1) * self._rowHeight + 10
                         : maxHeight;
 
-                    // Store the result set from the event
+                    // Store the result set from the event);
```
Charles-Gagnon: >); [](start = 58, length = 2)

remove

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -670,11 +686,17 @@ export class AppComponent implements OnInit, AfterViewChecked {
...
                 setTimeout(() => {
-                    this.slickgrids.toArray()[0].setActive();
+                    self.slickgrids.toArray()[0].setActive();
+                    self.cd.detectChanges();
                 });
+            } else {
+                // setTimeout(() => {
+                //     self.slickgrids.toArray()[0].setActive();
+                //     self.cd.detectChanges();
+                // });
```
Charles-Gagnon: Not needed?

src/views/htmlcontent/src/js/services/data.service.ts
```diff
@@ -2,65 +2,48 @@
...
-import { ResultSetSubset, ISelectionData,
-    IResultsConfig, WebSocketEvent } from './../interfaces';
-
+import { QueryEvent, ResultSetSubset, ISelectionData } from './../interfaces';
 import * as Constants from './../constants';
-
-const WS_URL = 'ws://localhost:' + window.location.port + '/';
+import { createProxy, IMessageProtocol, IServerProxy } from '../../../../../protocol.js';
+
+declare function acquireVsCodeApi(): { postMessage: (message: string) => void; };
```
Charles-Gagnon: >acquireVsCodeApi [](start = 17, length = 16)

So this gets set at run-time to the actual acquireVsCodeApi function provided by vscode?

src/views/htmlcontent/src/js/services/data.service.ts
```diff
@@ -146,68 +103,35 @@ export class DataService {
...
-        }
-        self.http.post(url, selection, { headers: headers }).subscribe();
+        this._proxy.copyResults(batchId, resultId, selection, includeHeaders);
     }
 
     /**
      * Sends a request to set the selection in the VScode window
-     * @param selection The selection range in the VSCode window
+     * @param batchId The associated batch for the batch result
      */
```
Charles-Gagnon: Update with correct params

gulpfile.js
Charles-Gagnon: Normally clean would be run before a full fresh build so it'd be expected that you'd be pulling down the node_modules again anyways - why are you just deleting the ts files? What is the purpose of doing that?

---
In reply to: [321910002](https://github.com/microsoft/vscode-mssql/pull/1282#discussion_r321910002) [](ancestors = 321910002)

tasks/htmltasks.js
```diff
@@ -32,20 +32,6 @@ gulp.task('html:lint', () => {
     .pipe(gulpTsLint.report())
 });
 
-// Compile TypeScript to JS
-gulp.task('html:compile-src', () => {
```
Charles-Gagnon: Where is this compilation being done now?

src/models/connectionCredentials.ts
Charles-Gagnon: Sure - although from my point of view if I see a PR go out that keeps having more commits added to it I'm going to assume it's not finished yet so tend to hold off looking at it until I hear it's ready. What I usually do if I have more work I want to continue doing is then create another branch off of the initial branch and continue doing my work there so the open PR isn't cluttered with later changes - then when that PR is merged in you can rebase the newer branch to master and continue working as normal from there. It makes the whole reviewing process a lot more straightforward and easier to keep track of for those involved. 

---
In reply to: [321911026](https://github.com/microsoft/vscode-mssql/pull/1282#discussion_r321911026) [](ancestors = 321911026)

src/models/sqlOutputContentProvider.ts
abist: It returns a void promise. Are you looking at the old signature?

Charles-Gagnon: I'm saying the callback itself could return anything since you define it as an any

queryCallback: any

I would suggest updating it to specify that it should have a void return type :

```
queryCallback: (queryRunner: QueryRunner) => void
```

---
In reply to: [322361648](https://github.com/microsoft/vscode-mssql/pull/1282#discussion_r322361648) [](ancestors = 322361648)

Charles-Gagnon: And that you shouldn't be returning the result of that call anyways since this method returns Promise<void> (so just calling return will be essentially the same as return Promise.resolve())

---
In reply to: [322382850](https://github.com/microsoft/vscode-mssql/pull/1282#discussion_r322382850) [](ancestors = 322382850,322361648)

gulpfile.js
abist: because it causes compilation issues and typescript projects aren't supposed to have .ts files in the compiled part

Charles-Gagnon: That's fine - I have no issues with deleting the files. I'm just saying that "clean" has a very specific meaning and that this isn't really doing a clean as I would consider it. I'd at least suggest updating the name to be something like remove-angular2-slickgrid-sourcefiles

.vscode/launch.json
```diff
@@ -2,13 +2,14 @@
 {
     "version": "0.1.0",
     "configurations": [
+
```
kburtram: these changes in this file seem unnecessary I think?

gulpfile.js
```diff
@@ -40,19 +41,41 @@ gulp.task('ext:lint', () => {
...
+                .pipe(nls.rewriteLocalizeCalls())
+                .pipe(nls.createAdditionalLanguageFiles(nls.coreLanguages, config.paths.project.root + '/localization/i18n', undefined, false))
+                .pipe(srcmap.write('.', {
+                   sourceRoot: function(file){ return file.cwd + '/src'; }
+                }))
+                .pipe(gulp.dest('out/src/'));
+});
+
+// Compile view source
+gulp.task('ext:compile-view', (done) => {
```
kburtram: for some reason I was under the impression we were removing the the separate build steps for the "view" code and just have one normal build?

gulpfile.js
```diff
@@ -133,23 +259,23 @@ gulp.task('ext:test', (done) => {
...
+    return gulp.src(['./package.json'])
                 .pipe(install());
 });
 
 gulp.task('watch', function(){
     return gulp.watch(config.paths.project.root + '/src/**/*', gulp.series('build'))
 });
 
-gulp.task('lint', gulp.series('ext:lint', 'html:lint'));
+// gulp.task('lint', gulp.series('ext:lint', 'html:lint'));
```
kburtram: dead code?

src/controllers/queryRunner.ts
```diff
@@ -222,6 +223,31 @@ export default class QueryRunner {
         this.eventEmitter.emit('batchComplete', batch);
     }
 
+    /**
+     * Refreshes the webview panel with the query results when tabs are changed
+     */
+    public async refreshQueryTab(): Promise<boolean> {
+        for (let batchId = 0; batchId < this.batchSets.length; batchId++) {
+            const batchSet = this.batchSets[batchId];
```
kburtram: I haven't heard of a concept as a "batchSet", I think it's just a batch?  Unless this variable contains a set of batches...

src/controllers/webviewController.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+function newResultPaneViewColumn(queryUri: string): vscode.ViewColumn {
+    // Find configuration options
+    let config = vscode.workspace.getConfiguration(Constants.extensionConfigSectionName, vscode.Uri.parse(queryUri));
+    let splitPaneSelection = config[Constants.configSplitPaneSelection];
+    let viewColumn: vscode.ViewColumn;
+
+
+    switch (splitPaneSelection) {
```
kburtram: I'm assuming this logic in this file is copy\paste from previous implementation?  If not, how did we work out these details?

src/models/connectionCredentials.ts
kburtram: @abist you should prefer to create different PRs for different changes rather than growing an existing PR.  If you need new changes based on pending changes then you can create the new PR off your dev branch and diff it against that for concurrent review.  Also, if you're not getting tracking on a review please send email or set up a code review meeting to drive to completion.  Thanks!

src/views/htmlcontent/src/js/services/data.service.ts
abist: yes

tasks/htmltasks.js
abist: This whole file isn't needed. There's only one file for the whole compilation. I'm gonna remove this file.

Issue #1287, opened 2019-09-05
CTRL+C greyed out and Not working


klukiyan:
- MSSQL Extension Version: 1.5.0
- VSCode Version: 1.38.0
- OS Version: Win10

Steps to Reproduce:

1. Execute any query
2. Select one or more cells in results
3. Press CTRL+C
4. Try to paste. Nothing is pasted.
Copying with right click with mouse works. And the `ctrl+c` ShortCut hint on right click with mouse is `greyed out`

It started to happen some time ago. I used mouse and hoped it will get fixed. But it doesn't and is already getting on my nerves. Please help.
![image](https://user-images.githubusercontent.com/27725051/64376270-318b6900-d028-11e9-9cd8-5555ba9ea50a.png)

Issue #1285, opened 2019-09-04
Copy results by keyboard shortcut doesn't work


berglundpetter:
- MSSQL Extension Version: 1.6.0
- VSCode Version: 1.37.1
- OS Version: OS X 10.14.6

Steps to Reproduce:

1. Run a simply query, SELECT 'test' 
2. Copy results using right-click and copy
3. Paste result in test 
4. Change query to test2 
5. Select the results cell, copy using ⌘ + c
6. Paste, giving random results from the query window 

This has worked previously, and while it works using the dropdown it's quite inconvenient.

The GIF shows the above steps.

![vscode_mssql_copy_bug](https://user-images.githubusercontent.com/63369/64268601-ad0ded00-cf38-11e9-8b9b-f14f89c70e31.gif)

Issue #1280, opened 2019-08-26
Feature request: "Add current connection to profile" command


cmeyertons:
Would be great if there was a "Add current connection to profile" command.

Typically in our multi-tenant SAAS environment, our workflow is:

- Connect to server
- Use database X

Adding that combination specifically to the profile requires copy/pasting both the server & database into the Add Connection wizard.  Would be great if it leveraged the current context and all we had to supply was a friendly name.

kburtram:
Thanks for this suggestion! I'm closing older feature requests without significant upvotes as part of routine backlog clean-up. I've added a label so we can reconsider when doing longer-term planning in this area.

Issue #1272, opened 2019-08-08
Cannot Adjust Size of SQL Results Window


jd-carroll:
Issue Type: <b>Bug</b>

When I run a SQL query using MSSQL extension, the results tab will automatically open to the right with 50% of the screen.  If I try to adjust the size (width) of the SQL window, it will grow in size perfectly.

However, if I try to reduce the size (width) of the results tab, it does not let me.  The mouse will separate from the vertical sash bar and just starts behaving very strangely.

NOTE: Not sure if this is the msssql extension or VS Code itself

Extension version: 1.6.0
VS Code version: Code 1.36.1 (2213894ea0415ee8c85c5eea0d0ff81ecc191529, 2019-07-08T22:59:35.033Z)
OS version: Windows_NT x64 10.0.16299

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz (4 x 2496)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>oop_rasterization: disabled_off<br>protected_video_decode: unavailable_off<br>rasterization: enabled<br>skia_deferred_display_list: disabled_off<br>skia_renderer: disabled_off<br>surface_synchronization: enabled_on<br>video_decode: enabled<br>viz_display_compositor: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|15.88GB (4.74GB free)|
|Process Argv|C:\Users\a639918\code|
|Screen Reader|no|
|VM|0%|
</details>

PR #1290, opened 2019-09-13
Bug fixes/customer asks (WIP)

abist:
Listing down all the bug fixes that came as part of the refactor and added some new customer asks (features)

Fixes -
- [x] https://github.com/microsoft/vscode-mssql/issues/1272
- [x] https://github.com/microsoft/vscode-mssql/issues/1260
- [x] https://github.com/microsoft/vscode-mssql/issues/1123
- [ ] https://github.com/microsoft/vscode-mssql/issues/1280
- [x] https://github.com/microsoft/vscode-mssql/issues/1287
- [x] https://github.com/microsoft/vscode-mssql/issues/1285
- [x] https://github.com/microsoft/vscode-mssql/issues/1189
- [x] https://github.com/microsoft/vscode-mssql/issues/1124
- [x] https://github.com/microsoft/vscode-mssql/issues/1121
- [x] https://github.com/microsoft/vscode-mssql/issues/1001
- [x] https://github.com/microsoft/vscode-mssql/issues/275
- [x] https://github.com/microsoft/vscode-mssql/issues/1000
- [x] https://github.com/microsoft/vscode-mssql/issues/833
- [x] https://github.com/microsoft/vscode-mssql/issues/1093
- [x] https://github.com/microsoft/vscode-mssql/issues/969


abist:
@yualan This list might help you in getting the list of bug fixes/features for the coming release.

src/controllers/webviewController.ts
```diff
@@ -61,6 +62,10 @@ export class WebviewPanelController implements vscode.Disposable {
                     // angular component to show up
                     setTimeout(async () => await this.queryRunner.refreshQueryTab(), 1700);
                 }
+            // if focus is on text editor
+            } else if (p.webviewPanel.visible && !p.webviewPanel.active) {
+                this._isPanelFocused = true;
```
kburtram: this code block seems weird to me.  The comment says "focus is on the text editor" the condition checks that the panel is not active, then we set panel is focused to true?

maybe I'm missing something but the three inconsistent.  Could you please explain this code block a little?  Thanks!

abist: There are 3 cases in this scenario -
1. The webview panel is visible and active. 
2. The webview panel is not visible and not active. (That means the results tab was changed)
2. The webview panel is visible but not active. (That means the editor has focus). 

I set the `isFocused` variable in the second and third cases because this event is propagated before the actual `active` and `focus` properties are set. The way the second case can happen is also 2 ways -
1. it just got focus from the text editor (which is case 2)
2. it just got focus from a tab change

I keep track of the first part by using the `isPanelFocused` variable to detect what kind of change it was. Perhaps I can name it to `didTabChange` for better readability.

PR #1296, opened 2019-09-18
Enable SQLCMD mode

abist:


src/controllers/connectionManager.ts
```diff
@@ -438,9 +438,18 @@ export default class ConnectionManager {
         });
     }
 
-    public onChooseLanguageFlavor(): Promise<boolean> {
+    public onChooseLanguageFlavor(isSqlCmd: boolean = false): Promise<boolean> {
```
udeeshagautam: quick question - when we change the sqlcmd mode, is there an event that will trigger this call?

abist: Yes, here - https://github.com/microsoft/vscode-mssql/pull/1296/files#diff-7ee4de034bfbbafda956e5159d6e82a7R254

I probably missed this when you first commented here, apologies

PR #1299, opened 2019-09-19
Bump SQL Tools Service to 2.0.0.release.14

kburtram:


kburtram:
@abist let's include this SQL Tools vbump in the test VSIX we'll be testing tomorrow.

PR #1301, opened 2019-09-20
demo bugs

abist:
Fixed bugs from the demo yesteday -

- [x] Null reference in concurrent when removing all connections
- [x] When no connections show a node with a text message.  And hook the click event to add new connection Azure extension.
- [x] All saved connections need be displayed in the Connections viewlet
- [x] All saved connections should show in the Connections viewlet
- [ ] Connected nodes should have connection icon (Blocked on UX team)
- [x] SELECT 1000 doesn’t list columns
- [x] SQLCMD status bar isn’t consistently
- [x] OE doesn’t refresh properly on
- [x] Rename viewlet to “SQL Server: CONNECTIONS”
- [x] Connect from viewlet
- [x] Connect from editor
- [x] Connect from editor to existing
- [x] Node should collapse if not provided a password
- [x] Make new connections for `New Query`
- [x] Add disconnect node context menu action


src/constants/constants.ts
```diff
@@ -13,6 +13,8 @@ export const connectionApplicationName = 'vscode-mssql';
 export const outputChannelName = 'MSSQL';
 export const connectionConfigFilename = 'settings.json';
 export const connectionsArrayName = 'connections';
+export const disconnectedServerLabel = 'disconnectedServer';
```
Charles-Gagnon: >disconnectedServerLabel [](start = 13, length = 23)

These aren't really labels (label to me would imply displayed text). NodeType seems more appropriate here

src/controllers/connectionManager.ts
```diff
@@ -523,7 +523,7 @@ export default class ConnectionManager {
                     self.connect(fileUri, connectionCreds)
                     .then(result => {
                         self.handleConnectionResult(result, fileUri, connectionCreds).then(() => {
-                            resolve(true);
+                            resolve(connectionCreds);
                         });
```
Charles-Gagnon: You never seem to actually use the returned creds so I'd just remove this whole .then() since it's not doing anything

src/controllers/mainController.ts
```diff
@@ -144,28 +146,52 @@ export default class MainController implements vscode.Disposable {
...
+                    // if the node is already connected
+                    if (treeNodeInfo.sessionId) {
+                        return self.runAndLogErrors(self.onNewQuery(treeNodeInfo.sessionId), 'onNewQueryOE');
+                    } else {
+                        const uri = await this._untitledSqlDocumentService.newQuery();
+                        // connect to the node if the command came from the context
+                        this._statusview.languageFlavorChanged(uri.toString(), Constants.mssqlProviderName);
+                        this._statusview.sqlCmdModeChanged(uri.toString(), false);
+                        return this.connectionManager.connect(uri.toString(), connectionCredentials);
+                    }
```
Charles-Gagnon: This should be await'ed

src/objectExplorer/objectExplorerService.ts
```diff
@@ -130,44 +266,71 @@ export class ObjectExplorerService {
...
-        };
-        const response = await this._connectionManager.client.sendRequest(CloseSessionRequest.type,
-            closeSessionParams);
-        if (response && response.success) {
-            this._sessionIdToConnectionCredentialsMap.delete(response.sessionId);
-            node.sessionId = undefined;
-            this._currentNode = node;
-            this._treeNodeToChildrenMap.set(this._currentNode, undefined);
-            this._currentNode.collapsibleState = TreeItemCollapsibleState.Collapsed;
+    public signInNodeServer(node: TreeNodeInfo): void {
```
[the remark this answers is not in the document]
abist: No there's no way to handle this. VSCode's API doesn't allow you to refresh specific nodes without refreshing the root node. Even the azure extension does error handling the same way. I worked on this specific issue all last week. This is also because how the objectExplorerService is written in tools service. It doesn't have a response event for OE requests, but has notifications. Since the VSCode API expects children every time the treeDataChanged event is fired, I ended up using Deferred promises to mimic correct Promise behavior. However, to force the nodes to show up, I have to refresh them. 

Not removing the node because removing from the OE also removes from saved connections.

We can go over this offline in detail if you want.

src/models/connectionStore.ts
```diff
@@ -153,6 +153,21 @@ export class ConnectionStore {
...
+    public lookupPassword(connectionCredentials: IConnectionCredentials): Promise<string> {
+        const credentialId = ConnectionStore.formatCredentialIdForCred(connectionCredentials);
+        return this._credentialStore.readCredential(credentialId).then((savedCredential) => {
+            if (savedCredential && savedCredential.password) {
+                return savedCredential.password;
+            } else {
+                return undefined;
+            }
+        }).catch((err) => Promise.reject(err));
+    }
```
Charles-Gagnon: Would prefer this refactored to use async/await.  It'd then look like this
```
const savedCredential = await this._credentialStore.readCredential(credentialId)
if (savedCredential && savedCredential.password) {
      return savedCredential.password;
} else {
      return undefined;
}
```

src/objectExplorer/accountSignInTreeNode.ts
```diff
@@ -0,0 +1,29 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+
+import * as vscode from 'vscode';
+import * as LocalizedConstants from '../constants/localizedConstants';
+import Constants = require('../constants/constants');
+import { TreeNodeInfo } from './treeNodeInfo';
```
Charles-Gagnon: Just use import from

src/objectExplorer/addConnectionTreeNode.ts
```diff
@@ -0,0 +1,24 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+import * as path from 'path';
+import * as vscode from 'vscode';
+import Constants = require('../constants/constants');
+import * as LocalizedConstants from '../constants/localizedConstants';
```
Charles-Gagnon: Just use import from

src/objectExplorer/objectExplorerProvider.ts
```diff
@@ -29,29 +31,36 @@ export class ObjectExplorerProvider implements vscode.TreeDataProvider<any> {
...
-            return Promise.resolve(children);
+            return children;
         }
     }
 
-    async createSession(): Promise<string> {
-        return await this._objectExplorerService.createSession();
+    async createSession(promise: Deferred<TreeNodeInfo>, connectionCredentials?: IConnectionCredentials): Promise<void> {
+        return this._objectExplorerService.createSession(promise, connectionCredentials);
     }
```
Charles-Gagnon: Just await the call here instead of returning so we're consistent with async/await style

src/objectExplorer/objectExplorerProvider.ts
```diff
@@ -29,29 +31,36 @@ export class ObjectExplorerProvider implements vscode.TreeDataProvider<any> {
...
 
     public getConnectionCredentials(sessionId: string): ConnectionCredentials {
```
Charles-Gagnon: >sessionId [](start = 36, length = 9)

This should be sessionId? if it's actually expected to be possibly undefined

Charles-Gagnon: And why are you even doing the check here? The OE Service already does this check so this shouldn't be necessary. 

---
In reply to: [327299650](https://github.com/microsoft/vscode-mssql/pull/1301#discussion_r327299650) [](ancestors = 327299650)

src/objectExplorer/objectExplorerProvider.ts
```diff
@@ -29,29 +31,36 @@ export class ObjectExplorerProvider implements vscode.TreeDataProvider<any> {
...
+    public removeObjectExplorerNode(node: TreeNodeInfo, isDisconnect: boolean = false): Promise<void> {
+        return this._objectExplorerService.removeObjectExplorerNode(node, isDisconnect);
     }
 
-    public refreshNode(node: TreeNodeInfo): Promise<boolean> {
+    public refreshNode(node: TreeNodeInfo): Promise<void> {
         return this._objectExplorerService.refreshNode(node);
     }
 
+    public signInNodeServer(node: TreeNodeInfo): void {
```
Charles-Gagnon: >signInNodeServer [](start = 11, length = 16)

This needs to be renamed or have a doc comment - I have no idea what this is even doing from the name. It sounds like a property but doesn't return anything.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -43,10 +58,46 @@ export class ObjectExplorerService {
         const handler = (result: SessionCreatedParameters) => {
             if (result.success) {
                 let nodeLabel = this._nodePathToNodeLabelMap.get(result.rootNode.nodePath);
-                self._currentNode = TreeNodeInfo.fromNodeInfo(result.rootNode, result.sessionId, self.currentNode, nodeLabel);
-                self._rootTreeNodeArray.push(self.currentNode);
+                // if no node label, check if it has a name in saved profiles
+                // in case this call came from new query
```
Charles-Gagnon: Shouldn't you be checking if nodeLabel is undefined at this point so you don't have to do the logic if we already have a label?

src/objectExplorer/objectExplorerService.ts
```diff
@@ -56,70 +107,155 @@ export class ObjectExplorerService {
...
+                rootTreeNode.label === node.label) {
+                    const index = this._rootTreeNodeArray.indexOf(rootTreeNode);
+                    this._rootTreeNodeArray[index] = node;
+                    return;
+            }
+        }
+        this._rootTreeNodeArray.push(node);
+    }
+
+    async getChildren(element?: TreeNodeInfo): Promise<vscode.TreeItem[]> {
```
Charles-Gagnon: >async [](start = 4, length = 5)

All functions should have visibility modifiers so it's explicit what it should be

src/objectExplorer/objectExplorerService.ts
```diff
@@ -56,70 +107,155 @@ export class ObjectExplorerService {
...
+            if (this._treeNodeToChildrenMap.has(this._currentNode)) {
                 return this._treeNodeToChildrenMap.get(this._currentNode);
             } else {
-                // expansion
-                await this.expandNode(element, element.sessionId);
-                return [];
+                // check if session exists
+                if (element.sessionId) {
+                    // clean created session promise
+                    this._sessionIdToPromiseMap.delete(element.sessionId);
```
Charles-Gagnon: So by deleting these without completing you're going to leave dangling promises all over the place. It would really be better practice to resolve them with just undefined or something as appropriate and make sure the callers awaiting them handle those situations

src/objectExplorer/objectExplorerService.ts
```diff
@@ -56,70 +107,155 @@ export class ObjectExplorerService {
...
+                            return children;
+                        }
+                    });
+                } else {
+                    // start node session
+                    let promise = new Deferred<TreeNodeInfo>();
+                    this.createSession(promise, element.connectionCredentials);
+                    return promise.then((node) => {
+                        // If password wasn't given
+                        if (!node) {
```
Charles-Gagnon: How does node being undefined mean that the password wasn't given?

src/objectExplorer/objectExplorerService.ts
abist: How's it dangling? Where is the case where the promise isn't resolved or rejected?

src/views/connectionUI.ts
```diff
@@ -182,6 +183,26 @@ export class ConnectionUI {
...
+                name: LocalizedConstants.passwordPrompt,
+                message: LocalizedConstants.passwordPrompt,
+                placeHolder: LocalizedConstants.passwordPlaceholder
+            };
+            self._prompter.promptSingle(question).then((result: string) => {
+                resolve(result);
+            }).catch(err => {
+                reject(err);
+            });
+        });
```
Charles-Gagnon: Would prefer this done using async/await

```
let question: IQuestion = {
      type: QuestionTypes.password,
      name: LocalizedConstants.passwordPrompt,
      message: LocalizedConstants.passwordPrompt,
      placeHolder: LocalizedConstants.passwordPlaceholder
};
return await self._prompter.promptSingle(question);
```

test/mainController.test.ts
```diff
@@ -205,18 +205,18 @@ suite('MainController Tests', () => {
     test('onNewQuery should call the new query and new connection' , () => {
```
Charles-Gagnon: No actual test changes? They all pass? That signifies a pretty large gap in the test matrix if so. Do you have any plans on what you'd like to do around that given the renewed focus on fundamentals and quality?

src/objectExplorer/objectExplorerService.ts
Charles-Gagnon: mainController.ts

```
                    let promise = new Deferred<TreeNodeInfo>();
                    await self._objectExplorerProvider.createSession(promise);
                    return promise.then(() => {
                        this._objectExplorerProvider.refresh(undefined)
                    });
```

The .then would never be called and thus the Promise would never be fully resolved. I don't believe this would ever cause an actual runtime issue (haven't investigated your usage of these fully) but regardless we'd currently be leaking promises which is something we should avoid.

---
In reply to: [327304232](https://github.com/microsoft/vscode-mssql/pull/1301#discussion_r327304232) [](ancestors = 327304232)

abist:
@Charles-Gagnon Addressing in coming commits.

PR #1304, opened 2019-09-23
Scripting

abist:
Added scriptSelect to extension and opened APIs for further scripting actions.


src/controllers/connectionManager.ts
```diff
@@ -722,6 +735,23 @@ export default class ConnectionManager {
         });
     }
 
+    /**
+	 * Gets the connection string to send to the middleware
```
Charles-Gagnon: >	  [](start = 0, length = 2)

nit: spacing (you can visualize this in VS Code with View: Toggle Render Whitespace command to make this easy to spot)

src/controllers/connectionManager.ts
```diff
@@ -722,6 +735,23 @@ export default class ConnectionManager {
...
+	 */
+    public async getConnectionString(connection: Interfaces.IConnectionCredentials): Promise<string> {
+		let connectionString: string;
+
+		if (connection.authenticationType === 'Integrated') {
+            connectionString =
+            `Data Source=${connection.server + (connection.port ? `,${connection.port}` : '')};Initial Catalog=${connection.database};Integrated Security=True`;
+		} else {
```
Charles-Gagnon: Is this going to be correct if a port is specified? Don't you need to add the , in or is that already added in that case?

abist: I copied this code from ADS

src/controllers/mainController.ts
```diff
@@ -189,18 +189,27 @@ export default class MainController implements vscode.Disposable {
...
+                    if (!this.connectionManager.isConnected(uri.toString())) {
+                        const connectionCreds = node.connectionCredentials;
+                        const databaseName = `${escapeCharacters(self.getDatabaseName(node))}`;
+                        node.connectionCredentials.database = databaseName;
+                        this._statusview.languageFlavorChanged(uri.toString(), Constants.mssqlProviderName);
+                        this._statusview.sqlCmdModeChanged(uri.toString(), false);
+                        await this.connectionManager.connect(uri.toString(), connectionCreds);
+                        const selectStatement = await this._scriptingService.scriptSelect(node, uri.toString());
+                        self.onNewQuery(node.sessionId, selectStatement).then((result) => {
+                            if (result) {
```
Charles-Gagnon: This is already async - suggest just using await unless you have a specific reason for not wanting to wait for onNewQuery to complete

src/objectExplorer/objectExplorerService.ts
```diff
@@ -68,6 +68,7 @@ export class ObjectExplorerService {
                     }
                 }
                 // set connection and other things
+                // let credentials = self._currentNode.connectionCredentials;
                 if (self._currentNode) {
```
Charles-Gagnon: remove?

src/objectExplorer/objectExplorerUtils.ts
```diff
@@ -5,6 +5,7 @@
 import * as path from 'path';
 import { TreeNodeInfo } from './treeNodeInfo';
 import { IConnectionProfile } from '../models/interfaces';
+import Constants = require('../constants/constants');
```
Charles-Gagnon: Just use import

src/objectExplorer/objectExplorerUtils.ts
```diff
@@ -15,8 +16,13 @@ export class ObjectExplorerUtils {
             return path.join(ObjectExplorerUtils.rootPath, `${label}.svg`);
         }
     }
-
     public static getNodeUri(node: TreeNodeInfo): string {
```
Charles-Gagnon: >getNodeUri [](start = 18, length = 10)

Your changes aren't really doing this anymore though - you're always getting the Server node URI. Why is this change necessary?

src/scripting/scriptingService.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    constructor(
+        private _connectionManager: ConnectionManager,
+        private _vscodeWrapper: VscodeWrapper
+    ) {
+        this._client = this._connectionManager.client;
+    }
+
+    // map for the version of SQL Server (default is 140)
+    readonly scriptCompatibilityOptionMap = {
```
Charles-Gagnon: Bleh...we should really be supporting 150. But in general I'm not a fan of us having these hard coded values everywhere since it makes vBump a huge pain. Nothing I expect you to fix right now but if you could be thinking about ways to make this not be duplicated as much that'd be good.

src/views/connectionUI.ts
```diff
@@ -454,6 +454,9 @@ export class ConnectionUI {
     private validateAndSaveProfile(profile: Interfaces.IConnectionProfile): PromiseLike<Interfaces.IConnectionProfile> {
         const self = this;
         let uri = self.vscodeWrapper.activeTextEditorUri;
+        // if (self.vscodeWrapper.isEditingSqlFile) {
+        //     uri = self.vscodeWrapper.activeTextEditorUri;
+        // }
```
Charles-Gagnon: remove

src/controllers/connectionManager.ts
Charles-Gagnon: Doesn't mean it's correct. You should verify yourself to make sure it's doing the right thing because to me it seems like it isn't. 

---
In reply to: [327252219](https://github.com/microsoft/vscode-mssql/pull/1304#discussion_r327252219) [](ancestors = 327252219)

src/objectExplorer/objectExplorerUtils.ts
abist: https://github.com/microsoft/vscode-mssql/pull/1301/files#diff-9401975d07f4df26ce1df2d776b8c746R277

PR #1310, opened 2019-09-24
Fix SQLCMD mode

abist:
Fixed issue in SQLCMD mode when you would try to toggle SQL CMD without a query runner. Also show SQLCMD only when text editor is connected.

Also cleaned up OE nodes when removed.


src/controllers/mainController.ts
```diff
@@ -288,35 +289,40 @@ export default class MainController implements vscode.Disposable {
...
     }
 
+    /**
+     * Helper function to toggle SQLCMD mode
+     */
+    private async toggleSqlCmdMode(isSqlCmd: boolean): Promise<boolean> {
+        return this._outputContentProvider.toggleSqlCmd(this._vscodeWrapper.activeTextEditorUri).then(async () => {
+            await this._connectionMgr.onChooseLanguageFlavor(true, !isSqlCmd);
+            return Promise.resolve(true);
+        });
```
Charles-Gagnon: Why are you not using async/await for the whole thing?

```
await this._outputContentProvider.toggleSqlCmd(this._vscodeWrapper.activeTextEditorUri);
await this._connectionMgr.onChooseLanguageFlavor(true, !isSqlCmd);
return true;
```

I also don't see the point in returning a boolean here since you never return false.

src/controllers/mainController.ts
```diff
@@ -288,35 +289,40 @@ export default class MainController implements vscode.Disposable {
...
     /**
      * Handles the command to enable SQLCMD mode
      */
     private async onToggleSqlCmd(): Promise<boolean> {
         let isSqlCmd: boolean;
-        const queryRunner = this._outputContentProvider.getQueryRunner(this._vscodeWrapper.activeTextEditorUri);
+        const uri = this._vscodeWrapper.activeTextEditorUri;
+        const queryRunner = this._outputContentProvider.getQueryRunner(uri);
         const promise = new Promise<boolean>(async (resolve, reject) => {
+            // if a query runner exists, use it
```
Charles-Gagnon: Shouldn't need to wrap this in a promise - it's already async.

src/controllers/mainController.ts
```diff
@@ -288,35 +289,40 @@ export default class MainController implements vscode.Disposable {
...
+            return Promise.resolve(true);
+        });
+    }
+
+
     /**
      * Handles the command to enable SQLCMD mode
      */
     private async onToggleSqlCmd(): Promise<boolean> {
         let isSqlCmd: boolean;
```
Charles-Gagnon: The only place this is called doesn't even use the return value so I don't really see the point in having it

src/controllers/mainController.ts
```diff
@@ -672,22 +678,31 @@ export default class MainController implements vscode.Disposable {
...
+                            connectionCreds.password = password;
+                        }
+                    }
                     this._statusview.languageFlavorChanged(uri.toString(), Constants.mssqlProviderName);
+                    await this.connectionManager.connect(uri.toString(), connectionCreds);
                     this._statusview.sqlCmdModeChanged(uri.toString(), false);
-                    let result = await this.connectionManager.connect(uri.toString(), connectionCreds);
-                    return Promise.resolve(result);
+                    return Promise.resolve(true);
                 }
```
Charles-Gagnon: Just return true, no point in wrapping it in a Promise.resolve

src/controllers/webviewController.ts
```diff
@@ -60,7 +60,7 @@ export class WebviewPanelController implements vscode.Disposable {
                     // give a 2 sec delay because the the webview visible event is fired
                     // before the angular component is actually built. Give time for the
                     // angular component to show up
-                    setTimeout(async () => await this.queryRunner.refreshQueryTab(), 1700);
+                    setTimeout(async () => await this.queryRunner.refreshQueryTab(), 2000);
                 }
```
Charles-Gagnon: Why? How do you know this new value is high enough?

Charles-Gagnon:
>         bar.sqlCmdMode.command = Constants.cmdToggleSqlCmd;

This should be localized

---
Refers to: src/views/statusView.ts:215 in c109e32. [](commit_id = c109e3288caa421f6ea421f376d9c314e04ff115, deletion_comment = False)

src/controllers/webviewController.ts
abist: This came from testing myself initially from the initial PR, and hence the comment on the top. I made some perf changes, which led it query results loading faster, but it takes more time on script as select, for example

abist:
> > ```
> >     bar.sqlCmdMode.command = Constants.cmdToggleSqlCmd;
> > ```
> 
> This should be localized
> 
> Refers to: src/views/statusView.ts:215 in [c109e32](https://github.com/microsoft/vscode-mssql/commit/c109e3288caa421f6ea421f376d9c314e04ff115). [](commit_id = [c109e32](https://github.com/microsoft/vscode-mssql/commit/c109e3288caa421f6ea421f376d9c314e04ff115), deletion_comment = False)

It's an extension constant, don't think it needs to.

abist:
> Looks fine logic-wise but it'd be great if you could be cleaning up all this old Promise-style async code as you're fixing these issues. There's a lot of unnecessary cruft and it makes it a lot more annoying to follow the logic vs async/await.
> 
> Also in case you haven't noticed the appveyor builds have been failing intermittenly quite a bit recently due to hitting the 1 hour limit our account has for the build. I would suggest looking into improving the build time at some point (looking into the build caching stuff Anythony just added to ADS seems like a good place to start) since not having PR validation a lot of the time is definitely not a great state to be in.

I would want to move to async/await as well, but a lot of the logic already written is in Thenables and I don't want to touch them without breaking stuff since I'm pretty much the only one testing the extension. I want to own the extension as a committer and work on the steps you mentioned above. However, I'm reluctant to make huge changes since it has side effects on all features.

Charles-Gagnon:
Shoot, sorry about the localize comment. Forgot that GitHub does that when you select lines that are outside the displayed sections in the PR. 

Anyways that's referring to https://github.com/microsoft/vscode-mssql/blob/c109e3288caa421f6ea421f376d9c314e04ff115/src/views/statusView.ts#L214

PR #1313, opened 2019-09-24
add script select for views too

abist:


abist:
@Charles-Gagnon @kburtram Will merge later this afternoon if I don't hear anything because I want to test all features in conjunction as soon as I can.

Issue #1324, opened 2019-09-24
Contents not redrawn when switching between multiple results windows


kenvanhyning:
Testing for September release on Windows x64 pkg.

Steps to Reproduce:

1.  Create new query and attach to server 1, type a simple query
2.  Execute query
3. Create another new query and attach to server 2, type another simple query
4. Execute second query

UI will have 2 results windows.  Results from #4 are visible on top of #2 

5. Click on the tab for #2 
6. UI displays: 
![image](https://user-images.githubusercontent.com/7833721/65553343-84e03100-dedb-11e9-8dde-8a49b35fcba6.png)

No results are ever redrawn.

Console in developer tools has this error:
mainThreadExtensionService.ts:65 SyntaxError: Unexpected token r in JSON at position 0
	at JSON.parse (<anonymous>)
	at MessageProxy.onReceive (c:\Users\kenvh\.vscode\extensions\ms-mssql.mssql-1.7.0\out\src\protocol.js:69:30)
	at constructor.self.protocol.onMessage.message.disposables.push.self.protocol.onMessage.val (c:\Users\kenvh\.vscode\extensions\ms-mssql.mssql-1.7.0\out\src\protocol.js:41:79)
	at u.fire (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:50:207)
	at d.$onMessage (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:611:836)
	at d._doInvokeHandler (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:623:396)
	at d._invokeHandler (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:623:88)
	at d._receiveRequest (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:621:783)
	at d._receiveOneMessage (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:620:605)
	at define.constructor._protocol.onMessage.e (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:618:791)
[...]

PR #1330, opened 2019-09-25
retain results when hidden

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1324

Instead of having the logic for refreshing the views, retain the webview content.

### DIFF
--- a/src/controllers/webviewController.ts
+++ b/src/controllers/webviewController.ts
@@ -34,40 +34,16 @@ export class WebviewPanelController implements vscode.Disposable {
         uri: string,
         title: string,
         serverProxy: IServerProxy,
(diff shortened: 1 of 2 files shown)


Charles-Gagnon:
>         const proxy: IServerProxy = {

You should be able to remove queryRunner from this as a param too - it doesn't appear to be used anywhere else here

---
Refers to: src/models/sqlOutputContentProvider.ts:143 in 0ad1177. [](commit_id = 0ad117722868bbebfd5636f53a2b4c4a2257946e, deletion_comment = False)

abist:
Closing this as we won't be going this route most likely.

Issue #1326, opened 2019-09-24
Switching between results tabs breaks view


Charles-Gagnon:
1. Open two SQL editors and execute queries in both of them
2. Results appear as expected
3. Switch the results tab to the currently non-selected one

Instead of showing the results for the other tab it displays a blank page for a couple of seconds and then shows a loading spinner indefinitely

![image](https://user-images.githubusercontent.com/28519865/65553818-970e9f00-dedc-11e9-973c-371c8b7e07d2.png)

Issue #1321, opened 2019-09-24
Moving results window breaks it


Charles-Gagnon:
1. Execute query
2. Move results window over to same side of editor as original query

It displays as blank for a while and then eventually just displays an **Executing query...** spinner endlessly.

![image](https://user-images.githubusercontent.com/28519865/65552868-6463a700-deda-11e9-895f-699fb63b0c30.png)

Trying to re-run the query does nothing either (no new results window is opened up). Closing the results editor tab and re-running query results in new results window being opened up.

PR #1331, opened 2019-09-25
Fixed tab switching without retaining context

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1326 and  https://github.com/microsoft/vscode-mssql/issues/1324 and 
https://github.com/microsoft/vscode-mssql/issues/1321

without using the "retainContextWhenHidden" property 

The issue was that sometimes the `WebviewPanel` changed state event occurred before the angular app was loaded and hence the query runner events weren't being listened to, by angular, to show the results on the canvas.

Fixed the issue by sending a ready event from angular when `ngOnInit()` is called. When a `WebviewPanel` is not active, the `WebviewPanel'`s `Webview` is disposed and thus the angular app is built every time. Thus, `ngOnInit()` will only be called the very first time or if the tab was switched (`WebviewPanel` is not active).

### DIFF
--- a/package.json
+++ b/package.json
@@ -261,6 +261,10 @@
         {
           "command": "extension.disconnectObjectExplorerNode",
           "when": "enablePreviewFeatures && viewItem != null"
+        },
+        {
+          "command": "extension.toggleSqlCmd",
+          "when": "editorFocus"
         }
       ]
     },
--- a/src/models/sqlOutputContentProvider.ts
+++ b/src/models/sqlOutputContentProvider.ts
@@ -299,7 +299,7 @@ export class SqlOutputContentProvider {
         if (panelController.isActive && queryRunner.hasCompleted) {
             return queryRunner.refreshQueryTab();
         }
-        return;
+        return false;
     }
 
     private setRunnerDeletionTimeout(uri: string): NodeJS.Timer {
--- a/src/models/sqlOutputContentProvider.ts
+++ b/src/models/sqlOutputContentProvider.ts
@@ -153,9 +153,10 @@ export class SqlOutputContentProvider {
                 this.saveResultsRequestHandler(uri, batchId, resultId, format, selection),
             setEditorSelection: (selection: ISelectionData) => this.editorSelectionRequestHandler(uri, selection),
             showError: (message: string) => this.showErrorRequestHandler(message),
-            showWarning: (message: string) => this.showWarningRequestHandler(message)
+            showWarning: (message: string) => this.showWarningRequestHandler(message),
+            sendReadyEvent: () => this.sendReadyEvent(uri)
         };
-        const controller = new WebviewPanelController(uri, title, proxy, this.context.extensionPath, queryRunner);
+        const controller = new WebviewPanelController(uri, title, proxy, this.context.extensionPath);
         this._panels.set(uri, controller);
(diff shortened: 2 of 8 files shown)


Charles-Gagnon:
What about https://github.com/microsoft/vscode-mssql/pull/1330? Were there issues with using that approach?

abist:
> What about #1330? Were there issues with using that approach?

It could possibly have some performance/memory overhead - 

https://code.visualstudio.com/api/extension-guides/webview#retaincontextwhenhidden

Charles-Gagnon:
So are you going with this approach then? If so please close the other PR so people know which one to review.

abist:
@Charles-Gagnon The plan is to discuss both approaches in the shiproom and see which way we want to go.

abist:
@Charles-Gagnon @kburtram @anthonydresser Will merge later this afternoon if I don't hear anything because I want to test all features in conjunction as soon as I can.

package.json
```diff
@@ -259,6 +259,10 @@
         {
           "command": "extension.disconnectObjectExplorerNode",
           "when": "viewItem != null"
+        },
+        {
+          "command": "extension.toggleSqlCmd",
+          "when": "editorFocus"
         }
```
Charles-Gagnon: Shouldn't this also only be available when the editor is connected? What happens currently if you do it for a disconnected action?

abist: It will prompt for a connection and run the query if not connected or just run

src/models/sqlOutputContentProvider.ts
```diff
@@ -283,6 +288,19 @@ export class SqlOutputContentProvider {
...
+     * Ready event sent by the angular app
+     * @param uri
+     */
+    private async sendReadyEvent(uri: string): Promise<boolean> {
+        // in case of a tab switch
+        const panelController = this._panels.get(uri);
+        const queryRunner = this.getQueryRunner(uri);
+        if (panelController.isActive && queryRunner.hasCompleted) {
+            return queryRunner.refreshQueryTab();
+        }
```
Charles-Gagnon: You're not returning anything if this check is false

package.json
Charles-Gagnon: Awesome. And it'll correctly be in SQLCMD mode for that query that's ran?

abist: Yes

src/models/sqlOutputContentProvider.ts
```diff
@@ -283,6 +288,20 @@ export class SqlOutputContentProvider {
...
+        return;
```
Charles-Gagnon: You should be returning false - this returns a boolean

src/protocol.ts
```diff
@@ -20,6 +20,7 @@ export interface IServerProxy {
     showWarning(message: string): void;
     showError(message: string): void;
     getLocalizedTexts(): Promise<{ [key: string]: any }>;
+    sendReadyEvent(uri: string);
```
Charles-Gagnon: return type for this

package.json
abist: @Charles-Gagnon Actually, the behavior now is even better. SQL CMD is a switch that can be on or off when editing a sql file, and that option is respected for that file until a connection is made/query is run.

Issue #1332, opened 2019-09-25
Executing a query does bring results window to foreground


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Open two SQL scripts, say script1.sql and script2.sql
2. Execute script1.sql
3. Execute script2.sql
4. Execute script1.sql

Expected: Results for script1 execution in step 4 are in the foreground

Actual: Results from scipt2 execution in step 3 are still in the foreground

Issue #1327, opened 2019-09-24
Executing query on non-connected editor connects but never executes


Charles-Gagnon:
1. Open non-connected editor
2. Attempt to run query
3. It prompts for a connection

After this it successfully connects but the query is not ran. Executing the query again does result in the query successfully running.

chlafreniere:
This and #1315 may be related

Issue #1319, opened 2019-09-24
SQLCMD Mode Status Item not visible


Charles-Gagnon:
1. Open existing SQL file
2. SQLCMD mode status item doesn't appear and it's not clear how to go about getting it

And then related - https://github.com/microsoft/vscode-mssql/issues/1317 makes it so even if you try to toggle the SQLCMD mode through the command it often just opens a new window. But that window also isn't set to SQLCMD mode and trying to set it results in the same behavior.

abist:
@Charles-Gagnon Can you confirm this issue or give a repro?

Charles-Gagnon:
I gave the repro steps, are they not clear enough? I'll see if I can still repro it later but I was hitting this consistently just playing around with SQLCMD mode. It would show up occasionally but at least half the time there would be no status item displayed.

abist:
I'll be waiting on your update here. If you still see this happening, let me know.

Charles-Gagnon:
Have you tried this yourself? Are you not seeing this behavior?

abist:
Of course. I don't see any of the behavior mentioned - no new window opening or sqlcmd not shown

abist:
@Charles-Gagnon Did you test before this PR was merged - https://github.com/microsoft/vscode-mssql/pull/1310

PR #1334, opened 2019-09-25
bring results of respective editor when query is run

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1332

Addresses https://github.com/microsoft/vscode-mssql/issues/1319 by showing the "Toggle SQLCMD Mode" only when editor has focus, and thus not being confusing any more. Then there are 3 cases -
1. Editor file is .sql and connected - command runs
2. Editor file is .sql and not connected - editor asks to connect to run command
3. Editor file is not .sql - show info message that a .sql file is required to run the command

Fixes https://github.com/microsoft/vscode-mssql/issues/1327

### DIFF
--- a/package.json
+++ b/package.json
@@ -261,6 +261,10 @@
         {
           "command": "extension.disconnectObjectExplorerNode",
           "when": "enablePreviewFeatures && viewItem != null"
+        },
+        {
+          "command": "extension.toggleSqlCmd",
+          "when": "editorFocus"
         }
       ]
     },
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -417,7 +417,7 @@ export default class MainController implements vscode.Disposable {
      */
     public onNewConnection(): Promise<boolean> {
         if (this.canRunCommand() && this.validateTextDocumentHasFocus()) {
-            this._connectionMgr.onNewConnection().then((result) => {
+            return this._connectionMgr.onNewConnection().then((result) => {
                 if (result) {
                     if (this._previewEnabled) {
                         this._objectExplorerProvider.objectExplorerExists = false;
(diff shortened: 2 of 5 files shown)


Charles-Gagnon:
>     private _isDisposed: boolean = false;

This is already a disposable - you should be calling this._register to register the disposables instead. Then adding your own dispose method is unnecessary

---
Refers to: src/controllers/webviewController.ts:30 in 3207fff. [](commit_id = 3207fffc1dc19d47c14dc9b8e9b715119c443cf8, deletion_comment = False)

Charles-Gagnon:
>     private _isDisposed: boolean = false;

Since you're using the isDisposed as well there's two options to solve that :

1. Override dispose() and add setting isDisposed but then still call super.dispose();
2. Find out where the controller is being disposed and remove it from the map at that point instead of waiting until the query callback is called (this seems like the more correct approach)

---
In reply to: [535534137](https://github.com/microsoft/vscode-mssql/pull/1334#issuecomment-535534137) [](ancestors = 535534137)

---
Refers to: src/controllers/webviewController.ts:30 in 3207fff. [](commit_id = 3207fffc1dc19d47c14dc9b8e9b715119c443cf8, deletion_comment = False)

abist:
@Charles-Gagnon This class is implementing it, not extending it. The only way the panel gets disposed is when it's closed. So, `super.dispose()` doesn't work as it's our class.

Charles-Gagnon:
Huh. Why are you implementing it instead of extending? Seems like we're just duplicating code for no real reason.

abist:
@Charles-Gagnon this is `vscode.Disposable` which is different from the `Disposable` class we use in ADS which is which a wrapper for this class - https://github.com/microsoft/azuredatastudio/blob/master/extensions/mssql/src/util/dispose.ts

Issue #1329, opened 2019-09-24
OE doesn't show the database name in the connection node


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Add a connection to a database
2. Add a connection to another database on the same server

Expected: Two nodes in OE for each connection, and the nodes are clearly distinguishable

Actual: The nodes don't have sufficient details to tell them apart.  Also, the DB connections behave like Server connections, having a child node Databases folder that shows all folders.  We should review the design here since this is different than what we do in Azure Data Studio.
   
![image](https://user-images.githubusercontent.com/599935/65556482-9f6ad800-dee4-11e9-8f09-b1dd6723ec98.png)

Charles-Gagnon:
Yup, was just going to file an issue on this. We should definitely do something like DB name + auth type like ADS has

![image](https://user-images.githubusercontent.com/28519865/65699747-63925880-e033-11e9-8946-569e3f9f3127.png)

Issue #1328, opened 2019-09-24
OE shows connections not in settings


kburtram:
- MSSQL Extension Version: 1.7
Steps to Reproduce:

1. Add a connection to a server
2. Connect to a database on the server, perhaps through an editor, but don't save it
3. Restart VS Code

Expected: Only 1 connection in OE

Actual: Two connections in OE.  One appears to be coming from MRU the other from settings.

![image](https://user-images.githubusercontent.com/599935/65556207-ae04bf80-dee3-11e9-994e-c7350b661347.png)

Issue #1314, opened 2019-09-24
Removing Connections behaves inconsistently


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Add Server 1
2. Add Server 2
3. Delete Server 2
4. Delete Server 1

Expected: No servers in viewlet or settings

Actual: Various bugs that result in mismatch between settings and viewlet.  The below screenshots show a couple examples.

*Case 1*
![image](https://user-images.githubusercontent.com/599935/65550308-311e1980-ded4-11e9-842b-32a9888b29d0.png)

*Case 2*
![image](https://user-images.githubusercontent.com/599935/65550405-590d7d00-ded4-11e9-9add-49f98aa38e9d.png)

Issue #1311, opened 2019-09-24
"Add Connection" button should skip "Choose connection" prompt


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Click the `Add Connection` button

Expected: Prompted to add a connection

Actual: Prompted to choose a connection or create a connection.  Since user already said "Create Connection" they should be taken directly to Servername prompt in the create workflow.

![image](https://user-images.githubusercontent.com/599935/65544618-6fadd700-dec8-11e9-81ba-5f37b28526bc.png)

PR #1337, opened 2019-09-26
OE enable/bugs

abist:
1. Enables OE
2. Fixes https://github.com/microsoft/vscode-mssql/issues/1329
3. Fixes https://github.com/microsoft/vscode-mssql/issues/1314
4. Fixes https://github.com/microsoft/vscode-mssql/issues/1311
5. Fixes https://github.com/microsoft/vscode-mssql/issues/1328
6. Fixes https://github.com/microsoft/vscode-mssql/issues/1307
7. https://github.com/microsoft/vscode-mssql/issues/1305 - The behavior now is that if you run a query from a different database from the server profile, it will make a new profile because it has a different database, and it shows up on OE as well. Will add another PR to avoid that behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,8 +170,7 @@
       "objectExplorer": [
         {
           "id": "objectExplorer",
-          "name": "%extension.connections%",
-          "when": "enablePreviewFeatures"
+          "name": "%extension.connections%"
         }
       ]
     },
@@ -211,56 +210,55 @@
         {
           "command": "extension.addObjectExplorer",
           "title": "%extension.addObjectExplorer%",
-          "group": "navigation",
-          "when": "enablePreviewFeatures"
+          "group": "navigation"
         }
       ],
       "view/item/context": [
         {
           "command": "extension.objectExplorerNewQuery",
-          "when": "enablePreviewFeatures && viewItem =~ /^(disconnectedServer|Server|Database|Table)$/",
+          "when": "viewItem =~ /^(disconnectedServer|Server|Database|Table)$/",
           "group": "MS_SQL@1"
         },
         {
           "command": "extension.removeObjectExplorerNode",
-          "when": "enablePreviewFeatures && viewItem =~ /^(disconnectedServer|Server)$/",
+          "when": "viewItem =~ /^(disconnectedServer|Server)$/",
           "group": "MS_SQL@3"
         },
         {
           "command": "extension.refreshObjectExplorerNode",
-          "when": "enablePreviewFeatures && view == objectExplorer",
+          "when": "view == objectExplorer",
           "group": "MS_SQL@2"
         },
         {
           "command": "extension.disconnectObjectExplorerNode",
-          "when": "enablePreviewFeatures && viewItem == Server",
+          "when": "viewItem == Server",
           "group": "MS_SQL@4"
         },
         {
           "command": "extension.scriptSelect",
-          "when": "enablePreviewFeatures && viewItem == Table"
+          "when": "viewItem == Table"
         }
       ],
       "commandPalette": [
         {
           "command": "extension.objectExplorerNewQuery",
-          "when": "enablePreviewFeatures && view == objectExplorer"
+          "when": "view == objectExplorer"
         },
         {
           "command": "extension.removeObjectExplorerNode",
-          "when": "enablePreviewFeatures && viewItem != null"
+          "when": "viewItem != null"
         },
         {
           "command": "extension.refreshObjectExplorerNode",
-          "when": "enablePreviewFeatures && viewItem != null"
+          "when": "viewItem != null"
         },
         {
           "command": "extension.scriptSelect",
-          "when": "enablePreviewFeatures && viewItem == Table"
+          "when": "viewItem == Table"
         },
         {
           "command": "extension.disconnectObjectExplorerNode",
-          "when": "enablePreviewFeatures && viewItem != null"
+          "when": "viewItem != null"
         }
       ]
     },
@@ -383,7 +381,7 @@
         "command": "extension.addObjectExplorer",
         "key": "ctrl+shift+o",
         "mac": "cmd+shift+o",
-        "when": "enablePreviewFeatures && view == objectExplorer"
+        "when": "view == objectExplorer"
       }
     ],
     "configuration": {
@@ -401,11 +399,6 @@
           "default": 5,
           "description": "%mssql.maxRecentConnections%",
           "scope": "window"
-        },
-        "mssql.enablePreviewFeatures": {
-          "type": "boolean",
-          "default": false,
-          "description": "%mssql.enablePreviewFeatures%"
         },
         "mssql.connections": {
           "type": "array",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -211,56 +210,55 @@
...
         {
           "command": "extension.objectExplorerNewQuery",
-          "when": "enablePreviewFeatures && view == objectExplorer"
+          "when": "view == objectExplorer"
         },
         {
           "command": "extension.removeObjectExplorerNode",
-          "when": "enablePreviewFeatures && viewItem != null"
+          "when": "viewItem != null"
         },
```
Charles-Gagnon: So I assume this is supposed to only show up when an OE item is selected? Did you verify that works? Although shouldn't this only be for server connection nodes? In general I would expect these when clauses to match the corresponding context contributions.

src/objectExplorer/objectExplorerProvider.ts
```diff
@@ -49,11 +49,11 @@ export class ObjectExplorerProvider implements vscode.TreeDataProvider<any> {
         return undefined;
     }
 
-    public removeObjectExplorerNode(node: TreeNodeInfo, isDisconnect: boolean = false): Promise<void> {
+    public async removeObjectExplorerNode(node: TreeNodeInfo, isDisconnect: boolean = false): Promise<void> {
```
Charles-Gagnon: >async [](start = 11, length = 5)

Why are you making these async? It's not necessary since removeObjectExplorerNode already returns a promise

src/objectExplorer/objectExplorerProvider.ts
```diff
@@ -49,11 +49,11 @@ export class ObjectExplorerProvider implements vscode.TreeDataProvider<any> {
...
         return this._objectExplorerService.removeObjectExplorerNode(node, isDisconnect);
     }
 
-    public refreshNode(node: TreeNodeInfo): Promise<void> {
+    public async refreshNode(node: TreeNodeInfo): Promise<void> {
         return this._objectExplorerService.refreshNode(node);
```
Charles-Gagnon: So this is calling this._objectExplorerService.refreshNode which then calls return this._objectExplorerProvider.refresh(node)...that seems a bit unnecessary don't you think? Why aren't callers just calling _objectExplorerService.refreshNode themselves directly instead of getting routed through here first?

Charles-Gagnon:
Possible issue I noticed while looking at this - on https://github.com/microsoft/vscode-mssql/blob/3f58659e1b2337a68a4f8b9dfff3127095ec54c1/src/objectExplorer/objectExplorerService.ts#L137 you're first firing off the expand request and only then adding the promise to the map. So there's a race condition here - if the expand response comes back first it could potentially get into the expand callback first and not have the promise in the map yet so fail to resolve correctly.

You should be adding it to the map first (and then correctly cleaning it up if an error occurs sending the expand request)

src/objectExplorer/objectExplorerService.ts
```diff
@@ -179,34 +182,34 @@ export class ObjectExplorerService {
...
-                            // clean expand session promise
-                            for (const key of this._expandParamsToPromiseMap.keys()) {
-                                if (key.sessionId === element.sessionId &&
-                                    key.nodePath === element.nodePath) {
-                                    this._expandParamsToPromiseMap.delete(key);
-                                }
+                    await this.expandNode(element, element.sessionId, promise);
+                    let children = await promise;
+                    if (children) {
+                        // clean expand session promise
```
Charles-Gagnon: Why are you only removing the promise if children is undefined? The promise is done at this point - there's nothing else awaiting on it - so there's no point keeping it around even if no children were returned

src/objectExplorer/objectExplorerService.ts
```diff
@@ -255,7 +259,7 @@ export class ObjectExplorerService {
                 // look up saved password
                 let password = await this._connectionManager.connectionStore.lookupPassword(connectionCredentials);
                 if (!password) {
-                    let password = await this._connectionManager.connectionUI.promptForPassword();
+                    password = await this._connectionManager.connectionUI.promptForPassword();
                     if (!password) {
```
Charles-Gagnon: I would strongly suggest enabling [https://palantir.github.io/tslint/rules/no-shadowed-variable/](https://palantir.github.io/tslint/rules/no-shadowed-variable/) so this doesn't happen

src/objectExplorer/objectExplorerService.ts
```diff
@@ -319,6 +317,20 @@ export class ObjectExplorerService {
         }
     }
 
+    private createNodeLabel(credentials: IConnectionCredentials): string {
+        const database = credentials.database;
+        const server = credentials.server;
+        if (database && database !== '') {
+            return `${server} (${database})`;
+        } else {
```
Charles-Gagnon: You should include auth type/username as well

src/views/connectionUI.ts
```diff
@@ -73,10 +73,15 @@ export class ConnectionUI {
 
     // Helper to let user choose a connection from a picklist
     // Return the ConnectionInfo for the user's choice
-    public showConnections(): Promise<IConnectionCredentials> {
+    public showConnections(fromObjectExplorer: boolean = false): Promise<IConnectionCredentials> {
         const self = this;
```
Charles-Gagnon: Why not just have the command from OE call self.createAndSaveProfile directly instead?

If you're really against that for some reason then I'd at least suggest making this name more generic - something like showExistingConnections = true

src/objectExplorer/objectExplorerService.ts
abist: We don't do that in ADS.

Charles-Gagnon: Yes we do

image

abist: oops

src/objectExplorer/objectExplorerProvider.ts
abist: Because you need to call it via the tree provider for the tree view to refresh

src/objectExplorer/objectExplorerProvider.ts
abist: its cleaner to use async/await

Charles-Gagnon: But you're not awaiting anything - so this is adding nothing. Returning a Promise happens regardless and making it async or not doesn't change that. Anyways it's not a huge deal since it just doesn't do anything - but you should really consider cleaning up pointless stuff like this unless it's needed since it just adds extra overhead.

abist: https://github.com/microsoft/vscode-mssql/blob/master/src/controllers/mainController.ts#L176

I need to await here - https://github.com/microsoft/vscode-mssql/blob/master/src/controllers/mainController.ts#L176

I don't want to have objectExplorerService to be exposed directly to the controller, so I call it via the provider. Please look at the usage of code before calling it useless.

Charles-Gagnon: You can await a function that returns a promise. It being asyc doesn't affect that. If your method signature was

```
public removeObjectExplorerNode(node: TreeNodeInfo, isDisconnect: boolean = false): Promise<void> {
```

it would be exactly the same logic-wise as it is now.

Issue #1305, opened 2019-09-23
Right click “new query” doesn’t set the connection to the selected database


pensivebrian:
- MSSQL Extension Version: "\\scratch2\scratch\adbist\vscode-mssql\mssql-1.7.0-win7-x64.vsix"
- VSCode Version: 1.38.0
- OS Version: Windows Server 2016

Steps to Reproduce:

1. In object explorer, connect to a server with two database: database1 and database 2
2. Expand database1
2. Right click on database2, and select “new query”
3. Run SELECT DB_Name()

Expected: database2 output
Actual: database1 output

pensivebrian:
Hmm... I can't repro this anymore. I'll see if I can distill the repro a bit more.

pensivebrian:
Okay, my guess is, if you click around in database2 before and async action completes for database1 completes, the connection context for database1 is set for database 2, so any right click operation is not running against the wrong database.

PR #1338, opened 2019-09-27
Bug fixes

abist:
1. Fixes Object Explorer's `New Query` context menu action. Fixes https://github.com/microsoft/vscode-mssql/issues/1305
It already used to connect to another database, but now a node wouldn't show up on OE even if the database name is different.

2. Add back script select for Views that got removed and rearranged the order to match ADS.
3. Remove `Refresh` option from disconnected nodes and match the command palette when condition to match the context menu's.
4. Added icons that show whether a server is connected (green) or not (red).


abist:
@Charles-Gagnon @kburtram @pensivebrian Will merge this PR EOD if I don't hear anything.

Charles-Gagnon:
@abist FYI it would be great if you could update the title to be more descriptive - that's going to be the commit message and so having a commit be "bug fixes" doesn't really help if you're looking through history.

abist:
@Charles-Gagnon that's the title of the PR not the commit. I have listed the bug fixes in the details with the commits.

src/controllers/mainController.ts
```diff
@@ -191,9 +188,9 @@ export default class MainController implements vscode.Disposable {
                         Constants.cmdScriptSelect, async (node: TreeNodeInfo) => {
                     const uri = await this._untitledSqlDocumentService.newQuery();
                     if (!this.connectionManager.isConnected(uri.toString())) {
-                        const connectionCreds = node.connectionCredentials;
+                        let connectionCreds = node.connectionCredentials;
                         const databaseName = `${escapeCharacters(self.getDatabaseName(node))}`;
```
Charles-Gagnon: Why did you change this? const just means you can't change what connectionCreds is pointing to - you're still free to modify the properties of a const.

src/controllers/mainController.ts
```diff
@@ -160,14 +160,11 @@ export default class MainController implements vscode.Disposable {
                     vscode.commands.registerCommand(
                         Constants.cmdObjectExplorerNewQuery, async (treeNodeInfo: TreeNodeInfo) => {
                     const connectionCredentials = treeNodeInfo.connectionCredentials;
-                    if (connectionCredentials) {
-                        const databaseName = `${escapeCharacters(self.getDatabaseName(treeNodeInfo))}`;
-                        if (databaseName !== connectionCredentials.database) {
-                            connectionCredentials.database = databaseName;
-                        }
+                    const databaseName = `${escapeCharacters(self.getDatabaseName(treeNodeInfo))}`;
+                    if (databaseName !== connectionCredentials.database) {
```
Charles-Gagnon: So this is actually a bug here - we shouldn't be escaping the characters (I'll file an issue separately)

src/controllers/mainController.ts
```diff
@@ -202,6 +199,7 @@ export default class MainController implements vscode.Disposable {
                         editor.edit(editBuilder => {
                             editBuilder.replace(editor.selection, selectStatement);
                         }).then(() => this.onRunQuery());
+                        return this.connectionManager.connectionStore.removeRecentlyUsed(<IConnectionProfile>connectionCreds);
                     }
```
Charles-Gagnon: So you're removing it because you don't want it to show up in the MRU? Why not just add a flag to the call to connect that tells it not to save it in the first place? That seems a lot cleaner. Right now you might even have the weird situation where it appears briefly in the MRU list until this call here is actually made.

Charles-Gagnon:
>             }

nit: but this could just be simplified to 

```
picklist = showExistingConnections ? self._connectionStore.getPickListItems() : [];
```

---
Refers to: src/views/connectionUI.ts:84 in bba864c. [](commit_id = bba864c1559b51b54c4fdaeeae2f8b293613c6f9, deletion_comment = False)

Charles-Gagnon:
> > ```
> >         }
> > ```
> 
> nit: but this could just be simplified to
> 
> ```
> picklist = showExistingConnections ? self._connectionStore.getPickListItems() : [];
> ```
> 
> Refers to: src/views/connectionUI.ts:84 in [bba864c](https://github.com/microsoft/vscode-mssql/commit/bba864c1559b51b54c4fdaeeae2f8b293613c6f9). [](commit_id = [bba864c](https://github.com/microsoft/vscode-mssql/commit/bba864c1559b51b54c4fdaeeae2f8b293613c6f9), deletion_comment = False)

It did it again....this is referencing https://github.com/microsoft/vscode-mssql/pull/1338/files#diff-ccc3cfd5f3eb67d20ae81c28d59adf0dR80-R83

src/controllers/mainController.ts
abist: I don't want to change the existing `User Settings` and `MRu` logic as it can be prone to many side effects. Also, we will never have that issue because the object explorer updates the connections when refreshed.

Charles-Gagnon:
> @Charles-Gagnon @kburtram @pensivebrian Will merge this PR EOD if I don't hear anything.



> @Charles-Gagnon that's the title of the PR not the commit. I have listed the bug fixes in the details with the commits. It doesn't make sense to put all the bug fixes in the title

Yes, but when you squash merge it'll take that and make it the title of the commit. So if you were planning on updating that when you merged then great - but regardless in general it's helpful to make the title descriptive since it's the first thing people see and having to go into the PR to see what's going on is a bit of an annoyance.

src/controllers/mainController.ts
kburtram: It seems like a "code smell" to have code that does something then has to call another function to partially undo what was just previously done.  If we have code we're afraid to change due to potential side-effects that seems like a problem.  Presumably some refactoring and additional tests would be a prudent mitigation?

kburtram:
@abist please add unrelated changes to new PRs.  Adding new commits (https://github.com/microsoft/vscode-mssql/pull/1338/commits/e4be6218979d27e48f96d4fc70b301d295fe0f1a) make it difficult to review as the intent of the PR keeps changing.

abist:
> @abist please add unrelated changes to new PRs. Adding new commits ([e4be621](https://github.com/microsoft/vscode-mssql/commit/e4be6218979d27e48f96d4fc70b301d295fe0f1a)) make it difficult to review as the intent of the PR keeps changing.

I didn't know you started reviewing it. I'll add the commit in a new PR.

kburtram:
@abist thanks!

Issue #1347, opened 2019-09-28
Object Explorer sort order seems arbitrary


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Add several servers to OE

Expected: Servers are in some predictable sort order.  I think either alphabetical or matching the order in settings could make sense.

Actual: Servers are in a seemingly arbitrary order.

![image](https://user-images.githubusercontent.com/599935/65809320-36d46300-e151-11e9-991d-537975b7183e.png)

kburtram:
After restarting vscode the order changed in a seemly "random" order.

![image](https://user-images.githubusercontent.com/599935/65809359-8b77de00-e151-11e9-9773-89ccea006f87.png)

PR #1355, opened 2019-09-30
Sort server names

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1347

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -169,6 +169,48 @@ export class ObjectExplorerService {
         }
     }
 
+    /**
+     * Sort the array based on server names
+     * @param array array that needs to be sorted
+     */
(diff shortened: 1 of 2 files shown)


src/objectExplorer/objectExplorerService.ts
```diff
@@ -235,10 +235,16 @@ export class ObjectExplorerService {
                     this._rootTreeNodeArray.push(node);
                 });
                 this._objectExplorerProvider.objectExplorerExists = true;
-                return this._rootTreeNodeArray;
+                const sortedNodeArray = this._rootTreeNodeArray.sort((a, b) => {
```
kburtram: is it possible to refactor this method into a some smaller more easily digestible pieces?  A good rule of thumb is that methods should be less than ~20 lines of code.

also, could you please add a test for this scenario?

Issue #1348, opened 2019-09-28
Connection password saved with "savePassword" is false


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Add a connection profile
2. Select "No" when asked to save password
3. Restart vs code
4. Expand save connection node in OE

Expected: Prompted to enter password

Actual: Node expand without prompting for password

![image](https://user-images.githubusercontent.com/599935/65809554-31781800-e153-11e9-8ef4-15b3ed9dc63b.png)

kburtram:
This is still repro'ing for me on the latest build from master.

abist:
Could you try to clear any cached extensions in `\.vscode\extensions`? I cannot repro this.

Issue #1343, opened 2019-09-27
connection in OE does not reprompt for password if wrong password typed in


kenvanhyning:
- MSSQL Extension Version: 1.7 (9/27 build)
- VSCode Version: 1.38.1
- OS Version: Windows 10

Setup:
- have a saved connection with sql auth but no p/w saved

Repro
1.  Launch VS Code, switch to OE and click on connection
2.  Mistype the password for that server
3. Error message is displayed
4. Try to connect again by clicking on the connection in OE

Results:
Node state appears to think it is connected (chevron switches between expanded collapsed and right-click menu is displayed. 

Expected:
Node should remain in the "unconnected state" such that a subsequent click attempts to connect again.

PR #1356, opened 2019-09-30
Fix saved password bug

abist:
Fixed a bug in MRU where it would not respect the `savePassword` option. Also removed password from node is `savePassword` is false, so a user will be prompted again if disconnected. 

Fixes https://github.com/microsoft/vscode-mssql/issues/1348 and https://github.com/microsoft/vscode-mssql/issues/1343


src/objectExplorer/objectExplorerService.ts
```diff
@@ -302,6 +302,10 @@ export class ObjectExplorerService {
             const connectionDetails = ConnectionCredentials.createConnectionDetails(connectionCredentials);
             const response = await this._connectionManager.client.sendRequest(CreateSessionRequest.type, connectionDetails);
             if (response) {
+                // remove password from node if saved password is false
+                if (!(<IConnectionProfile>connectionCredentials).savePassword) {
+                    connectionCredentials.password = '';
```
Charles-Gagnon: This shouldn't be necessary - at the top of this function the shouldPromptForPassword should handle the logic correctly to prompt if disconnected/reconnected. Is that not happening?

src/models/connectionStore.ts
```diff
@@ -158,11 +158,9 @@ export class ConnectionStore {
      * @param connectionCredentials Connection credentials of profile for password lookup
      */
     public async lookupPassword(connectionCredentials: IConnectionCredentials): Promise<string> {
-        const databaseName = connectionCredentials.database === '' ? 'master' :
-            connectionCredentials.database;
         const credentialId = ConnectionStore.formatCredentialId(
-            connectionCredentials.server, databaseName,
-            connectionCredentials.user, ConnectionStore.CRED_MRU_USER);
+            connectionCredentials.server, connectionCredentials.database,
+            connectionCredentials.user, ConnectionStore.CRED_PROFILE_USER);
```
Charles-Gagnon: >CRED_PROFILE_USER [](start = 56, length = 17)

So it looks like we're saving the password in either the MRU or PROFILE stores - so shouldn't the lookup be looking in both possibly? Although I guess I'd wonder why we'd even need to store the password from the MRU separately - aren't all connections added to the profile anyways?

src/objectExplorer/objectExplorerService.ts
abist: This logic doesn't have to do anything with `shouldPromptForPassword`. This is to respect the `savePassword` option for the nodes, while getting the password from the prompt. With this, we will have the password per session.

src/models/connectionStore.ts
abist: I'm not sure what the design was when this was written, but the MRU saved every single connection made to it and didn't respect `savePassword`. I fixed that bug now, and get the password from User Profile, if it were saved.

src/objectExplorer/objectExplorerService.ts
Charles-Gagnon: I don't get why this is necessary though. save password would be more referring to saving it in the connection store in my eyes. Whether the node has the password or not really shouldn't matter. Is there a bug you're fixing here?

---
In reply to: [329843754](https://github.com/microsoft/vscode-mssql/pull/1356#discussion_r329843754) [](ancestors = 329843754)

abist: The `savePassword` option will be respected when the connection is made using that profile. The handleConnectionSuccessNotificationHandler saves any password if `savePassword` was set to true. 

However, when since each node has a connectionCredential, we don't want to populate its password field if `savePassword` is true, but rather look up the credential store for a saved password instead.

Charles-Gagnon: Well - I would expect that when we made the connection we'd do the lookup for the password then and if it wasn't saved to the credentialStore then it'd prompt the user at that point. So it really shouldn't matter.

I guess in the end this is fine but again - why are you making this change? I don't really see a problem with it or anything but I'm trying to figure out if you're fixing a bug here. 

And then in addition - why would it be necessary to also clear it in the disconnect logic? Is there ever a scenario where the password would be still there if you're stripping it out when you connect here?

abist: The line you quoted is just solving the issue you asked about. We need to keep node.connectionCredentials different from connectionCredentials, because we don't want to save that password on the node but to fetch it from credential store.

abist: I removed this as it was redundant, you were right. Didn't really need it.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -212,6 +212,15 @@ export class ObjectExplorerService {
         }
     }
 
+    /**
+     * Helper to show the Add Connection node
+     */
+    private getAddConnectionNode(): AddConnectionTreeNode[] {
+        this._rootTreeNodeArray = [];
+        this._objectExplorerProvider.objectExplorerExists = true;
```
Charles-Gagnon: Few non-blocking things :

- Why is this separated out? It's only called here so having a separate function for something this simple doesn't seem super useful
- The function name doesn't really describe what's happening - in particular setting oeExists = true and clearing the cached node array
- Why return this as an array? Seems clearer to just return the single node

abist: It's separated out because it's being used twice. Don't want reduntant code in the `getChildren` function. Will change the name.

Issue #1345, opened 2019-09-28
Default database connections should be labeled <default>


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Add a connection with no database specified.

Expected: Connection profiles with no database specified should say `servername, <default>` similar to Azure Data Studio.

Actual: Label says `servername, master`.  There is no reason to assume the default database is master since this is a per-user configurable setting.

![image](https://user-images.githubusercontent.com/599935/65809226-05a76300-e150-11e9-9259-bf789a067371.png)

PR #1358, opened 2019-09-30
Change default database name

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1345

### DIFF
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -45,7 +45,6 @@ export const azureDatabase = 'Azure';
 export const defaultPortNumber = 1433;
 export const sqlAuthentication = 'SqlLogin';
 export const defaultDatabase = 'master';
-export const defaultDatabaseName = '<default>';
 export const errorPasswordExpired = 18487;
 export const errorPasswordNeedsReset = 18488;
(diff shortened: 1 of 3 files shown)


src/constants/constants.ts
```diff
@@ -45,6 +45,7 @@ export const azureDatabase = 'Azure';
 export const defaultPortNumber = 1433;
 export const sqlAuthentication = 'SqlLogin';
 export const defaultDatabase = 'master';
+export const defaultDatabaseName = '<default>';
```
Charles-Gagnon: Shouldn't this be localized? It looks like we don't do this for ADS but I think that's wrong since this should just be a display value.

src/controllers/mainController.ts
```diff
@@ -351,7 +351,7 @@ export default class MainController implements vscode.Disposable {
             }
             node = node.parentNode;
         }
-        return Constants.defaultDatabase;
+        return LocalizedConstants.defaultDatabaseLabel;
     }
```
Charles-Gagnon: And you've verified that this doesn't break new query? It looks like that uses this function to update the connection so be careful to check that it's not going to try and set the DB to <default>.

Issue #1352, opened 2019-09-28
Removing a connection profile with "Remove Profile" command doesn't refresh OE


kburtram:
- MSSQL Extension Version:
- VSCode Version: 1.7

Steps to Reproduce:

1. Add a couple connection profiles
2. Use "Manage Connection Profiles" -> "Remove" command to remove one

Expected: Connection removed from OE when removed using this command.

Actual: It is not removed from OE.

![image](https://user-images.githubusercontent.com/599935/65809726-7309c280-e155-11e9-83bd-bc2e89f0a3dc.png)

Issue #1346, opened 2019-09-28
Adding connections manually in user settings does not updated OE


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Open user settings JSON
2. Manually add a `mssql.connection` and save settings

Expected: OE is updated to reflect current state of settings when the file is saved.

Actual: OE is not updated and is out-of-date with current user connection metadata.

PR #1359, opened 2019-10-01
Remove profile command now refreshes OE

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1352

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -121,7 +121,7 @@ export default class MainController implements vscode.Disposable {
                 this.registerCommand(Constants.cmdManageConnectionProfiles);
                 this._event.on(Constants.cmdRunCurrentStatement, () => { self.onRunCurrentStatement(); });
                 this.registerCommand(Constants.cmdRunCurrentStatement);
-                this._event.on(Constants.cmdManageConnectionProfiles, () => { self.runAndLogErrors(self.onManageProfiles(), 'onManageProfiles'); });
+                this._event.on(Constants.cmdManageConnectionProfiles, async () => { await self.onManageProfiles(); });
                 this.registerCommand(Constants.cmdChooseDatabase);
                 this._event.on(Constants.cmdChooseDatabase, () => { self.runAndLogErrors(self.onChooseDatabase(), 'onChooseDatabase') ; } );
                 this.registerCommand(Constants.cmdChooseLanguageFlavor);
@@ -398,12 +398,13 @@ export default class MainController implements vscode.Disposable {
     /**
      * Manage connection profiles (create, edit, remove).
      */
-    private onManageProfiles(): Promise<boolean> {
+    private async onManageProfiles(): Promise<void> {
         if (this.canRunCommand()) {
             Telemetry.sendTelemetryEvent('ManageProfiles');
-            return this._connectionMgr.onManageProfiles();
+            await this._connectionMgr.onManageProfiles();
+            return this._objectExplorerProvider.refresh(undefined);
         }
-        return Promise.resolve(false);
+        return Promise.resolve();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/controllers/mainController.ts
```diff
@@ -398,12 +398,13 @@ export default class MainController implements vscode.Disposable {
...
+    private async onManageProfiles(): Promise<void> {
         if (this.canRunCommand()) {
             Telemetry.sendTelemetryEvent('ManageProfiles');
-            return this._connectionMgr.onManageProfiles();
+            await this._connectionMgr.onManageProfiles();
+            return this._objectExplorerProvider.refresh(undefined);
         }
-        return Promise.resolve(false);
+        return Promise.resolve();
     }
```
Charles-Gagnon: Just remove this line - it's not needed with the method being async and a void return type

src/controllers/mainController.ts
```diff
@@ -398,12 +398,13 @@ export default class MainController implements vscode.Disposable {
...
      * Manage connection profiles (create, edit, remove).
      */
-    private onManageProfiles(): Promise<boolean> {
+    private async onManageProfiles(): Promise<void> {
         if (this.canRunCommand()) {
             Telemetry.sendTelemetryEvent('ManageProfiles');
-            return this._connectionMgr.onManageProfiles();
+            await this._connectionMgr.onManageProfiles();
+            return this._objectExplorerProvider.refresh(undefined);
         }
```
Charles-Gagnon: This can be removed as well. Since this function has a void return type the return would typically be on the next line after the call to refresh, but in this case it's not even necessary since there's nothing else that's going to execute so no need to early return.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -257,19 +257,24 @@ export class ObjectExplorerService {
...
             if ((!this._objectExplorerProvider.objectExplorerExists ||
-                savedConnections.length !== this._rootTreeNodeArray.length) &&
-                savedConnections.length > 0) {
-                this._rootTreeNodeArray = [];
-                this.getSavedConnections();
-                this._objectExplorerProvider.objectExplorerExists = true;
-                return this.sortByServerName(this._rootTreeNodeArray);
-            } else {
-                if (this._rootTreeNodeArray.length > 0) {
+                savedConnections.length !== this._rootTreeNodeArray.length)) {
```
Charles-Gagnon: >savedConnections.length !== this._rootTreeNodeArray.length [](start = 16, length = 58)

What about if there's the same number of saved connections but they were changed? You should probably be doing a deeper check here to verify that.

src/controllers/mainController.ts
abist: refresh returns a void Promise, and that's what I return here

Charles-Gagnon: No it doesn't - it returns just a void
https://github.com/microsoft/vscode-mssql/blob/master/src/objectExplorer/objectExplorerProvider.ts#L26

(the async handles wrapping it in the Promise for you)

While that's still technically fine the practice should be if your return type is void you only ever use `return;` It just makes the intent clearer and reduces chance of confusion.

abist: That's what I meant with the async, since it's wrapped in a promise, returning on the same line should be fine, but sure I'll add the return on the next line if you want.

src/objectExplorer/objectExplorerService.ts
abist: I will handle that when fixing this - https://github.com/microsoft/vscode-mssql/issues/1346
since editing connections or manually adding connections needs an event for workspace settings change

src/controllers/mainController.ts
Charles-Gagnon: Thanks. It's definitely a nit but we should really be consistent with return usage in all our code and this is the style we general use elsewhere.

Issue #1353, opened 2019-09-28
Object Explorer shows connections not in the settings file


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Add a connection to OE
2. Open a new query windows for the connection
3. Run "USE DATABASE" command from Command Palette to select a different DB
4. Restart VS Code

Expected: Only one node in OE

Actual: Two connection nodes in OE.  One from the settings and one apparently from MRU entry from the USE command.

![image](https://user-images.githubusercontent.com/599935/65809770-e6abcf80-e155-11e9-98d1-14ad2d1129e4.png)

PR #1361, opened 2019-10-01
Pick only user settings connections for OE and its features

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1353

Interact OE with only user settings connections and keep MRU away from OE.


abist:
@kburtram @Charles-Gagnon gentle ping :)

src/models/connectionStore.ts
```diff
@@ -391,11 +391,14 @@ export class ConnectionStore {
     }
 
     // Load connections from user preferences
-    public loadAllConnections(): IConnectionCredentialsQuickPickItem[] {
+    public loadAllConnections(addRecentConnections: boolean = true): IConnectionCredentialsQuickPickItem[] {
```
kburtram: it seems like all invocations of this method are calling with false.  Maybe we should make the default false?

kburtram:
@abist regarding the "ping", please keep in mind that the code review SLA we're targeting is ~24 hour turn around time.  Expecting a review within a couple hours of posting a PR may ending up being a bit much.  Though it doesn't hurt to ask I suppose 😄.

src/models/connectionStore.ts
abist: Only in the case of OE features. For queries, people still might use the MRU.

PR #1362, opened 2019-10-01
Refresh OE when user settings file is changed manually

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1346


abist:
@kburtram @Charles-Gagnon gentle ping :)

Issue #1368, opened 2019-10-02
Ctrl-A in grid doesn't select all


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Run a query
2. Run Command-A in the grid

Expected: Select all

Actual: Everything gets selected in the window, not the grid content

<img width="693" alt="Screen Shot 2019-10-02 at 12 20 14 PM" src="https://user-images.githubusercontent.com/599935/66074907-2c2e1b00-e50f-11e9-8f1a-b926bad4eb35.png">

PR #1369, opened 2019-10-02
Fix keyboard shortcuts and added support for select all inside grid

abist:
Fix https://github.com/microsoft/vscode-mssql/issues/1368

### DIFF
--- a/package.json
+++ b/package.json
@@ -593,7 +593,7 @@
             "event.copyWithHeaders": "",
             "event.copyHeaders": "",
             "event.maximizeGrid": "",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -593,7 +593,7 @@
             "event.copyWithHeaders": "",
             "event.copyHeaders": "",
             "event.maximizeGrid": "",
-            "event.selectAll": "",
+            "event.selectAll": "ctrl+A",
```
kburtram: will this be Command-A on mac?

PR #1371, opened 2019-10-02
Check for default db when making a new query

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1370

Don't change the database name if the database name is `<default>`

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -160,7 +160,8 @@ export default class MainController implements vscode.Disposable {
                         Constants.cmdObjectExplorerNewQuery, async (treeNodeInfo: TreeNodeInfo) => {
                     const connectionCredentials = treeNodeInfo.connectionCredentials;
                     const databaseName = self.getDatabaseName(treeNodeInfo);
-                    if (databaseName !== connectionCredentials.database) {
+                    if (databaseName !== connectionCredentials.database &&
+                        databaseName !== LocalizedConstants.defaultDatabaseLabel) {
                         connectionCredentials.database = databaseName;
                     }
                     await self.onNewQuery(treeNodeInfo);


src/controllers/mainController.ts
```diff
@@ -160,7 +160,8 @@ export default class MainController implements vscode.Disposable {
                         Constants.cmdObjectExplorerNewQuery, async (treeNodeInfo: TreeNodeInfo) => {
                     const connectionCredentials = treeNodeInfo.connectionCredentials;
                     const databaseName = self.getDatabaseName(treeNodeInfo);
-                    if (databaseName !== connectionCredentials.database) {
+                    if (databaseName !== connectionCredentials.database &&
+                        databaseName !== LocalizedConstants.defaultDatabaseLabel) {
```
Charles-Gagnon: What if the user has a database with the name `<default>`? We really shouldn't be doing string comparisons here - having a var that says whether a connection is using the default DB or not would be much safer.

abist: If the user has a database with the name `<default>` then it would still work because that's the database name we will get from `getDatabaseName`

Charles-Gagnon: @abist What is connectionCredentials.database at this point for a connection that's using the default DB?

The scenario I'm concerned about is the user has a DB named `<default>` and then makes a connection set to use the actual default database defined for the user (let's say it's `master` in this case). 

They right click on the `<default>` database node and select new query. The logic comes here and sets databaseName to `<default>` which is then not equal  to `connectionCredentials.database` but because you added the check here it never updates the connection DB name and so connects to the actual default database (`master` or whatever)

Just try these steps :

1. Create database named `<default>`
2. Connect to server with connection that's using default DB
3. Right click on `<default>` node and select `New Query`

Is the new connection correctly on the `<default>` database or is it something else?

abist: Just tried it, it connects to `master`

Issue #1350, opened 2019-09-28
Cannot read property edit error when SELECT 1000 several times


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Run SELECT 1000 on a bunch of tables quickly

Expected: Everything works

Actual: Often something goes wrong the the query doesn't actually execute and get results.  Sometimes I see the below error notification.

![image](https://user-images.githubusercontent.com/599935/65809617-10fc8d80-e154-11e9-8321-33fb25744ce5.png)

Issue #1367, opened 2019-10-02
SELECT 1000 sometimes doesn't generate script content or resultset


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Run `SELECT 1000` on several tables quickly in a row

Expected: All script files and resultsets are generated

Actual: Sometimes the text file is blank, other times there are not result windows, other times the result window has duplicate resultsets.  It seems like there are some race conditions running this action several times quickly in a row.

The below show the output from running `SELECT 1000` several times quickly in a row.  Note the blank editor and the fact that many script windows don't have result sets.

<img width="1056" alt="Screen Shot 2019-10-02 at 12 04 27 PM" src="https://user-images.githubusercontent.com/599935/66073969-0869d580-e50d-11e9-8e07-c4802364db76.png">

Issue #1366, opened 2019-10-02
Queries seem to show double resultsets sometimes


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Sometimes when I run `SELECT 1000` I see two resultset grids produced

Note: It doesn't consistently repro but I've seen a few times so far in basic testing

<img width="1060" alt="Screen Shot 2019-10-02 at 12 01 38 PM" src="https://user-images.githubusercontent.com/599935/66073753-8d082400-e50c-11e9-947a-301c89980efd.png">

PR #1372, opened 2019-10-02
No more multiple result set for queries

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1350 and https://github.com/microsoft/vscode-mssql/issues/1366 and https://github.com/microsoft/vscode-mssql/issues/1367

Return one promise for scripting action and hence repeatedly clicking on Select Top 1000 doesn't show multiple resultset for one editor (URI)

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -185,30 +185,25 @@ export default class MainController implements vscode.Disposable {
                 this._scriptingService = new ScriptingService(this._connectionMgr, this._vscodeWrapper);
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
-                        Constants.cmdScriptSelect, (node: TreeNodeInfo) => {
-                    return this._untitledSqlDocumentService.newQuery().then((uri) => {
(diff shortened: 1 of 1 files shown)


src/controllers/mainController.ts
```diff
@@ -185,22 +185,30 @@ export default class MainController implements vscode.Disposable {
...
-                            editBuilder.replace(editor.selection, selectStatement);
-                        }).then(() => this.onRunQuery());
-                        return this.connectionManager.connectionStore.removeRecentlyUsed(<IConnectionProfile>connectionCreds);
-                    }
+                        if (editor) {
+                            let connectionCreds = node.connectionCredentials;
+                            const databaseName = self.getDatabaseName(node);
+                            connectionCreds.database = databaseName;
+                            this._statusview.languageFlavorChanged(uri.toString(), Constants.mssqlProviderName);
+                            return this.connectionManager.connect(uri.toString(), connectionCreds).then(() => {
```
kburtram: I would recommend rewriting this code to use `await`.  I think it would be significantly cleaner.

PR #1373, opened 2019-10-02
Reprompt pass if wrong password

abist:
Reset the node password when there's a create session failure so clicking on `Sign In` re-prompts for password.

Fixes https://github.com/microsoft/vscode-mssql/issues/1343

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -19,7 +19,7 @@ import CodeAdapter from '../prompts/adapter';
 import Telemetry from '../models/telemetry';
 import VscodeWrapper from './vscodeWrapper';
 import UntitledSqlDocumentService from './untitledSqlDocumentService';
-import { ISelectionData, IConnectionProfile } from './../models/interfaces';
+import { ISelectionData, IConnectionProfile, IConnectionCredentials } from './../models/interfaces';
 import * as path from 'path';
 import fs = require('fs');
 import { ObjectExplorerProvider } from '../objectExplorer/objectExplorerProvider';
@@ -205,9 +205,14 @@ export default class MainController implements vscode.Disposable {
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
                         Constants.cmdObjectExplorerNodeSignIn, async (node: AccountSignInTreeNode) => {
-                    this._objectExplorerProvider.signInNodeServer(node.parentNode);
-                    const profile = <IConnectionProfile>node.parentNode.connectionCredentials;
-                    return this.connectionManager.connectionUI.promptForRetryCreateProfile(profile);
+                    let profile = <IConnectionProfile>node.parentNode.connectionCredentials;
+                    profile = await self.connectionManager.connectionUI.promptForRetryCreateProfile(profile);
+                    if (profile) {
+                        node.parentNode.connectionCredentials = <IConnectionCredentials>profile;
+                        self._objectExplorerProvider.updateNode(node.parentNode);
+                        self._objectExplorerProvider.signInNodeServer(node.parentNode);
+                        return self._objectExplorerProvider.refresh(undefined);
+                    }
                 }));
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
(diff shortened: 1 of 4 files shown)


src/objectExplorer/objectExplorerService.ts
```diff
@@ -89,7 +89,10 @@ export class ObjectExplorerService {
                 }
                 return promise.resolve(self._currentNode);
             } else {
-                // failure
+                // create session failure
+                if (self._currentNode.connectionCredentials.password) {
+                    self._currentNode.connectionCredentials.password = '';
+                }
```
Charles-Gagnon: What if the user had "save password" selected but the connection failed for some other reason? Wouldn't this reprompt them at that point? That doesn't seem like the correct behavior.

abist:
@Charles-Gagnon This error is for all connection errors. We need to reset password regardless of what specific error it was with the connection

Charles-Gagnon:
So the current behavior in SSMS/ADS is that if the connection fails it'll open the connection dialog with the password field filled out (assuming save password was checked). I'd say we should keep that behavior here - if the connection fails have it walk through the create connection flow with all the appropriate fields prefilled.

abist:
@Charles-Gagnon Now asks to recreate the profile like ADS.

src/controllers/mainController.ts
```diff
@@ -205,8 +205,14 @@ export default class MainController implements vscode.Disposable {
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
                         Constants.cmdObjectExplorerNodeSignIn, async (node: AccountSignInTreeNode) => {
-                    this._objectExplorerProvider.signInNodeServer(node.parentNode);
-                    return this._objectExplorerProvider.refresh(undefined);
+                    let profile = <IConnectionProfile>node.parentNode.connectionCredentials;
+                    profile = await self.connectionManager.connectionUI.promptForRetryCreateProfile(profile);
```
Charles-Gagnon: It looks like this is going to prompt every time? Wouldn't it be better to try and connect first and only if that fails then prompt for re-entering information?

abist: This is the command for the `Sign In` node. So it will only prompt if you click on it.

Charles-Gagnon: I'm saying that it'll prompt you to go through the flow every time you click on the button. But if we have enough information stored to connect (either windows auth or correct username/password) we shouldn't go through that flow and instead just directly connect.

abist: Well, we don't know what's wrong so we do exactly what ADS does. Retry the profile with all the things pre-populated. We can't break flows in the extension because it doesn't have a dialog like ADS.

Charles-Gagnon: This is the flow I'm imagining :

1. User clicks Sign In
2. We attempt to connect using the current credentials. If the connection is successful then great - we're done here (this could easily be the case if it was a transient error or something where the connection string was correct but something else happened)
3. Otherwise catch the error (and display it) then do the re-prompting flow like you have now

Does that make sense?

abist: The flow is exactly the same, except in the second part, we don't connect with the same credentials because we knew it was wrong the first time and hence got to the `Sign In` node. Otherwise, everything's the way you described.

Charles-Gagnon: But as you said we don't know why we failed to connect. (Well...we do. We could be passing the actual error code back from the connection error. But that's beyond what we need to do here I think). So we shouldn't assume that it's something wrong with the connection information. If it's a transient issue (network down or whatever) then trying to do a reconnect may just work and save the user having to click enter a whole bunch of times.

abist: @Charles-Gagnon Error specific UX can be an enhancement for the next iteration, but for now I think this works fine. Would like to get this in for this release.

Charles-Gagnon: Alright, although I don't think this should really be that much extra work and the experience as it is isn't great. But if you have other bugs to fix then just file an issue and get back to it later.

abist: I think the UX now is better than connecting to redundant incorrect credentials. Merging now and will look into better UX with PMs for later releases.

Issue #1365, opened 2019-10-02
Connections added in disconnected state


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Add a Connection to OE

Expected: Connection is added in a connected state

Actual: Connections are added in disconnected state

<img width="472" alt="Screen Shot 2019-10-02 at 11 54 16 AM" src="https://user-images.githubusercontent.com/599935/66073195-7d3c1000-e50b-11e9-9726-d651cbea4512.png">

kburtram:
This looks fixed in the 1.7 build.  Thanks!

PR #1374, opened 2019-10-03
Connect the node when adding connection

abist:
The nodes are in a connected state when added from a command (`Add Object Explorer`/ `New Connection`/ `New Query`)  etc. 

However, when a connection is added manually by editing the settings file, the node isn't automatically connected because a connection isn't made for it. It will be added in a disconnected state.

Fixes https://github.com/microsoft/vscode-mssql/issues/1365

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -160,7 +160,8 @@ export default class MainController implements vscode.Disposable {
                         Constants.cmdObjectExplorerNewQuery, async (treeNodeInfo: TreeNodeInfo) => {
                     const connectionCredentials = treeNodeInfo.connectionCredentials;
                     const databaseName = self.getDatabaseName(treeNodeInfo);
-                    if (databaseName !== connectionCredentials.database) {
+                    if (databaseName !== connectionCredentials.database &&
+                        databaseName !== LocalizedConstants.defaultDatabaseLabel) {
                         connectionCredentials.database = databaseName;
                     }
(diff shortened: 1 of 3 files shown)


src/controllers/mainController.ts
```diff
@@ -49,6 +49,7 @@ export default class MainController implements vscode.Disposable {
     private _untitledSqlDocumentService: UntitledSqlDocumentService;
     private _objectExplorerProvider: ObjectExplorerProvider;
     private _scriptingService: ScriptingService;
+    private _connectionFromCommand: boolean = false;
```
Charles-Gagnon: Having this be shared seemed dangerous - what if someone starts a connection that takes a while and then while it's spinning in the background tries to make a new connection? They'll each be overwriting this value  at different times.

abist: @Charles-Gagnon A node is only connected when a session is created for it. When connections are added, I have to create the session and then refresh the tree, otherwise we just get the cached disconnected nodes.

Charles-Gagnon: I'm not sure how that's relevant to my concern. Even if that's the case what's stopping two nodes from being connected at nearly the same time and having the same issue?

---
In reply to: [331256285](https://github.com/microsoft/vscode-mssql/pull/1374#discussion_r331256285) [](ancestors = 331256285)

abist:
Will add a new PR

PR #1375, opened 2019-10-04
Ask for password every time save password is false

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1348

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -303,18 +303,19 @@ export class ObjectExplorerService {
         }
         if (connectionCredentials) {
             // show password prompt if SQL Login and password isn't saved
-            const shouldPromptForPassword = ConnectionCredentials.shouldPromptForPassword(connectionCredentials);
-            if (shouldPromptForPassword) {
-                // look up saved password
-                let password = await this._connectionManager.connectionStore.lookupPassword(connectionCredentials);
+            let password = connectionCredentials.password;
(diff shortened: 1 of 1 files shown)


src/objectExplorer/objectExplorerService.ts
```diff
@@ -302,16 +302,19 @@ export class ObjectExplorerService {
...
-            // show password prompt if SQL Login and password isn't saved
-            const shouldPromptForPassword = ConnectionCredentials.shouldPromptForPassword(connectionCredentials);
-            if (shouldPromptForPassword) {
-                // look up saved password
-                let password = await this._connectionManager.connectionStore.lookupPassword(connectionCredentials);
-                if (!password) {
+            if (ConnectionCredentials.isPasswordBasedCredential(connectionCredentials)) {
+                // show password prompt if SQL Login and password isn't saved
+                let password = connectionCredentials.password;
+                // if password isn't saved
```
Charles-Gagnon: No need to set this since you always overwrite it in the below lines anyways

abist:
@Charles-Gagnon This issue occurs when a user makes a profile and saves a password, but then decides to clear App data cache. That deletes the user connection but not the password. I'll put a PR to clear out all the credentials for the connections that don't exist in the settings.

Charles-Gagnon:
Interesting, yeah that's definitely a pretty weird scenario. Glad you were able to track it down! We should definitely be making sure to clear the password from the store whenever we detect that a profile has changed or been removed.

PR #1382, opened 2019-10-10
Fix issue with multiple result sets

abist:
There were a bunch of race conditions in the existing code, so fixed them while fixing this issue. Cleaned code to use `async`'s instead of `Thenables`. If you want me to put those changes in a different PR, let me know. It was confusing to read when fixing the race conditions so I changed it as a part of this PR.

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -641,7 +641,7 @@ export default class ConnectionManager {
     public connect(fileUri: string, connectionCreds: Interfaces.IConnectionCredentials, promise?: Deferred<boolean>): Promise<boolean> {
         const self = this;
 
-        return new Promise<boolean>((resolve, reject) => {
+        return new Promise<boolean>(async (resolve, reject) => {
             let connectionInfo: ConnectionInfo = new ConnectionInfo();
             connectionInfo.extensionTimer = new Utils.Timer();
             connectionInfo.intelliSenseTimer = new Utils.Timer();
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -400,69 +400,80 @@ export default class ConnectionManager {
     /**
      * Clear the recently used connections list in the connection store
      */
-    public async clearRecentConnectionsList(): Promise<void> {
-        await this.connectionStore.clearRecentlyUsed();
-        return;
+    public clearRecentConnectionsList(): Promise<void> {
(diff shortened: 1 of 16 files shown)


src/controllers/connectionManager.ts
```diff
@@ -99,6 +101,7 @@ export default class ConnectionManager {
         this._connections = {};
         this._connectionCredentialsToServerInfoMap =
             new Map<Interfaces.IConnectionCredentials, ConnectionContracts.ServerInfo>();
+        this._uriToConnectionPromiseMap = new Map<string, Deferred<boolean>>();
```
Charles-Gagnon: nit: but you could just create this above when declaring the var

src/controllers/untitledSqlDocumentService.ts
```diff
@@ -18,25 +18,16 @@ export default class UntitledSqlDocumentService {
...
+    public async newQuery(content?: string): Promise<vscode.TextEditor> {
+        // Open an untitled document. So the  file doesn't have to exist in disk
+        let doc = await this.vscodeWrapper.openMsSqlTextDocument(content);
+        // Show the new untitled document in the editor's first tab and change the focus to it.
+        const editor = await this.vscodeWrapper.showTextDocument(doc, 1, false);
+        const position = editor.selection.active;
+        let newPosition = position.with(position.line + 1, 0);
+        let newSelection = new vscode.Selection(newPosition, newPosition);
+        editor.selection = newSelection;
+        return Promise.resolve(editor);
```
Charles-Gagnon: You can just return the editor here, no need to wrap in the Promise.resolve

src/models/connectionStore.ts
```diff
@@ -279,18 +279,13 @@ export class ConnectionStore {
...
-                resolve(undefined);
-            }, err => {
-                reject(err);
-            });
-        });
+    public async clearRecentlyUsed(): Promise<void> {
+        // Update the MRU list to be empty
+        try {
+            await this._context.globalState.update(Constants.configRecentConnections, []);
+        } catch (error) {
```
Charles-Gagnon: The try/catch is unnecessary here with an async function- the returned promise will be rejected if an exception is thrown already.

abist:
@Charles-Gagnon We don't a want very big changes at the last minute, so I'm going to revert the `async`s to `Thenables` and have 2 PRs: one for the fix, and one cleaning stuff. I'm going to close this PR for now.

PR #1384, opened 2019-10-11
Fixes multiple results

abist:
Fixes multiple result tabs when scripting in rapid succession.


src/controllers/mainController.ts
```diff
@@ -185,26 +186,40 @@ export default class MainController implements vscode.Disposable {
                 this._scriptingService = new ScriptingService(this._connectionMgr, this._vscodeWrapper);
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
-                        Constants.cmdScriptSelect, async (node: TreeNodeInfo) => {
-                    let uri = await this._untitledSqlDocumentService.newQuery();
-                    let editor = this._vscodeWrapper.activeTextEditor;
-                    if (editor) {
+                    Constants.cmdScriptSelect, async (node: TreeNodeInfo) => {
+                    let actionPromise = new Promise<boolean>(async (resolve, reject) => {
+                        const nodeUri = ObjectExplorerUtils.getNodeUri(node);
```
Charles-Gagnon: This doesn't seem to be need to be wrapped in a Promise - the method is already async so you could just have the body of the promise called here directly

src/controllers/queryRunner.ts
```diff
@@ -137,30 +144,32 @@ export default class QueryRunner {
...
+                await this._client.sendRequest(QueryExecuteRequest.type, queryDetails).then(onSuccess, onError);
             });
     }
 
     // Pulls the query text from the current document/selection and initiates the query
-    private doRunQuery(selection: ISelectionData, queryCallback: any): Thenable<void> {
-        const self = this;
-        self._vscodeWrapper.logToOutputChannel(Utils.formatString(LocalizedConstants.msgStartedExecute, this._ownerUri));
+    private async doRunQuery(selection: ISelectionData, queryCallback: any): Promise<void> {
+        this._vscodeWrapper.logToOutputChannel(Utils.formatString(LocalizedConstants.msgStartedExecute, this._ownerUri));
```
Charles-Gagnon: nit - I know it was like this before but generally it's a good idea to type callbacks so you're sure that the function signature is as expected. So it'd look like this `queryCallback: (onSuccess: (result) => void, onError: (reason) => void) => void` (you might even be able to get the types for result and reason based on what sendRequest returns)

src/controllers/queryRunner.ts
```diff
@@ -280,42 +294,42 @@ export default class QueryRunner {
...
-            }, error => {
-                // TODO: Localize
-                self._vscodeWrapper.showErrorMessage('Something went wrong getting more rows: ' + error.message);
-                reject();
-            });
-        });
+        try {
+            const queryExecuteSubsetResult = await this._client.sendRequest(QueryExecuteSubsetRequest.type, queryDetails);
+            if (queryExecuteSubsetResult) {
+                return queryExecuteSubsetResult;
```
Charles-Gagnon: >queryExecuteSubsetResult [](start = 23, length = 24)

You're never returning anything if queryExecuteSubsetResult is falsy

src/controllers/queryRunner.ts
```diff
@@ -280,42 +294,42 @@ export default class QueryRunner {
...
+            }
+        } catch (error) {
+            // TODO: Localize
+            this._vscodeWrapper.showErrorMessage('Something went wrong getting more rows: ' + error.message);
```
Charles-Gagnon: >Something went wrong getting more rows [](start = 50, length = 38)

This error message isn't great, would suggest changing it to 'Unexpected error occurred retrieving more rows: ' + error.message.

src/controllers/queryRunner.ts
```diff
@@ -280,42 +294,42 @@ export default class QueryRunner {
...
+            Promise.reject(error);
+        }
```
Charles-Gagnon: Just rethrow the error here instead

`throw error`

src/controllers/queryRunner.ts
```diff
@@ -280,42 +294,42 @@ export default class QueryRunner {
...
-            });
-        });
+    public async dispose(): Promise<void> {
+        let disposeDetails = new QueryDisposeParams();
+        disposeDetails.ownerUri = this.uri;
+        try {
+            const result = await this._client.sendRequest(QueryDisposeRequest.type, disposeDetails);
+            if (result) {
+                return;
+            }
```
Charles-Gagnon: Unnecessary return here

src/controllers/queryRunner.ts
```diff
@@ -337,99 +351,88 @@ export default class QueryRunner {
...
+        await this._vscodeWrapper.clipboardWriteText(copyString);
+        if (process.platform === 'darwin') {
+            process.env['LANG'] = oldLang;
+        }
     }
 
 
-    public toggleSqlCmd(): Thenable<boolean> {
+    public async toggleSqlCmd(): Promise<boolean> {
         const queryExecuteOptions: QueryExecutionOptions = { options: new Map<string, any>() };
```
Charles-Gagnon: nit - but it's weird to have this return a bool when it'll never be able to return a false as it's currently written

src/controllers/untitledSqlDocumentService.ts
```diff
@@ -18,25 +18,16 @@ export default class UntitledSqlDocumentService {
...
+        // Open an untitled document. So the  file doesn't have to exist in disk
+        let doc = await this.vscodeWrapper.openMsSqlTextDocument(content);
+        // Show the new untitled document in the editor's first tab and change the focus to it.
+        const editor = await this.vscodeWrapper.showTextDocument(doc, 1, false);
+        const position = editor.selection.active;
+        let newPosition = position.with(position.line + 1, 0);
+        let newSelection = new vscode.Selection(newPosition, newPosition);
+        editor.selection = newSelection;
+        return Promise.resolve(editor);
     }
```
Charles-Gagnon: Can just return editor here

src/controllers/untitledSqlDocumentService.ts
```diff
@@ -18,25 +18,16 @@ export default class UntitledSqlDocumentService {
...
-            }
-        });
+    public async newQuery(content?: string): Promise<vscode.TextEditor> {
+        // Open an untitled document. So the  file doesn't have to exist in disk
+        let doc = await this.vscodeWrapper.openMsSqlTextDocument(content);
+        // Show the new untitled document in the editor's first tab and change the focus to it.
+        const editor = await this.vscodeWrapper.showTextDocument(doc, 1, false);
+        const position = editor.selection.active;
+        let newPosition = position.with(position.line + 1, 0);
+        let newSelection = new vscode.Selection(newPosition, newPosition);
```
Charles-Gagnon: I see lots of +1's around in this code but no explanation of why. Would be good to try and add comments since that's not something I'd consider normal and it's not clear why they're needed

src/controllers/webviewController.ts
```diff
@@ -28,16 +29,18 @@ export class WebviewPanelController implements vscode.Disposable {
     private _disposables: vscode.Disposable[] = [];
     private _isDisposed: boolean = false;
     private _isActive: boolean;
+    private _rendered: boolean = false;
 
     constructor(
+        private _vscodeWrapper: VscodeWrapper,
         uri: string,
```
Charles-Gagnon: So if you're using the wrapper I'd suggest updating all the current calls to use the wrapper as well (such as vscode.workspace.getConfiguration on line 41)

src/models/connectionStore.ts
```diff
@@ -279,18 +279,13 @@ export class ConnectionStore {
...
-                reject(err);
-            });
-        });
+    public async clearRecentlyUsed(): Promise<void> {
+        // Update the MRU list to be empty
+        try {
+            await this._context.globalState.update(Constants.configRecentConnections, []);
+        } catch (error) {
+            Promise.reject(error);
+        }
```
Charles-Gagnon: try/catch is unnecessary, thrown exceptions from the call to update will cause the Promise to be rejected already

src/objectExplorer/objectExplorerService.ts
```diff
@@ -80,6 +80,14 @@ export class ObjectExplorerService {
                     self._currentNode = TreeNodeInfo.fromNodeInfo(result.rootNode, result.sessionId,
                         undefined, credentials, nodeLabel);
                 }
+                // make a connection if not connected already
+                const nodeUri = ObjectExplorerUtils.getNodeUri(self._currentNode);
+                if (!this._connectionManager.isConnected(nodeUri) &&
```
Charles-Gagnon: What scenario is this needed for? It seems odd to me that a session that was just created doesn't have a connection associated with it

Charles-Gagnon:
Although it does look like you still have a failing test so you should fix that before merging in. 

 1) MainController Tests
       onNewQuery should call the new query and new connection:
     TypeError: Cannot read property 'uri' of undefined
      at MainController.<anonymous> (C:\projects\vscode-mssql\out\src\controllers\src\controllers\mainController.ts:707:41)
      at Generator.next (<anonymous>)
      at fulfilled (C:\projects\vscode-mssql\out\src\controllers\mainController.js:8:58)

src/objectExplorer/objectExplorerService.ts
abist: The object explorer session is different from a normal connection. This can happen foe example if you run `New Query` from a disconnected node.

src/controllers/queryRunner.ts
abist: Will do it as a part of the cleanup.

Issue #1387, opened 2019-10-11
leaf node in OE


alanrenmsft:
all leaf nodes are showed as expandable

![Screen Shot 2019-10-11 at 11 16 42 AM](https://user-images.githubusercontent.com/13777222/66674788-a7cb4e80-ec18-11e9-810e-99dcc0ac75d1.png)

PR #1389, opened 2019-10-11
non expandable leaf nodes from toolsservice

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1387


alanrenmsft:
@abist still need the branch?

Issue #1391, opened 2019-10-11
Inconsistent experience for connect/sign in


alanrenmsft:
I have 2 connections disconnected, but one shows Connect and the other one shows Sign In

![Screen Shot 2019-10-11 at 12 28 23 PM](https://user-images.githubusercontent.com/13777222/66679329-cafafb80-ec22-11e9-9aef-070357c984bf.png)

abist:
@alanrenmsft This is by design. One's to connect back a disconnected node and the other one is to actually sign in into the profile (making a connection)

alanrenmsft:
I think we should consolidate them, from user perspective, it is the same.

abist:
There's a difference though. For example, if you have disconnected from a node that has saved password, it would directly connect, but ask for password if it didn't have it saved. The sign in node will always ask for a password because the connection doesn't exist.

alanrenmsft:
sure, I understand the technical difference, to the user, it is the same, I will be prompted for the password for both scenarios. @yualan what do u think

abist:
Sure, I can change the text to both being `Sign In` or `Connect`. Is there a preference?

yualan:
Both being Connect makes sense to me

PR #1396, opened 2019-10-11
Changed sign in and connect nodes to both connect

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1391


src/objectExplorer/accountSignInTreeNode.ts
```diff
@@ -14,10 +14,10 @@ export class AccountSignInTreeNode extends vscode.TreeItem {
     constructor(
         private _parentNode: TreeNodeInfo,
     ) {
-        super(LocalizedConstants.msgSignIn, vscode.TreeItemCollapsibleState.None);
```
alanrenmsft: do we still need the msgSignIn, if not, please delete it.

Issue #1404, opened 2019-10-11
Expanding OE after disconnecting briefly shows the previous information


aaomidi:
![8bk8ft3FRe](https://user-images.githubusercontent.com/1465785/66687874-318b1400-ec39-11e9-9e8d-c085e0e5a657.gif)

PR #1418, opened 2019-10-12
Fix issue with disconnect node

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1404


abist:
@amitmsft How so?

aaomidi:
Wrong @ btw, but since you're doing recursion you're going to be filling the stack pretty fast.

If each node has 10 children and you're going 3 layers deep that is a 1000 element long stack and JS will not work with that.

abist:
We handle thousands of files recursively when using gulp tasks and packaging but that's fine. I'll put put an iterative one in a moment,

aaomidi:
Put this in chrome and see what happens:

```
function fibonacci(num) {
  if (num <= 1) return 1;

  return fibonacci(num - 1) + fibonacci(num - 2);
};

fibonacci(1000)
```

The recursive handling of files lead us to the EMFILE error yesterday. It's bad practice.

Issue #1399, opened 2019-10-11
SELECT TOP 1000 Results in Incorrect Query for System Versioned Tables


chlafreniere:
I right clicked on a table, and chose "Select Top 1000". A query editor "Untitled-1" opened with no query in it. After a few seconds, the query populated in the editor, and soon thereafter the query failed with a message.

SELECT TOP (1000) *   FROM [<redacted>].[Application].[Cities (System-Versioned)]

"(System-Versioned)" should not be included in the query!

PR #1420, opened 2019-10-13
Add metadata service for correct object names in scripting

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1399

### DIFF
--- a/src/scripting/scriptingService.ts
+++ b/src/scripting/scriptingService.ts
@@ -46,7 +46,7 @@ export class ScriptingService {
     /**
      * Helper to get the object name and schema name
      */
-    private async getObjectFromNode(node: TreeNodeInfo, uri: string): Promise<ScriptingObject> {
+    private getObjectFromNode(node: TreeNodeInfo, uri: string): ScriptingObject {
         let metadata = node.metadata;
         let scriptingObject: ScriptingObject = {
             type: metadata.metadataTypeName,
@@ -57,7 +57,7 @@ export class ScriptingService {
     }
 
     public async scriptSelect(node: TreeNodeInfo, uri: string): Promise<string> {
-        const scriptingObject = await this.getObjectFromNode(node, uri);
+        const scriptingObject = this.getObjectFromNode(node, uri);
         let serverInfo = this._connectionManager.getServerInfo(node.connectionCredentials);
         let scriptOptions: ScriptOptions = {
             scriptCreateDrop: 'ScriptSelect',
--- a/src/scripting/scriptingService.ts
+++ b/src/scripting/scriptingService.ts
@@ -8,24 +8,16 @@ import ConnectionManager from '../controllers/connectionManager';
 import { ScriptingRequest, ScriptingParams, ScriptOperation, ScriptingObject, ScriptOptions } from '../models/contracts/scripting/scriptingRequest';
 import { TreeNodeInfo } from '../objectExplorer/treeNodeInfo';
 import VscodeWrapper from '../controllers/vscodeWrapper';
-import { MetadataService } from '../metadata/metadataService';
-import { IConnectionCredentials } from '../models/interfaces';
-import { ObjectMetadata, MetadataType } from '../models/contracts/metadata/metadataRequest';
-import Utils = require('../models/utils');
 
 export class ScriptingService {
 
     private _client: SqlToolsServiceClient;
-    private _metdataService: MetadataService;
-    private _credentialsToMetadataMap: Map<IConnectionCredentials, ObjectMetadata[]>;
 
     constructor(
         private _connectionManager: ConnectionManager,
         private _vscodeWrapper: VscodeWrapper
     ) {
         this._client = this._connectionManager.client;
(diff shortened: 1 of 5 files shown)


src/scripting/scriptingService.ts
```diff
@@ -8,16 +8,24 @@ import ConnectionManager from '../controllers/connectionManager';
 import { ScriptingRequest, ScriptingParams, ScriptOperation, ScriptingObject, ScriptOptions } from '../models/contracts/scripting/scriptingRequest';
 import { TreeNodeInfo } from '../objectExplorer/treeNodeInfo';
 import VscodeWrapper from '../controllers/vscodeWrapper';
+import { MetadataService } from '../metadata/metadataService';
+import { IConnectionCredentials } from '../models/interfaces';
+import { ObjectMetadata, MetadataType } from '../models/contracts/metadata/metadataRequest';
+import Utils = require('../models/utils');
```
Charles-Gagnon: Any particular reason you're using require here? It's not wrong per-se but generally we should try to use import when we can.

src/scripting/scriptingService.ts
```diff
@@ -43,23 +51,62 @@ export class ScriptingService {
...
-     * @param node
      */
-    private getObjectNames(node: TreeNodeInfo): string[] {
-        let fullName = node.label;
-        let objects = fullName.split('.');
-        return objects;
+    private async getObjectFromNode(node: TreeNodeInfo, uri: string): Promise<ScriptingObject> {
+        const nodeCredentials = node.connectionCredentials;
+        let metadata: ObjectMetadata[];
+        for (let credential of this._credentialsToMetadataMap.keys()) {
```
Charles-Gagnon: nit: This is a bit simpler

`const metadata = Array.from(this._credentialsToMetadataMap.keys()).find(c => Utils.isSameConnection(c, nodeCredentials));`

src/scripting/scriptingService.ts
```diff
@@ -43,23 +51,62 @@ export class ScriptingService {
...
+                break;
+            }
+        }
+        if (!metadata) {
+            metadata = await this._metdataService.getMetadata(uri);
+            const newCredentials = Object.assign({}, nodeCredentials);
+            this._credentialsToMetadataMap.set(newCredentials, metadata);
+        }
+        for (const obj of metadata) {
+            const objectLabels = node.label.split('.');
```
Charles-Gagnon: This isn't going to work correctly if the schema/name has a . in its name

src/scripting/scriptingService.ts
```diff
@@ -43,23 +51,62 @@ export class ScriptingService {
...
     };
 
+    /**
+     * Helper to node name from label and metadata
+     */
+    private getObjectName(nameLabel: string, objectName: string, metadata: ObjectMetadata[]): string {
+        if (nameLabel === objectName) {
+            return nameLabel;
+        }
+        // if the names aren't the same, find the closest
```
Charles-Gagnon: I assume you're doing this because the labels are going to have other stuff like the `(System-Versioned)` in it?

Regardless I think there's a bug here if there's an object that has a name that matches some of the "extra" bit. For example if I have an object named `(System-Versioned)` and am searching for the metadata for a system-versioned node named `mydb` then the label will be `mydb (System-Versioned)` and it'll end up with the metadata for the `(System-Versioned)` object because the label contains that name and it's longer than the `mydb` name.

abist: Currently, there's no way to verify this because Object Explorer doesn't send the metadata with it. How do I go about checking this?

Charles-Gagnon:
Also this seems like something we're going to need to have more generally than just here. Is there anywhere else we're using the label and assuming it's the exact object name? ADS attaches the metadata to the node itself which seems like the correct thing to do here.

src/scripting/scriptingService.ts
Charles-Gagnon: By adding the metadata to the node itself as I suggested below seems like the correct way to me 😄

abist: This was explicitly removed from sqltoolsservice because it was making OE too slow, getting metadata with every node. That's the reason we had metadata service in the first place afaik

abist:
@Charles-Gagnon We get metadata for scriptable objects. I want to keep the metadata service for future, because it will be needed by other types of scripting operations

src/scripting/scriptingService.ts
```diff
@@ -45,21 +45,19 @@ export class ScriptingService {
...
     /**
      * Helper to get the object name and schema name
-     * @param node
      */
-    private getObjectNames(node: TreeNodeInfo): string[] {
-        let fullName = node.label;
-        let objects = fullName.split('.');
-        return objects;
+    private async getObjectFromNode(node: TreeNodeInfo, uri: string): Promise<ScriptingObject> {
+        let metadata = node.metadata;
```
Charles-Gagnon: Looks like this doesn't need to be async anymore

Charles-Gagnon:
(doing the run-through can be done after merging this in, no reason to tie it to this PR)

Issue #1414, opened 2019-10-11
Disconnected from server - new query


aaomidi:
![DP4enGXOxT](https://user-images.githubusercontent.com/1465785/66689869-b0844a80-ec41-11e9-913c-b693caf06300.gif)

If I disconnect from the server I shouldn't be able to open a new query

abist:
This is the same behavior as ADS. I'm putting a PR to make the server connected though.

Issue #1393, opened 2019-10-11
infinite password prompt loop when making connection with profile name


alanrenmsft:
1. connect to server with sql login and no profile name
2. connect to the same server with the same login with a profile name

expected:
a new connection established with the new profile name.

actual:
1. the existing connection is closed
2. I ran into a infinite password prompt loop

abist:
Tracking all password issues here, including https://github.com/microsoft/vscode-mssql/issues/1392

PR #1423, opened 2019-10-14
New added connections feature

abist:
Newly added connections now add in a "connected" state and expand when connected. 

Fixes https://github.com/microsoft/vscode-mssql/issues/1393 and https://github.com/microsoft/vscode-mssql/issues/1365

Also fixes https://github.com/microsoft/vscode-mssql/issues/1414, which although is the same behavior in ADS


src/objectExplorer/objectExplorerUtils.ts
```diff
@@ -25,21 +26,28 @@ export class ObjectExplorerUtils {
...
 
     public static getNodeUri(node: TreeNodeInfo): string {
-        if (node.nodeType === Constants.disconnectedServerLabel) {
-            return undefined;
+        const profile = <IConnectionProfile>node.connectionCredentials;
+        return ObjectExplorerUtils.getNodeUriFromProfile(profile);
+    }
+
+    public static getNodeUriFromProfile(profile: IConnectionProfile): string {
+        const uri = profile.server + '_' +
```
aaomidi: Nit. Using + to concatenate strings is pretty slow (although it doesn't really matter in this case.

Some alternatives are using the new ``` `${}` ``` syntax or `[profile.server, profile.database, profile.user, profile.profileName].join('_');`


Some more reading if you're interested: https://codecalamity.com/stop-using-plus-signs-to-concatenate-strings/

src/controllers/connectionManager.ts
```diff
@@ -195,6 +195,25 @@ export default class ConnectionManager {
...
+        const connectedCredentials = Object.keys(this._connections).map((uri) => this._connections[uri].credentials);
+        for (let connectedCredential of connectedCredentials) {
+            if (Utils.isSameConnection(credential, connectedCredential)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public getUriForConnection(connection: Interfaces.IConnectionCredentials): string {
```
Charles-Gagnon: >string [](start = 79, length = 6)

string | undefined

src/controllers/connectionManager.ts
```diff
@@ -616,11 +635,18 @@ export default class ConnectionManager {
         });
     }
 
+    /**
+     * Delete a credential from the credential store
+     */
+    public async deleteCredential(profile: Interfaces.IConnectionProfile): Promise<boolean> {
+        let result = await this._connectionStore.deleteCredential(profile);
```
Charles-Gagnon: nit - is there a reason you define a var here? If it's for debugging you can see the return value for a function by setting an inline breakpoint on the return statement. Not a huge deal but it's just unnecessary clutter and adds extra work creating the var

src/controllers/mainController.ts
```diff
@@ -710,31 +699,24 @@ export default class MainController implements vscode.Disposable {
...
+                if (!node.sessionId) {
+                    // connect it first
+                    await this.createObjectExplorerSession(node.connectionCredentials);
                 }
+                this._statusview.languageFlavorChanged(uri.toString(), Constants.mssqlProviderName);
+                await this.connectionManager.connect(uri.toString(), connectionCreds);
+                this._statusview.sqlCmdModeChanged(uri.toString(), false);
+                await this.connectionManager.connectionStore.removeRecentlyUsed(<IConnectionProfile>connectionCreds);
+                return Promise.resolve(true);
             } else {
```
Charles-Gagnon: Can just return true

src/controllers/mainController.ts
```diff
@@ -865,9 +847,56 @@ export default class MainController implements vscode.Disposable {
...
         if (e.affectsConfiguration(Constants.extensionName)) {
-            this._objectExplorerProvider.refresh(undefined);
+            let needsRefresh = false;
+            // user connections is a super set of object explorer connections
+            let userConnections: any[] = this._vscodeWrapper.getConfiguration(Constants.extensionName).get(Constants.connectionsArrayName);
+            let objectExplorerConnections = this._objectExplorerProvider.rootNodeConnections;
+
+            // if a connection(s) was/were manually removed
+            if (objectExplorerConnections.length >= userConnections.length) {
+                let staleConnections = objectExplorerConnections.filter((oeConn) => {
```
Charles-Gagnon: THe connections can change without the size actually changing - you should filter all the connections regardless I'd say

src/controllers/mainController.ts
```diff
@@ -865,9 +847,56 @@ export default class MainController implements vscode.Disposable {
...
+                    await this.connectionManager.connectionStore.removeRecentlyUsed(profile);
+                    if (profile.authenticationType === Constants.sqlAuthentication &&
+                        profile.savePassword) {
+                            await this.connectionManager.deleteCredential(profile);
+                        }
+                }
+                // remove them from object explorer
+                await this._objectExplorerProvider.removeStaleConnectionNodes(staleConnections);
+                needsRefresh = true;
+            } else if (userConnections.length > objectExplorerConnections.length) {
```
Charles-Gagnon: >objectExplorerConnections [](start = 48, length = 25)

and then for this you would just take filter all the userConnections that aren't in objectExplorerConnections and add them

src/models/connectionStore.ts
```diff
@@ -293,13 +293,19 @@ export class ConnectionStore {
...
+        return new Promise<void>(async (resolve, reject) => {
             // Get all profiles
             let configValues = self.getRecentlyUsedConnections();
 
             // Remove the connection from the list if it already exists
             configValues = configValues.filter(value => !Utils.isSameProfile(<IConnectionProfile>value, conn));
 
+            // Remove any saved password
+            if (conn.savePassword) {
+                let credentialId = ConnectionStore.formatCredentialId(conn.server, conn.database, conn.user, ConnectionStore.CRED_MRU_USER);
```
Charles-Gagnon: Do we really want to do that here? Removing something from the MRU doesn't necessarily mean we should remove the entire connection normally. I would expect that whatever is calling this should determine if the password needs to be cleared (such as if we're completely removing a connection from all the places it's stored such as the MRU and Connection Store)

src/models/utils.ts
```diff
@@ -254,7 +255,9 @@ export function isSameConnection(conn: interfaces.IConnectionCredentials, expect
         expectedConn.server === conn.server
         && isSameDatabase(expectedConn.database, conn.database)
         && isSameAuthenticationType(expectedConn.authenticationType, conn.authenticationType)
-        && expectedConn.user === conn.user
+        && (conn.authenticationType === LocalizedConstants.authTypeSql ?
```
Charles-Gagnon: >authenticationType [](start = 17, length = 18)

The authenticationType stored in the profile shouldn't be localized. If it is I'd consider that a bug

src/controllers/mainController.ts
abist: What do you mean by all the connections? User connections are basically all the connections, unless I misunderstand what you're saying.

src/controllers/mainController.ts
abist: that's what it's doing right now

src/objectExplorer/objectExplorerService.ts
```diff
@@ -350,18 +353,40 @@ export class ObjectExplorerService {
...
         }
-        const nodeUri = ObjectExplorerUtils.getNodeUri(node);
-        await this._connectionManager.disconnect(nodeUri);
         this._nodePathToNodeLabelMap.delete(node.nodePath);
         this.cleanNodeChildren(node);
-        if (isDisconnect) {
-            this._treeNodeToChildrenMap.set(this._currentNode, [new ConnectTreeNode(this._currentNode)]);
+    }
+
+    public async removeStaleConnectionNodes(staleConnections: IConnectionCredentials[]): Promise<void> {
```
Charles-Gagnon: >removeStaleConnectionNodes [](start = 17, length = 26)

I'd suggest calling this something like removeConnectionNodes instead - there's nothing in the logic that makes this specifically for removing stale nodes only and something like this might be useful elsewhere

src/objectExplorer/objectExplorerUtils.ts
```diff
@@ -25,21 +26,28 @@ export class ObjectExplorerUtils {
...
+    public static getDatabaseName(node: TreeNodeInfo): string {
+        if (node.nodeType === Constants.serverLabel) {
+            return node.connectionCredentials.database;
         }
         while (node) {
-            if (node.nodeType === Constants.serverLabel) {
-                break;
+            if (node.nodeType === Constants.databaseString) {
+                return node.label;
             }
```
Charles-Gagnon: Isn't this going to have the same issue as you just fixed with the system versioned dbs?

src/controllers/mainController.ts
Charles-Gagnon: Yes, but again you're checking the length first though. If someone manually deleted a connection and added a new one in its place then this logic wouldn't correctly add the new node. 

You should never be using the length to determine if anything changed - that's not always going to be an accurate representation.

src/objectExplorer/objectExplorerUtils.ts
abist: Not really. This function sets the name to `<default>` is no database is found. We set it back to '' in the fix I sent out. Empty databases are handled as default in connections.

src/controllers/mainController.ts
Charles-Gagnon: As in just remove the length check - it's unnecessary and potentially buggy. Just do this

`const staleConnections = objectExplorerConnections.filter( oeConn => !userConnections.some(userConn => Utils.isSameConnection(oeConn, userConn)));
for(const conn of staleConnections) {
...
`

If there aren't any stale connections it just won't do anything which is fine. 

---
In reply to: [334677441](https://github.com/microsoft/vscode-mssql/pull/1423#discussion_r334677441) [](ancestors = 334677441)

src/models/connectionStore.ts
abist: Fixed.

src/controllers/mainController.ts
```diff
@@ -865,9 +847,55 @@ export default class MainController implements vscode.Disposable {
...
+                }
+                await this.connectionManager.connectionStore.removeRecentlyUsed(profile);
+                if (profile.authenticationType === Constants.sqlAuthentication &&
+                    profile.savePassword) {
+                        await this.connectionManager.deleteCredential(profile);
+                    }
+            }
+            // remove them from object explorer
+            await this._objectExplorerProvider.removeConnectionNodes(staleConnections);
+            needsRefresh = true;
```
Charles-Gagnon: >needsRefresh [](start = 12, length = 12)

You're always refreshing now. I'd change this to `needsRefresh = staleConnections.count > 0`;

src/objectExplorer/objectExplorerUtils.ts
Charles-Gagnon: I'm saying in that you return node.label here - but as we saw before the label isn't always the actual DB name. You should be using node.metadata.name.

And actually I filed a bug in another place I noticed this - [https://github.com/microsoft/vscode-mssql/issues/1426](https://github.com/microsoft/vscode-mssql/issues/1426) - which brings up the question why we have two separate functions doing the same thing. 

So my suggestion is for you to just leave this as-is for now and then do a follow-up PR to consolidate those two into a single method and fix the issue there. 

---
In reply to: [334679533](https://github.com/microsoft/vscode-mssql/pull/1423#discussion_r334679533) [](ancestors = 334679533)

abist: For databases, the node label is always the database name. The issue arose only for objects such as tables, sprocs, view etc. But I'll consolidate the logic anyway.

Charles-Gagnon:
And just noticed that there's a failing test now too - https://ci.appveyor.com/project/kburtram/vscode-mssql/builds/28106943

I'm not going to reset my vote but you should fix that before merging in.

src/objectExplorer/objectExplorerUtils.ts
Charles-Gagnon: Thank you. The reason why we'd want to do this is that if we ever wanted to add something to the label in the future then this would just start silently start breaking. Using the metadata name here is definitely the correct approach.

Issue #1394, opened 2019-10-11
OE Refresh function is not reliable


alanrenmsft:
I ran into this multiple times but does not always repro.
from the gif you can tell, the database has been deleted but refreshing OE doesn't remove it.

![Oct-11-2019 12-57-25](https://user-images.githubusercontent.com/13777222/66681245-2af3a100-ec27-11e9-97a5-d5e4c67858db.gif)

PR #1428, opened 2019-10-15
Refresh the tree when a node is refreshed

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1394

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -371,7 +371,7 @@ export class ObjectExplorerService {
             nodePath: node.nodePath
         };
         await this._connectionManager.client.sendRequest(RefreshRequest.type, refreshParams);
-        return this._objectExplorerProvider.refresh(node);
+        return this._objectExplorerProvider.refresh(undefined);
     }
 
     public signInNodeServer(node: TreeNodeInfo): void {


src/objectExplorer/objectExplorerService.ts
```diff
@@ -396,7 +396,7 @@ export class ObjectExplorerService {
             nodePath: node.nodePath
         };
         await this._connectionManager.client.sendRequest(RefreshRequest.type, refreshParams);
-        return this._objectExplorerProvider.refresh(node);
+        return this._objectExplorerProvider.refresh(undefined);
```
Charles-Gagnon: Why wouldn't you just refresh the tree after the node is deleted? This seems really heavy handed to refresh the whole tree every time we call refreshNode (which given the name would imply to me that it would only refresh the node itself)

abist: That's what I'm doing, refreshing the tree in the end. The `refreshNode` here means the refresh from backend. The `this._objectExplorerProvider.refresh(undefined)` is refreshing the Tree View. They are unrelated.

Charles-Gagnon: What I'm saying is that this function appears to be called when you right click on a node and select refresh. In that case I don't think we should be refreshing the whole tree - that's extremely inefficient.

Why wouldn't you just refresh the tree after deleting a node? Like have the code that's actually deleting the node call refresh on the tree afterwards.

abist: It already does. But the delete `XXX` node from UI is different from running `drop database XXX` from query editor.

PR #1429, opened 2019-10-15
Fix bug with new nodes when DB is changed

abist:
Instead of changing node credentials, make a connection with an object so that a new node isn't made when we connect to a new database.


Charles-Gagnon:
What was the bug here? It'd be great to at least describe the bug in the description if you don't link an actual issue so that reviewers have context on what is being fixed

abist:
The node's connection changed into that database, and so when you disconnected and connected the node back, you'd get a node of that database instead.

Issue #1431, opened 2019-10-15
Unexpected error when wrong password entered for New Query


Charles-Gagnon:
1. Create SQL login profile and set **Save Password** to no
2. Disconnect the server node
3. Right click the profile and select **New Query**
4. Either enter the wrong password or hit esc

This error appears (along with the expected error about not being able to log in)

![image](https://user-images.githubusercontent.com/28519865/66845191-74dccf80-ef24-11e9-8975-2b66345bb9aa.png)

PR #1434, opened 2019-10-15
Fix UX for create session issue

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1431.

`ObjectExplorerService.createSession` now rejects if there were no credentials i.e. if connection prompt failed. After that the deferred promise resolves a node if the `createSession` was successful in the notification handler. 

Also shows the correct node for just prompting password in the case where savePassword is false. (instead of retry creating the profile, which is shown when connection prompt failed)

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -321,13 +321,19 @@ export default class MainController implements vscode.Disposable {
      */
     private async createObjectExplorerSession(connectionCredentials?: IConnectionCredentials): Promise<void> {
         let createSessionPromise = new Deferred<TreeNodeInfo>();
-        await this._objectExplorerProvider.createSession(createSessionPromise, connectionCredentials);
-        const newNode = await createSessionPromise;
-        this._objectExplorerProvider.refresh(undefined);
-        let expandSessionPromise = new Deferred<TreeNodeInfo[]>();
-        await this._objectExplorerProvider.expandNode(newNode, newNode.sessionId, expandSessionPromise);
-        await expandSessionPromise;
-        this._objectExplorerProvider.refresh(undefined);
+        try {
+            await this._objectExplorerProvider.createSession(createSessionPromise, connectionCredentials);
+            const newNode = await createSessionPromise;
+            if (newNode) {
+                this._objectExplorerProvider.refresh(undefined);
+                let expandSessionPromise = new Deferred<TreeNodeInfo[]>();
+                await this._objectExplorerProvider.expandNode(newNode, newNode.sessionId, expandSessionPromise);
+                await expandSessionPromise;
+                this._objectExplorerProvider.refresh(undefined);
+            }
+        } catch (err) {
+            throw err;
+        }
     }
 
 
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -322,14 +322,16 @@ export default class MainController implements vscode.Disposable {
      */
     private async createObjectExplorerSession(connectionCredentials?: IConnectionCredentials): Promise<void> {
         let createSessionPromise = new Deferred<TreeNodeInfo>();
-        await this._objectExplorerProvider.createSession(createSessionPromise, connectionCredentials);
-        const newNode = await createSessionPromise;
-        if (newNode) {
-            this._objectExplorerProvider.refresh(undefined);
-            let expandSessionPromise = new Deferred<TreeNodeInfo[]>();
-            await this._objectExplorerProvider.expandNode(newNode, newNode.sessionId, expandSessionPromise);
-            await expandSessionPromise;
-            this._objectExplorerProvider.refresh(undefined);
+        const sessionId = await this._objectExplorerProvider.createSession(createSessionPromise, connectionCredentials);
+        if (sessionId) {
+            const newNode = await createSessionPromise;
+            if (newNode) {
+                this._objectExplorerProvider.refresh(undefined);
+                let expandSessionPromise = new Deferred<TreeNodeInfo[]>();
+                await this._objectExplorerProvider.expandNode(newNode, newNode.sessionId, expandSessionPromise);
+                await expandSessionPromise;
+                this._objectExplorerProvider.refresh(undefined);
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/controllers/mainController.ts
```diff
@@ -321,13 +321,19 @@ export default class MainController implements vscode.Disposable {
...
+            await this._objectExplorerProvider.createSession(createSessionPromise, connectionCredentials);
+            const newNode = await createSessionPromise;
+            if (newNode) {
+                this._objectExplorerProvider.refresh(undefined);
+                let expandSessionPromise = new Deferred<TreeNodeInfo[]>();
+                await this._objectExplorerProvider.expandNode(newNode, newNode.sessionId, expandSessionPromise);
+                await expandSessionPromise;
+                this._objectExplorerProvider.refresh(undefined);
+            }
+        } catch (err) {
```
Charles-Gagnon: Why are you catching and rethrowing? Just let the uncaught exception bubble up into a rejected promise.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -341,6 +349,9 @@ export class ObjectExplorerService {
                 this._sessionIdToPromiseMap.set(response.sessionId, promise);
                 return;
             }
+        } else {
+            // no connection was made
+            return Promise.reject();
```
Charles-Gagnon: Instead of returning Promise.reject it'd be better to actually return an error.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -267,14 +267,22 @@ export class ObjectExplorerService {
...
+                        let node = await promise;
+                        // if password failed
+                        if (!node) {
+                            const connectNode = new ConnectTreeNode(element);
+                            this._treeNodeToChildrenMap.set(element, [connectNode]);
+                            return [connectNode];
+                        }
+                    } catch (err) {
+                        // If node create session failed
                         const signInNode = new AccountSignInTreeNode(element);
```
Charles-Gagnon: Do we really want to re-create the node each time? If the sign in node was already there it'd be good not to recreate it.

src/objectExplorer/objectExplorerService.ts
abist: But there's no error here, it's just that the user pressed `Esc`. I don't think we should return an error.

src/objectExplorer/objectExplorerService.ts
abist: This is run only once because once this node is made, the code path is different.

src/objectExplorer/objectExplorerService.ts
Charles-Gagnon: Rejecting is implying there's an error so you're already doing that. That's why you added the try/catch around the call above.

 I generally agree with you though and would suggest just returning instead and making sure that the caller handles an undefined tree node.

aaomidi: It's a scenario like this https://github.com/microsoft/azuredatastudio/pull/7711/files

Unfortunately just as @Charles-Gagnon said the infrastructure here is pretty meh

src/objectExplorer/objectExplorerService.ts
```diff
@@ -310,7 +317,7 @@ export class ObjectExplorerService {
      * OE out of
      * @param connectionCredentials Connection Credentials for a node
      */
-    public async createSession(promise: Deferred<vscode.TreeItem>, connectionCredentials?: IConnectionCredentials): Promise<void> {
+    public async createSession(promise: Deferred<vscode.TreeItem>, connectionCredentials?: IConnectionCredentials): Promise<string> {
```
Charles-Gagnon: the promise should be `Deferred<vscode.TreeItem | undefined>` and the return type `Promise<string | undefined>`

src/objectExplorer/objectExplorerService.ts
```diff
@@ -339,8 +347,11 @@ export class ObjectExplorerService {
             if (response) {
                 this._sessionIdToConnectionCredentialsMap.set(response.sessionId, connectionCredentials);
                 this._sessionIdToPromiseMap.set(response.sessionId, promise);
-                return;
+                return response.sessionId;
             }
+        } else {
+            // no connection was made
+            return undefined;
```
Charles-Gagnon: You should be calling promise.resolve(undefined) here as well

src/controllers/mainController.ts
```diff
@@ -232,6 +232,7 @@ export default class MainController implements vscode.Disposable {
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
                         Constants.cmdConnectObjectExplorerNode, async (node: ConnectTreeNode) => {
+                        self._objectExplorerProvider.currentNode = node.parentNode;
```
aaomidi: Just so we know, any chance a node won't have a parent?

src/controllers/mainController.ts
```diff
@@ -321,13 +322,17 @@ export default class MainController implements vscode.Disposable {
...
         let createSessionPromise = new Deferred<TreeNodeInfo>();
-        await this._objectExplorerProvider.createSession(createSessionPromise, connectionCredentials);
-        const newNode = await createSessionPromise;
-        this._objectExplorerProvider.refresh(undefined);
-        let expandSessionPromise = new Deferred<TreeNodeInfo[]>();
-        await this._objectExplorerProvider.expandNode(newNode, newNode.sessionId, expandSessionPromise);
-        await expandSessionPromise;
-        this._objectExplorerProvider.refresh(undefined);
+        const sessionId = await this._objectExplorerProvider.createSession(createSessionPromise, connectionCredentials);
+        if (sessionId) {
```
aaomidi: What happens if no sessionId?

src/controllers/mainController.ts
```diff
@@ -321,13 +322,17 @@ export default class MainController implements vscode.Disposable {
...
+            const newNode = await createSessionPromise;
+            if (newNode) {
```
aaomidi: what happens if no newNode?

src/controllers/mainController.ts
abist: not for a `ConnectTreeNode`

src/controllers/mainController.ts
abist: nothing

src/controllers/mainController.ts
abist: Nothing. There's no change so no need to refresh.

Issue #1455, opened 2019-10-21
Add Connection action of mssql extension is added to other extension's Explorer in Side Bar


left-code:
- MSSQL Extension Version: 1.7.0
- VSCode Version: 1.39.2
- OS Version: Windows 10

Steps to Reproduce:

1. Install GitHub Pull Requests, or GitLens, or Remote-SSH 
2. Click on any of this extensions in Activity Bar
3. Add Connection from mssql extension will be visible in each Explorer of other extensions

Example below from Remote - SSH extension

![image](https://user-images.githubusercontent.com/4906671/67244818-dd630980-f45a-11e9-8c9e-1a01a25169c4.png)

![image](https://user-images.githubusercontent.com/4906671/67244883-02577c80-f45b-11e9-8532-f794644ec449.png)

Issue #1451, opened 2019-10-21
Refresh context menu of mssql extension is added to other extension's treeview


blackchoey:
- MSSQL Extension Version:1.7.0
- VSCode Version: 1.38.1
- OS Version: windows 10

Steps to Reproduce:

1. Install VS Code Azure Functions extension
2. Expand the treeview, right click on a subscription

Actual: There're 2 refresh context menus. One of them is from mssql extension. After disabled mssql extension, one of the refresh menu disappeared.
![image](https://user-images.githubusercontent.com/16605901/67173770-54b08300-f3f2-11e9-9f12-cb5a27d2242a.png)

Expected: Only one refresh context menu which is contributed by Azure Functions extension should exist.

PR #1456, opened 2019-10-22
Fix context menu clause

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1455 and https://github.com/microsoft/vscode-mssql/issues/1451


Charles-Gagnon:
FYI you need to use multiple fixes clauses if you have multiple items in order to close them all

e.g.

`Fixes #1455 Fixes #1451`

Issue #1454, opened 2019-10-21
Number of rows affected no longer in messages


jdpepperman:
- MSSQL Extension Version: 1.7.0
- VSCode Version: 1.39.2
- OS Version: Windows 7 Pro Version 6.1 (Build 7601: Service Pack 1)

Steps to Reproduce:

1. Execute a query. The messages window used to to display:

> Started executing query at Line 1
> (n rows affected)
> Batch execution time: 00:00:01.000
> Total execution time: 00:00:01.000

Now instead of (n rows affected) I'm just seeing "Commands completed successfully." I thought I may have accidentally changed a setting but I reinstalled MSSQL extension v1.6.0 and I got the row count again. I tested this using a SELECT statement and an UPDATE statement and neither gave a number of rows affected.

PR #1457, opened 2019-10-22
Update tools service with updated sqlclient

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1454

### DIFF
--- a/src/configurations/dev.config.json
+++ b/src/configurations/dev.config.json
@@ -1,7 +1,7 @@
 {
     "service": {
         "downloadUrl": "https://github.com/Microsoft/sqltoolsservice/releases/download/v{#version#}/microsoft.sqltools.servicelayer-{#fileName#}",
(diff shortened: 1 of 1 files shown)


Charles-Gagnon:
Are you going to update https://github.com/microsoft/vscode-mssql/blob/383200bfdbfab85ae0874f41e15f2862e1802a3d/src/configurations/production.config.json separately?

(is there any particular reason we do this anyways? Seems like needless overhead to upload to download center)

abist:
@Charles-Gagnon the version in that file doesn’t matter apparently and is only used for signed downloads of the service for production bits.

Charles-Gagnon:
So we just update the zip there instead of creating a new fwlink? What about users who don't update right away - couldn't that break them if they try to redownload the service?

abist:
The link to the zip is created by some other process that Karl drives. For the new release, he will have to create new links.

Issue #1440, opened 2019-10-17
Query results line link opens empty editor if original editor is closed


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Open an untitled, unsaved SQL query editor document
2. Execute a query
3. Close the untitled document
4. Click the line link in the Messages pane

Expected: Either nothing should happen, or the untitled editor is reopened with the original contents that executed the query

Actual: A new untitled editor document is opened with no contents in Plain Text mode.

PR #1459, opened 2019-10-22
Look for opened doc when selecting text

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1440

### DIFF
--- a/src/models/sqlOutputContentProvider.ts
+++ b/src/models/sqlOutputContentProvider.ts
@@ -75,7 +75,10 @@ export class SqlOutputContentProvider {
     }
 
     public editorSelectionRequestHandler(uri: string, selection: ISelectionData): void {
-        this._queryResultsMap.get(uri).queryRunner.setEditorSelection(selection);
+        const docExists = this._vscodeWrapper.textDocuments.find(doc => doc.uri.toString() === uri);
+        if (docExists) {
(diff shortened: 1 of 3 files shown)


anthonydresser:
I think this is a user setting, so I'm not sure if we want to fix thix?

abist:
@anthonydresser What user setting is this?

anthonydresser:
image

abist:
@anthonydresser I'm not sure if that setting is related to the issue. The issue here is that a new doc is opened when the line selection is made from a results tab whose corresponding editor was closed.

anthonydresser:
o, nvm i thought the issue was different.

src/models/sqlOutputContentProvider.ts
```diff
@@ -75,7 +75,10 @@ export class SqlOutputContentProvider {
     }
 
     public editorSelectionRequestHandler(uri: string, selection: ISelectionData): void {
-        this._queryResultsMap.get(uri).queryRunner.setEditorSelection(selection);
+        const docExists = this._vscodeWrapper.textDocuments.find(doc => doc.uri.toString() === uri);
```
anthonydresser: Does it make more sense to move this check into the query runner, since that is where the document is actually attempted to open?

Issue #1415, opened 2019-10-11
Copying a selection


aaomidi:
If you have multiple small selections and ctrl+c it does copy something, but you can't copy with the context menu.

![AAEeMGWLDM](https://user-images.githubusercontent.com/1465785/66689953-2b4d6580-ec42-11e9-8ab5-cde56c9a8abd.gif)

PR #1463, opened 2019-10-24
Feature: Copy multiple selection

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1415 (Enable context menu for multiple selections as well)

Example - 
Capture

I think ADS should have the same implementation to maintain row order (Issue I opened for this - https://github.com/microsoft/azuredatastudio/issues/7931)

### DIFF
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -431,9 +431,6 @@ export default class QueryRunner {
             copyString += os.EOL;
         }
 
-        // Remove any extra EOLs and tabs
-        copyString.trim();
-
         let oldLang: string;
         if (process.platform === 'darwin') {
             oldLang = process.env['LANG'];
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -376,8 +376,8 @@ export default class QueryRunner {
             };
             let columnHeaders = this.getColumnHeaders(batchId, resultId, columnRange);
             copyString += columnHeaders.join('\t');
+            copyString += os.EOL;
         }
-        copyString += os.EOL;
 
         // sort the selections by row to maintain copy order
         selection.sort((a, b) => a.fromRow - b.fromRow);
@@ -431,8 +431,9 @@ export default class QueryRunner {
         }
 
         // Remove the last extra new line
-        copyString = copyString.substring(0, copyString.length - os.EOL.length);
-
+        if (copyString.length > 1) {
+            copyString = copyString.substring(0, copyString.length - os.EOL.length);
+        }
 
         let oldLang: string;
         if (process.platform === 'darwin') {
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -430,8 +430,9 @@ export default class QueryRunner {
             copyString += os.EOL;
         }
 
-        // Remove any extra lines
-        copyString.trim();
+        // Remove the last extra new line
+        copyString = copyString.substring(0, copyString.length - 1);
+
 
         let oldLang: string;
         if (process.platform === 'darwin') {
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -20,7 +20,8 @@ import { BatchSummary, QueryExecuteParams, QueryExecuteRequest,
     QueryExecuteBatchNotificationParams,
     QueryExecuteOptionsRequest,
     QueryExecutionOptionsParams,
-    QueryExecutionOptions} from '../models/contracts/queryExecute';
+    QueryExecutionOptions,
+    DbCellValue} from '../models/contracts/queryExecute';
 import { QueryDisposeParams, QueryDisposeRequest } from '../models/contracts/queryDispose';
 import { QueryCancelParams, QueryCancelResult, QueryCancelRequest } from '../models/contracts/queryCancel';
 import { ISlickRange, ISelectionData, IResultMessage } from '../models/interfaces';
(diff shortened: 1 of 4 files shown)


src/controllers/queryRunner.ts
```diff
@@ -380,38 +381,81 @@ export default class QueryRunner {
...
+                }
+            }
+            let columnRange: ISlickRange = {
+                fromCell: firstCol,
+                toCell: lastCol,
+                fromRow: undefined,
+                toRow: undefined
+            };
+            let columnHeaders = this.getColumnHeaders(batchId, resultId, columnRange);
+            copyString += columnHeaders.join('\t');
```
Charles-Gagnon: Make sure you aren't having the same issue as ADS where you're appending an extra tab onto the end of the copied row

src/controllers/queryRunner.ts
```diff
@@ -380,38 +381,81 @@ export default class QueryRunner {
...
+        }
+        copyString += os.EOL;
```
Charles-Gagnon: You should also be trimming off the last EOL

src/controllers/queryRunner.ts
```diff
@@ -380,38 +381,81 @@ export default class QueryRunner {
...
-                        : cellObjects.map(x => x.displayValue);
-                    copyString += cells.join('\t');
-                    if (rowIndex < result.resultSubset.rows.length - 1) {
-                        copyString += os.EOL;
+                for (let row of result.resultSubset.rows) {
+                    let rowNumber = row[0].rowId + range.fromRow;
+                    if (rowIdToSelectionMap.has(rowNumber)) {
+                        let rowSelection = rowIdToSelectionMap.get(rowNumber);
+                        rowSelection.push(range);
+                        rowIdToSelectionMap.set(rowNumber, rowSelection);
```
Charles-Gagnon: This set is unnecessary - you're already modifying the array directly

src/controllers/queryRunner.ts
```diff
@@ -380,38 +381,81 @@ export default class QueryRunner {
...
+                    copyString += ' \t';
+                }
+                let cellObjects = row.slice(rowSelection.fromCell, (rowSelection.toCell + 1));
+                // Remove newlines if requested
+                let cells = this.shouldRemoveNewLines()
+                ? cellObjects.map(x => this.removeNewLines(x.displayValue))
+                : cellObjects.map(x => x.displayValue);
+                copyString += cells.join('\t');
+            }
+            copyString += os.EOL;
```
Charles-Gagnon: Same comments with the tabs and EOL's here

test/queryRunner.test.ts
```diff
@@ -727,6 +727,153 @@ suite('Query Runner tests', () => {
...
+                testuri,
+                testuri,
+                testStatusView.object,
+                testSqlToolsServerClient.object,
+                testQueryNotificationHandler.object,
+                testVscodeWrapper.object
+            );
+            queryRunner.uri = testuri;
+            queryRunner.copyResults(testRange, 0, 0).then(() => {
+                testVscodeWrapper.verify<void>(x => x.clipboardWriteText(TypeMoq.It.isAnyString()), TypeMoq.Times.once());
```
Charles-Gagnon: This test would be improved by actually checking the pasted value matches what you expect

test/queryRunner.test.ts
```diff
@@ -727,6 +727,153 @@ suite('Query Runner tests', () => {
...
+                testStatusView.object,
+                testSqlToolsServerClient.object,
+                testQueryNotificationHandler.object,
+                testVscodeWrapper.object
+            );
+            queryRunner.uri = testuri;
+            // Call handleResult to ensure column header info is seeded
+            queryRunner.handleQueryComplete(result);
+            return queryRunner.copyResults(testRange, 0, 0).then(() => {
+                testVscodeWrapper.verify<void>(x => x.clipboardWriteText(TypeMoq.It.isAnyString()), TypeMoq.Times.once());
```
Charles-Gagnon: You never actually check whether the headers were copied or not so this test and the ones below don't seem super useful as is

src/controllers/queryRunner.ts
abist: why? we want the column headers to be in separate line

src/controllers/queryRunner.ts
abist: `.join('\t')` doesn't append the joining string at the end

src/controllers/queryRunner.ts
abist: I'm assuming you mean trimming off tabs and EOL's at the end of the copyString generation, which I agree with. However, I don't think we should be trimming off for every row.

src/controllers/queryRunner.ts
Charles-Gagnon: Sorry, missed that this was for the header row only.

Let's make sure that if there's no data we're still trimming off this last EOL though to be consistent.

src/controllers/queryRunner.ts
```diff
@@ -380,38 +381,84 @@ export default class QueryRunner {
...
+                ? cellObjects.map(x => this.removeNewLines(x.displayValue))
+                : cellObjects.map(x => x.displayValue);
+                copyString += cells.join('\t');
+            }
+            copyString.trim();
+            copyString += os.EOL;
+        }
+
+        // Remove any extra EOLs and tabs
+        copyString.trim();
```
Charles-Gagnon: Don't do this otherwise we trim off user content if they have whitespace characters in their content. You should be safe to just remove the last EOL - as you pointed out there won't be extra tabs so the EOL is the only thing that should be extra there.

src/controllers/queryRunner.ts
```diff
@@ -355,63 +356,81 @@ export default class QueryRunner {
...
+                    copyString += ' \t';
+                }
+                let cellObjects = row.slice(rowSelection.fromCell, (rowSelection.toCell + 1));
+                // Remove newlines if requested
+                let cells = this.shouldRemoveNewLines()
+                ? cellObjects.map(x => this.removeNewLines(x.displayValue))
+                : cellObjects.map(x => x.displayValue);
+                copyString += cells.join('\t');
+            }
+            copyString.trim();
```
Charles-Gagnon: Isn't this going to have the same problem with possibly trimming off user content? Why is the trim necessary here anyways?

src/controllers/queryRunner.ts
```diff
@@ -355,63 +356,80 @@ export default class QueryRunner {
...
+                }
+                let cellObjects = row.slice(rowSelection.fromCell, (rowSelection.toCell + 1));
+                // Remove newlines if requested
+                let cells = this.shouldRemoveNewLines()
+                ? cellObjects.map(x => this.removeNewLines(x.displayValue))
+                : cellObjects.map(x => x.displayValue);
+                copyString += cells.join('\t');
+            }
+            copyString += os.EOL;
+        }
```
Charles-Gagnon: After this you should remove the last EOL, otherwise we'll end up with an extra one on the end (ADS users complained a lot about this when we accidentally added that in)

src/controllers/queryRunner.ts
```diff
@@ -355,63 +356,83 @@ export default class QueryRunner {
...
+
+        // Remove any extra lines
+        copyString.trim();
```
Charles-Gagnon: Again, you can't use trim otherwise you risk removing user content. Just specifically remove the last EOL - we know there'll always be one (because of the header row) and we only care about removing that last one.

abist: Good point, didn't consider that.

src/controllers/queryRunner.ts
```diff
@@ -430,8 +430,9 @@ export default class QueryRunner {
             copyString += os.EOL;
         }
 
-        // Remove any extra lines
-        copyString.trim();
+        // Remove the last extra new line
+        copyString = copyString.substring(0, copyString.length - 1);
```
Charles-Gagnon: EOL is two characters on windows. Use EOL.length instead of 1

src/controllers/queryRunner.ts
```diff
@@ -355,63 +356,84 @@ export default class QueryRunner {
...
+            let columnRange: ISlickRange = {
+                fromCell: firstCol,
+                toCell: lastCol,
+                fromRow: undefined,
+                toRow: undefined
+            };
+            let columnHeaders = this.getColumnHeaders(batchId, resultId, columnRange);
+            copyString += columnHeaders.join('\t');
         }
+        copyString += os.EOL;
```
Charles-Gagnon: Oops, one more thing. So just noticed that we actually check if headers should be included. So two things :

1. We shouldn't always be adding on the EOL in that case (having an extra line above copied content would be annoying)
2. In that case we need to check at the end that copystring.length > 1 before trimming off the last EOL since it'd be possible for the string to be empty

Issue #1435, opened 2019-10-15
New Query on a database node uses server node database for connection


kburtram:
- MSSQL Extension Version: 1.7
Steps to Reproduce:

1. Create a connection to a SQL Server with default DB
2. Expand Databases node
3. Execute New Query on a database node

Expected: New query is connected to the specified database

Actual: New query is connect to the default database

PR #1464, opened 2019-10-24
Fix database context for New Query

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1435

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -162,6 +162,7 @@ export default class MainController implements vscode.Disposable {
                     } else if (databaseName === LocalizedConstants.defaultDatabaseLabel) {
                         connectionCredentials.database = '';
                     }
+                    treeNodeInfo.connectionCredentials = connectionCredentials;
                     await self.onNewQuery(treeNodeInfo);
                 }));
 


src/controllers/mainController.ts
```diff
@@ -162,6 +162,7 @@ export default class MainController implements vscode.Disposable {
                     } else if (databaseName === LocalizedConstants.defaultDatabaseLabel) {
                         connectionCredentials.database = '';
                     }
+                    treeNodeInfo.connectionCredentials = connectionCredentials;
```
Charles-Gagnon: This doesn't seem right to do it here. By fixing up the credentials here that means any other actions on the node will see different connection credentials based on whether new query has been ran first.

A couple ways to handle this that I think are better :
1. Don't store connection credentials on EVERY node. Instead have each node implement a getter where most nodes just ask for the parents connection and then have the server node actually return the connection. This is what SMO does and it works nicely since it's really simple for most nodes and then you can easily have the database node then modify the database it gets from the server in its getter and pass it along so all child nodes get the correct connection
2. Update the connection credentials for the node correctly when it's created

abist: That's intended. The node's database won't change, so what's wrong with keeping it set
 instead of looking up the parents credentials every time and taking a perf hit?

Charles-Gagnon: Because you're only setting it when the user selects New Query. So any other action taken on the node will be using the connection without an updated DB if the user didn't run New Query first.

And the perf hit would be minimal - I don't think that should be a reason to not do it that way. It's just a simple walking up the tree which isn't going to be more than 4-5 nodes.

Issue #1437, opened 2019-10-17
Connections to a specific database shows all the server's databases


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Connect to a specific database on a server

Expected: Only see the contents of that database

Actual: See the content of the entire server.  If we are going to show the entire server contents why prompt for a specific database name during connection?

![image](https://user-images.githubusercontent.com/599935/67031070-a771fe00-f0c5-11e9-96bd-8fef2d9bd514.png)

msftrncs:
This is normal for most SQL applications.  Even in an active query window that is already connected to a specific 'starting container', you can use `USE another_database` to use another database in the same server instance.

Issue #1426, opened 2019-10-14
getDatabaseName isn't using the metadata for DB name


Charles-Gagnon:
https://github.com/microsoft/vscode-mssql/blob/f836c23ce4be6b0081ab75de453f49c582b4e7d2/src/controllers/mainController.ts#L378 

This shouldn't be using the label as that can have other stuff appended to it. The TreeNodeInfo.metadata that was recently should be used instead.

PR #1473, opened 2019-11-05
Separate database connections from server connections

abist:
Separate database connections from server connections.
Fixes https://github.com/microsoft/vscode-mssql/issues/1437
Fixes https://github.com/microsoft/vscode-mssql/issues/1426

Capture


src/objectExplorer/objectExplorerService.ts
```diff
@@ -65,21 +65,22 @@ export class ObjectExplorerService {
...
+                            nodeLabel = connection.label;
+                            break;
+                        }
                     }
                 }
                 // set connection and other things
                 if (self._currentNode && (self._currentNode.sessionId === result.sessionId)) {
-                    nodeLabel = nodeLabel === result.rootNode.nodePath ?
-                    self.createNodeLabel(self._currentNode.connectionCredentials) : nodeLabel;
+                    nodeLabel = !nodeLabel ? self.createNodeLabel(self._currentNode.connectionCredentials) : nodeLabel;
```
Charles-Gagnon: nit: but you can shorten this slightly by doing 
`nodeLabel = nodeLabel || self.createNodeLabel(self._currentNode.connectionCredentials);`

src/objectExplorer/objectExplorerService.ts
```diff
@@ -65,21 +65,22 @@ export class ObjectExplorerService {
                 let nodeConnection = this._sessionIdToConnectionCredentialsMap.get(result.sessionId);
                 for (let connection of savedConnections) {
                     if (Utils.isSameConnection(connection.connectionCreds, nodeConnection)) {
-                        nodeLabel = connection.label;
-                        break;
+                        // if it's not the defaul label
+                        if (connection.label !== connection.connectionCreds.server) {
```
Charles-Gagnon: What is this fixing?

Also don't you want to break out if the connection is the same regardless of whether the label is the default one or not?

abist: This is fixing the fact that root nodes can now be databases as well, not just servers. So, we only change the label if there's an actual profile name, not just the server name.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -151,18 +152,22 @@ export class ObjectExplorerService {
...
         this._expandParamsToPromiseMap.set(expandParams, promise);
         const response = await this._connectionManager.client.sendRequest(ExpandRequest.type, expandParams);
-        if (!response) {
+        if (response) {
+            this._currentNode = node;
+            return response;
+        } else {
             this._expandParamsToPromiseMap.delete(expandParams);
+            promise.resolve(undefined);
+            return undefined;
```
Charles-Gagnon: Either return false or change the return type of this to be `Promise<boolean | undefined>` (returning false seems better)

src/models/connectionCredentials.ts
```diff
@@ -56,6 +56,7 @@ export class ConnectionCredentials implements IConnectionCredentials {
             details.options['server'] += (',' + credentials.port);
         }
         details.options['database'] = credentials.database;
+        details.options['databaseDisplayName'] = credentials.database;
```
Charles-Gagnon: Where is this used?

abist: in the backend to get the database node

src/objectExplorer/objectExplorerService.ts
abist: There's no bug here. We are separating database connections from server connections. I'll add a screenshot of 2 nodes.

abist:
@Charles-Gagnon 
Capture

Both these nodes would be like the first node in the screenshot without this change.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -144,25 +145,29 @@ export class ObjectExplorerService {
...
```
Charles-Gagnon: Not sure why you'd want to return undefined here. False should result in the same behavior and is just cleaner

src/objectExplorer/objectExplorerUtils.ts
```diff
@@ -40,8 +45,10 @@ export class ObjectExplorerUtils {
             return node.connectionCredentials.database;
         }
         while (node) {
-            if (node.nodeType === Constants.databaseString) {
-                return node.label;
+            if (node.metadata) {
```
Charles-Gagnon: Hm, can you check if there's a bug here for DB-specific connections? The issue I'm seeing is if you're in a child node of the server tree that isn't under a database node you're going to get the wrong DB name.

- Server connection is to DB `mydb`
- Call getDatabaseName on a child node (such as Tables folder)
- It isn't a server node so goes into the while loop
- It isn't a database node and so walks up the tree, eventually getting to the server node
- We then return the default database label instead of the actual database for the connection

Seems like we need to still check if the server connection is to a specified db after falling out of the while loop

src/objectExplorer/objectExplorerService.ts
abist: @Charles-Gagnon sqltoolsservice looks at the term 'databaseDisplayName` in the connection details options when listing out the object explorer nodes for it. This line specifically - https://github.com/microsoft/vscode-mssql/pull/1473/files/e000bbc9411d369a8e699c9fd650c036ba998573#diff-40da5394330734421e8030834675cd46R59

src/objectExplorer/objectExplorerUtils.ts
abist: @Charles-Gagnon That case wouldn't occur since we are looking at the metadata. Even though the rootNode is technically a server, the metadata will be a database. There wouldn't ever be a case when we are traversing a node's parent and we don't find a node's whose metadata isn't a `Database`

Charles-Gagnon: Ah ok, well that's not intuitive at all! Thanks for looking into it though.

src/objectExplorer/objectExplorerService.ts
Charles-Gagnon: Interesting, but it also seems to look at databaseName as well. Was that not working correctly before?

https://github.com/Microsoft/sqltoolsservice/blob/ec09ad21d291b33ac04b9e5ac4714ae958fbcfdb/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs#L488

Charles-Gagnon: Although I guess you're setting database, not databaseName. Is that maybe the issue here?

abist: Perhaps. I basically tried to see what's different in the extension and ADS and the `databaseDisplayName` was the only terms that wasn't being set on the extension side. Adding that made OE work as expected.

Charles-Gagnon: Pretty odd. I'm not even sure what a "database display name is". We don't have display names for databases AFAIK - it's just the database name. I feel like something else is wrong here, but if this fixes the issue and there's no other problems seen in your testing I guess it's fine.

PR #1474, opened 2019-11-06
Scripting

abist:
Added Script as Create, Drop, Execute and Alter


src/controllers/mainController.ts
```diff
@@ -252,6 +258,45 @@ export default class MainController implements vscode.Disposable {
...
+            const queryUriPromise = new Deferred<boolean>();
+            await this.connectionManager.connect(uri, connectionCreds, queryUriPromise);
+            await queryUriPromise;
+            this._statusview.languageFlavorChanged(uri, Constants.mssqlProviderName);
+            this._statusview.sqlCmdModeChanged(uri, false);
+            if (executeScript) {
+                const queryPromise = new Deferred<boolean>();
+                await this._outputContentProvider.runQuery(this._statusview, uri, undefined, title, queryPromise);
+                await queryPromise;
+                await this.connectionManager.connectionStore.removeRecentlyUsed(<IConnectionProfile>connectionCreds);
```
Charles-Gagnon: Why not just have runQuery take in a param that tells it whether to add it to recently used? That's what we do in ADS and it makes it much better than having to remember to remove it each time you didn't want to save it (plus gets rid of any possible state issues where things query for connections between when it's added and when it's removed)

src/scripting/scriptingService.ts
```diff
@@ -55,11 +55,27 @@ export class ScriptingService {
...
+    /**
+     * Helper to create scripting params
+     */
+    public createScriptingParams(node: TreeNodeInfo, uri: string, operation: ScriptOperation): ScriptingParams {
         const scriptingObject = this.getObjectFromNode(node);
         let serverInfo = this._connectionManager.getServerInfo(node.connectionCredentials);
+        let scriptCreateDropOption: string;
+        switch (operation) {
+            case (ScriptOperation.Select):
+                scriptCreateDropOption = 'ScriptSelect';
```
Charles-Gagnon: Bleh - why is this not an enum. Anyways I know this is just a string value on the service side but I would suggest making either a string enum or a string literal type to reduce the chance of errors in the future.

src/scripting/scriptingService.ts
```diff
@@ -55,11 +55,27 @@ export class ScriptingService {
...
+                break;
+            case (ScriptOperation.Delete):
+                scriptCreateDropOption = 'ScriptDrop';
+                break;
+            case (ScriptOperation.Create):
+                scriptCreateDropOption = 'ScriptCreate';
+            default:
+                scriptCreateDropOption = 'ScriptCreate';
```
Charles-Gagnon: You can just remove the create option if you're doing to default to ScriptCreate anyways (although realistically I would say you should throw an error instead - there's no good reason to default to create I'd think)

test/scriptingService.test.ts
```diff
@@ -73,7 +72,88 @@ suite('Scripting Service Tests', () => {
...
+            metadataType: MetadataType.SProc,
+            metadataTypeName: 'StoredProcedure',
+            urn: undefined,
+            schema: 'dbo',
+            name: 'test_proc'
+        };
+        const testNode = new TreeNodeInfo('test_table (System Versioned)', undefined, undefined,
+            undefined, undefined, 'Table', undefined, undefined, undefined, testNodeMetadata);
+        scriptingService = new ScriptingService(connectionManager.object);
+        const script = await scriptingService.script(testNode, 'test_uri', ScriptOperation.Execute);
```
Charles-Gagnon: It would be more useful to ensure that the correct path is being taken by also checking that the mock SqlToolsServiceClient is getting the correct params sent for the calls to sendRequest. Right now this is only vaguely checking that things are returning as expected.

test/scriptingService.test.ts
```diff
@@ -73,7 +72,88 @@ suite('Scripting Service Tests', () => {
...
+
+    test('Test Script Alter function', async () => {
+        const testNodeMetadata: ObjectMetadata = {
+            metadataType: MetadataType.SProc,
+            metadataTypeName: 'StoredProcedure',
+            urn: undefined,
+            schema: 'dbo',
+            name: 'test_sproc'
+        };
+        const testNode = new TreeNodeInfo('test_table (System Versioned)', undefined, undefined,
```
Charles-Gagnon: The params here should match the metadata

src/controllers/mainController.ts
```diff
@@ -252,6 +258,45 @@ export default class MainController implements vscode.Disposable {
...
+        let actionPromise = new Promise<boolean>(async (resolve, reject) => {
+            const nodeUri = ObjectExplorerUtils.getNodeUri(node);
+            let connectionCreds = Object.assign({}, node.connectionCredentials);
+            const databaseName = ObjectExplorerUtils.getDatabaseName(node);
+            // if not connected or different database
+            if (!this.connectionManager.isConnected(nodeUri) ||
+                connectionCreds.database !== databaseName) {
+                // make a new connection
+                connectionCreds.database = databaseName;
+                if (!this.connectionManager.isConnecting(nodeUri)) {
```
Charles-Gagnon: If it is connecting shouldn't you wait on it to finish connecting before you proceed?

Issue #1444, opened 2019-10-18
Result grid doesn't have minimum size and doesn't render correctly with many results


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Run a query with multiple resultsets, for example:

```SQL
SELECT 1
GO 20
```
Expected: twenty resultsets that are sized correctly and usable.

Actual: resultsets are forced to fit visible area and are sized too small.

**Output from vscode**
<img width="1140" alt="Screen Shot 2019-10-17 at 11 22 28 PM" src="https://user-images.githubusercontent.com/599935/67070787-d031dc00-f135-11e9-9443-4e829c6f299e.png">

**Output from azure data studio**
<img width="1053" alt="Screen Shot 2019-10-17 at 11 23 58 PM" src="https://user-images.githubusercontent.com/599935/67070791-d45df980-f135-11e9-9019-ad4acaf1226b.png">

PR #1476, opened 2019-11-07
Show results expanded

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1444
Also fixes https://github.com/microsoft/vscode-mssql/issues/1488

Capture


src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -369,7 +369,7 @@ export class AppComponent implements OnInit, AfterViewChecked {
                     // Precalculate the max height and min height
                     let maxHeight = resultSet.rowCount < self._defaultNumShowingRows
                         ? Math.max((resultSet.rowCount + 1) * self._rowHeight, self.dataIcons.length * 30) + 10
-                        : 'inherit';
+                        : (self.dataIcons.length * 30) + 10;
```
kburtram: it's hard for me to tell from the code if this will have side-effects.  If it fixes a high-pri bug then let's push and do a test pass.  I'm assuming you've already run through pre-checkin validation that this works correctly in the various cases regarding resultset count, row count, and window size.

abist:
@kburtram waiting on build to complete before merging

Issue #1436, opened 2019-10-17
Missing results export icons on insiders build


kburtram:
- MSSQL Extension Version: 1.7
- VSCode Version: Insider 10/17

Steps to Reproduce:

1. Run a query

Expected: 3 export icons

Actual: 1 export icon

Note: There are now also double scrollbars on the results document where there aren't on stable.

![image](https://user-images.githubusercontent.com/599935/67027580-a12c5380-f0be-11e9-8cf7-b6f37acfa32e.png)

kburtram:
@abist I'm seeing this again now with the latest 1.7.1 RC build.  Are you able to repro?  I'm on Windows with the latest VS Code in case that matters.

abist:
Waiting on VS Code team to repond to this - https://github.com/microsoft/vscode/issues/84543

gdvallance:
This issue is present in non-insiders build too.

I raise the matter here so that there are not too many bug reports. However, would it be better to raise a separate ticket for a non-insider build?

The details of system etc is below: 
Version: 1.40.0 (system setup)
Commit: 86405ea23e3937316009fc27c9361deee66ffbf5
Date: 2019-11-06T17:02:13.381Z
Electron: 6.1.2
Chrome: 76.0.3809.146
Node.js: 12.4.0
V8: 7.6.303.31-electron.0
OS: Windows_NT x64 10.0.17134

Issue #1277, opened 2019-08-15
Feedback on vscode's Webview.toWebviewResource API proposal


mjbvz:
## Overview
VS Code 1.37 added a proposed api to help webviews create uris for local resources: https://github.com/microsoft/vscode/blob/192e15b613dd4f4dab9c735b7556e175580c6011/src/vs/vscode.proposed.d.ts#L1166

The main addition is a `toWebviewResource` method property on webviews that translates a local file uri into a webview uri (i.e. one that uses the special `vscode-resource:` scheme). This api is being added primarily to support running VS Code on the web where the existing `vscode-resource:` scheme will not work

## Request
As a popular user of the VS Code webview API, we'd appreciate your feedback on this api proposal. Does it work as advertised? Does the documentation for it make sense? Is it missing any functionality? Feel free to leave your feedback on this issue or on https://github.com/microsoft/vscode/issues/79242

You can try out the proposed api in VS Code 1.38 insiders by following [these instructions](https://code.visualstudio.com/api/advanced-topics/using-proposed-api). Please let me know if you have any questions about testing it out

*PS: There are no plans at the moment to remove the `vscode-resource:` scheme so your extension should continue to work on desktop VS Code even if it does not adopt this new API*

mjbvz:
Did you adopt this API?

abist:
We are already using the `vscode-resource`, but it's generated this way - https://github.com/microsoft/vscode-mssql/blob/8696a7bb8c5644dca3435f200d152547784c2bf6/src/controllers/webviewController.ts#L101

mjbvz:
Please switch to use the new api

abist:
@mjbvz Is it still in proposed or is this in the actual API now?

abist:
@mjbvz Using this API breaks the existing behavior that used to work fine. The webview doesn't show `svg`'s any more

abist:
@mjbvz From the docs here - https://code.visualstudio.com/api/extension-guides/webview

```
      // Get path to resource on disk
      const onDiskPath = vscode.Uri.file(
        path.join(context.extensionPath, 'media', 'cat.gif')
      );

      // And get the special URI to use with the webview
      const catGifSrc = panel.webview.asWebviewUri(onDiskPath);
```
However, when I convert using the same method, I get something like -

```vscode-resource://file///c%3A...```

PR #1478, opened 2019-11-08
Use new webview api to get resource

abist:
Update VS Code engine and use the new webview api to get a resource instead of manipulating it. 
Fixes https://github.com/microsoft/vscode-mssql/issues/1277
Fixes https://github.com/microsoft/vscode-mssql/issues/1436

### DIFF
--- a/test/connectionStore.test.ts
+++ b/test/connectionStore.test.ts
@@ -51,9 +51,9 @@ suite('ConnectionStore tests', () => {
             savePassword: true
         });
 
-        context = TypeMoq.Mock.ofType(stubs.TestExtensionContext);
+        context = TypeMoq.Mock.ofType<stubs.TestExtensionContext>();
         globalstate = TypeMoq.Mock.ofType(stubs.TestMemento);
-        context.object.globalState = globalstate.object;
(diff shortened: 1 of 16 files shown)


test/connectionStore.test.ts
```diff
@@ -51,9 +51,9 @@ suite('ConnectionStore tests', () => {
             savePassword: true
         });
 
-        context = TypeMoq.Mock.ofType(stubs.TestExtensionContext);
+        context = TypeMoq.Mock.ofType<stubs.TestExtensionContext>();
         globalstate = TypeMoq.Mock.ofType(stubs.TestMemento);
-        context.object.globalState = globalstate.object;
+        context.setup(c => c.globalState).returns(() => globalstate.object);
```
abist: Had to change the test because the property is readonly in the new API

abist:
`mainController` and `initialization` tests were failing with the new vscode engine update. I've disabled them for now to get a release build out with the driver update since these aren't product bugs. Tracking the issue to re-enable/rewrite the tests here - https://github.com/microsoft/vscode-mssql/issues/1485

PR #1487, opened 2019-11-12
Convert inline action icons to base64

abist:
Convert inline icons to base64


package.json
```diff
@@ -118,7 +118,7 @@
     "typescript": "3.5.3",
     "uglify-js": "mishoo/UglifyJS2#harmony-v2.8.22",
     "vsce": "^1.58.0",
-    "vscode": "^1.1.36",
+    "vscode": "^1.1.33",
```
kburtram: why are we doing this?

abist: We upgraded it earlier, just reverting it since it's not necessary. I can let it be too.

Issue #1492, opened 2019-11-13
Results don't show up for results not in screen when there are multiple batches


abist:
Run a query with 20 batches. Scroll to the bottom-most result set.

Actual - Result set doesn't load 
Expected - Resultset should load when scrolled down

![Capture](https://user-images.githubusercontent.com/6411451/68811674-12abe300-0626-11ea-81a4-84dc7d6fbcaf.JPG)

PR #1494, opened 2019-11-13
Fix bug with multiple result sets

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1492

### DIFF
--- a/src/views/htmlcontent/src/js/components/app.component.ts
+++ b/src/views/htmlcontent/src/js/components/app.component.ts
@@ -676,6 +676,7 @@ export class AppComponent implements OnInit, AfterViewChecked {
                     self.placeHolderDataSets[i].resized.emit();
                 }
             } else {
+                self.scrollEnabled = true;
                 let gridHeight = self._el.nativeElement.getElementsByTagName('slick-grid')[0].offsetHeight;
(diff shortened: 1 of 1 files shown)


src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -676,6 +676,7 @@ export class AppComponent implements OnInit, AfterViewChecked {
                     self.placeHolderDataSets[i].resized.emit();
                 }
             } else {
+                self.scrollEnabled = true;
```
kburtram: I don't understand why this fixes the problem.  Could you please explain a bit?  Thanks!

abist: When the number of result sets is less than the threshold, we were setting scrollEnabled to false here - https://github.com/microsoft/vscode-mssql/pull/1494/files/013d5f97dbfcfa1125084595dd898d1f88b5938b#diff-1df1f9bd5f3b0cf63b472db70bd235bfR673.

However, we never set it to true ever again, which was a bug. This needs to be true here - https://github.com/microsoft/vscode-mssql/pull/1494/files/013d5f97dbfcfa1125084595dd898d1f88b5938b#diff-1df1f9bd5f3b0cf63b472db70bd235bfR50

for the resultsets to emit the resize event.

PR #1500, opened 2019-11-14
Remove dependencies and move to yarn

abist:
General extension clean up: 
- [x] Removeunused files and modules.
- [x] Move to yarn package manager.
- [x] Fixes reported from[ LGTM](https://lgtm.com/projects/g/microsoft/vscode-mssql/?mode=list&severity=error%2Crecommendation%2Cwarning)


package.json
```diff
@@ -82,29 +82,18 @@
...
-    "gulp-remote-src": "0.4.1",
     "gulp-rename": "^1.2.2",
     "gulp-shell": "^0.5.2",
     "gulp-sourcemaps": "^1.6.0",
     "gulp-tslint": "^8.1.4",
     "gulp-typescript": "^3.1.4",
     "gulp-uglify": "^2.0.0",
     "istanbul": "^0.4.5",
-    "jasmine-core": "~2.4.1",
-    "karma": "^1.3.0",
```
anthonydresser: this was used for the frontend tests, if you plan on bringing that back, you will need to keep this.

gulpfile.js
```diff
@@ -1,9 +1,7 @@
-﻿"use strict"
```
anthonydresser: any reason you are removing this use strict? In js files it actually serves a function.

abist: lgtm complained about it and doesn't look like it was doing anything

anthonydresser: do you have a link to it complaining? it shouldn't. It is always recommended to use this for js files. (and typescript, but the ts compiler adds that for us). https://www.w3schools.com/js/js_strict.asp

abist:
@aaomidi Sorry I didn't see your PR. I did some more cleaning here in addition to moving to yarn.

Issue #1507, opened 2019-11-21
Double scroll bars error when messages pane is resized 


abist:
Steps for repro - 

1. Run any query.
2. Resize messages pane to make it smaller. 

Expected - Double scroll bars should not appear for the first time when message pane is resized.

Actual - Double scroll bars appear if the messages pane is resized for the first time. Results look fine if messages pane is toggled and resized. until messages pane is toggled and then it's fixed. 

![scroll-error](https://user-images.githubusercontent.com/6411451/69292372-d3980780-0bba-11ea-8be2-7a0d50f281fa.gif)

PR #1508, opened 2019-11-21
Fix double scroll bars when resizing

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1507

scroll-fix


src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -147,6 +147,7 @@ export class AppComponent implements OnInit, AfterViewChecked {
     private Constants = Constants;
     private Utils = Utils;
     private _messagesPaneHeight: number;
+    private _resultsPaneHeight: string;
```
Charles-Gagnon: This never appears to be used

PR #1510, opened 2019-11-23
Feature: Add firewall rules

abist:
Added feature to add firewall rule when there's a firewall error.

EDIT: WIP since need to add tests for this area.


localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -224,9 +230,48 @@
...
+        </trans-unit>
+        <trans-unit id="msgPromptAzureExtensionActivatedSignedIn">
+            <source xml:lang="en">Azure Account extension activated and signed in.</source>
+        </trans-unit>
+        <trans-unit id="msgInvalidIpAddress">
+            <source xml:lang="en">Invalid IP Address </source>
+        </trans-unit>
+        <trans-unit id="msgInvalidEndIpAddress">
+            <source xml:lang="en">Invalid IP Address </source>
+        </trans-unit>
```
Charles-Gagnon: Is this purposely a duplicate of the one above? I'd just have a single string if so - no reason to duplicate it

src/controllers/mainController.ts
```diff
@@ -50,6 +51,7 @@ export default class MainController implements vscode.Disposable {
     private _untitledSqlDocumentService: UntitledSqlDocumentService;
     private _objectExplorerProvider: ObjectExplorerProvider;
     private _scriptingService: ScriptingService;
+    private _firewallService: FirewallService;
```
Charles-Gagnon: Why is this here? Doesn't seem to be used at all

src/controllers/vscodeWrapper.ts
```diff
@@ -337,4 +337,33 @@ export default class VscodeWrapper {
...
+    public get azureAccountExtensionActive(): boolean {
+        return this.azureAccountExtension && this.azureAccountExtension.isActive;
+    }
+
+    /**
+     * Returns whether an azure account is signed in
+     */
+    public get isAccountSignedIn(): boolean {
+        return this.azureAccountExtension.exports.status === 'LoggedIn';
+    }
```
Charles-Gagnon: Should probably be safe and check for azureAccountExtension being undefined as well

src/controllers/vscodeWrapper.ts
```diff
@@ -337,4 +337,33 @@ export default class VscodeWrapper {
...
+     */
+    public get onDidChangeExtensions(): vscode.Event<void> {
+        return vscode.extensions.onDidChange;
+    }
+
+    /**
+     * Gets the Azure Account extension
+     */
+    public get azureAccountExtension(): vscode.Extension<any> {
+        return vscode.extensions.getExtension(Constants.azureAccountExtensionId);
```
Charles-Gagnon: vscode.Extension<any> | undefined

src/firewall/firewallService.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    public get account(): Account {
+        return this._account;
+    }
+
+    private convertToAzureAccount(azureSession: any): Account {
+        let tenant = {
+            displayName: Constants.tenantDisplayName,
+            id: this._session.tenantId,
```
Charles-Gagnon: >_session [](start = 21, length = 8)

You're already passing in the session - it would be better to either just use azureSession or this._session throughout this function

src/firewall/firewallService.ts
```diff
@@ -0,0 +1,120 @@
...
+import SqlToolsServiceClient from '../languageservice/serviceclient';
+import Constants = require('../constants/constants');
+import { Deferred } from '../protocol';
+import VscodeWrapper from '../controllers/vscodeWrapper';
+
+export class FirewallService {
+
+    private _isSignedIn: boolean = false;
+    private _session: any = undefined;
+    private _account: Account = undefined;
```
Charles-Gagnon: It would be good to add the typings for the azure account extension so you don't have to have these any's all over the place [https://github.com/microsoft/vscode-azure-account/blob/master/src/azure-account.api.d.ts](https://github.com/microsoft/vscode-azure-account/blob/master/src/azure-account.api.d.ts)

src/firewall/firewallService.ts
```diff
@@ -0,0 +1,120 @@
...
+            securityTokenMappings: await this.createSecurityTokenMapping()
+        }
+        return params;
+    }
+
+    public set isSignedIn(value: boolean) {
+        this._isSignedIn = value;
+        if (value) {
+            this._session = this._vscodeWrapper.azureAccountExtension.exports.sessions[0];
+            this._account = this.convertToAzureAccount(this._session);
```
Charles-Gagnon: Do you want to clear session and account when the user logs out?

src/languageservice/serviceclient.ts
```diff
@@ -119,6 +126,14 @@ export default class SqlToolsServiceClient {
         this._client = client;
     }
 
+    private get resourceClient(): LanguageClient {
+        return this._resourceClient;
+    }
+
+    private set resourceClient(client: LanguageClient) {
+        this._resourceClient = client;
+    }
```
Charles-Gagnon: No real point in having getters/setters if all you're doing is just modifying the private var directly

src/models/utils.ts
```diff
@@ -370,3 +379,79 @@ export function parseNumAsTimeString(value: number): string {
...
+    return vscode.workspace.getConfiguration(Constants.extensionConfigSectionName);
+}
+
+export function getConfigTracingLevel(): string {
+	let config = getConfiguration();
+	if (config) {
+		return config.get(configTracingLevel);
+	}
+	else {
+		return undefined;
```
Charles-Gagnon: >undefined [](start = 9, length = 9)

Not sure you want to be returning undefined for these config settings - it's unlikely the service exe's will be able to handle that correctly. I would suggest logging an error and then just not including that option at all if the config is invalid for some reason.

src/views/connectionUI.ts
```diff
@@ -456,25 +456,72 @@ export class ConnectionUI {
...
-                        return self.validateAndSaveProfile(updatedProfile);
+                // Check whether the error was for firewall rule or not
+                if (self.connectionManager.failedUriToFirewallIpMap.has(uri)) {
+                    // Firewall rule error
+                    const clientIp = self.connectionManager.failedUriToFirewallIpMap.get(uri);
+
+                    // Check whether the azure account extension is installed and active
+                    if (self._vscodeWrapper.azureAccountExtensionActive) {
+                        // Sign in to azure account
+                        return self.promptForAccountSignIn().then(async (signedIn) => {
```
Charles-Gagnon: Instead of having this huge amount of nested thens it would be greatly beneficial to just convert this into an async function for readability purposes...

src/views/connectionUI.ts
```diff
@@ -490,9 +537,20 @@ export class ConnectionUI {
         return ConnectionProfile.createProfile(this._prompter);
     }
 
-    public promptForRetryCreateProfile(profile: IConnectionProfile): PromiseLike<IConnectionProfile> {
+    private async promptToRetryAndSaveProfile(profile: IConnectionProfile, isFirewallError: boolean = false): Promise<IConnectionProfile> {
+        return this.promptForRetryCreateProfile(profile, isFirewallError).then(updatedProfile => {
```
Charles-Gagnon: This is already async - you can get rid of the .then and just await on promptForRetryCreateProfile

src/views/connectionUI.ts
```diff
@@ -501,6 +559,110 @@ export class ConnectionUI {
...
+                if (result === LocalizedConstants.signInLabel) {
+                    // show firewall dialog with all sign-in options
+                    return this.showSignInOptions();
+                } else {
+                    return false;
+                }
+            });
+        } else {
+            this.connectionManager.firewallService.isSignedIn = true;
+            return Promise.resolve(true);
```
Charles-Gagnon: Just return true - no reason to wrap it in a promise resolve

localization/xliff/enu/constants/localizedConstants.enu.xlf
abist: Whoops, good catch

src/firewall/firewallService.ts
abist: Is it worth adding the typings and add 3 new dependencies for the declarations? I could add this to interfaces, but some fields such as `Subscription` will stay as `any`

src/constants/constants.ts
```diff
@@ -77,8 +82,10 @@ export const gettingStartedGuideLink = 'https://aka.ms/mssql-getting-started';
 export const changelogLink = 'https://aka.ms/vscode-mssql-changelog';
 export const integratedAuthHelpLink = 'https://aka.ms/vscode-mssql-integratedauth';
 export const sqlToolsServiceCrashLink = 'https://github.com/Microsoft/vscode-mssql/wiki/SqlToolsService-Known-Issues';
+export const azureAccountExtensionId = 'ms-vscode.azure-account';
 export const databaseString = 'Database';
 export const localizedTexts = 'localizedTexts';
+export const ipAddressRegex = /\b(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\b/;
```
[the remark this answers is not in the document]
abist: This is the same regex in sqltoolsserivce, and I wanted to keep it consistent. Perhaps we should take in a change for sqltoolsservice for IP V6 as well

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -374,6 +419,24 @@
         <trans-unit id="msgConnect">
             <source xml:lang="en">Connect</source>
         </trans-unit>
+        <trans-unit id="azureSignIn">
+            <source xml:lang="en">Azure: Sign In</source>
```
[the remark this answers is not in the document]
abist: @kburtram  I copied the `Command Palette options` from the account extension to show all the commands at once since each option has a different command. It looks something like this -

Capture

I can change `Azure` to `MSSQL` if we want.

src/firewall/firewallService.ts
Charles-Gagnon: Up to you in the end - but I know it's a huge annoyance for me working in a codebase that doesn't have typings since then I have to hunt around for what the types are and what properties they have. You're ignoring a huge part of typescript (the type safety) by doing this.

src/constants/constants.ts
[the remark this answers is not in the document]
Charles-Gagnon: I thought that Azure DBs don't support ipv6 at the moment. I'm having trouble finding any sort of documentation for this at the moment but Azure isn't listed as supported here : https://docs.microsoft.com/en-us/sql/tools/configuration-manager/connecting-using-ipv6?view=sql-server-ver15

Issue #1432, opened 2019-10-15
Results font size setting isn't used


Charles-Gagnon:
Changing the results font size in the settings doesn't seem to actually have an effect on the text in the results tabs.

PR #1511, opened 2019-11-27
Fix results font size not respected from config

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1432

Get font size from the config.

Capture

### DIFF
--- a/src/models/sqlOutputContentProvider.ts
+++ b/src/models/sqlOutputContentProvider.ts
@@ -33,6 +33,7 @@ export class QueryRunnerState {
 class ResultsConfig implements Interfaces.IResultsConfig {
     shortcuts: { [key: string]: string };
     messagesDefaultOpen: boolean;
+    resultsFontSize: number;
 }
 
 export class SqlOutputContentProvider {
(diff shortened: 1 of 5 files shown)


src/models/sqlOutputContentProvider.ts
```diff
@@ -33,6 +33,7 @@ export class QueryRunnerState {
 class ResultsConfig implements Interfaces.IResultsConfig {
     shortcuts: { [key: string]: string };
     messagesDefaultOpen: boolean;
+    resultsFontSize: number;
```
Charles-Gagnon: You shouldn't need to define these here - they're already a part of the interface definition

Charles-Gagnon:
It looks like the header is longer than the cell below it in your image though - is that expected?

src/models/sqlOutputContentProvider.ts
abist: I added to the interface in the same PR

Charles-Gagnon: I'm saying that because it's defined in the interface you don't need to redefine it here - it'll inherit them from the interface already. You would only need to do something if you're adding actual implementation such as a method definition.

abist: This is implementing the interface, not extending it, so we do need to specify the field

Charles-Gagnon: Argh, yeah you're completely right. Dunno what I was thinking - sorry!

Issue #1438, opened 2019-10-17
Connecting to server in OE doesn't remove "saved password" from settings file


kburtram:
- MSSQL Extension Version: 1.7

Steps to Reproduce:

1. Manually create a connection profile in the user's settings.json file with `savedPassword: true` and provide a `password` directly in the JSON
2. Connect to the server in OE

Expected: The saved password will be removed from the settings file and stored in the secure key storage.  This is the behavior when a connection is made to the connection profile through the Command Palette `MS SQL: Connect` command.

Actual: The OE node is connected but the password is left in the settings file.

PR #1512, opened 2019-11-28
Settings now removes password when a connection is added manually

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1438


src/objectExplorer/objectExplorerService.ts
```diff
@@ -244,6 +244,37 @@ export class ObjectExplorerService {
...
+    private async handleNodeCreationFailure(element: TreeNodeInfo): Promise<AccountSignInTreeNode[]> {
+        const signInNode = new AccountSignInTreeNode(element);
+        this._treeNodeToChildrenMap.set(element, [signInNode]);
+        return [signInNode];
+    }
+
+    /**
+     * Handles an incorrect password failure when creating a session
+     */
+    private async handleNodePasswordFailure(element: TreeNodeInfo): Promise<ConnectTreeNode[]> {
```
Charles-Gagnon: There doesn't appear to be any reason for this to be async

src/objectExplorer/objectExplorerService.ts
```diff
@@ -244,6 +244,37 @@ export class ObjectExplorerService {
...
+        const newProfile = Object.assign({}, profile);
+        newProfile.password = '';
+        newProfile.savePassword = false;
+        await this._connectionManager.connectionStore.saveProfile(newProfile);
+    }
+
+    /**
+     * Handles a generic create session failure
+     */
+    private async handleNodeCreationFailure(element: TreeNodeInfo): Promise<AccountSignInTreeNode[]> {
```
Charles-Gagnon: There doesn't appear to be any reason for this to be async

src/objectExplorer/objectExplorerService.ts
```diff
@@ -277,15 +308,19 @@ export class ObjectExplorerService {
                         let node = await promise;
                         // if password failed
                         if (!node) {
-                            const connectNode = new ConnectTreeNode(element);
-                            this._treeNodeToChildrenMap.set(element, [connectNode]);
-                            return [connectNode];
+                            // remove password if it's saved in the credential store
```
Charles-Gagnon: How do you know that the failure was because of a wrong password here? What happens if something else is wrong - such as a user doesn't have connect permissions, wrong username, etc?

src/controllers/vscodeWrapper.ts
```diff
@@ -339,6 +339,13 @@ export default class VscodeWrapper {
     }
 
     /**
+     * Change a configuration setting
+     */
+    public setConfiguration(extensionName: string, resource: string, value: any): Thenable<void> {
```
Charles-Gagnon: Why is this a thenable instead of just using promises?

src/controllers/mainController.ts
```diff
@@ -930,10 +930,22 @@ export default class MainController implements vscode.Disposable {
...
                     !this.connectionManager.isConnecting(uri)) {
-                    // add a disconnected node for it
+                    // if the added connection is SQL Auth and has password
+                    // save the password and remove it from settings
+                    if (conn.authenticationType === Constants.sqlAuthentication &&
+                        !Utils.isEmpty(conn.password)) {
+                        await this.connectionManager.connectionStore.saveProfile(newConnectionProfile);
+                        const newConnectionProfileCopy = Object.assign({}, newConnectionProfile);
+                        newConnectionProfileCopy.password = '';
+                        this._vscodeWrapper.setConfiguration(Constants.extensionConfigSectionName,
```
Charles-Gagnon: Isn't this going to overwrite the entire array of connections with just the single one?

src/objectExplorer/objectExplorerService.ts
```diff
@@ -277,15 +308,19 @@ export class ObjectExplorerService {
...
+                            let profile = element.connectionCredentials as IConnectionProfile;
+                            let password = await this._connectionManager.connectionStore.lookupPassword(profile);
+                            if (password) {
+                                await this.removeProfilePassword(profile);
+                                return this.handleNodeCreationFailure(element);
+                            } else {
+                                return this.handleNodePasswordFailure(element);
```
Charles-Gagnon: These function names seem backwards to me - if we had a saved password we're calling handleNodeCreationFailure? But if we don't we call handleNodePasswordFailure? Is that intended?

src/controllers/vscodeWrapper.ts
abist: That comes from the vscode api

src/objectExplorer/objectExplorerService.ts
abist: Changed the comment

Charles-Gagnon: Ok, but why are you removing the password in all cases? What if the connection fails because of a network failure for example? I wouldn't expect (or want) the stored password to be removed in that case.

abist: Because if a connection wasn't made, there's no need to save password. We should only save password if a connection was made atleast once (a profile was made).

Charles-Gagnon: What about an existing connection that I'm reconnecting? So I successfully make and save a connection and then later on disconnect and try to reconnect but some general network error occurs. Won't it strip the saved password at that point so if I try to connect again it'll prompt me for the password even though the error had nothing to do with a wrong password?

Charles-Gagnon: Ok, so looking through it more closely it'll remove it from the profile as long as it was created and stored in the credential store. That seems fine then. Although I'm curious why you need to check if it's saved in the credential store - would you ever expect there to be a password in the profile that isn't then stored in the credential store? 

A scenario where that might be the case is if someone put a password field for an integrated auth profile. Not sure what would happen but I would expect we wouldn't necessarily store anything in the credential store. But in that case wouldn't it still be correct to strip out the password from the profile anyways? I'm not sure myself - it's a weird edge case. But it's something to consider since in general I think we should just avoid ever having passwords stored in that file (even if users enter it in themselves)

src/objectExplorer/objectExplorerService.ts
```diff
@@ -275,17 +306,21 @@ export class ObjectExplorerService {
...
-                            this._treeNodeToChildrenMap.set(element, [connectNode]);
-                            return [connectNode];
+                            // remove password if it's saved in the credential store
+                            let profile = element.connectionCredentials as IConnectionProfile;
+                            let password = await this._connectionManager.connectionStore.lookupPassword(profile);
+                            if (password) {
+                                await this.removeProfilePassword(profile);
+                                return this.handleNodeCreationFailure(element);
+                            } else {
+                                return this.handleNodeConnectionFailure(element);
```
Charles-Gagnon: I don't understand why you have these two node types based on whether a password was stored in the credential store. Isn't that something separate from whether a connection was successfully made? As mentioned in the other thread what about password-less connections such as integrated auth/AAD?

src/controllers/mainController.ts
abist: No, it actually updates the existing configuration instead of replacing it

Charles-Gagnon: What do you mean "updates"? Isn't the configuration in this case an array of profiles? How does it know which one to "update"?

src/objectExplorer/objectExplorerService.ts
abist: Yes, it was fixing this case - https://github.com/microsoft/vscode-mssql/issues/1438

You manually add a connection in the settings file with save password and it makes a connection. In that case, we need to remove the password from the `json`. This part is the error handling for that feature.

src/objectExplorer/objectExplorerService.ts
abist: Because we have 2 kinds of UX - one where only a password prompt is shown (was a login error) and one where the whole connection was unable to be made (server not found etc), in which case we go through the whole UX of recreating a profile with the same profile name.

src/controllers/mainController.ts
abist: https://github.com/microsoft/vscode-mssql/pull/1512/files#diff-59d53de3ec5d967a3079fee08a1391d7R345

abist: https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration

Charles-Gagnon: Yes. So first you get the configuration under the `mssql` section with the id `connections`. This should be an array of profiles
image

 And then you update the value of that `connections` config and set it to a single profile - the new one you cloned here. Am I missing something?

It should be simple to test. Have multiple profiles saved, then add the password field to one of them such that it runs through this logic. What does the config look like afterwards? 

If you say you've tested this and it works as expected I'll take your word for it - but my understanding of what is happening here doesn't line up with what you're saying.

abist: Yes, I tested with multiple connections already. From what I know and have seen, the update method updates the section with the new object. If you want to get rid of any object, you need to explicitly update with `undefined` - 

```
Note 2: To remove a configuration value use undefined, like so: config.update('somekey', undefined)
```

Charles-Gagnon: Especially since you're doing this kind of logic already but there you're calling update with an array of profiles instead.

https://github.com/Microsoft/vscode-mssql/blob/2e745d3d24344216b453c9f8db334775a9bae2a4/src/connectionconfig/connectionconfig.ts#L153
```
private writeProfilesToSettings(profiles: IConnectionProfile[]): Promise<void> {
        // Save the file
        const self = this;
        return new Promise<void>((resolve, reject) => {
            self._vscodeWrapper.getConfiguration(Constants.extensionName).update(Constants.connectionsArrayName, profiles, true).then(() => {
                resolve();
            }, err => {
                reject(err);
            });
        });
    }
```

abist: I'll update the function just to be safe, though I tested with multiple profiles

src/objectExplorer/objectExplorerService.ts
Charles-Gagnon: Ok - so it really isn't about the type of error that occurred is what I'm understanding. It's just that one flow is ONLY for re-entering a password to try again and the other is recreating the whole connection so is what you use for password-less connections.

The names here are really confusing and don't indicate that at all. "Node Creation Failure" and "Node Connection Failure" isn't really what's happening here at all - in both cases you failed to connect for some reason (which may or may not be related to a wrong password). I'd just call them "CreateSignInChildNode" and "CreateConnectChildNode"  or something to make it super clear what's happening and then explain in comments why you're doing the different ones here based on the password.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -244,6 +244,37 @@ export class ObjectExplorerService {
...
+    /**
+     * Removes password from a saved profile
+     */
+    private async removeProfilePassword(connection: IConnectionCredentials): Promise<void> {
+        // if the password is saved in the credential store, remove it
+        let profile = connection as IConnectionProfile;
+        await this._connectionManager.connectionStore.removeProfile(profile);
+        const newProfile = Object.assign({}, profile);
+        newProfile.password = '';
+        newProfile.savePassword = false;
```
Charles-Gagnon: So let's say I have save password set to true and then connect, but it fails due to some other issue (network issues or whatnot). Isn't this going to wipe out the password and then also make it not save the password that I enter in again? Why are you changing savePassword anyways?

src/objectExplorer/objectExplorerService.ts
```diff
@@ -244,6 +244,37 @@ export class ObjectExplorerService {
         return [new AddConnectionTreeNode()];
     }
 
+    /**
+     * Removes password from a saved profile
+     */
+    private async removeProfilePassword(connection: IConnectionCredentials): Promise<void> {
```
Charles-Gagnon: This seems like it should be something done in the connectionStore class since that should really be handling all the connection-specific logic.

src/objectExplorer/objectExplorerService.ts
abist: Ok, so when you make a new Object Explorer session, you sent an OE request, which responds with a boolean on whether that create OE call was successful or not. That's the `NodeCreationFailure` part (if `createSession` has an `undefined` response). However, this is not all. We then receive another notification for the result node for the OE. This can fail too (wrong login, firewall etc). That's the `NodeConnectionFailure` case. In this case, `createSession` does have a response, but it fails in the notification handler. 

I have a comment here - https://github.com/microsoft/vscode-mssql/pull/1512/files#diff-9401975d07f4df26ce1df2d776b8c746R309

Charles-Gagnon: How does this then get translated into the password existing in the store or not though? 

The call to createSession coming back with an undefined sessionId makes sense for what you're saying but I'm saying specifically here where you successfully made a session but the node itself wasn't created for some reason (as you said a connection error of some sort). 

Are you just reusing logic here or are these two situations (having a password stored vs not) actually somehow tied to whether a creation/connection failure occurred?

src/objectExplorer/objectExplorerService.ts
abist: It's because we prompt for password only if `savePassword` is true when creating a session - https://github.com/microsoft/vscode-mssql/pull/1512/files#diff-9401975d07f4df26ce1df2d776b8c746R371

So, when you have a disconnected node with an incorrect password, clicking on it to expand would prompt you for password again.

src/objectExplorer/objectExplorerService.ts
abist: The `handleNodeCreationFailure` case will give you a node under the connection, which when clicked will let you through the whole "Connection Retry" experience, which requires you to fill out the server, db etc all over again (since the connection was never made)

The `handleNodeConnectionFailure` case will give you a node under the connection, which when clicked will let you enter the password again (the only case I've noticed when we get an OE success response but OE notification failure). In all other cases, it failed in the initial create OE response itself. For the first create OE session call, it just looks at whether the connection details that are sent are null or not, and responds with a uri - https://github.com/microsoft/sqltoolsservice/blob/master/src/Microsoft.SqlTools.ServiceLayer/ObjectExplorer/ObjectExplorerService.cs#L182

src/controllers/vscodeWrapper.ts
Charles-Gagnon: It'd be good to update https://github.com/microsoft/vscode-mssql/blob/2e745d3d24344216b453c9f8db334775a9bae2a4/src/connectionconfig/connectionconfig.ts#L153 to use this as well

src/connectionconfig/connectionconfig.ts
```diff
@@ -150,7 +150,7 @@ export class ConnectionConfig implements IConnectionConfig {
         // Save the file
         const self = this;
         return new Promise<void>((resolve, reject) => {
-            self._vscodeWrapper.getConfiguration(Constants.extensionName).update(Constants.connectionsArrayName, profiles, true).then(() => {
+            self._vscodeWrapper.getConfiguration(Constants.extensionName).update(Constants.connectionsArrayName, profiles, vscode.ConfigurationTarget.Global).then(() => {
```
Charles-Gagnon: I meant update this to call _vscodeWrapper.setConfiguration directly

src/objectExplorer/objectExplorerService.ts
```diff
@@ -244,6 +244,26 @@ export class ObjectExplorerService {
         return [new AddConnectionTreeNode()];
     }
 
+    /**
+     * Handles a generic OE create session failure by creating a
```
Charles-Gagnon: Update these function descriptions

src/connectionconfig/connectionconfig.ts
abist: I did that and the tests started to fail for some reason so I reverted it

src/objectExplorer/objectExplorerService.ts
abist: it's already updated

PR #1514, opened 2019-12-02
Use correct database for new query when called from OE

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1435

Database connections are now different from server connections (https://github.com/microsoft/vscode-mssql/commit/16045d2be75e5f773a0cc928cf0f361204a0be40), so we won't have the issue of getting a new node for a database connection when starting a new query

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -166,6 +166,7 @@ export default class MainController implements vscode.Disposable {
                     } else if (databaseName === LocalizedConstants.defaultDatabaseLabel) {
                         connectionCredentials.database = '';
                     }
+                    treeNodeInfo.connectionCredentials = connectionCredentials;
                     await self.onNewQuery(treeNodeInfo);
                 }));
 


src/controllers/mainController.ts
```diff
@@ -166,6 +166,7 @@ export default class MainController implements vscode.Disposable {
                     } else if (databaseName === LocalizedConstants.defaultDatabaseLabel) {
                         connectionCredentials.database = '';
                     }
+                    treeNodeInfo.connectionCredentials = connectionCredentials;
```
Charles-Gagnon: I had this comment in another PR which I don't recall the resolution of, but why are you setting the connection credentials at this point? Wouldn't it be better to set it on whatever creates the TreeNodeInfo and have that set the database correctly there?

The issue I see here is that assuming at some point we add another context menu item to database nodes then we'll have to duplicate the logic being done here so that it also gets the correct connection credentials, which isn't great.

Charles-Gagnon: Yeah - that was in the previous you had for this https://github.com/microsoft/vscode-mssql/pull/1464 which was doing the same thing

abist: Yeah, but since then I have merged the database connections being different from server connections. It caused an issue earlier as it would create a new node when a new connection with a database was made. That's not the case any more.

Charles-Gagnon: That's fine - but again why is this logic being done here instead of when the TreeNode is created? If you hit refresh multiple times with this change made do the second and later calls then get a TreeNodeInfo with the database already set? (or does it create a new TreeNodeInfo every time?)

abist: So every node should have a database in their connection credentials, except server level connections with default connections. However, the way the registered commands work, we need to send in an arg, `currentNode` in this case (based on which node was right clicked). Most of the times if a node does have a database, it will run fine. This case is only when we right click a database node under a default server node and switch from a node with a specific database to a `default` server node. And thus, it just makes it easier for future commands as well.

Charles-Gagnon: That's reasonable - but again why is this being done in the refresh action handler? What happens when a new action is added and a user right clicks on that one first? If you didn't duplicate the logic there then it wouldn't have the correct info, right? That's why I'm saying this should be done in a more general location (such as when the TreeNodeInfo is created)

abist: It's being done in the `New Query` handler not refresh. But you're right, for any new context action, this will have to be done. One way to avoid this is to set the database names of every node when they are retrieved initially. 

I think I'll go with that solution in another PR because I don't want to make a lot of changes in the connections area right before release since I want to get testing done.

Charles-Gagnon: Hm, alright. I don't think it'd be that big of a change to do it there instead but I'll defer to you on that. Let's just hope there aren't other issues that stem from the info being "wrong" here. One concern I have is that if you're updating the info after refresh is called and something else is also using the info then that other thing will get different connections based on whether refresh was called first or not which isn't great (and is likely broken anyways like this was)

PR #1516, opened 2019-12-02
Add more tests 

abist:


test/firewallService.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        };
+        vscodeWrapper.setup(v => v.azureAccountExtension).returns(() => mockExtension);
+        firewallService = new FirewallService(client.object, vscodeWrapper.object);
+    });
+
+    test('isSignedIn Test', () => {
+        let isSignedIn = firewallService.isSignedIn;
+        assert.isNotTrue(isSignedIn, 'Firewall Service should not be signed in initially');
+        firewallService.isSignedIn = true;
+        assert.isTrue(firewallService.isSignedIn, 'Firewall Service should be signed in once set');
```
Charles-Gagnon: Not sure what the value in testing a getter/setter like this would be unless it's doing something more complicated in the function. I assume there's something else that sets the isSignedIn property once it detects that the user is signed in. Why aren't you testing that that sets it correctly?

test/initialization.test.ts
```diff
@@ -34,9 +34,9 @@ suite('Initialization Tests', () => {
 
     test('Connection manager is initialized properly', () => {
         // Wait for the extension to activate
-        ensureExtensionIsActive().then(() => {
+        ensureExtensionIsActive().then(async () => {
             // Verify that the connection manager was initialized properly
-            let controller: MainController = Extension.getController();
+            let controller: MainController = await Extension.getController();
             let connectionManager: ConnectionManager = controller.connectionManager;
             assert.notStrictEqual(undefined, connectionManager.client);
```
Charles-Gagnon: You need to make this test function async or use the done function to tell it that it's done. Right now it'll always pass because it'll finish immediately after calling ensureExtensionIsActive

test/languageService.test.ts
```diff
@@ -0,0 +1,68 @@
...
+            };
+            let testLogger: ILogger = {
+                logDebug: undefined,
+                increaseIndent: undefined,
+                decreaseIndent: undefined,
+                append: undefined,
+                appendLine: undefined
+
+            };
+            decompressProvider.decompress(testPackage, testLogger).catch((err) => {
```
Charles-Gagnon: Same here - this needs to be an async test function or use done

test/mainController.test.ts
```diff
@@ -99,9 +54,32 @@ suite('MainController Tests', async () => {
         // Resetting call back variables
         docUriCallback = '';
         newDocUriCallback = '';
-    });
 
-    mainController = await setupMaincontroller();
+
+        // Using the mainController that was instantiated with the extension
+        mainController = await Extension.getController();
```
Charles-Gagnon: Why are you refactoring this anyways? Was there an issue with how it was done before?

test/firewallService.test.ts
abist: I wanetd 100% code coverage for files so I need the getters and setters.

test/mainController.test.ts
abist: Yeah, the mainController tests fail otherwise

test/firewallService.test.ts
Charles-Gagnon: Code coverage should include code called further down the stack. So if you make other tests that change these that should get you that coverage as well - you don't have to test every method directly. 

Adding tests for code coverage is fine - but there's so much more useful tests you could be doing here. As an example I see promptForAccountSignIn sets this value to true if the account is signed in. That seems like a more valuable thing to test - that after calling promptForAccountSignIn when vscode reports that the account is signed in that the service is also updated correctly. Since that's a condition which could easily change and is probably important for your app to function. (and also conveniently tests that setting isSignedIn to true will result in that reporting as true correctly). 

Anyways I'm not going to go through and critique every single test - but adding tests just for the sake of pushing a number higher isn't really what we should be aiming for. Every test should have a purpose and the goal here should be to make every test cover a useful piece of functionality that you want to verify works as expected.

test/mainController.test.ts
Charles-Gagnon: And why did they fail? I assume they weren't failing before you made these changes...

abist: They were that's why I changed them. There's no other reason to change the tests otherwise.

test/firewallService.test.ts
abist: As you know these tests are unit tests. Some things like signing in is not possible without integration tests. To get full coverage, I had to use the getters/setters. It would be great to do everything in an ideal way, but when that's not possible, you need to resort to this. The goal of this PR is to get code coverage up to 70%. Adding integration tests is not in the scope of this PR and so you'll see getters and setters all around the code which need connections etc. for code coverage like any other unit test has.

Charles-Gagnon: That's what mocks are for 😄

test/mainController.test.ts
```diff
@@ -154,6 +114,23 @@ suite('MainController Tests', async () => {
         }
     });
 
+    // Renamed file event test
+    test('onDidCloseTextDocument should call renamedDoc function when rename occurs' , done => {
```
Charles-Gagnon: done is unnecessary here since this is a sync method. Just remove the try/catch and if an error is thrown it'll fail the test.

test/languageService.test.ts
```diff
@@ -0,0 +1,70 @@
...
+            let testLogger: ILogger = {
+                logDebug: undefined,
+                increaseIndent: undefined,
+                decreaseIndent: undefined,
+                append: undefined,
+                appendLine: undefined
+
+            };
+            try {
+                await decompressProvider.decompress(testPackage, testLogger);
```
Charles-Gagnon: You need to fail the test after the away - otherwise even if an error isn't throw the test will pass

test/initialization.test.ts
```diff
@@ -32,13 +32,14 @@ suite('Initialization Tests', () => {
         Telemetry.disable();
     });
 
-    test('Connection manager is initialized properly', () => {
+    test('Connection manager is initialized properly', (done) => {
         // Wait for the extension to activate
-        ensureExtensionIsActive().then(() => {
+        ensureExtensionIsActive().then(async () => {
```
Charles-Gagnon: If you're changing this to async anyways you might as well make the whole test function async and then can remove the done callback (same with the other similar tests)

test/firewallService.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        firewallService.isSignedIn = true;
+        let mockToken = {
+            expiresOn: new Date(),
+            resource: undefined,
+            tokenType: 'test',
+            accessToken: 'test_token'
+        }
+        firewallService.token = mockToken;
+        let result = await firewallService.createFirewallRule(server, startIpAddress, endIpAddress);
+        assert.isNotNull(result, 'Create Firewall Rule request is sent successfully');
```
Charles-Gagnon: I'd update the message to `Create Firewall Rule request wasn't sent successfully` - the message should be describing what the issue was since that's what's printed when it fails (same with the firewall test above)

test/mainController.test.ts
abist: The way the tests have been written, the done is required.

Charles-Gagnon: Why? I don't see how this wouldn't accomplish the same thing since you're never awaiting anything and don't have any callbacks.

```
// A renamed doc constitutes an openDoc event directly followed by a closeDoc event
mainController.onDidOpenTextDocument(newDocument);
mainController.onDidCloseTextDocument(document);

// Verify renameDoc function was called
connectionManager.verify(x => x.transferFileConnection(TypeMoq.It.isAny(), TypeMoq.It.isAny()), TypeMoq.Times.once());
assert.equal(docUriCallback, document.uri.toString());
assert.equal(newDocUriCallback, newDocument.uri.toString());
```

abist: I don't want to change old tests, but when I tried removing the `done`, the tests timed out. The mocha interface is also very flaky (especially the TDD interface we use, and there are tons of bugs on the repo), so I don't want to change them. If you can figure out how to not time out the tests while using the `done`, feel free to make the change.

Charles-Gagnon: Well, there's no reason to have it be async so I'd just remove that and what I had above should be fine. But up to you if you don't want to try and fix it.

abist:
@Charles-Gagnon I'm going to add more tests, this is the first iteration.

Issue #1517, opened 2019-12-03
Firewall rule dialog not shown when connecting to disconnect removed firewall server


abist:
Firewall service right now only shows the dialog when creating a connection.

Steps to repro -
1. Add a connection which needs a firewall rule.
2. Add firewall rule.
3. Remove firewall rule from Azure Portal.
4. Disconnect the node.
5. Connect node again.

Expected - Firewall dialog to show up
Actual - Only error is shown.

PR #1520, opened 2019-12-05
Improved UX for firewall scenarios

abist:
1. Now shows a notification when a firewall rule is sucessfully created.
2. Fixes https://github.com/microsoft/vscode-mssql/issues/1517

### DIFF
--- a/localization/xliff/enu/constants/localizedConstants.enu.xlf
+++ b/localization/xliff/enu/constants/localizedConstants.enu.xlf
@@ -252,7 +252,7 @@
             <source xml:lang="en">Azure Account extension activated and signed in.</source>
         </trans-unit>
         <trans-unit id="msgPromptFirewallRuleCreated">
-            <source xml:lang="en">Firewall successfully created.</source>
+            <source xml:lang="en">Firewall rule successfully created.</source>
         </trans-unit>
         <trans-unit id="msgInvalidIpAddress">
             <source xml:lang="en">Invalid IP Address </source>
--- a/localization/xliff/enu/constants/localizedConstants.enu.xlf
+++ b/localization/xliff/enu/constants/localizedConstants.enu.xlf
@@ -251,6 +251,9 @@
         <trans-unit id="msgPromptAzureExtensionActivatedSignedIn">
             <source xml:lang="en">Azure Account extension activated and signed in.</source>
         </trans-unit>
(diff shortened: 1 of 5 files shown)


localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -251,6 +251,9 @@
         <trans-unit id="msgPromptAzureExtensionActivatedSignedIn">
             <source xml:lang="en">Azure Account extension activated and signed in.</source>
         </trans-unit>
+        <trans-unit id="msgPromptFirewallRuleCreated">
+            <source xml:lang="en">Firewall successfully created.</source>
```
Charles-Gagnon: Firewall rule successfully created

would be better I think

src/objectExplorer/objectExplorerUtils.ts
```diff
@@ -54,4 +54,8 @@ export class ObjectExplorerUtils {
         }
         return LocalizedConstants.defaultDatabaseLabel;
     }
+
+    public static isFirewallError(errorMessage: string): boolean {
+        return errorMessage.includes(Constants.firewallErrorMessage);
+    }
```
Charles-Gagnon: This is really starting to get to be a huge annoyance that we don't return the error code. Have you looked into adding the error code to the session response? That would be way more preferable than doing stuff like this...for example what happens if the error message we get back is localized? The message is from the engine right? So it might not be in English in that case...

src/views/connectionUI.ts
```diff
@@ -470,52 +470,13 @@ export class ConnectionUI {
...
-                                            }
-                                        });
-                                    });
-                                }
-                                return undefined;
-                            });
-                        }
+                    const clientIp = this.connectionManager.failedUriToFirewallIpMap.get(uri);
+                    let success = await this.handleFirewallError(uri, profile, clientIp);
+                    if (success) {
```
Charles-Gagnon: You should be returning something here (undefined?) if not successful.

src/objectExplorer/objectExplorerUtils.ts
abist: That is something I planned to do right after making this PR. It would need some work on the tools service side so I don't want to do it now, but after this release I would definitely add error codes.

PR #1523, opened 2019-12-05
Check for ip address from firewall result

abist:
Found a bug in my own testing for firewall rules. Check explicitly if result was true and if we get an ipAddress back

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -389,7 +389,7 @@ export default class ConnectionManager {
                 Utils.showErrorMsg(Utils.formatString(LocalizedConstants.msgConnectionError, result.errorNumber, result.errorMessage));
                 // check whether it's a firewall rule error
                 let firewallResult = await this.firewallService.handleFirewallRule(result.errorNumber, result.errorMessage);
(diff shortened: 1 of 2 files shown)


src/controllers/connectionManager.ts
```diff
@@ -389,7 +389,7 @@ export default class ConnectionManager {
                 Utils.showErrorMsg(Utils.formatString(LocalizedConstants.msgConnectionError, result.errorNumber, result.errorMessage));
                 // check whether it's a firewall rule error
                 let firewallResult = await this.firewallService.handleFirewallRule(result.errorNumber, result.errorMessage);
-                if (firewallResult) {
+                if (firewallResult.result && firewallResult.ipAddress) {
```
Charles-Gagnon: What result would we get that didn't also have an ip address?

abist: false

Charles-Gagnon: As in we'd get a firewallResult object whose result property was set to false?

abist: Yes, that's right

Issue #1453, opened 2019-10-21
Intellisense stops working when query is saved


HSCelahaye:
- MSSQL Extension Version: 1.70
- VSCode Version: 1.39.2
- OS Version: Windows 10 Enterprise build version 1709

Steps to Reproduce:

1. Create New Query, do a standard select statement from DB
ex: SELECT *
FROM MEMBERS
2. Check if Intellisense works on individual columns, this should prepopulate the names
ex: SELECT MEMNAME, MEMLASTNAME
FROM MEMBERS 
3. Save the Query (this is where it breaks, the IntelliSense doesn't prepopulate anything for the DB, as if I disconnected)

I was able to roll back 1.60 and my IntelliSense works on saved queries there, but this last update broke it.

wzlwit:
And seems Visual Studio IntelliCode is required to be 1.19 to keep MSSQL 1.60 works.

mikelogaciuk:
I have the same issue. Intellisense isn't working...

jkarr:
I have this same issue.  MSSQL 1. 7.0, Visual Studio IntelliCode is 1.2.0 though, and it's only change description:
Added support for MSSQL.  And a link to a blog post that redirects to  Microsoft home page (https://aka.ms/vsicsql)

DanwithArduino:
Same issue for me, Intellisense is broken. MSSQL 1.7.1 and VSC 1.40.1

HSCelahaye:
> Same issue for me, Intellisense is broken. MSSQL 1.7.1 and VSC 1.40.1
> Definitely has something to do with when the file is saved... Guess I'll just copy and paste between two tabs for a while...
@DanwithArduino 
I personally recommend rolling back to MSSQL 1.6.0 for the time being, the previous version works fine for me.

DrDanRyan:
Any updates here? I have been pinned to 1.6.0 since this bug was introduced. I can't work without functioning Intellisense...

kburtram:
@abist have we confirmed this is fixed in 1.8.0 release?

HSCelahaye:
Thanks guys!

PR #1529, opened 2019-12-10
Fixed intellisense bug

abist:
Fixed intellisense stopped working after saving file - https://github.com/microsoft/vscode-mssql/issues/1453

1. `doc.uri.toString(true)` disables encoding when given a parameter `true`, so the connection URI would be the same as the file URI.
2. Remove `sql` from the `Resource Provider Language Client`, because it was sending double language changes to the server along with the actual `Sql Tools Service Client`.

### DIFF
--- a/src/languageservice/serviceclient.ts
+++ b/src/languageservice/serviceclient.ts
@@ -312,17 +312,11 @@ export default class SqlToolsServiceClient {
     }
 
     private createResourceClient(resourcePath: string): LanguageClient {
-        // Options to control the language client
-        let clientOptions: LanguageClientOptions = {
-            // documentSelector: ['sql'],
-            // synchronize: {
-            //     configurationSection: 'mssql'
-            // },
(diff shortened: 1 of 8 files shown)


src/languageservice/serviceclient.ts
```diff
@@ -314,11 +314,11 @@ export default class SqlToolsServiceClient {
     private createResourceClient(resourcePath: string): LanguageClient {
         // Options to control the language client
         let clientOptions: LanguageClientOptions = {
-            documentSelector: ['sql'],
-            synchronize: {
-                configurationSection: 'mssql'
-            },
-            errorHandler: new LanguageClientErrorHandler(this._vscodeWrapper)
+            // documentSelector: ['sql'],
```
Charles-Gagnon: Suggest just removing this - and possibly leaving a comment explaining why you aren't setting any options.

abist: Not sure how it showed up again, I removed this...

Issue #1540, opened 2019-12-10
Running a query steals focus from the editor


kburtram:
- MSSQL Extension Version: 1.8

Steps to Reproduce:

1. Open a query editor and run a query

Expected: Cursor is focused in the text editor after the query results are displayed.

Actual: Cursor focus is moved to the result grid.  If the result window is already visible when running the query (from a previous query execution) the cursor behavior appears correct.

PR #1543, opened 2019-12-11
Fix query focus for all scenarios

abist:
Retain focus on the editor when running query for the first time or when refreshing.

Fixes https://github.com/microsoft/vscode-mssql/issues/1540

### DIFF
--- a/src/models/sqlOutputContentProvider.ts
+++ b/src/models/sqlOutputContentProvider.ts
@@ -102,11 +102,12 @@ export class SqlOutputContentProvider {
         await this.runQueryCallback(statusView ? statusView : this._statusView, uri, title,
             async (queryRunner) => {
                 if (queryRunner) {
-                    // if the panel isn't active, bring it to foreground
-                    if (!this._panels.get(uri).isActive) {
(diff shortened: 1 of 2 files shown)


src/models/sqlOutputContentProvider.ts
```diff
@@ -298,7 +299,7 @@ export class SqlOutputContentProvider {
         const queryRunner = this.getQueryRunner(uri);
         // in case of a tab switch
         // and if it has rendered before
-        if (panelController.isActive &&
+        if (panelController.isActive !== undefined &&
```
Charles-Gagnon: What is this fixing? Seems odd to be fine with both true and false as values here.

abist: This is telling the webview to refresh if `panelController.isActive` isn't undefined, then it's being refreshed. 
This is fixing the scenario when let's say you have 2 editor tabs open - Untilted-1 and Untitled-2, with results for Untilted-2 open on the side. Then, running a query from Untitled-1 should bring the correct corresponding results to foreground, while maintaining focus on the editor.

Issue #1505, opened 2019-11-20
Copy from Resultview adds Tab before value


lucafurrer:
- MSSQL Extension Version: 1.7.1
- VSCode Version: 1.41.1
- OS Version: Windows 10 Enterprise 1809

Steps to Reproduce:

1. In Results view select a value which is not in the first column, copy value using context menu or Ctrl-C
2. Paste into text file

kburtram:
@abist do you know if this is fixed in 1.8?

abist:
@kburtram This is actually be design, since we preserve the columns and rows ordering if a disjointed or partial range is selected on the table. We can change the design to ignore this for the initial columns and rows.

lucafurrer:
This would be great. It is really a pain if you want to copy a value into Excel and it deletes the content of serveral columns to puts the value in a different cell than intended.

PR #1544, opened 2019-12-11
Fixed logic for copy paste

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1505


abist:
The PR validation seems to be failing because of a 503 error in the AppVeyor step.

PR #1545, opened 2019-12-11
Remove extra step for activating the extension

abist:
Show the `Create Firewall Rule` dialog directly and remove the `Activation` UX dialog from the Firewall feature.


abist:
The PR's seem to be stuck in the `license/cla Expected — Waiting for status to be reported` step

PR #1546, opened 2019-12-12
Increase code coverage

abist:
Will open for review once it hits 70.
Current coverage - 70.1%

Capture

Build - https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=49741&view=codecoverage-tab


test/serverStatus.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    let serverStatusView: ServerStatusView;
+
+    setup(() => {
+        serverStatusView = new ServerStatusView();
+    });
+
+    test('Test installing service status', () => {
+        serverStatusView.installingService();
+        setTimeout(() => {
```
Charles-Gagnon: This test is always going to pass. It'll call setTimeout but that returns immediately at which point the test will finish and pass. (you can test this by adding a throws in the callback). You can either wrap the setTimeout in a Promise and then resolving/rejecting within the setTimeout callback based on that or pass in done and then call done() at the end of the callback.

test/utils.test.ts
```diff
@@ -88,3 +88,48 @@ suite('Utility Tests - isSameConnection', () => {
...
+        assert.notEqual(signInToAzureCloudItem.description, undefined);
+        assert.equal(signInToAzureCloudItem.command, Constants.cmdAzureSignInToCloud);
+    });
+});
+
+suite('Utility tests - Timer Class', () => {
+    let timer = new Utils.Timer();
+
+    test('timer should start when initiated', () => {
+        setTimeout(() => {
```
Charles-Gagnon: Same issue here with these timer tests - they'll always pass.

PR #1550, opened 2019-12-17
Add build pipeline

Charles-Gagnon:
Part of microsoft/azuredatastudio#8528


build/build.yml
```diff
@@ -0,0 +1,116 @@
+pool:
+  name: Hosted macOS
+  demands:
+  - npm
+  - node.js
+
+steps:
+- task: ArchiveFiles@1
```
anthonydresser: ok in this instance these _actually_ are useless. At least in ads it was capturing compiled files. Here you can see the same code by just looking at the commit.

build/build.yml
```diff
@@ -0,0 +1,116 @@
...
+    archiveType: tar
+    archiveFile: '$(Build.BinariesDirectory)/source.tar.gz'
+
+- task: PublishBuildArtifacts@1
+  displayName: 'Publish Artifact: source'
+  inputs:
+    PathtoPublish: '$(Build.BinariesDirectory)/source.tar.gz'
+    ArtifactName: source
+
+- task: Npm@1
```
anthonydresser: We are using yarn now
We should replace all the npm commands with their equivalent yarn commands. Let me know if you need help with this.

build/build.yml
```diff
@@ -0,0 +1,116 @@
...
+  inputs:
+    SourceFolder: '$(Build.SourcesDirectory)'
+    Contents: '*.vsix'
+    TargetFolder: '$(Build.ArtifactStagingDirectory)'
+
+- task: PublishBuildArtifacts@1
+  displayName: 'Publish Artifact: drop'
+
+- script: |
+   brew update
```
anthonydresser: We should move to vscode's tester so we don't have to do this, but i guess thats a separate issue.

build/build.yml
Charles-Gagnon: Yeah - we should do a cleanup of these since there's a lot of useless/duplicate tasks that have built up over time. Was planning on doing that in a separate PR though as committer stuff later on.

build/build.yml
Charles-Gagnon: I'll check this in as is since that's what the build is currently doing - but I'm planning on spending some time cleaning this up during committer time.

Issue #1551, opened 2019-12-17
Move to vscode's new tester


anthonydresser:
@types/vscode for typings

vscode-testrunnner for testing

PR #1552, opened 2019-12-17
move to vscode-tester

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1551

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -238,7 +238,7 @@ gulp.task('ext:localization', gulp.series('ext:localization:xliff-to-ts', 'ext:l
 
 gulp.task('ext:build', gulp.series('ext:localization', 'ext:copy', 'ext:clean-library-ts-files', 'ext:compile', 'ext:compile-view')); // removed lint before copy
 
-gulp.task('ext:test', async (done) => {
+gulp.task('ext:test', async () => {
     let workspace = process.env['WORKSPACE'];
     if (!workspace) {
         workspace = process.cwd();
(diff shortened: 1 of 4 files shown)


gulpfile.js
```diff
@@ -237,24 +238,28 @@ gulp.task('ext:localization', gulp.series('ext:localization:xliff-to-ts', 'ext:l
...
+            extensionTestsPath: extensionTestsPath,
+            launchArgs: args
+        });
+    } catch (error) {
         console.log(`stdout: ${stdout}`);
         console.log(`stderr: ${stderr}`);
-        done();
-    });
+        console.error(`exec error: ${error}`);
+        process.exit(1);
```
anthonydresser: I don't think you need to process.exit anymore, I would just throw the error again and gulp will record it as failed

gulpfile.js
```diff
@@ -237,24 +238,28 @@ gulp.task('ext:localization', gulp.series('ext:localization:xliff-to-ts', 'ext:l
...
     process.env.JUNIT_REPORT_PATH = workspace + '/test-reports/ext_xunit.xml';
     var args = ['--verbose', '--disable-gpu', '--disable-telemetry', '--disable-updates', '-n'];
-    cproc.exec(`code --extensionDevelopmentPath="${workspace}" --extensionTestsPath="${workspace}/out/test" ${args.join(' ')}`,
-        { maxBuffer: 1024 * 500 },
-        (error, stdout, stderr) => {
-        if (error) {
-            console.error(`exec error: ${error}`);
-            process.exit(1);
-        }
+    let vscodePath = await vscodeTest.downloadAndUnzipVSCode('1.38.0');
```
anthonydresser: I would read this from our package.json rather than a hardcoded number

Issue #1506, opened 2019-11-20
Feature Request: Row Count on Status Bar


jwhite-sp:
Simply requesting a feature that would be of large benefit I use this add on and love it but, when performing queries it would be nice to have a row count of the current queried results in the status bar. There is some information in the messages but when performing queries I keep that minimized to save on my view and having a count on the status bar would save some trouble.

PR #1557, opened 2019-12-19
Add row count to status bar

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1506

Capture


src/views/statusView.ts
```diff
@@ -217,6 +226,19 @@ export default class StatusView implements vscode.Disposable {
         this.showStatusBarItem(fileUri, bar.sqlCmdMode);
     }
 
+    public showRowCount(fileUri: string, message?: string): void {
+        let bar = this.getStatusBar(fileUri);
+        if (message) {
+            bar.rowCount.text = message.slice(1, -1);
```
Charles-Gagnon: Might want to explain why you're doing this - I assume because there's a leading space or something?

src/views/statusView.ts
```diff
@@ -124,6 +132,7 @@ export default class StatusView implements vscode.Disposable {
         this.showStatusBarItem(fileUri, bar.statusQuery);
         this.showStatusBarItem(fileUri, bar.statusLanguageService);
         this.showStatusBarItem(fileUri, bar.sqlCmdMode);
+        this.showStatusBarItem(fileUri, bar.rowCount);
```
Charles-Gagnon: Is this going to look fine if the row count item doesn't have any text? i.e. if an error occurs or you switch away from a tab before executing a query?

src/controllers/queryRunner.ts
```diff
@@ -294,6 +294,11 @@ export default class QueryRunner {
 
         // Send the message to the results pane
         this.eventEmitter.emit('message', message);
+
+        // Set row count on status bar if there are no errors
+        if (!obj.message.isError) {
```
Charles-Gagnon: Will this be correctly cleared if you have a row count from a previously executed query and then run again but an error occurs?

src/controllers/webviewController.ts
```diff
@@ -52,17 +54,20 @@ export class WebviewPanelController implements vscode.Disposable {
             }
         ));
         this._panel.onDidDispose(() => {
+            this.statusView.hideRowCount(this.uri);
```
Charles-Gagnon: Shouldn't disposing the panel dispose the whole status bar?

abist: it's handled by the onCloseDocument event

src/views/statusView.ts
abist: Added comment

Issue #651, opened 2017-01-29
[FEAT-REQ] Command execution event history


eByte23:
I would think this would be a useful feature.

Have an output tray of 'SQL Execution History' List all commands that have run in order Desc order, most recent first and being able to view the result.

Possibly save result to json on disk? 
For some people this may be an issue if there is sensitive data so maybe a config option to allow this.

But it would be great to be able to see previous query results.

Thanks

kevcunnane:
Thanks for this suggestion. This is definitely something to be considered in the longer term, and discussions internally show there have been requests for something similar in other SQL tools.

As you mentioned, key considerations would be how many queries to keep, how to clean up given the possibility of multiple sessions being open at once. However it's technically feasible to implement this. 

Note that this is also a good candidate for external contributions if anyone wishes to pick this up - the core logic is in the SqlToolsService which would need updating, but that can be discussed since it would mostly be around disposal and state management for multiple queries.

PR #1561, opened 2020-01-09
Feature: Query History

abist:
Implemented Query History feature. 
Fixes https://github.com/microsoft/vscode-mssql/issues/651


abist:
@kburtram @Charles-Gagnon Gentle ping. Ideally would like to get this in before Thursday so we can do a design review.

Charles-Gagnon:
Please include screenshots of the feature.

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+        }
+        // return the query history sorted by timestamp
+        this._queryHistoryNodes.sort((a, b) => {
+            return (b as QueryHistoryNode).timeStamp.getTime()-
+                (a as QueryHistoryNode).timeStamp.getTime();
+        });
+        this._onDidChangeTreeData.fire();
+    }
+
+    getTreeItem(node: QueryHistoryNode): any {
```
Charles-Gagnon: return type should be QueryHistoryNode

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+        return node;
+    }
+
+    async getChildren(element?: any): Promise<vscode.TreeItem[]> {
+        if (this._queryHistoryNodes.length == 0) {
+            this._queryHistoryNodes.push(new EmptyHistoryNode());
```
Charles-Gagnon: Instead of making a new node each time why not just make one in the constructor and then if queryHistoryNodes is empty return an array that contains just that node instead? That saves you having to do the extra logic to clear it when adding a new node too.

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+                (a as QueryHistoryNode).timeStamp.getTime();
+        });
+        this._onDidChangeTreeData.fire();
+    }
+
+    getTreeItem(node: QueryHistoryNode): any {
+        return node;
+    }
+
+    async getChildren(element?: any): Promise<vscode.TreeItem[]> {
```
Charles-Gagnon: This doesn't need to be async

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+        if (this._queryHistoryNodes.length == 0) {
+            this._queryHistoryNodes.push(new EmptyHistoryNode());
+        }
+        return this._queryHistoryNodes;
+    }
+
+    /**
+     *
```
Charles-Gagnon: Missing doc comment

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+     */
+    public async showQueryHistoryCommandPalette(): Promise<void> {
+        const options = this._queryHistoryNodes.map(node => this._queryHistoryUI.convertToQuickPickItem(node));
+        let queryHistoryQuickPickItem = await this._queryHistoryUI.showQueryHistoryCommandPalette(options);
+        this.openQueryHistoryEntry(queryHistoryQuickPickItem.node, queryHistoryQuickPickItem.action ===
```
Charles-Gagnon: If the user cancels out of the dialog is queryHistoryPickItem going to be undefined?

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+        let credentials = this._connectionManager.getConnectionInfo(node.ownerUri).credentials;
+        await this._connectionManager.connect(uri, credentials, queryUriPromise);
+        await queryUriPromise;
+        this._statusView.languageFlavorChanged(uri, Constants.mssqlProviderName);
+        this._statusView.sqlCmdModeChanged(uri, false);
+        if (isExecute) {
+            const queryPromise = new Deferred<boolean>();
+            await this._outputContentProvider.runQuery(this._statusView, uri, undefined, title, queryPromise);
+            await queryPromise;
+            await this._connectionManager.connectionStore.removeRecentlyUsed(<IConnectionProfile>credentials);
```
Charles-Gagnon: This again...it really would be great to make it so when you run the query you tell it whether you want to add it to the recently used list instead of relying on removing it manually each time.

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+
+    /**
+     * Deletes a query history entry for a URI
+     */
+    public deleteQueryHistoryEntry(ownerUri: string): void {
```
Charles-Gagnon: Isn't the ownerUri going to be per-editor only? What happens if you execute multiple different queries from the same editor and then try to delete one - my guess looking at this logic is that it's just going to delete the first one it finds (not necessarily the one you selected)

If so couldn't you just pass in the actual node instead and save having to look it up via the URI?

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+     * Getters
+     */
+    public get queryHistoryNodes(): vscode.TreeItem[] {
+        return this._queryHistoryNodes;
+    }
+
+    /**
+     * Limits the size of a string with ellipses in the middle
+     */
+    public static limitStringSize(input: string, forCommandPalette: boolean = false): string {
```
Charles-Gagnon: This seems like something that can go in a generic utils file instead (if you do probably want to make the values params instead of being hardcoded like that. Explanations of why you chose these numbers would be useful too. What happens if you change the font size - do they still display correctly?)

src/views/queryHistoryUI.ts
```diff
@@ -0,0 +1,83 @@
...
+        let options = [{ label: LocalizedConstants.msgOpenQueryHistoryListing },
+            { label: LocalizedConstants.msgRunQueryHistoryListing }];
+        let question: IQuestion = {
+            type: QuestionTypes.expand,
+            name: 'question',
+            message: LocalizedConstants.msgChooseQueryHistoryAction,
+            choices: options
+        };
+        return this._prompter.promptSingle(question).then((answer: vscode.QuickPickItem) => {
+            if (answer) {
```
Charles-Gagnon: Would be good to explicitly return undefined (and update the return type) if an answer isn't chosen.

src/views/queryHistoryUI.ts
```diff
@@ -0,0 +1,83 @@
...
+                return this.showQueryHistoryActions(answer.node).then((actionAnswer: string) => {
+                    if (actionAnswer === LocalizedConstants.msgOpenQueryHistoryListing) {
+                        answer.action = QueryHistoryAction.OpenQueryHistoryAction;
+                    } else if ( actionAnswer === LocalizedConstants.msgRunQueryHistoryListing) {
+                        answer.action = QueryHistoryAction.RunQueryHistoryAction;
+                    }
+                    return answer;
+                })
+            }
```
Charles-Gagnon: Same here, explicitly return undefined if an answer isn't chosen

test/queryRunner.test.ts
```diff
@@ -72,6 +72,10 @@ suite('Query Runner tests', () => {
         // ... Mock up the view and VSCode wrapper to handle requests to update view
         testStatusView.setup(x => x.executingQuery(TypeMoq.It.isAnyString()));
         testVscodeWrapper.setup( x => x.logToOutputChannel(TypeMoq.It.isAnyString()));
+        let testDoc: vscode.TextDocument = {
```
Charles-Gagnon: Add tests

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+        if (this._queryHistoryNodes.length === 1) {
+            if (this._queryHistoryNodes[0] instanceof EmptyHistoryNode) {
+                this._queryHistoryNodes = [];
+            }
+        }
+        this._queryHistoryNodes.push(node);
+        // Push out the first listing if it crosses limit to maintain
+        // an LRU order
+        if (this._queryHistoryNodes.length > this._queryHistoryLimit) {
+            this._queryHistoryNodes.shift();
```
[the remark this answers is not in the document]
Charles-Gagnon: I guess this probably isn't critical given the implementation detail that this method is how nodes are added (so you should only ever overrun by 1 unless something else is modifying the array). I still think it's better to slice regardless to be safe.

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -251,9 +251,24 @@
...
             <source xml:lang="en">Firewall rule successfully created.</source>
         </trans-unit>
+        <trans-unit id="msgChooseQueryHistoryListing">
+            <source xml:lang="en">Choose Query History listing</source>
+        </trans-unit>
+        <trans-unit id="msgChooseQueryHistoryAction">
+            <source xml:lang="en">Choose An Action</source>
+        </trans-unit>
+        <trans-unit id="msgOpenQueryHistoryListing">
+            <source xml:lang="en">Open Query History Listing</source>
```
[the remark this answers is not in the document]
abist: I explicitly didn't want the names `Open Query` and `Run Query` because they clash with the normal commands to them. In order for the actions to have context, I think it's necessary to tell the user which query is being run or open. The context menu actions get their text from here - https://github.com/microsoft/vscode-mssql/pull/1561/files#diff-6e5249b55e8d29bc438070c22a9c9e34R11

What you are talking about is the text on command palette. Since this is the first pass so I can get this to the design review on Thursday, this is something that should come up on Thursday in the review with the PM's.

abist:
@Charles-Gagnon I'm going to make changes where necessary and dismiss the changes requested since I don't think any of these changes are blockers. Moreover, this isn't the final check-in, I want a design review with reviews from everyone.

Charles-Gagnon:
I see at least 2 potentially big issues that I would consider check-in blockers. 

1. What happens if a user cancels out of the quick pick dialog? Per my comment I'm not sure you're handling that correctly so will get an error
1. Deleting nodes doesn't seem like it'll be working correctly

The rest I mostly agree aren't super critical but I'd rather we spend the time now fixing any potential issues before merging in. Otherwise things tend to be forgotten and that just builds up more debt.

localization/xliff/enu/constants/localizedConstants.enu.xlf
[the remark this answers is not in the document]
Charles-Gagnon: That's fine - you can ignore the string comments until you have the design review then so that we have the full context of what the experience is going to look like.

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    /**
+     * Deletes a query history entry for a URI
+     */
+    public deleteQueryHistoryEntry(node: QueryHistoryNode): void {
+        let index = this._queryHistoryNodes.findIndex(n => {
+            let historyNode = n as QueryHistoryNode;
+            return (historyNode.ownerUri === node.ownerUri) &&
+                historyNode.timeStamp === node.timeStamp;
```
Charles-Gagnon: Still not sure why you're having this extra logic. Is the node object being returned not the same instance as the one in the nodes list? i.e. couldn't you just do n === historyNode? Using the timestamp is probably ok I guess but it seems more hacky than should be necessary

abist: I saw some weird behavior earlier, but I don't see it now. I'll remove the node itself.

test/queryRunner.test.ts
abist: Will add tests once the feature is fully in after the review.

abist:
@Charles-Gagnon screenshot

Capture

abist:
@Charles-Gagnon Would like to get this in and address the tests/concerns in a new PR since this one's growing.

Charles-Gagnon:
@abist There's still a lot of open comments I have - in particular have you looked at https://github.com/microsoft/vscode-mssql/pull/1561#discussion_r366995685 at all? If you're planning on fixing these at some point I don't see why you wouldn't just do that now so that they don't get lost in the mix.

Regardless I'm not going to sign off on this until someone else also looks at it. If @kburtram or someone thinks this is fine then I'll sign off but right now I think this needs some more looking at (especially without any tests to verify the behavior).

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+        this._onDidChangeTreeData.fire();
+    }
+
+    refresh(ownerUri: string, timeStamp: Date, hasError): void {
+        const timeStampString = timeStamp.toLocaleString();
+        const historyNodeLabel = this.createHistoryNodeLabel(ownerUri, timeStampString);
+        const tooltip = this.createHistoryNodeTooltip(ownerUri, timeStampString);
+        const queryString = this.getQueryString(ownerUri);
+        const connectionLabel = this.getConnectionLabel(ownerUri);
+        const node = new QueryHistoryNode(historyNodeLabel, tooltip, queryString,
```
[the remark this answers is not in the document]
abist: Because refresh function refreshes the tree.

abist:
@Charles-Gagnon Idk why GitHub wasn't showing all your comments, to me it seemed I addressed them all until you linked your comment. I'll address those. I need one sign off to get features in because I don't want to undergo the process of bringing huge features and breaking stuff. I'm going to merge this as soon as I get an approval and then follow up with a PR for tests/any bugs found/suggested UI changes.

src/queryHistory/queryHistoryProvider.ts
abist: There will be a rejected promise in that case. I'll add a check just to be sure.

src/queryHistory/queryHistoryProvider.ts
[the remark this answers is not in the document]
Charles-Gagnon: But this is actually creating a new node and adding it. Sure, doing that should probably also refresh the tree but by itself this is doing a lot more than the name "refresh" indicates.

abist:
Gentle Ping

gulpfile.js
```diff
@@ -36,13 +36,21 @@ gulp.task('ext:lint', () => {
...
+gulp.task('ext:copy-OE-assets', (done) => {
     return gulp.src([
         config.paths.project.root + '/src/objectExplorer/objectTypes/*'
     ])
     .pipe(gulp.dest('out/src/objectExplorer/objectTypes'));
 });
 
+// Copy icons for Query History
+gulp.task('ext:copy-queryHistory-assets', (done) => {
+    return gulp.src([
```
kburtram: I wonder if it would be better to have a `copy-icons` or `copy-resources` task instead, so when we have more icons\resources we don't have to keep adding tasks.  Then instead of putting icons under query history folder there could be a `media` folder or something that help all this type of asset.  It may work better as the extension grows?

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -251,9 +251,24 @@
         <trans-unit id="msgPromptFirewallRuleCreated">
             <source xml:lang="en">Firewall rule successfully created.</source>
         </trans-unit>
+        <trans-unit id="msgChooseQueryHistoryListing">
+            <source xml:lang="en">Choose Query History listing</source>
```
kburtram: the casing on `listing` is inconsistent with the spelling in the item below.  Is that intentional?

kburtram: ..also, not sure that "listing" is a great noun here.  Where did it some from?  I would probably go with something like "script" or similar.

package.nls.json
```diff
@@ -8,9 +8,18 @@
 "mssql.scriptDelete":"Script as Drop",
 "mssql.scriptExecute":"Script as Execute",
 "mssql.scriptAlter":"Script as Alter",
+"mssql.openQueryHistory":"Open Query",
+"mssql.runQueryHistory":"Run Query",
+"mssql.deleteQueryHistory":"Delete",
+"mssql.clearAllQueryHistory":"Clear All History",
```
kburtram: we seem to say "Query History" everywhere else, but just say "History" here.  I guess from the context in the UI it is clear that this text applies only to query history?

src/controllers/queryRunner.ts
```diff
@@ -160,6 +162,18 @@ export default class QueryRunner {
                     querySelection: selection
                 };
 
+                const doc = await this._vscodeWrapper.openTextDocument(this._vscodeWrapper.parseUri(this._ownerUri));
+                let queryString = doc.getText();
```
kburtram: personally I would put the `getText` call in the `else` of the below conditions.  It looks like if there is a selection we'll toss out the return value, which causes an unnecessary, potentially expensive, IPC to the main process.

src/controllers/queryRunner.ts
```diff
@@ -216,7 +230,9 @@ export default class QueryRunner {
             this._uriToQueryPromiseMap.delete(result.ownerUri);
         }
         this._statusView.executedQuery(result.ownerUri);
-        this.eventEmitter.emit('complete', Utils.parseNumAsTimeString(this._totalElapsedMilliseconds));
+        let hasError = false;
```
kburtram: nit: combine this line with the subsequent.  Why init to false and then immediately reassign a vale?

src/models/utils.ts
```diff
@@ -274,13 +274,13 @@ export function isSameConnection(conn: IConnectionCredentials, expectedConn: ICo
  * Check if a file exists on disk
  */
 export function isFileExisting(filePath: string): boolean {
-        try {
-            fs.statSync(filePath);
-            return true;
-        } catch (err) {
-            return false;
-        }
+    try {
```
kburtram: I'd recommend to send reformatting changes separately from functional changes.  It makes the functional change PRs unnecessarily complex, increasing the likelihood an important change will be missed.

abist: These changes came in due to linting before pre-commit, will remove next time

package.nls.json
abist: I got the text from ADS for it, but you make a good point. I'll add `Query History` for the command palette text, and let the context menu action string be the same.

gulpfile.js
abist: Let me address this in a separate PR.

Issue #1548, opened 2019-12-16
Copy from grid missing leading tabs with multi-select


kburtram:
- MSSQL Extension Version: 1.8

Steps to Reproduce:

1. Run a query that generate a result set
2. Select two different regions with different column sets using Ctrl-Select
3. Copy\Paste

Expected: Rows with missing values for leading columns get tabs so values are properly aligned

Actual: Leading tabs are removed.  For example, in the screenshot below there should be a leading tab in rows 7-10 in the editor so the LoginID values are correctly aligned with the LoginID header.

![image](https://user-images.githubusercontent.com/599935/70937592-6335a800-1ff9-11ea-8e78-064480ea0294.png)

PR #1562, opened 2020-01-13
fixed leading tabs when copying multiple selections

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1548

### DIFF
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -419,7 +419,9 @@ export default class QueryRunner {
 
         // Go through all rows and get selections for them
         let allRowIds = rowIdToRowMap.keys();
-        let firstColumn = this.getFirstSelectionColumn(rowIdToRowMap, rowIdToSelectionMap);
+        const endColumns = this.getSelectionEndColumns(rowIdToRowMap, rowIdToSelectionMap);
+        const firstColumn = endColumns[0];
+        const lastColumn = endColumns[1];
         for (let rowId of allRowIds) {
             let row = rowIdToRowMap.get(rowId);
             const rowSelections = rowIdToSelectionMap.get(rowId);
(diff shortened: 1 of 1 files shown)


src/controllers/queryRunner.ts
```diff
@@ -518,6 +516,23 @@ export default class QueryRunner {
         }
     }
 
+    /**
+     * Gets the first column where the selection started from
+     */
+    private getFirstSelectionColumn(rowIdToRowMap: Map<number, DbCellValue[]>, rowIdToSelectionMap: Map<number, ISlickRange[]>): number {
```
Charles-Gagnon: You should do the same for the trailing tabs as well.

abist: we don't hae trailing spaces after the last column, so this is not an issue there

Charles-Gagnon: I'm saying adding tabs for rows that don't have the "last" column selected. 

Repro :

image

I expect the copied text to be

```
name<tab>database_id
master<tab>
<tab>2
```

But with your change it will only this

```
name<tab>database_id
master
<tab>2
```

This isn't as big of a deal as the leading tabs, but we should be consistent with how we treat "empty" columns.

abist: Added tabs in the end as well

src/controllers/queryRunner.ts
```diff
@@ -428,20 +431,25 @@ export default class QueryRunner {
...
 
-            // start copy paste from left-most column
-            const firstColumn = rowSelections[0].fromCell;
-
             for (let i = 0; i < rowSelections.length; i++) {
                 let rowSelection = rowSelections[i];
+
+                // Add tabs starting from the first column of the selection
                 for (let j = firstColumn; j < rowSelection.fromCell; j++) {
                     copyString += ' \t';
```
Charles-Gagnon: Do you have any idea why we're adding an extra space here? Does that show up in the copied text after your fix? It doesn't seem like that should be included

abist: I'm not sure why it was there, good catch!

Issue #1555, opened 2019-12-19
Expanding two connections at same time causes incorrect nodes


kburtram:
- MSSQL Extension Version: 1.8

Steps to Reproduce:

1. Have multiple connections saved in Connections view
2. Expand a couple saved connections at the same time

Expected: Contents of each connection correctly displayed under respective connection node.

Actual: Contents from one connection shown under the wrong connection node.  For example, in the below screenshot note how a DB connections table contents are shown on a Server connection node.  The server connection should be showing a list of databases in this case.

![image](https://user-images.githubusercontent.com/599935/71133955-5360c400-21b1-11ea-8019-1bea8bcd416a.png)

PR #1568, opened 2020-01-16
Fix  expanding 2 nodes at once

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1555. 

There was a race condition where `self._currentNode` wasn't being set before we got the notification about the children. Explicitly look for parent now.

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -135,6 +135,7 @@ export class ObjectExplorerService {
                 const credentials = self._sessionIdToConnectionCredentialsMap.get(result.sessionId);
                 const children = result.nodes.map(node => TreeNodeInfo.fromNodeInfo(node, result.sessionId,
                     self._currentNode, credentials));
+                self._currentNode = self._rootTreeNodeArray.find(n => n.nodePath === result.nodePath);
                 self._treeNodeToChildrenMap.set(self._currentNode, children);
                 const expandParams: ExpandParams = {
                     sessionId: result.sessionId,
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -133,9 +133,9 @@ export class ObjectExplorerService {
         const handler = (result: ExpandResponse) => {
             if (result && result.nodes) {
                 const credentials = self._sessionIdToConnectionCredentialsMap.get(result.sessionId);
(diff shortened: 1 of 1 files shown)


src/objectExplorer/objectExplorerService.ts
```diff
@@ -135,6 +135,7 @@ export class ObjectExplorerService {
                 const credentials = self._sessionIdToConnectionCredentialsMap.get(result.sessionId);
                 const children = result.nodes.map(node => TreeNodeInfo.fromNodeInfo(node, result.sessionId,
                     self._currentNode, credentials));
+                self._currentNode = self._rootTreeNodeArray.find(n => n.nodePath === result.nodePath);
```
Charles-Gagnon: This isn't really going to solve the inherent problem that you have a huge potential for race conditions here. There's other callbacks that use self._currentNode so depending on the order the events are handled that could result in a lot of subtle bugs. 

In the end I'm not even sure what "currentNode" in this context would mean. The last one to react to an event? The currently selected one in the tree? Something else? I feel like the definition is pretty vague which is why you're having problems like this so would suggest looking at fixing that instead of doing one-off fixes like this (that might result in other issues popping up if things aren't expecting currentNode to be changed by this handler)

src/objectExplorer/objectExplorerService.ts
```diff
@@ -135,6 +135,7 @@ export class ObjectExplorerService {
                 const credentials = self._sessionIdToConnectionCredentialsMap.get(result.sessionId);
                 const children = result.nodes.map(node => TreeNodeInfo.fromNodeInfo(node, result.sessionId,
                     self._currentNode, credentials));
```
Charles-Gagnon: You use _currentNode here but your fix seems to indicate that the "currentNode" might not actually be the one you want. In that case isn't this going to result in you getting the children for a different node than the new one you find with the code you added and thus result in the same problem as before?

src/objectExplorer/objectExplorerService.ts
```diff
@@ -135,7 +135,8 @@ export class ObjectExplorerService {
...
-                self._treeNodeToChildrenMap.set(self._currentNode, children);
+                let parentNode = self._rootTreeNodeArray.find(n => n.nodePath === result.nodePath);
```
Charles-Gagnon: You're still setting the parent node of the children to self._currentNode which doesn't seem correct

PR #1572, opened 2020-01-21
Proper fix for race condition in expanding nodes

abist:
I realized that https://github.com/microsoft/vscode-mssql/pull/1568 was not a correct fix for this bug - https://github.com/microsoft/vscode-mssql/issues/1555.

This is a proper fix. The fix above was incorrect because it only looked for parents in the server nodes. I didn't realize I didn't `await` any of those scripting operations and hence was getting a race condition.

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -223,32 +223,32 @@ export default class MainController implements vscode.Disposable {
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
                     Constants.cmdScriptSelect, async (node: TreeNodeInfo) => {
-                        this.scriptNode(node, ScriptOperation.Select, true);
+                        await this.scriptNode(node, ScriptOperation.Select, true);
                     }));
 
                 // Script as Create
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
                     Constants.cmdScriptCreate, async (node: TreeNodeInfo) =>
-                    this.scriptNode(node, ScriptOperation.Create)));
+                    await this.scriptNode(node, ScriptOperation.Create)));
 
                 // Script as Drop
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
                     Constants.cmdScriptDelete, async (node: TreeNodeInfo) =>
-                    this.scriptNode(node, ScriptOperation.Delete)));
+                    await this.scriptNode(node, ScriptOperation.Delete)));
 
                 // Script as Execute
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
                     Constants.cmdScriptExecute, async (node: TreeNodeInfo) =>
-                    this.scriptNode(node, ScriptOperation.Execute)));
+                    await this.scriptNode(node, ScriptOperation.Execute)));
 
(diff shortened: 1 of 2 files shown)


src/objectExplorer/objectExplorerService.ts
```diff
@@ -133,10 +133,9 @@ export class ObjectExplorerService {
         const handler = (result: ExpandResponse) => {
             if (result && result.nodes) {
                 const credentials = self._sessionIdToConnectionCredentialsMap.get(result.sessionId);
-                let parentNode = self._rootTreeNodeArray.find(n => n.nodePath === result.nodePath);
                 const children = result.nodes.map(node => TreeNodeInfo.fromNodeInfo(node, result.sessionId,
-                    parentNode, credentials));
-                self._treeNodeToChildrenMap.set(parentNode, children);
+                    self._currentNode, credentials));
```
abist: Need to set `self._currentNode` because that's the selected node when it's expanded. Looking into just rootNodeArray would fail as we keep expanding and keep going deeper.

Charles-Gagnon: I still really don't like storing _currentNode like this. What exactly is preventing a race condition with the _currentNode where simultaneous actions update the _currentNode before something like this has finished executing?

src/controllers/mainController.ts
```diff
@@ -223,32 +223,32 @@ export default class MainController implements vscode.Disposable {
                 this._context.subscriptions.push(
                     vscode.commands.registerCommand(
                     Constants.cmdScriptSelect, async (node: TreeNodeInfo) => {
-                        this.scriptNode(node, ScriptOperation.Select, true);
+                        await this.scriptNode(node, ScriptOperation.Select, true);
```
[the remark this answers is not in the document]
abist: I ust verified it. Basically it wasn't waiting before setting the `self._currentNode`

src/objectExplorer/objectExplorerService.ts
abist: because it's set right when a node is clicked on and hence the `await` will wait for the notification handler to finish doing it work before setting the next `_currentNode`

abist:
Gentle Ping

src/objectExplorer/objectExplorerService.ts
```diff
@@ -128,24 +130,36 @@ export class ObjectExplorerService {
         return handler;
     }
 
+    private getParentFromExpandParams(params: ExpandParams): TreeNodeInfo {
```
Charles-Gagnon: TreeNodeInfo | undefined

src/objectExplorer/objectExplorerService.ts
```diff
@@ -160,12 +174,14 @@ export class ObjectExplorerService {
             nodePath: node.nodePath
         };
         this._expandParamsToPromiseMap.set(expandParams, promise);
+        this._expandParamsToTreeNodeInfoMap.set(expandParams, node);
```
Charles-Gagnon: Instead of 2 separate maps I have a slight preference to just a single map where the value contains the promise and node. Just less to keep track of when deleting/updating the map.

Issue #1472, opened 2019-11-05
connection string can't work finely


lokiiart:
- MSSQL Extension Version: Docker-mssql-lasted
- VSCode Version: 1.39.2
- OS Version: Linux x64 5.3.8-arch

Steps to Reproduce:

1. add different connections with connection string
2. connections link only the last instance

It happens only when you connect them by connection string. if set them step by step , it works fine.

Issue #862, opened 2017-05-03
Encrypt connection strings


benrr101:
It is imperative from a security perspective that we encrypt sensitive information that users input into our tools. With the newly added support for creating connections using plain connection strings, it is likely that some users (particularly Azure users) will copy and paste in connection strings that contain passwords. The current implementation would store the sensitive connection string in the user's preferences in plain text. Although complex solutions to encrypt just the password would be ideal, simply encrypting the entire connection string would be suitable, and we already have a credential storage service for this purpose.

Proposed implementation is:
If a connection string is used to connect to a database, store the entire connection string in the credentials service. Retrieve it when connecting and pass to the tools service.

PR #1577, opened 2020-01-28
Support OE features with connection strings

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1472

Also fixes https://github.com/microsoft/vscode-mssql/issues/862

The issue was that it was expecting a password for credentials with a connection string.


src/objectExplorer/objectExplorerService.ts
```diff
@@ -377,7 +377,8 @@ export class ObjectExplorerService {
             connectionCredentials = await connectionUI.showConnections(false);
         }
         if (connectionCredentials) {
-            if (ConnectionCredentials.isPasswordBasedCredential(connectionCredentials)) {
+            if (ConnectionCredentials.isPasswordBasedCredential(connectionCredentials) &&
```
Charles-Gagnon: If the connection string doesn't have a password but is using SQL auth what does that experience look like? I would expect it to still query for the password at that point since we don't necessarily want to require users putting their password in the connection string.

abist: This assumes that if the connection is SQL Auth, it will have the credentials in the string. We just send the connection string as is to the sqltoolsservice just like ADS. Querying for the password in the connection string is something we can look into for future for both the extension and ADS.

Charles-Gagnon: What's the experience like? I assume if they don't an error will pop up telling them that they need to enter a password or something similar?

abist:
@Charles-Gagnon Made an update to convert a profile using connection strings to a normal profile and saving the connection string in the credential service.

src/models/connectionCredentials.ts
```diff
@@ -281,6 +281,14 @@ export class ConnectionCredentials implements IConnectionCredentials {
         return authenticationType === utils.authTypeToString(AuthenticationTypes.SqlLogin);
     }
 
+    public static isPasswordBasedConnectionString(connectionString: string): boolean {
+        return connectionString.toLowerCase().includes('password');
+    }
+
+    public static isConnectionStringSaved(connectionString: string): boolean {
+        return connectionString.includes(ConnectionStore.CRED_CONNECTION_STRING_PREFIX);
```
Charles-Gagnon: Would rather we use a separate field for this (like you do with the password). The phrase "connectionString:" could end up in a normal connection string (such as if a database contained that string). And given our discussion of the security concerns I would say it's worth making this more robust for that case.

src/models/connectionStore.ts
```diff
@@ -324,13 +329,21 @@ export class ConnectionStore {
...
+    public saveProfileWithConnectionString(profile: IConnectionProfile): Promise<boolean> {
+        if (!profile.connectionString) {
+            return Promise.resolve(true);
+        }
+        return this.doSavePassword(profile, CredentialsQuickPickItemType.Profile, true);
+    }
+
+    private doSavePassword(conn: IConnectionCredentials, type: CredentialsQuickPickItemType, isConnectionString: boolean = false): Promise<boolean> {
         let self = this;
+        let password = isConnectionString ? conn.connectionString : conn.password;
```
Charles-Gagnon: When you make changes like this it would really be great if you also updated the function and variable names to properly reflect the changes. For example this is no longer specifically for a password so I'd say something like doSaveCredential is more appropriate and will reduce the chance someone misinterprets what this method does in the future.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -377,23 +377,32 @@ export class ObjectExplorerService {
...
-                    if (!(<IConnectionProfile>connectionCredentials).savePassword) {
-                        // prompt for password
-                        password = await this._connectionManager.connectionUI.promptForPassword();
-                        if (!password) {
-                            promise.resolve(undefined);
-                            return undefined;
+            // connection string based credential
+            if (connectionCredentials.connectionString) {
+                if (ConnectionCredentials.isConnectionStringSaved(connectionCredentials.connectionString)) {
+                    // look up password
```
Charles-Gagnon: You're looking up a connection string here, not a password

src/models/connectionCredentials.ts
```diff
@@ -281,6 +281,14 @@ export class ConnectionCredentials implements IConnectionCredentials {
         return authenticationType === utils.authTypeToString(AuthenticationTypes.SqlLogin);
     }
 
+    public static isPasswordBasedConnectionString(connectionString: string): boolean {
+        return connectionString.toLowerCase().includes('password');
```
Charles-Gagnon: Similar to the connectionString check below - password is something I could easily see a user having in a database/server name. This is a bit trickier to solve though - I don't know offhand if there's any libraries we could use for this but I would suggest looking around to see how others are solving this problem.

Issue #1322, opened 2019-09-24
Closing SQL editor doesn't close associated results editor


Charles-Gagnon:
1. Execute query so that results editor appears
2. Close original SQL editor

Results editor stays open and must be manually closed as well

PR #1579, opened 2020-01-29
Close results when closing associated editor

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1322

### DIFF
--- a/src/models/sqlOutputContentProvider.ts
+++ b/src/models/sqlOutputContentProvider.ts
@@ -294,6 +294,7 @@ export class SqlOutputContentProvider {
                 value.timeout = this.setRunnerDeletionTimeout(key);
             }
         }
+        this.closeResultsTab(doc.uri.toString());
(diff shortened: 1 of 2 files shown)


src/models/sqlOutputContentProvider.ts
```diff
@@ -403,6 +403,16 @@ export class SqlOutputContentProvider {
...
     }
 
+    /**
+     * Closes the associated results tab when a query editor is closed
+     */
+    public closeResultsTab(uri: string): void {
+        let controller = this._panels.get(uri);
+        controller.dispose();
+        this._panels.delete(uri);
+        this._queryResultsMap.delete(uri);
```
Charles-Gagnon: Is this not being called properly? https://github.com/microsoft/vscode-mssql/blob/master/src/models/sqlOutputContentProvider.ts#L285

PR #1596, opened 2020-02-18
Remove all custom events

Charles-Gagnon:


abist:
Why are we removing these?

Issue #1591, opened 2020-01-31
Extension has a lot of output channels


Charles-Gagnon:
I'm seeing 4 separate channels, which seems like too many (at the very least the duplicate MSSQL one shouldn't be there)

![image](https://user-images.githubusercontent.com/28519865/73576011-4ed52f80-442e-11ea-90d7-75f405b75bb4.png)

PR #1599, opened 2020-02-20
Merge output channels to one MSSQL output channel

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1591


src/prompts/adapter.ts
```diff
@@ -19,9 +19,7 @@ export default class CodeAdapter implements IPrompter {
     constructor(
         private vscodeWrapper: VscodeWrapper
     ) {
-        // TODO Decide whether output channel logging should be saved here?
-        this.outChannel = this.vscodeWrapper.createOutputChannel(Constants.outputChannelName);
-        // this.outChannel.clear();
+        this.outChannel = this.vscodeWrapper.outputChannel;
```
Charles-Gagnon: I'm not understanding how this is removing any channels. Both this and the vscodeWrapper were using the same channel name - MSSQL

Charles-Gagnon: Oh this is just getting rid of the duplicate MSSQL one. Ok, that's fine. I'm not sure what the standard is for extensions contributing channels, but 3 seems like a lot for a single extension. The SqlToolsService initialization errors in particular seem like something that would be better to just have in a log file or something instead - that's rarely going to be something users need/want to check.

abist: It's always been like this, and I don't want to make a change unless we got a customer issue about it since the users are probably used to it. The duplicate `MSSQL` was clearly a bug however.

Issue #1585, opened 2020-01-31
Can't run queries with profiles made from connection strings


Charles-Gagnon:
1. Create new profile from connection string
2. Close VS Code
3. Re-open and select New Query from that profile
4. The query window never connects, and trying to run a query with that profile results in this in the console log

![image](https://user-images.githubusercontent.com/28519865/73568205-cd28d600-441c-11ea-9e81-51ead2942871.png)

PR #1600, opened 2020-02-21
Fix connection to connection string based credentials

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1585


src/objectExplorer/objectExplorerUtils.ts
```diff
@@ -32,6 +32,11 @@ export class ObjectExplorerUtils {
 
     public static getNodeUriFromProfile(profile: IConnectionProfile): string {
         let uri: string;
+        if (profile.connectionString) {
+            let fields = profile.connectionString.split(';').filter(s => !s.toLowerCase().includes('password'));
```
Charles-Gagnon: Edge case - but what if the password contains a semi-colon? Reading around it seems like {} is used to escape it (and then double up }'s to escape those). 

Might be worth looking into a library to do this for you. I'm not a huge fan of adding dependencies for one small edge case but possibly you could at least see what they do and handle it the same way.

abist: Will add additional logic for edge case in a separate PR

Issue #1606, opened 2020-02-25
Refreshing OE right after creating object does not show new object


abist:
Steps to repro -

1. Create/drop a new database/table.
2. Refresh the parent (or any node) up in the OE hierarchy.

Expected - New object should appear or be removed from the OE.
Actual - Nothing happens

PR #1607, opened 2020-02-25
Fix refresh mechanism to not use current node

abist:
Remove the old `currentNode` mechanism to refresh a node, and use the node directly when right clicking. 

Fixes https://github.com/microsoft/vscode-mssql/issues/1606

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -467,8 +467,11 @@ export class ObjectExplorerService {
             sessionId: node.sessionId,
             nodePath: node.nodePath
         };
-        await this._connectionManager.client.sendRequest(RefreshRequest.type, refreshParams);
-        return this._objectExplorerProvider.refresh(undefined);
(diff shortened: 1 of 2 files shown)


src/objectExplorer/objectExplorerService.ts
```diff
@@ -467,8 +467,11 @@ export class ObjectExplorerService {
             sessionId: node.sessionId,
             nodePath: node.nodePath
         };
-        await this._connectionManager.client.sendRequest(RefreshRequest.type, refreshParams);
-        return this._objectExplorerProvider.refresh(undefined);
+        let response = await this._connectionManager.client.sendRequest(RefreshRequest.type, refreshParams);
+        if (response) {
+            this._treeNodeToChildrenMap.delete(node);
```
Charles-Gagnon: Why are you adding this? It's not clear how this is tied to your changes.

abist: It's not needed, but good to have

Issue #1613, opened 2020-02-27
Disconnected server in OE with no firewall rule doesn't show popup


abist:
There was a regression where connecting to existing disconnected server wouldn't pop up a Firewall Rule dialog. 

Steps to repro -
1. Add server with no firewall rule to OE.
2. Add firewall rule using the extension.
3. Disconnect the server.
4. Remove the firewall rule from the portal.
5. Re-connect to server in the extension.

Expected - Firewall rule dialog should show up.
Actual - Only the firewall error pops up.

PR #1614, opened 2020-02-27
Fix firewall issue for disconnected server

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1613

This was a regression when connecting to an existing connection in OE which doesn't have firewall rule set up.

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -293,6 +293,10 @@ export class ObjectExplorerService {
 
     async getChildren(element?: TreeNodeInfo): Promise<vscode.TreeItem[]> {
         if (element) {
+            // set current node for very first expansion of disconnected node
+            if (this._currentNode !== element) {
+                this._currentNode = element;
+            }
             // get cached children
             if (this._treeNodeToChildrenMap.has(element)) {
                 return this._treeNodeToChildrenMap.get(element);


src/objectExplorer/objectExplorerService.ts
```diff
@@ -293,6 +293,10 @@ export class ObjectExplorerService {
 
     async getChildren(element?: TreeNodeInfo): Promise<vscode.TreeItem[]> {
         if (element) {
+            // set current node for very first expansion of disconnected node
```
Charles-Gagnon: So the comment isn't really accurate since it'll do this for every single node that children are requested of. So would be preferable to update that.

But in general - how is this fixing the issue? You set the currentNode whenever a new item is selected in the tree - is that not happening when expanding a node? I would expect it would since it does actually select the node it expands.

abist: @Charles-Gagnon The issue is that it does set the current node, but that's after the `getChildren` call, which needs its value.

Charles-Gagnon: Where is it using the currentNode that's failing? You have the "current" node already (element param) so why wouldn't you just be using that like the rest of the getChildren method is doing?

abist: @Charles-Gagnon That's exactly the issue. When a node with a command is clicked or right clicked for context menu, we get the node parameter. This is the case where there's a collapsed disconnected server, which when clicked on goes to the `getChildren` call from the provider. 

Now, when there's a failure (firewall rule in this case), it goes here - https://github.com/microsoft/vscode-mssql/pull/1614/files#diff-9401975d07f4df26ce1df2d776b8c746R105

Since object explorer gives notifications about sessions instead of callbacks. Now, at this point, the current node isn't set, and hence this fails. The fix addresses this issue.

Issue #1499, opened 2019-11-14
NULL values show as link in XML columns


kburtram:
- MSSQL Extension Version: 1.7.1

Steps to Reproduce:

1. Run a query with XML columns where some values are null.  For example, the below query with AdventureWorks2014 sample.

```SQL
SELECT [CatalogDescription] ,[Instructions]
FROM [AdventureWorks2014].[Production].[ProductModel]
```

Expected: NULL values are not links

Actual: NULL values are links. Clicking opens untitled doc with text `NULL` in it.

![image](https://user-images.githubusercontent.com/599935/68897605-b5299c00-06e2-11ea-85f6-6f5a3520b4e4.png)

PR #1621, opened 2020-02-28
Remove link for NULL cells of type XML

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1499


src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -656,7 +660,7 @@ export class AppComponent implements OnInit, AfterViewChecked {
     /**
      * Format all text to replace all new lines with spaces and performs HTML entity encoding
      */
-    textFormatter(row: number, cell: any, value: any, columnDef: any, dataContext: any): string {
+    static textFormatter(row: number, cell: any, value: DbCellValue, columnDef: any, dataContext: any): string {
```
abist: Had to make it static because slickgrid doesn't have reference to `this` or `self`

PR #1625, opened 2020-03-03
Fix results resizing when resizing the window

abist:
This PR fixes double scrollbars to appear when resizing the whole VSCode window. The old PR https://github.com/microsoft/vscode-mssql/pull/1559/files only fixed issues with resizing using the resizing handle and toggling the results and messages pane.

### DIFF
--- a/src/views/htmlcontent/src/js/components/app.component.ts
+++ b/src/views/htmlcontent/src/js/components/app.component.ts
@@ -827,9 +827,8 @@ export class AppComponent implements OnInit, AfterViewChecked {
      * Resizes the results pane
      */
     resizeResults(): void {
-        const self = this;
         let scrollableHeight = $('.results.vertBox.scrollable').get(0).clientHeight;
-        $('.horzBox').get(0).style.height = `${scrollableHeight - self._resultsPaneBoundary}px`;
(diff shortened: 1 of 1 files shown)


src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -138,7 +142,7 @@ export class AppComponent implements OnInit, AfterViewChecked {
     private selectionModel = 'DragRowSelectionModel';
     private slickgridPlugins = ['AutoColumnSize'];
     private _rowHeight = 29;
-    private _resultsPaneBoundary = 2;
+    private _resultsPaneBoundary = 22;
```
abist: This was a bug

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -817,6 +823,16 @@ export class AppComponent implements OnInit, AfterViewChecked {
         });
     }
 
+    /**
+     * Resizes the results pane
+     */
+    resizeResults(): void {
+        const self = this;
```
kburtram: not sure why we need to add self in this case, since we're not using any lambdas or callbacks.  Also inconsistent self\this usage below.

PR #1630, opened 2020-03-04
fix: string highlighting containing escaped characters

sukano:
Issue : When a string enclose by single quote contains an escaped single quote, the higlight of the string stop at the escaped single quote and not at the end of the string (show below).

image

### DIFF
--- a/syntaxes/SQL.plist
+++ b/syntaxes/SQL.plist
@@ -654,7 +654,7 @@
 					<key>comment</key>
 					<string>this is faster than the next begin/end rule since sub-pattern will match till end-of-line and SQL files tend to have very long lines.</string>
 					<key>match</key>
-					<string>(N)?(')[^']*(')</string>
+					<string>(N)?(')(?:[^'\\]|\\.)*(')</string>
 					<key>name</key>
 					<string>string.quoted.single.sql</string>
 				</dict>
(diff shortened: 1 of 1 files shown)


abist:
@sukano Thanks for opening this PR! Looks good to me!

gontadu:
It seems the issue still exists - and I didn;t have the issue in 1.8.0
The string below:
select REPLACE(SES.LOGIN_NAME,'UNIZA\','') AS LOGIN_NAME

doesn't highlight properly:
image

If the \ is removed then it seems fine:
image

This suggests there is some incorrect recognition of escape characters here.

sukano:
@gontadu can you test with `select REPLACE(SES.LOGIN_NAME,'UNIZA','*****') AS LOGIN_NAME` and provide a print Screen.

thanks.

AO-TERRA:
I am also seeing the same behaviour as @gontadu 
As seen here:
Code_16-03_16-42-38

In longer code especially if this occurs near the top it renders the rest of the code completely as a string (brown colouring). 
I have confirmed that reverting back the V1.8.0 of the extension will fix the issue.

gontadu:
> @gontadu can you test with `select REPLACE(SES.LOGIN_NAME,'UNIZA','*****') AS LOGIN_NAME` and provide a print Screen.
> 
> thanks.

This seems to be fine.
image

PR #1632, opened 2020-03-05
Query history for non refreshed queries

abist:
This PR fixes an issue where when you switched (refreshed) a tab, it would appear on the query history tree.


abist:
@aaomidi We already have a complete event, which is fired when a query is complete (even refresh) so we don't want to confuse with that.

PR #1644, opened 2020-03-10
Update packages

abist:
Update/Substitute packages


src/languageservice/decompressProvider.ts
```diff
@@ -7,18 +7,59 @@
 
 import {IDecompressProvider, IPackage} from './interfaces';
 import  {ILogger} from '../models/interfaces';
-const decompress = require('decompress');
+import { PlatformInformation } from '../models/platform';
+
+const DecompressZip = require('decompress-zip');
```
Charles-Gagnon: Why are you using require here? Imports are the preferred way

src/languageservice/decompressProvider.ts
```diff
@@ -7,18 +7,59 @@
...
+            onwarn: (warn) => {
+                if (warn.data && !warn.data.recoverable) {
+                    logger.appendLine(`[ERROR] ${warn.message}`);
+                }
+            }
+        }, () => { logger.appendLine(`Done! ${totalFiles} files unpacked.\n`); });
+    }
+
+    public decompress(pkg: IPackage, logger: ILogger): Promise<void> {
+        if (this._isWindows) {
```
Charles-Gagnon: Why are you relying on the platform for determining what type of decompression to do? Looking at the file extension instead is the usual way of doing this.

src/languageservice/decompressProvider.ts
Charles-Gagnon: Any reason not to use yauzl like the rest of our stuff?

src/languageservice/decompressProvider.ts
abist: We cannot use the file extension here because it's a temporary file with a `.tmp` extension

src/languageservice/decompressProvider.ts
abist: Not really, I just used the most used libraries

src/languageservice/decompressProvider.ts
Charles-Gagnon: I would suggest doing what I did for service-downloader and instead of using .tmp use the same file extension as the downloaded file (I assume the logic is the same here). It just makes more sense to do that anyways - there's no good reason we'd want to rename it to .tmp.

Charles-Gagnon: (and the main reason I'm asking is because relying on the platform is making a race condition possible here if this function is called before the platform query returns)

src/languageservice/serviceDownloadProvider.ts
```diff
@@ -99,11 +99,14 @@ export default class ServiceDownloadProvider {
             this._logger.appendLine(`${Constants.serviceInstallingTo} ${installDirectory}.`);
             const urlString = this.getGetDownloadUrl(fileName);
 
+            const isZipFile: boolean = path.extname(fileName) === '.zip';
```
Charles-Gagnon: nit: but I'd suggest just having this be fileExt and have the calling code determine what they want to do with it. Just a bit more generic (there's more archive types than just zip/tar)

Issue #1640, opened 2020-03-06
Ctrl + A should have better behavior


abist:
Ctrl + A behavior for grid selection also selects the messages for a brief moment which is bad UX.

PR #1651, opened 2020-03-17
Angular app fixes

abist:
1. Removed `Http` module from angular since we don't use websockets any more
2. Fixes https://github.com/microsoft/vscode-mssql/issues/1640

Old behavior (selected message for a brief moment as well) - 

old-select-all

New behavior -

new-select-all

### DIFF
--- a/src/views/htmlcontent/src/js/components/app.component.ts
+++ b/src/views/htmlcontent/src/js/components/app.component.ts
@@ -79,7 +79,8 @@ const template = `
         <span class="shortCut"> {{messageShortcut}} </span>
     </div>
     <div id="messages" class="scrollable messages" [class.hidden]="!messageActive && dataSets.length !== 0"
-        (contextmenu)="openMessagesContextMenu($event)">
+        (contextmenu)="openMessagesContextMenu($event)"
+        (mousedown)="onMouseDown($event)">
         <br>
         <table id="messageTable">
             <colgroup>
@@ -188,6 +189,7 @@ export class AppComponent implements OnInit, AfterViewChecked {
             this.magnify(this.activeGrid);
         },
         'event.selectAll': () => {
+            $('#messages').find('*').css('user-select', 'none');
             this.slickgrids.toArray()[this.activeGrid].selection = true;
         },
         'event.saveAsCSV': () => {
@@ -417,7 +419,7 @@ export class AppComponent implements OnInit, AfterViewChecked {
                     self.onScroll(0);
                     break;
                 default:
-                    console.error('Unexpected web socket event type "' + event.type + '" sent');
+                    console.error('Unexpected proxy event type "' + event.type + '" sent');
                     break;
             }
         });
@@ -561,6 +563,10 @@ export class AppComponent implements OnInit, AfterViewChecked {
         event.preventDefault();
         let selectedRange: IRange = this.getSelectedRangeUnderMessages();
         this.messagesContextMenu.show(event.clientX, event.clientY, selectedRange);
+    }
+
+    onMouseDown(event: any): void {
+        $('#messages').find('*').css('user-select', 'text');
     }
 
     getSelectedRangeUnderMessages(): IRange {
--- a/src/views/htmlcontent/src/js/components/app.component.ts
+++ b/src/views/htmlcontent/src/js/components/app.component.ts
@@ -120,7 +120,6 @@ const template = `
     selector: 'my-app',
     host: { '(window:keydown)': 'keyEvent($event)',
         '(window:gridnav)': 'keyEvent($event)',
-        '(window:keyup)' : 'keyUpEvent($event)',
         '(window:resize)' : 'resizeResults()'
      },
     template: template,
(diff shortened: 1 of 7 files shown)


abist:
@Charles-Gagnon Setting a user-select for every container isn't viable since angular needs the html event to be selected. Secondly, we never had select all in messages, the behavior you saw has been always there. Anyway,  I'll close this PR and leave this as a task then.

abist:
@Charles-Gagnon Are you suggesting a new enhancement of adding `Select All` from messages?

Charles-Gagnon:
I would generally say that select all on the messages is a very useful/desired feature, so yeah if that wasn't intended to be a supported feature currently then adding that as an enhancement would be good.

(although just testing it out now in ADS it's pretty bad there too hah - we end up selecting the DB name from the dropdown as well for example 😞 )

abist:
@Charles-Gagnon I'll add a `Select All` for messages and re-open this PR in that case :)

Charles-Gagnon:
I mean I also think there's gotta be a better way to handle selecting the grid stuff too than what you did if you were planning on leaving that alone. It seems wrong to have to modify the user-select manually for these events to work properly...

abist:
@Charles-Gagnon New behavior : 

newest-select-all

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -793,34 +794,34 @@ export class AppComponent implements OnInit, AfterViewChecked {
...
      *
      */
-    keyEvent(e: KeyboardEvent): void {
-        const self = this;
-        let eString = this.shortcuts.buildEventString(e);
-        this.shortcuts.getEvent(eString).then((result) => {
+    async keyEvent(e: any): Promise<void> {
+        if (e.target.classList.contains('slick-cell')) {
+            $('#messages').css('user-select', 'none');
+            const self = this;
```
Charles-Gagnon: If you're going to use self be consistent and use it everywhere - don't use this like you are below.

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -793,34 +794,34 @@ export class AppComponent implements OnInit, AfterViewChecked {
     /**
      *
      */
-    keyEvent(e: KeyboardEvent): void {
-        const self = this;
-        let eString = this.shortcuts.buildEventString(e);
-        this.shortcuts.getEvent(eString).then((result) => {
+    async keyEvent(e: any): Promise<void> {
+        if (e.target.classList.contains('slick-cell')) {
+            $('#messages').css('user-select', 'none');
```
Charles-Gagnon: So you're saying that putting a user-select: contain on the results and messages container doesn't work? It's still HTML in the end and my understanding is that while you hook up to the events for selecting the grids properly that you could at least prevent the selection of the messages text by containing the select-all functionality.

abist: having user-select any other than none when selecting all from the grid will also select the messages unfortunately (and also the context menu if you right click before selecting all)

Charles-Gagnon: That's odd. Wonder if it's because of the slickgrid stuff...anyways this is messy as heck and I really don't like it but I don't have a better solution at the moment so if you've tried some other things and can't get anything else to work I guess this is fine. I'd suggest running it by @anthonydresser  though since it involves the slickgrid stuff and he's a lot more familiar with that...

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -793,34 +794,33 @@ export class AppComponent implements OnInit, AfterViewChecked {
...
-    /**
-     * Helper to deselect messages when Ctrl + A is pressed
-     * inside the grid
-    */
-    keyUpEvent(e: KeyboardEvent): void {
-        let eString = this.shortcuts.buildEventString(e);
-        this.shortcuts.getEvent(eString).then((result) => {
+    async keyEvent(e: any): Promise<void> {
+        if (e.target.classList.contains('slick-cell')) {
+            $('#messages').css('user-select', 'none');
```
Charles-Gagnon: What about at the very least setting the user-select back to text once you're done with the selection instead of doing that on every other event? That seems a like a lot of extra processing that most of the time won't actually be necessary.

src/views/htmlcontent/src/js/services/shortcuts.service.ts
```diff
@@ -93,16 +93,13 @@ export class ShortcutService {
         }
     }
 
-    getEvent(shortcut: string): Promise<string> {
+    async getEvent(shortcut: string): Promise<string> {
         const self = this;
         if (this.shortcuts) {
-            return Promise.resolve(this.getEventInternal(shortcut));
+            return this.getEventInternal(shortcut);
```
Charles-Gagnon: Mixing self and this again...you shouldn't need to use self here anymore.

src/views/htmlcontent/src/js/components/app.component.ts
abist: The selected text get unselected in that case. This is doing less work in fact, since it only switches the user selection once instead of twice

src/views/htmlcontent/src/js/components/app.component.ts
anthonydresser: I don't remember anything about this since i think this was added after I left the team. But select all in the grid should be able to entirely be handled by the grid, we shouldn't have to worry about the messages.

abist:
I'm going to merge this PR soon, if no one has any issues.

PR #1678, opened 2020-05-11
Fix build and upgrade vscode API

abist:
The VSCode API changed one of it's function parameters to be non-optional, causing the builds to break. This isn't VSCode's fault that we always break for minor engine upgrades since their APIs are always backwards compatible, but it's because we use a wrapper around VSCode's API for testing and any addition to the VSCode API breaks the wrapper.


abist:
Going ahead and merging this so we get a clean build out of it tonight.

Issue #1700, opened 2020-06-22
Backslash character last in string breaks syntax highlighting


johanrex:
- MSSQL Extension Version:
v1.9.0

- VSCode Version:
1.46.1

- OS Version:
Windows 10

Steps to Reproduce:

1. Paste this text in an sql file:
```
select 'syntax highlighting works fine if the \ character is not last'
select 'syntax highlighting breaks if the \ character is last, like this\'
select 'syntax highlighting is now broken.'
```
Picture of problem:
![bild](https://user-images.githubusercontent.com/6969298/85292213-db46c400-b49b-11ea-88e8-3462b6617963.png)

For reference, here's the correct syntax highlighting as rendered by SSMS. 
![bild](https://user-images.githubusercontent.com/6969298/85369555-93707d00-b52d-11ea-97f6-ab3ac5f99492.png)

SJMakin:
Duplicate of #1680

abist:
Fixed by https://github.com/microsoft/vscode-mssql/pull/1701

PR #1701, opened 2020-06-25
Fix syntax highlighting for backslashes at end of string

### LINKED ISSUE #1680: Syntax Colorizations misread CHARINDEX('\')
- MSSQL Extension Version: 1.9.0
- VSCode Version: 1.45.0
- OS Version: Windows 10 Version	10.0.17134 Build 17134

Steps to Reproduce:

1. Create text/code block to demonstrate syntax colorization
2. Insert CHARINDEX('\') before or in the middle of the block
3. Note that the colorizations to not consider the '\' string closed.

CHARINDEX('\') returns the correct character identifier when run but does not color accordingly in mssql.

SJMakin:
After noting in one of the issues that this bug wasnt present in 1.8 I did differential analysis and found the bug was introduced in 0af2cc6dc303225e18aa71b67642f7797fd4e07 syntaxes/sql.configuration.json.  

I noticed the rules here differ from the rules in src\languageservice\serviceclient.ts initializeLanguageConfiguration(), with the major difference being the surroundingPairs.  

While removing this surroundingPairs array from syntaxes/sql.configuration.json appears to fix the issue I am not absolutely sure it is the correct thing to do.  Would greatly appreciate a review from someone that knows the code base better than me!

Tested against a collection of samples from prior issues and in all cases highlighting and parsing (as seen in Inspect Editor Tokens and Scopes) appears correct.  Would be nice to add some unit tests but no idea how!

Resolves #1680 #1700

Testbed:
```
SELECT '/'
FROM SomeTable

SELECT CHARINDEX(bob, '\', 1)
FROM [Bob]

SELECT CHARINDEX(bob, '\''', 1)
FROM Bob

select REPLACE(SES.LOGIN_NAME,'UNI/''/ZA/','') AS LOGIN_NAME
select REPLACE(SES.LOGIN_NAME,'UNI''ZA/','') AS LOGIN_NAME

SELECT * FROM dbo.Person
WHERE firstname = 'Ashleigh\//\/\/'
-- example: in the Nord theme the above opening quote is white, closing is green
ORDER BY birthdate DESC

SET @cmd = 'powershell.exe -c "Move-Item C:\Example\*' + (SELECT something FROM @tableVar) + '.txt ''\\Example\''Path with Spaces\Example'''

insert into bpb(col1, col2, col3) Values ('This data will break the color because of this escaped character \', 'everything behind is scrambled','data');

select * from bob where col1='\asd' and col2=2
```


abist:
@SJMakin Could you also include the case that breaks for you without this change? Without your changes, all the highlighting seems fine to me -

Screen Shot 2020-06-30 at 10 39 24 AM

SJMakin:
@abist - Strange. This is what I see:

image

Please can you double check?

Extension version: ms-mssql.mssql v1.9.0

**Visual Studio Code info:**
Version: 1.46.1 (system setup)
Commit: cd9ea6488829f560dc949a8b2fb789f3cdc05f5d
Date: 2020-06-17T21:13:20.174Z (2 wks ago)
Electron: 7.3.1
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Windows_NT x64 10.0.18362

SJMakin:
@abist Any news? You using the same set up as me? Am I missing something?

kburtram:
@abist I'm seeing the same output as @SJMakin?  Could you please review and merge this if there aren't side-effects to the change?

abist:
@SJMakin I don't see the issues in the extension, so I'm not sure how I can verify this fix. Could you use the latest, clean installation and see if you get this error you have the fix for?

SJMakin:
@abist please can you confirm your OS? Wondering if this is Windows only.   If you can post the output from the help/about, as I have above, that would be helpful.  Thanks

abist:
@SJMakin I retried with a fresh install and it seems that the highlighting is correct initially but changes to the incorrect one as in your screenshot, so you're right. I'll pull your fixes and verify it. Here's my system btw - 
Electron: 7.3.2
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Windows_NT x64 10.0.18363

SJMakin:
@abist Any news on this one mate?

abist:
@SJMakin This is actually not a trivial issue upon my investigation. It seems that the behavior changes during dev and the compiled `.vsix` extension. I'm currently not working on this, but this is something I have planned for September end. If you find any leads yourself until then, feel free to post them here or update the PR accordingly!

SJMakin:
@abist Thanks for the update.  When I was working with this I was compiling the vsix and reinstalling to test.  

Could be the language server is not restating properly in dev?  Can you clarify exactly what behaviour is changing, and what you are expecting? 

Happy to lend a hand where possible.  Keen to bring this to a close!

abist (approved):

PR #1706, opened 2020-07-16
fix build

abist:
Fix build

### DIFF
--- a/test/stubs.ts
+++ b/test/stubs.ts
@@ -31,6 +31,7 @@ class TestExtensionContext implements vscode.ExtensionContext {
     logPath: string;
     extensionUri: vscode.Uri = vscode.Uri.parse('test_uri');
     environmentVariableCollection: vscode.EnvironmentVariableCollection = undefined;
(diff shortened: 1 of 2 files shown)


test/stubs.ts
```diff
@@ -31,6 +31,7 @@ class TestExtensionContext implements vscode.ExtensionContext {
     logPath: string;
     extensionUri: vscode.Uri = vscode.Uri.parse('test_uri');
     environmentVariableCollection: vscode.EnvironmentVariableCollection = undefined;
+    extensionMode: vscode.ExtensionMode.Test
```
Charles-Gagnon, line 34: So one thing that would solve this that I suggested Maddy do is to add the extension context as an API export from the extension, then the tests can call activate on the extension and get the actual extension context.

This also has the benefits of stuff like having the actual extension path and access to storage stuff.

abist: Oh I didn't know you could do that now. That sounds like a good plan, but something I'll work on in future. I'll merge this to get the builds green first.

PR #1711, opened 2020-07-24
Refactor firewall class

### LINKED ISSUE #1532: Create Firewall rule fails to locate SQL DB in subscription
- MSSQL Extension Version: 1.8

Setup:
- Have two Azure subscriptions, say Sub1 and Sub2
- Have a SQL DB in one of the subscriptions, say Sub2

Steps to Reproduce:

1. Sign into Sub1
2. Try to connect to SQL DB and Create Firewall Rule
3. Get error that DB is not in subscription.
4. Use "Azure Sign In" to switch to Sub2
5. Try to connect to SQL DB and Create Firewall Rule

Expected: Firewall rule created and server connected

Actual: Still get error message about not being able to locate server in subscription.  Need to restart VS Code and connect again to successfully create rule.

image

cssuh:
Refactored firewallService class to use the accountService class, and separated account interfaces from firewallRequest into accountInterfaces 
Fixes #1532


src/azure/accountService.ts
```diff
@@ -0,0 +1,107 @@
...
+            tokenType: this._token.tokenType,
+            token: this._token.accessToken
+        };
+        return mapping;
+    }
+
+    public set isSignedIn(value: boolean) {
+        this._isSignedIn = value;
+        if (value) {
+            this._session = this._vscodeWrapper.azureAccountExtension.exports.sessions[0];
```
abist, line 99: This won't be applicable now since now you will be supporting multiple accounts, so you need to select the correct session here.

src/connectionconfig/connectionconfig.ts
```diff
@@ -3,7 +3,6 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 'use strict';
-import * as vscode from 'vscode';
```
abist, line 6: Thanks for removing unused imports!

PR #1719, opened 2020-07-28
The interaction required code

aaomidi:


lib/adal-aad-node/src/i18n.ts/defaultI18n.ts
```diff
@@ -0,0 +1,24 @@
+import { StringLookup, InteractionRequiredContext } from "../models";
```
cssuh, line 1: What is this file for? are we getting rid of src/errors/errors.ts?

aaomidi: This is for normal string lookup, rather than error lookup. Might eventually combine them into one.

aaomidi: Don't want to cause merge conflicts right now.

PR #1721, opened 2020-07-28
fix brackets

cssuh:


aaomidi:
Think I may have fixed this in my push or other PRs

Issue #1685, opened 2020-05-27
A11y_mssql for VSCode_ Select top 1000 Results window_Screenreader:  After pressing Ctrl+Alt+R key for expand and collapse result menu, NVDA announce wrong status.(Reads R R R R)


harshgangrade111:
[“Check out Accessibility Insights!](https://accessibilityinsights.io/) - Identify accessibility bugs before check-in and make bug fixing faster and easier.” 
 
**GitHubTags**: #A11y_mssql for VSCode Object Explorer;#A11ySev2;#A11yTCS;#A11yMAS;#DesktopApp; #MAS1.3.1;#Narrator;#Win10;#Visual Studio Code;#mssql for VSCode;#NVDA;#JAWS;#FTP; #SH_OBjectExplorer_MAY2020;#Benchmark; #STP;

**Environment Details:**
Application Name: mssql for VSCode Object Explorer
Windows Version: Windows10
Screen Reader: NVDA, JAWS 2020

**Repro Steps:** 

1. Open Visual studio code 
2. Click on extension button and search for "mssql"
3. Click on install and move to the SQL Server button
4. Click on Add Connection Button and insert the valid credentials
5. after adding the server move to database expand it -->System database expand-->Master expand-->Tables expand -->System table expand and right click on it and select top 1000 option
6. Select top 1000 window will open navigate to Resluts window
7. Start NVDA and press Ctrl+Al+R  key.

**Actual:**

After pressing Ctrl+Alt+R key for expand and collapse result menu, NVDA announce wrong status. (Reads R R R R)

**Same issue is also observed with Message Window.**
**Same Issue is also observed with JAWS.**

**Expected:**

After pressing Ctrl+Alt+R key for expand and collapse result menu, NVDA should announce expand result and collapse result. It should not announce wrong content.

**User Impact:**

Screen reader user will not receive correct status of the expand and collapse result menu.

**Attachment link for Reference:**
[1685_A11y_mssql for VSCode Object Explorer_ Select top 1000 Results window_Screenreader.zip](https://github.com/microsoft/vscode-mssql/files/4686666/1685_A11y_mssql.for.VSCode.Object.Explorer_.Select.top.1000.Results.window_Screenreader.zip)

abist:
Please use the following build to verify - https://mssqltools.visualstudio.com/_apis/resources/Containers/2339179/drop?itemPath=drop%2Fmssql-1.9.0.vsix

Roopesh-T:
[...]

PR #1724, opened 2020-07-29
add header narration

abist:
Fixes https://github.com/microsoft/vscode-mssql/issues/1685

### DIFF
--- a/src/views/htmlcontent/src/js/components/app.component.ts
+++ b/src/views/htmlcontent/src/js/components/app.component.ts
@@ -40,6 +40,7 @@ const template = `
     <div *ngIf="dataSets.length > 0" role="button" id="resultspane" tabIndex="0" class="boxRow header collapsible"
         [class.collapsed]="!resultActive"
         (click)="toggleResultsPane()"
+        (keydown)="handleResultsKeydown($event)"
         [attr.aria-label]="Constants.resultPaneLabel"
         [attr.aria-expanded]="!resultActive">
         <span> {{Constants.resultPaneLabel}} </span>
@@ -85,6 +86,7 @@ const template = `
         [attr.aria-label]="Constants.messagePaneLabel"
         [attr.aria-expanded]="!messageActive"
         (click)="toggleMessagesPane()"
+        (keydown)="handleMessagesKeydown($event)"
         style="position: relative">
         <div id="messageResizeHandle" class="resizableHandle"></div>
         <span> {{Constants.messagePaneLabel}} </span>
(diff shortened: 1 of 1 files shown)


src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -37,12 +37,18 @@ export interface IGridDataSet {
 // tslint:disable:max-line-length
 const template = `
 <div class="fullsize vertBox">
-    <div *ngIf="dataSets.length > 0" id="resultspane" class="boxRow header collapsible" [class.collapsed]="!resultActive" (click)="toggleResultsPane()">
+    <div *ngIf="dataSets.length > 0" role="button" id="resultspane" tabIndex="0" class="boxRow header collapsible"
```
alanrenmsft, line 40: why not just use button element? it will handle the keypress as well.

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -37,12 +37,18 @@ export interface IGridDataSet {
...
+        [class.collapsed]="!resultActive"
+        (click)="toggleResultsPane()"
+        [attr.aria-label]="Constants.resultPaneLabel"
+        [attr.aria-expanded]="!resultActive">
         <span> {{Constants.resultPaneLabel}} </span>
         <span class="shortCut"> {{resultShortcut}} </span>
     </div>
-    <div id="results" *ngIf="renderedDataSets.length > 0" class="results vertBox scrollable"
-         (onScroll)="onScroll($event)" [scrollEnabled]="scrollEnabled" [class.hidden]="!resultActive">
+    <div id="results" tabIndex="0" *ngIf="renderedDataSets.length > 0" class="results vertBox scrollable"
```
alanrenmsft, line 48: you are enabling keyboard focus on this element, but i don't see any action you can do here, am i missing something?

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -74,7 +80,12 @@ const template = `
     </div>
     <context-menu #contextmenu (clickEvent)="handleContextClick($event)"></context-menu>
     <msg-context-menu #messagescontextmenu (clickEvent)="handleMessagesContextClick($event)"></msg-context-menu>
-    <div id="messagepane" class="boxRow header collapsible" [class.collapsed]="!messageActive" (click)="toggleMessagesPane()" style="position: relative">
+    <div id="messagepane" role="button" tabIndex="1" class="boxRow header collapsible"
```
alanrenmsft, line 83: same as above, consider using native button element

src/views/htmlcontent/src/js/components/app.component.ts
abist: it wasn't able to get focus earlier because `div`s don't support focus unless they have a tabIndex. You couldn't reach just the header just with keyboard.

src/views/htmlcontent/src/js/components/app.component.ts
abist: I'm not sure if angular supports these directives with buttons. We have it all styled, the only reason I put the role as a button was to support `aria-expanded`

src/views/htmlcontent/src/js/components/app.component.ts
abist: native button element doesnt support directives with angular

src/views/htmlcontent/src/js/components/app.component.ts
abist: It's the same reason it's not a button in ADS either

alanrenmsft: I do agree there are some styles need to be adjust if you change it to button, it is ok if you want to minimize the changes. As i pointed out earlier, you should handle the keypress event as well to treat space key and enter key as click.

abist: @alanrenmsft done

src/views/htmlcontent/src/js/components/app.component.ts
```diff
@@ -466,6 +479,38 @@ export class AppComponent implements OnInit, AfterViewChecked {
...
+        }
+        return false;
+    }
+
+    /**
+     * Handles toggling messages via key event
+     */
+    private handleMessagesKeydown(event: KeyboardEvent): void {
+        if (this.handleKeydown(event)) {
+            this.toggleMessagesPane();
```
alanrenmsft, line 500: also, we should stop the event propagation if matches the expected key.

alanrenmsft: event.stopPropagation();

PR #1727, opened 2020-07-29
adal-aad-node library

aaomidi:


lib/adal-aad-node/src/engine/AzureCodeGrant.ts
```diff
@@ -0,0 +1,104 @@
...
+        const authCode = await this.authRequest.getAuthorizationCode(state);
+
+
+        const response = await this.getTokenWithAuthorizationCode(tenant, resource, {
+            authCode,
+            redirectUri: this.providerSettings.redirectUri,
+            codeVerifier
+        });
+        if (!response) {
+            throw new AzureAuthError(10, this.errorLookup.getSimpleError(10));
```
abist, line 52: where does this 10 come from?

abist: I see now. Adding a comment stating error code 10 is for auth error will help here

lib/adal-aad-node/src/engine/AzureDeviceCode.ts
```diff
@@ -0,0 +1,112 @@
...
+
+        const accessTokenString = finalDeviceLogin.access_token;
+        const refreshTokenString = finalDeviceLogin.refresh_token;
+
+        const currentTime = new Date().getTime() / 1000;
+        const expiresOn = `${currentTime + finalDeviceLogin.expires_in}`;
+        const result = await this.getTokenHelper(tenant, resource, accessTokenString, refreshTokenString, expiresOn);
+
+        if (!result) {
+            throw new AzureAuthError(11, this.errorLookup.getSimpleError(11));
```
abist, line 46: Same here, please add a comment about the error code 11

lib/adal-aad-node/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "adal-aad-node",
+  "version": "0.0.0",
```
abist, line 3: might want to version it 😄

lib/adal-aad-node/src/errors/errors.ts
```diff
@@ -0,0 +1,29 @@
+import { ErrorLookup, Error1Context } from "../models/error";
+
+type ErrorMapping = {
+    [errorCode: number]: string;
+};
+
+const simpleErrorMapping: ErrorMapping = {
+    2: 'Something failed with the authentication, or your tokens have been deleted from the system. Please try adding your account to Azure Data Studio again',
```
abist, line 8: Are these errors always in this language and not affected by localization?

lib/adal-aad-node/package.json
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "devDependencies": {
+    "@types/qs": "^6.9.4",
+    "@typescript-eslint/eslint-plugin": "^3.7.1",
+    "@typescript-eslint/parser": "^3.7.1",
+    "eslint": "^7.5.0",
+    "typescript": "^3.9.7"
+  },
+  "dependencies": {
+    "@types/node": "^14.0.26",
```
Charles-Gagnon, line 21: This should be a devDependency

lib/adal-aad-node/src/engine/AzureAuth.ts
```diff
@@ -0,0 +1,458 @@
...
+		return this.userInteraction.askForConsent(this.stringLookup.getInteractionRequiredString({ tenant, resource }));
+
+	}
+	//#endregion
+
+	//#region data modeling
+
+	public createAccount(tokenClaims: TokenClaims, key: string, tenants: Tenant[]): AzureAccount {
+		// Determine if this is a microsoft account
+		let accountType: 'microsoft' | 'work_school';
```
Charles-Gagnon, line 366: Would suggest making this into an enum or type

lib/adal-aad-node/src/engine/AzureAuth.ts
```diff
@@ -0,0 +1,458 @@
...
+
+	//#endregion
+
+	//#region utils
+	protected getTokenClaims(accessToken: string): TokenClaims | undefined {
+		try {
+			const split = accessToken.split('.');
+			return JSON.parse(Buffer.from(split[1], 'base64').toString('binary'));
+		} catch (ex) {
+			throw new Error('Unable to read token claims: ' + JSON.stringify(ex));
```
Charles-Gagnon, line 446: For all the errors - are you ensuring these are being logged somewhere too?

lib/adal-aad-node/src/errors/errors.ts
```diff
@@ -0,0 +1,29 @@
...
+    3: 'Token retrival failed with an error. Open developer tools to view the error',
+    4: 'No access token returned from Microsoft OAuth',
+    5: 'The user had no unique identifier within AAD',
+    6: 'Error retrieving tenant information',
+    7: 'Error when getting your account from the cache',
+    8: 'Error when parsing your account from the cache',
+    9: 'Error when adding your account to the cache',
```
Charles-Gagnon, line 15: Would suggest having an enum that maps to these numeric values so you can use those in the code here instead of hardcoding numbers

lib/adal-aad-node/src/engine/AzureDeviceCode.ts
```diff
@@ -0,0 +1,112 @@
+import { AzureAuth } from "./AzureAuth";
```
Charles-Gagnon, line 1: nit: But you're inconsistent with the casing of your files. Some are capitalized, while some are camelCased

lib/adal-aad-node/src/engine/AzureAuth.ts
aaomidi: The caller needs to catch them basically

lib/adal-aad-node/src/errors/errors.ts
aaomidi: Isn't using enums that are can randomly change usually considered a code_smell? I was thinking maybe just using string based keys.

lib/adal-aad-node/src/engine/AzureDeviceCode.ts
aaomidi: Usually for single class files I use capital letters but I can change that.

aaomidi:
For tests I think the plan I had discussed with Chris was to write the tests as we start using the library. But they're going to come :) Hopefully 80-90% code coverage considering the code is tiny.

lib/adal-aad-node/src/errors/errors.ts
Charles-Gagnon: What do you mean randomly change? I wouldn't expect the number values to actually change...or is that not the case?

And to be clear - I'm suggesting making an enum for the numbers themselves.

```
enum ErrorCodes {
  MyError = 1
  MyOtherError = 2
  SomeOtherReallyBadError = 3
...
}
```

and then you'd keep this mapping and just use those enum values instead of random numbers all over the place.

lib/adal-aad-node/src/engine/AzureAuth.ts
Charles-Gagnon: The library should log all errors though I would think. So yeah - leave it up to the caller to handle (and log themselves if they want) but if an error happens we should make sure that the adal log file has an entry for that too.

lib/adal-aad-node/src/errors/errors.ts
aaomidi: I guess this is more of an issue in languages that can actually iterate over enum values, but this was a major issue with for example Bukkit using a Material enum, where every update introduced new enum values and broke a lot of other people using the Bukkit library - so it kinda stuck with me to not use enums for values that can change in future updates since it can break API contracts.

Either way I do like your suggestion and will prob switch to it.

aaomidi:
Postgre team wants us to add a test covering getting the OssRdbms token as well.

lib/adal-aad-node/src/engine/AzureAuth.ts
```diff
@@ -0,0 +1,469 @@
...
+			loginComplete = result?.authComplete;
+			if (!result?.response) {
+				this.logger.error('Authentication failed');
+				return undefined;
+			}
+			const account = await this.hydrateAccount(result.response.accessToken, result.response.tokenClaims);
+			loginComplete?.resolve();
+			return account;
+		} catch (ex) {
+			if (ex instanceof AzureAuthError) {
```
Charles-Gagnon, line 48: Are we logging this error if it happens? Should maybe move the log up above

lib/adal-aad-node/src/engine/AzureAuth.ts
```diff
@@ -0,0 +1,469 @@
...
+		let loginComplete: Deferred<void> | undefined;
+		try {
+			const result = await this.login(this.commonTenant, this.providerSettings.resources.windowsManagementResource);
+			loginComplete = result?.authComplete;
+			if (!result?.response) {
+				this.logger.error('Authentication failed');
+				return undefined;
+			}
+			const account = await this.hydrateAccount(result.response.accessToken, result.response.tokenClaims);
+			loginComplete?.resolve();
```
Charles-Gagnon, line 45: There's a couple paths here that aren't resolve/rejecting the loginComplete promise - is that intentional?

lib/adal-aad-node/src/engine/AzureAuth.ts
```diff
@@ -0,0 +1,469 @@
...
+			if (!tokenClaims) {
+				account.isStale = true;
+				return account;
+			}
+			return await this.hydrateAccount(tokenResult, tokenClaims);
+		} catch (ex) {
+			account.isStale = true;
+			return account;
+			// Let caller deal with it too.
+			throw ex;
```
Charles-Gagnon, line 88: This throw isn't reachable?

lib/adal-aad-node/src/engine/AzureAuth.ts
```diff
@@ -0,0 +1,469 @@
...
+
+	public async getAccountSecurityToken(account: AzureAccount, tenantId: string, azureResource: AADResource): Promise<Token | undefined> {
+		if (account.isStale === true) {
+			this.logger.log('Account was stale. No tokens being fetched.');
+			return undefined;
+		}
+		const tenant = account.properties.tenants.find(t => t.id === tenantId);
+
+		if (!tenant) {
+			throw new AzureAuthError(1, this.errorLookup.getTenantNotFoundError({ tenantId }), undefined);
```
Charles-Gagnon, line 106: What's the 1 here? I don't see an error code with 1. We shouldn't be using magic numbers like this regardless - the enums make it a lot easier to work with

lib/adal-aad-node/src/engine/AzureDeviceCode.ts
```diff
@@ -0,0 +1,115 @@
...
+                }
+                clearTimeout(timeout);
+                clearInterval(timer);
+                resolve(x);
+            }, info.interval * 1000);
+
+            timeout = setTimeout(() => {
+                clearInterval(timer);
+                // Error when getting access token for DeviceCodeLogin
+                reject(new AzureAuthError(11, this.errorLookup.getSimpleError(11)));
```
Charles-Gagnon, line 78: Use enum

lib/adal-aad-node/src/errors/AzureAuthError.ts
```diff
@@ -0,0 +1,16 @@
...
+	constructor(private errorCode: ErrorCodes, private readonly errorMessage: string, private readonly originalException?: any) {
+        super(errorMessage);
+    }
+
+	getPrintableString(): string {
+		return JSON.stringify({
+            errorCode: this.errorCode,
+			errorMessage: this.errorMessage,
+			originalException: this.originalException ?? ''
+		}, undefined, 2);
```
abist, line 14: use ErrorCode enum here

lib/adal-aad-node/src/errors/errors.ts
```diff
@@ -0,0 +1,42 @@
...
+    Tenant = 6,
+    GetAccount = 7,
+    ParseAccount = 8,
+    AddAccount = 9,
+    GetAccessTokenAuthCodeGrant = 10,
+    GetAccessTokenDeviceCodeLogin = 11,
+    TimedOutDeviceCode = 12
+}
+const simpleErrorMapping: ErrorMapping = {
+    [ErrorCodes.AuthError]: 'Something failed with the authentication, or your tokens have been deleted from the system. Please try adding your account to Azure Data Studio again',
```
abist, line 21: Should this error be generalized for the extension too?

lib/adal-aad-node/src/errors/errors.ts
```diff
@@ -0,0 +1,42 @@
...
+    [ErrorCodes.TokenRetrieval]: 'Token retrieval failed with an error. Open developer tools to view the error',
```
abist, line 22: Is that the only way to see the error? Do we not get an error pop up on ADS?

lib/adal-aad-node/src/i18n/defaultI18n.ts
```diff
@@ -0,0 +1,24 @@
+import { StringLookup, InteractionRequiredContext } from '../models';
+
+type StringMapping = {
+    [stringCode: number]: string;
+};
+
+const simpleStringMapping: StringMapping = {
+    2: 'Something failed with the authentication, or your tokens have been deleted from the system. Please try adding your account to Azure Data Studio again',
```
abist, line 8: Need to make the strings generalized for the library.

kburtram:
Closing stale PRs

PR #1729, opened 2020-07-31
errorcode, filename, pr review changes

cssuh:


lib/adal-aad-node/src/models/account.ts
```diff
@@ -34,6 +34,11 @@ export enum AzureAuthType {
     DeviceCode = 1
 }
 
+export enum AccountType {
+    Microsoft = 0,
```
Charles-Gagnon, line 38: Consider using the string values here instead so you don't have to hardcode those separately?

PR #1750, opened 2020-09-23
Fix build

abist:
This will fix the builds which are failing now and will fix this issue for future breaks with vscode extension context API changes.


Charles-Gagnon (approved):
:shipit:

PR #1767, opened 2020-10-12
Fix unit test failures

cssuh:
Fixes the failing unit tests in our build


test/connectionProfile.test.ts
```diff
@@ -4,6 +4,7 @@
  * ------------------------------------------------------------------------------------------ */
 
 'use strict';
+import vscode = require('vscode');
```
kburtram, line 7: why not use `import * as vscode from 'vscode';` syntax for this import?

abist (approved):

test/connectionProfile.test.ts
```diff
@@ -14,9 +15,10 @@ import { ConnectionUI } from '../src/views/connectionUI';
 import { ConnectionStore } from '../src/models/connectionStore';
 import ConnectionManager from '../src/controllers/connectionManager';
 import VscodeWrapper from '../src/controllers/vscodeWrapper';
-
+import Constants = require('../src/constants/constants');
```
kburtram, line 18: same here.  I think in typescript we generally prefer an `import..from` variation to calling require directly.  Is there a reason to use require directly here?

kburtram (approved):
LGTM

test/connectionProfile.test.ts
cssuh: I think I had just copied those imports from elsewhere, I can change them to the `import from` form

PR #1771, opened 2020-10-15
fix module 'ads-adal-library'

cssuh:
Fixing this issue:
Screen Shot 2020-10-14 at 10 22 06 PM


abist (approved):
remove the yarn.lock file and it looks good

PR #1772, opened 2020-10-17
moved token logic to azureController

cssuh:
Moved token logic out of the `ConnectionProfile.ts` file to a more central location, as it will be accessed from multiple places


cssuh:
this is included in the refactor firewall service pull request

PR #1773, opened 2020-10-19
Refactored firewall service to remove Azure Account dependency

cssuh:
Refactored the firewall service to use ads-adal-library, thus removing the Azure Account dependency.

### DIFF
--- a/src/azure/accountService.ts
+++ b/src/azure/accountService.ts
@@ -3,27 +3,35 @@
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 
+import vscode = require('vscode');
 import { IAccount, IAccountKey, IAccountDisplayInfo } from '../models/contracts/azure/accountInterfaces';
 import SqlToolsServiceClient from '../languageservice/serviceclient';
 import { IAzureSession } from '../models/interfaces';
-import { Deferred } from '../protocol';
 import Constants = require('../constants/constants');
 import VscodeWrapper from '../controllers/vscodeWrapper';
+import { AzureController } from './azureController';
+import { AccountStore } from './accountStore';
+import providerSettings from '../azure/providerSettings';
 
 export class AccountService {
 
-    private _session: IAzureSession = undefined;
     private _account: IAccount = undefined;
     private _token = undefined;
     private _isStale: boolean;
 
     constructor(
         private _client: SqlToolsServiceClient,
-        private _vscodeWrapper: VscodeWrapper
+        private _vscodeWrapper: VscodeWrapper,
+        private _context: vscode.ExtensionContext,
+        private _accountStore: AccountStore
     ) {}
 
     public get account(): IAccount {
         return this._account;
+    }
+
+    public setAccount(account: IAccount): void {
+        this._account = account;
     }
 
     public get client(): SqlToolsServiceClient {
@@ -65,40 +73,17 @@ export class AccountService {
     }
 
     public async createSecurityTokenMapping(): Promise<any> {
-        if (!this._token) {
-            let promise = new Deferred();
-            this._token = this._session.credentials.getToken((error, result ) => {
-                if (result) {
-                    this._isStale = false;
-                    this._token = result;
-                }
-                if (error) {
-                    this._isStale = true;
-                }
-                promise.resolve();
-            });
-            await promise;
-        }
         let mapping = {};
-        mapping[this._session.tenantId] = {
-            expiresOn: this._token.expiresOn.toISOString(),
-            resource: this._token.resource,
-            tokenType: this._token.tokenType,
-            token: this._token.accessToken
+        mapping[this.account.properties.tenants[0].id] = {
+            token: await this.refreshToken(this.account)
         };
         return mapping;
     }
 
-    public initializeSessionAccount(): void {
-        if (this._vscodeWrapper.azureAccountExtension.exports.sessions.length === 1) {
-            this._session = this._vscodeWrapper.azureAccountExtension.exports.sessions[0];
-        } else if (this._vscodeWrapper.azureAccountExtension.exports.filters) {
-            this._session = this._vscodeWrapper.azureAccountExtension.exports.filters[0].session;
-        }
-        this._account = this.convertToAzureAccount(this._session);
+    public async refreshToken(account): Promise<string> {
+        let azureController = new AzureController(this._context);
+        return await azureController.refreshToken(account, this._accountStore, providerSettings.resources.azureManagementResource);
     }
-
-
 
 
 }
--- a/src/azure/accountService.ts
+++ b/src/azure/accountService.ts
@@ -12,12 +12,17 @@ import VscodeWrapper from '../controllers/vscodeWrapper';
 import { AzureController } from './azureController';
 import { AccountStore } from './accountStore';
 import providerSettings from '../azure/providerSettings';
+import { Tenant } from 'ads-adal-library';
 
 export class AccountService {
 
     private _account: IAccount = undefined;
     private _token = undefined;
     private _isStale: boolean;
+    protected readonly commonTenant: Tenant = {
+        id: 'common',
+        displayName: 'common'
+    };
 
     constructor(
         private _client: SqlToolsServiceClient,
@@ -74,7 +79,7 @@ export class AccountService {
 
     public async createSecurityTokenMapping(): Promise<any> {
         let mapping = {};
-        mapping[this.account.properties.tenants[0].id] = {
+        mapping[this.getHomeTenant(this.account).id] = {
             token: await this.refreshToken(this.account)
         };
         return mapping;
@@ -83,6 +88,12 @@ export class AccountService {
     public async refreshToken(account): Promise<string> {
         let azureController = new AzureController(this._context);
         return await azureController.refreshToken(account, this._accountStore, providerSettings.resources.azureManagementResource);
+    }
+
+    public getHomeTenant(account: IAccount): Tenant {
+        // Home is defined by the API
+        // Lets pick the home tenant - and fall back to commonTenant if they don't exist
+        return account.properties.tenants.find(t => t.tenantCategory === 'Home') ?? account.properties.tenants[0] ?? this.commonTenant;
     }
 
 
--- a/src/azure/accountService.ts
+++ b/src/azure/accountService.ts
@@ -12,7 +12,7 @@ import VscodeWrapper from '../controllers/vscodeWrapper';
 import { AzureController } from './azureController';
 import { AccountStore } from './accountStore';
 import providerSettings from '../azure/providerSettings';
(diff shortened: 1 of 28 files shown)


src/azure/accountService.ts
```diff
@@ -65,40 +73,17 @@ export class AccountService {
...
-            });
-            await promise;
-        }
         let mapping = {};
-        mapping[this._session.tenantId] = {
-            expiresOn: this._token.expiresOn.toISOString(),
-            resource: this._token.resource,
-            tokenType: this._token.tokenType,
-            token: this._token.accessToken
+        mapping[this.account.properties.tenants[0].id] = {
```
abist, line 77: do we always take the first tenant? users can have multiple now right?

src/views/connectionUI.ts
```diff
@@ -566,49 +555,24 @@ export class ConnectionUI {
...
-                } else {
-                    return false;
-                }
-            });
-        } else {
-            this.connectionManager.accountService.initializeSessionAccount();
-            this.connectionManager.accountService.account.isSignedIn = true;
-            return true;
-        }
+        // prompt to "add account"
```
abist, line 560: Does this need to be uncommented?

abist (changes requested):
There's commented code which should be uncommented I think

abist:
@cssuh The builds are failing because they get timed out in the `gulp cover` step - https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=80865&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=6476f600-6ee7-510b-7199-f0c7eb1f3454. There's a test failure -
```
 1) Test Checkbox prompt
       Test checkbox prompt with simple question:
     Uncaught TypeError: Cannot read property 'checked' of undefined
      at /Users/runner/work/1/src/prompts/checkbox.ts:17:33
```

src/azure/accountService.ts
cssuh: yeah you're right, let me check what they do in ADS for this.

abist:
Fine to merge now, be sure to switch them on again

abist (approved):

src/views/connectionUI.ts
```diff
@@ -602,11 +603,13 @@ export class ConnectionUI {
         return this._vscodeWrapper.showInformationMessage(LocalizedConstants.msgPromptRetryFirewallRuleSignedIn,
             LocalizedConstants.createFirewallRuleLabel).then(async (result) => {
             if (result === LocalizedConstants.createFirewallRuleLabel) {
-                const firewallService = this.connectionManager.firewallService;
+                const firewallService = new FirewallService(this._connectionManager.accountService);
```
abist, line 606: Do we create a new FirewallService per account now?

src/azure/azureController.ts
```diff
@@ -83,4 +91,91 @@ export class AzureController {
...
+            accountStore.removeAccount(account.key.id);
+            throw new Error('account is stale. Please re-do login');
+        }
+        profile.azureAccountToken = await this.refreshToken(account, accountStore, settings);
+        profile.email = account.displayInfo.email;
+        profile.accountId = account.key.id;
+        return profile;
+    }
+
+    public async refreshToken(account: IAccount, accountStore: AccountStore, settings: AADResource): Promise<string> {
```
abist, line 136: The function signature says it returns `Promis<string>` but seems to me it's `Promise<Token|undefined>`. Am I missing something?

src/azure/accountService.ts
```diff
@@ -65,40 +78,23 @@ export class AccountService {
...
-            });
-            await promise;
-        }
         let mapping = {};
-        mapping[this._session.tenantId] = {
-            expiresOn: this._token.expiresOn.toISOString(),
-            resource: this._token.resource,
-            tokenType: this._token.tokenType,
-            token: this._token.accessToken
+        mapping[this.getHomeTenant(this.account).id] = {
```
abist, line 82: Please use a `Map<string, Token>` here and make the return type the same

src/azure/accountService.ts
```diff
@@ -65,40 +78,23 @@ export class AccountService {
...
     }
 
-    public initializeSessionAccount(): void {
-        if (this._vscodeWrapper.azureAccountExtension.exports.sessions.length === 1) {
-            this._session = this._vscodeWrapper.azureAccountExtension.exports.sessions[0];
-        } else if (this._vscodeWrapper.azureAccountExtension.exports.filters) {
-            this._session = this._vscodeWrapper.azureAccountExtension.exports.filters[0].session;
-        }
-        this._account = this.convertToAzureAccount(this._session);
+    public async refreshToken(account): Promise<string> {
```
abist, line 88: The signature here should be `Promise<Token>` from the helper too

src/views/connectionUI.ts
```diff
@@ -498,40 +503,25 @@ export class ConnectionUI {
...
-                        // Activate the Azure Account extension and call the function again
-                        if (this._vscodeWrapper.azureAccountExtension) {
-                            await this._vscodeWrapper.azureAccountExtension.activate();
-                            await this.handleExtensionActivation();
-                        }
-                    });
-                }
-                return false;
+            // If no match or no accountId present, need to add an azure account
+            let selection = await this._vscodeWrapper.showInformationMessage('Connection failed - firewall rule.', 'Add Azure Account');
```
abist, line 515: Please localize

src/views/connectionUI.ts
```diff
@@ -498,40 +503,25 @@ export class ConnectionUI {
...
+            if (selection === 'Add Azure Account') {
```
abist, line 516: We need to use string constants here. For example, if someone's using a localized version of vscode-mssql, this would fail.

src/views/connectionUI.ts
```diff
@@ -654,15 +599,17 @@ export class ConnectionUI {
...
         return this._vscodeWrapper.showInformationMessage(LocalizedConstants.msgPromptRetryFirewallRuleSignedIn,
             LocalizedConstants.createFirewallRuleLabel).then(async (result) => {
             if (result === LocalizedConstants.createFirewallRuleLabel) {
-                const firewallService = this.connectionManager.firewallService;
+                const firewallService = new FirewallService(this._connectionManager.accountService);
                 let ipRange = await this.promptForIpAddress(ipAddress);
                 if (ipRange) {
                     let firewallResult = await firewallService.createFirewallRule(serverName, ipRange.startIpAddress, ipRange.endIpAddress);
                     if (firewallResult.result) {
+                        this._vscodeWrapper.showInformationMessage('Firewall rule successfully created, please try reconnecting.');
```
abist, line 611: Localize string

src/azure/azureController.ts
```diff
@@ -83,4 +91,91 @@ export class AzureController {
...
+    }
+
+    public async refreshTokenWrapper(profile, accountStore, accountAnswer, settings: AADResource): Promise<ConnectionProfile> {
+        let account = accountStore.getAccount(accountAnswer.key.id);
+        if (!account) {
+            throw new Error('account not found');
+        }
+        if (account.isStale) {
+            accountStore.removeAccount(account.key.id);
+            throw new Error('account is stale. Please re-do login');
```
abist, line 128: You might want to throw these errors as information messages to the user so they know what's going on. Also, localize in that case. All strings that are visible in the UI need to be localized. Internal strings are fine.

abist (changes requested):

src/azure/azureController.ts
cssuh: it's returning `token.token` which is a string.  if it just returned `token` then that would be the Token type

src/azure/accountService.ts
cssuh: it will be a Map<string, string> since refreshToken returns a string

src/azure/azureController.ts
```diff
@@ -70,14 +72,19 @@ export class AzureController {
...
         this.context = context;
         if (!this.logger) {
             this.logger = new AzureLogger();
         }
+        if (!this._vscodeWrapper) {
+            this._vscodeWrapper = new VscodeWrapper();
+        }
     }
     public async init(): Promise<void> {
+        // const session = await vscode.authentication.getSession('microsoft', )
```
abist, line 87: dead code

src/views/connectionUI.ts
```diff
@@ -608,8 +609,7 @@ export class ConnectionUI {
                 if (ipRange) {
                     let firewallResult = await firewallService.createFirewallRule(serverName, ipRange.startIpAddress, ipRange.endIpAddress);
                     if (firewallResult.result) {
-                        this._vscodeWrapper.showInformationMessage('Firewall rule successfully created, please try reconnecting.');
-                        // this.promptForRetryCreateProfile(profile);
```
abist: dead code

src/azure/azureController.ts
```diff
@@ -83,4 +98,98 @@ export class AzureController {
...
+            profile.email = account.displayInfo.email;
+            profile.accountId = account.key.id;
+        }
+        return profile;
+    }
+
+    public async refreshTokenWrapper(profile, accountStore, accountAnswer, settings: AADResource): Promise<ConnectionProfile> {
+        let account = accountStore.getAccount(accountAnswer.key.id);
+        if (!account) {
+            throw new Error('account not found');
```
abist, line 131: Shouldn't this error be shown to the user? Or is it shown after this error is thrown?

src/models/connectionProfile.ts
```diff
@@ -94,37 +94,9 @@ export class ConnectionProfile extends ConnectionCredentials implements IConnect
...
+                    profile = await azureController.getTokens(profile, accountStore, providerSettings.resources.databaseResource);
                 } else {
-                    let account = accountStore.getAccount(accountAnswer.key.id);
-                    if (!account) {
-                        throw new Error('account not found');
-                    }
-                    profile.azureAccountToken = await profile.refreshToken(context, account, accountStore);
-                    profile.email = account.displayInfo.email;
-                    profile.accountId = account.key.id;
+                    profile = await azureController.refreshTokenWrapper(profile, accountStore, accountAnswer, providerSettings.resources.databaseResource);
```
abist, line 99: Doesn't look like we have any AAD token failure resolution mechanism here

src/views/connectionUI.ts
```diff
@@ -663,6 +609,8 @@ export class ConnectionUI {
                 if (ipRange) {
                     let firewallResult = await firewallService.createFirewallRule(serverName, ipRange.startIpAddress, ipRange.endIpAddress);
                     if (firewallResult.result) {
+                        this._vscodeWrapper.showInformationMessage(LocalizedConstants.msgPromptFirewallRuleCreated);
+                        // this.promptForRetryCreateProfile(profile);
```
abist, line 613: dead code

abist (approved):

abist:
adding @Charles-Gagnon since this is a big change and I'd like another set of eyes

PR #1803, opened 2020-12-08
Added error message

cssuh:
image


abist (approved):
Do we do the same in ADS? No way to actually point the user to the logs?

cssuh:
> Do we do the same in ADS? No way to actually point the user to the logs?

yeah, we have similar error messages in ADS.

PR #16912, opened 2020-12-14
Fixed build errors

cssuh:
Fixes build errors similar to this: 
`Argument of type 'string | TreeItemLabel' is not assignable to parameter of type 'string'.
  Type 'TreeItemLabel' is not assignable to type 'string'`

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -229,7 +229,15 @@ export class ObjectExplorerService {
      */
     public sortByServerName(array: TreeNodeInfo[]): TreeNodeInfo[] {
         const sortedNodeArray = array.sort((a, b) => {
-            return (a.label as string).toLowerCase().localeCompare((b.label as string).toLowerCase());
+            if (typeof a.label === 'string' && typeof b.label === 'string') {
+                return (a.label).toLowerCase().localeCompare((b.label).toLowerCase());
+            } else if (typeof a.label === 'string' && typeof b.label !== 'string') {
+                return (a.label).toLowerCase().localeCompare((b.label.label).toLowerCase());
+            } else if (typeof a.label !== 'string' && typeof b.label === 'string') {
+                return (a.label.label).toLowerCase().localeCompare((b.label).toLowerCase());
+            } else if (typeof a.label !== 'string' && typeof b.label !== 'string') {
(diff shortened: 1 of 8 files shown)


alanrenmsft:
are you sure the the label is always string? @cssuh

cssuh:
Yes, `clipboardWriteText` only accepts a string parameter, same with `toLowerCase`, `TreeNodeInfo` first parameter of the constructor should be a string, and `historyNodeLabel` should return a string.  Unless there is something I'm missing I believe they should all be strings.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -229,7 +229,15 @@ export class ObjectExplorerService {
      */
     public sortByServerName(array: TreeNodeInfo[]): TreeNodeInfo[] {
         const sortedNodeArray = array.sort((a, b) => {
-            return (a.label as string).toLowerCase().localeCompare((b.label as string).toLowerCase());
+            if (typeof a.label === 'string' && typeof b.label === 'string') {
```
alanrenmsft, line 232: to simplify the code, I would first get the string value for a and b and then do the compare

src/objectExplorer/objectExplorerService.ts
```diff
@@ -484,10 +492,17 @@ export class ObjectExplorerService {
             if (!(<IConnectionProfile>node.connectionCredentials).savePassword) {
                 node.connectionCredentials.password = '';
             }
+            let disconnectedNode;
             // make a new node to show disconnected behavior
-            let disconnectedNode = new TreeNodeInfo(node.label as string, Constants.disconnectedServerLabel,
-                node.collapsibleState, node.nodePath, node.nodeStatus, Constants.disconnectedServerLabel,
-                undefined, node.connectionCredentials, node.parentNode);
+            if (typeof node.label === 'string') {
```
alanrenmsft, line 497: same here:
`
const label = typeof node.label === 'string' ? node.label:node.label.label;
`

alanrenmsft: giving the fact that it is being used so many times, you can create a util function to get the label from a node.

cssuh: ok that makes a lot of sense, I knew there had to be a simpler way

alanrenmsft (approved):

Issue #16930, opened 2021-01-08
Error reconnecting to saved SQL Login auth type connection - "Cannot set the AccessToken property..."


dbreese-keystone:
- MSSQL Extension Version: 1.10.0
- VSCode Version: 1.52.1
- OS Version: Windows 10 Pro

Steps to Reproduce:

1.  Open a .sql file in VS Code.
2.  Attempt to execute a SQL query (CTRL+SHIFT+E).
3.  Select "Create new connection" and fill in the prompts to connect to an on-prem SQL server with authentication type as SQL Login.  (Do not save password.)
4.  Give the connection profile a name and save it.
5.  View the query results, it runs successfully.
6.  Then minimize VS Code for 5-10 minutes and/or close VS Code and re-open it.
7.  Open the same .sql file and try to run the same query again, using the saved connection profile.
8.  Enter the connection password when prompted.

Actual result:

Get the following error message:

```
mssql: Failed to connect: System.InvalidOperationException: Cannot set the AccessToken property if 'UserID', 'UID', 'Password', or 'PWD' has been specified in connection string.
   at Microsoft.Data.SqlClient.SqlConnection.CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessToken(SqlConnectionString connectionOptions)
   at Microsoft.Data.SqlClient.SqlConnection.set_AccessToken(String value)
   at Microsoft.SqlTools.ServiceLayer.Connection.ReliableConnection.ReliableSqlConnection..ctor(String connectionString, RetryPolicy connectionRetryPolicy, RetryPolicy commandRetryPolicy, String azureAccountToken) in D:\a\1\s\src\Microsoft.SqlTools.ManagedBatchParser\ReliableConnection\ReliableSqlConnection.cs:line 77
   at Microsoft.SqlTools.ServiceLayer.Connection.SqlConnectionFactory.CreateSqlConnection(String connectionString, String azureAccountToken) in D:\a\1\s\src\Microsoft.SqlTools.ServiceLayer\Connection\SqlConnectionFactory.cs:line 25
   at Microsoft.SqlTools.ServiceLayer.Connection.ConnectionService.TryOpenConnection(ConnectionInfo connectionInfo, ConnectParams connectionParams) in D:\a\1\s\src\Microsoft.SqlTools.ServiceLayer\Connection\ConnectionService.cs:line 532
```

Expected result:

[...]

PR #16932, opened 2021-01-13
fixed broken connections

cssuh:
Fixes https://github.com/microsoft/vscode-mssql/issues/16930

Modified the connection store section to only add the `azureAccountToken` field if it is an AzureMFA connection - users reported that this field was getting added for their SQL Login/Integrated Auth connection scenarios, and removing that field fixed the issue.

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -231,7 +231,7 @@ export default class MainController implements vscode.Disposable {
     /**
      * remove azureAccountToken from settings.json
      */
-    private async sanitizeProfiles() {
+    private async sanitizeProfiles(): Promise<void> {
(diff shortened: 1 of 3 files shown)


abist (approved):

etkeys:
I side loaded this update by using the vsix file provided by [this comment](https://github.com/microsoft/vscode-mssql/issues/16930#issuecomment-759711399). The update works correctly when adding new connection profiles. However, for existing profiles, the `azureAccessToken` element is not removed and attempting to make a connection using an existing profile will still fail.

If these changes get released, is there a cleanup process that will take place during upgrade that will remove `azureAccessTokens` for existing profiles if it will cause a conflict or will end users be responsible for removing that element themselves or knowing to deleting then recreate the profiles?

abist:
Looks good to me!

PR #16936, opened 2021-01-19
simplify if statements

cssuh:
combine if statements for readability


src/controllers/mainController.ts
```diff
@@ -234,11 +234,9 @@ export default class MainController implements vscode.Disposable {
     private async sanitizeProfiles(): Promise<void> {
         let userConnections: any[] = this._vscodeWrapper.getConfiguration(Constants.extensionName).get(Constants.connectionsArrayName);
         for (let conn of userConnections) {
-            if (conn.authenticationType !== 'AzureMFA') {
-                if (conn!.azureAccountToken === '') {
-                    // remove azure account token
-                    conn.azureAccountToken = undefined;
-                }
+            if (conn && conn.authenticationType !== 'AzureMFA' && conn.azureAccountToken === '') {
```
aasimkhan30, line 237: We probably don't need the second condition

aasimkhan30 (changes requested):

aasimkhan30 (approved):

Issue #16965, opened 2021-03-09
Syntax support for ANALYTIC & OFFSET window functions


life-droid:
- MSSQL Extension Version: SQL Sever 2016 (SP1-GDR) 13.0.4259.0
- VSCode Version:  1.54.1
- OS Version: Windows_NT x64 10.0.19042

Steps to Reproduce:

1. Use [ANALYTIC](https://docs.microsoft.com/en-us/sql/t-sql/functions/analytic-functions-transact-sql?view=sql-server-ver15) WINDOW functions such as FIRST_VALUE or CUME_DIST 
2. Use keyword WINDOW

Expectation is similar colouring as per other AGGREGATE or RANKING functions and other keywords like PARTITION

PR #16966, opened 2021-03-09
Updated grammar to support ANALYTIC WINDOW functions

life-droid:
Added syntax support for ANALYTIC WINDOW functions such as LAST_VALUE and CUME_DIST.
Added keyword WINDOW

Fixes issue #16965 

@alexr00 fixes vscode issue [#118500](https://github.com/microsoft/vscode/issues/118500)


alexr00:
Just pulled these updates into VS Code, they'll be available in version 1.61.

PR #17000, opened 2021-06-09
Adding patterns for hash comments

a-stewart:
The SQL syntax highlighting in VSCode recognises the whitespace leading up to a hash as proceeding a comment, but doesn't recognise the comment itself:

image

This seems to be because the patterns array is empty which is fixed in this PR.


a-stewart:
Added issue #17005

PR #17010, opened 2021-07-14
Export function to prompt user to select/create connection

Charles-Gagnon:
Exports the `promptForConnection` call (formerly `showConnections`) for external extensions to use. This meant moving the `IConnectionCredentials` interface to the typings file as well, which I renamed to `IConnectionInfo` since it contains more than just "credentials". 

I also removed the showExistingConnections parameter from the function since callers can just call the `createAndSaveProfile` function themselves if they just want to create a new connection.

### DIFF
--- a/src/views/connectionUI.ts
+++ b/src/views/connectionUI.ts
@@ -3,15 +3,14 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-'use strict';
-import vscode = require('vscode');
-import Constants = require('../constants/constants');
+import * as vscode from 'vscode';
+import * as constants from '../constants/constants';
 import LocalizedConstants = require('../constants/localizedConstants');
 import { ConnectionCredentials } from '../models/connectionCredentials';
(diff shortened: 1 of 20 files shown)


src/views/connectionUI.ts
```diff
@@ -81,37 +79,24 @@ export class ConnectionUI {
...
-                    });
-            }
-        });
+
+    /**
+     * Helper to let user choose a connection from a picklist, or to create a new connection.
+     * Return the ConnectionInfo for the user's choice
+     * @returns The connection picked or created.
+     */
+    public async promptForConnection(): Promise<IConnectionInfo | undefined> {
```
kburtram, line 88: it's hard to tell if this new method has the same UX as the previous implementation from looking at the code.  I'm assuming the connection process isn't being changed here?

kburtram (approved):

src/views/connectionUI.ts
Charles-Gagnon: Thanks for looking, yeah I didn't purposely change anything and in my testing it behaves the same as it did previously so there shouldn't be any issues.

PR #17011, opened 2021-07-15
Added more components

JericHunter:
The components that I have added are the text, inputbox, and dropdown components. For each component there is a implementation file for them. With each addition it meant changes for the interfaces typing file along with our Model Builder implementation file to import the newly added components.

### DIFF
--- a/src/dialogs/dropdownImpl.ts
+++ b/src/dialogs/dropdownImpl.ts
@@ -1,0 +1,66 @@
+import { ComponentImpl } from "./componentImpl";
+import { DropDownComponent, ModelComponentTypes, ComponentEventType, CategoryValue} from "./interfaces";
+import { Emitter } from "vscode-languageclient";
+import * as vscode from 'vscode';
+
+export class DropDownWrapper extends ComponentImpl implements DropDownComponent {
+
+	constructor(id: string) {
+		super(ModelComponentTypes.DropDown, id);
+		this.properties = {};
+		this._emitterMap.set(ComponentEventType.onDidChange, new Emitter<any>());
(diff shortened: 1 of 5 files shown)


src/dialogs/dropdownImpl.ts
```diff
@@ -0,0 +1,66 @@
+import { ComponentImpl } from "./componentImpl";
+import { DropDownComponent, ModelComponentTypes, ComponentEventType, CategoryValue} from "./interfaces";
+import { Emitter } from "vscode-languageclient";
+import * as vscode from 'vscode';
+
+export class DropDownWrapper extends ComponentImpl implements DropDownComponent {
```
kburtram, line 6: let's stick with the ...Impl naming convention instead of mixing with ...Wrapper

src/dialogs/inputboxImpl.ts
```diff
@@ -0,0 +1,116 @@
+import { Emitter } from "vscode-languageclient";
```
kburtram, line 1: add copywrite headers to all files

src/dialogs/modelBuilderImpl.ts
```diff
@@ -2,12 +2,14 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
-
```
kburtram, line 5: nit: would probably keep the blank line between header and imports for consistency with other files.

kburtram (approved):
LGTM

PR #17014, opened 2021-07-20
Expose listDatabases API

Charles-Gagnon:
Another thing to be used by the sql-database-projects extension.

We create a new connection because it's possible that a connection is being used that isn't actually associated with an editor - so we wouldn't have a URI for it already we could use.

As mentioned in the comment, a possible improvement if we wanted to make this as generic and re-usable as possible would be to optionally take in a URI instead so that we could re-use connections. But given the specific use case of this and the pooling nature of sqlclient I don't think this is a scenario we have to be too concerned about so am going with this for the time being until we need otherwise.

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -524,7 +524,7 @@ export default class ConnectionManager {
         const connectionPromise = new Deferred<boolean>();
         this.connect(uri, connectionInfo, connectionPromise);
         await connectionPromise;
-        let listParams = new ConnectionContracts.ListDatabasesParams();
+        const listParams = new ConnectionContracts.ListDatabasesParams();
         listParams.ownerUri = uri;
         const result = await this.client.sendRequest(ConnectionContracts.ListDatabasesRequest.type, listParams);
         return result.databaseNames;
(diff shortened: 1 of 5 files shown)


kisantia (approved):

src/controllers/connectionManager.ts
```diff
@@ -511,6 +510,26 @@ export default class ConnectionManager {
...
+     */
+    public async listDatabases(connectionInfo: IConnectionInfo): Promise<string[]> {
+        // Currently we're creating a new connection each time, this could be updated to take
+        // in a URI as well for re-using connections, but that case isn't something we need now
+        // so just leaving it simpler like this.
+        const uri = Utils.generateQueryUri().toString();
+        const connectionPromise = new Deferred<boolean>();
+        this.connect(uri, connectionInfo, connectionPromise);
+        await connectionPromise;
+        let listParams = new ConnectionContracts.ListDatabasesParams();
```
alanrenmsft, line 527: this variable can be a const.

alanrenmsft (approved):

abist (approved):

PR #17018, opened 2021-07-21
Update SQL Tools Service

Charles-Gagnon:
Bringing in the latest updates and fixes.

@abist @kburtram What's the process for uploading these binaries to download center for the [production config](https://github.com/microsoft/vscode-mssql/blob/main/src/configurations/production.config.json)?

I've ran through some of the basic scenarios and things seem to work as expected.

### DIFF
--- a/src/configurations/dev.config.json
+++ b/src/configurations/dev.config.json
@@ -1,19 +1,19 @@
 {
     "service": {
         "downloadUrl": "https://github.com/Microsoft/sqltoolsservice/releases/download/v{#version#}/microsoft.sqltools.servicelayer-{#fileName#}",
-        "version": "3.0.0-release.50",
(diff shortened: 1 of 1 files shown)


Charles-Gagnon:
Ping @kburtram @abist

kburtram (approved):

kburtram:
@Charles-Gagnon for SQL Tools Service updates we'll typically upload the binaries to MS download center prior to producing the final build, to avoid uploading multiple times for a single release.  We use the normal download uploader tool for this.

Charles-Gagnon:
Gotcha, don't know how to do that myself but guess I'll figure that out when it comes down to it!

Going to wait a little bit before merging this until I'm sure I won't need to do any other STS updates for now though.

PR #17019, opened 2021-07-22
Export connect API and have listDatabases use existing URI

Charles-Gagnon:
Turns out that we need a connection URI later on for the actual deployment - so I added and exported the API for doing that. I also modified listDatabases to take in a URI instead to reduce the number of duplicate connections being made.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -10,6 +10,8 @@ import * as LocalizedConstants from './constants/localizedConstants';
 import MainController from './controllers/mainController';
 import VscodeWrapper from './controllers/vscodeWrapper';
 import { IConnectionInfo, IExtension } from 'vscode-mssql';
+import { Deferred } from './protocol';
+import * as utils from './models/utils';
 
 let controller: MainController = undefined;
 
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -32,8 +34,24 @@ export async function activate(context: vscode.ExtensionContext): Promise<IExten
...
-        listDatabases: (connectionInfo: IConnectionInfo) => {
-            return controller.connectionManager.listDatabases(connectionInfo);
+        connect: async (connectionInfo: IConnectionInfo) => {
+
+            const uri = utils.generateQueryUri().toString();
+            const connectionPromise = new Deferred<boolean>();
+            // First wait for initial connection request to succeed
+            const requestSucceeded = await controller.connectionManager.connect(uri, connectionInfo, connectionPromise);
+            if (!requestSucceeded) {
+                throw new Error(`Connection request for ${JSON.stringify(connectionInfo)} failed`);
```
kisantia, line 44: do these error message need to be localized?

kisantia (approved):

src/extension.ts
Charles-Gagnon: Generally we shouldn't be localizing error messages - instead if callers want to display the errors they should catch them and display a localized error message themselves.

Charles-Gagnon: (but we're also not consistent about that at all so * shrug *. I'll look at the current connection functions and see what they do there)

Charles-Gagnon: Yeah most other errors here aren't localized either so I'll keep with that being consistent. Generally the failure isn't going to be coming from these anyways - if the connection itself fails the promise we pass in to connect is going to reject and so that's the error that will be sent up.

PR #17020, opened 2021-07-23
Add SqlTaskService

Charles-Gagnon:
Hooks into the SQL Task feature and displays notifications for ongoing tasks. The primary use for this currently is for the SQL DB Project to display the publish/generate script tasks but I made this generic enough to display progress for any tasks that are ongoing.

Currently it doesn't support other features such as canceling tasks or dealing with child tasks - those can be added later on as needed.

I'm testing it with the project scenarios outlined above, but it's fairly likely that if other features decide to use this that there will be oddities with the display strings. I noticed some weirdness with ordering of events - for example I'm getting a status of completed, followed by a status of in-progress with a message of "Completed" followed then by another one with status of completed except that one has a duration. I'll investigate those if they end up causing problems, but my intention here is just to get the basic framework working for at least the mainline scenarios and then fixing up further issues as we discover them. 

VsCodeTasks

### DIFF
--- a/src/services/sqlTasksService.ts
+++ b/src/services/sqlTasksService.ts
@@ -65,8 +65,7 @@ export class SqlTasksService {
 
     private _activeTasks = new Map<string, ActiveTaskInfo>();
 
-    constructor(
-        private _client: SqlToolsServiceClient) {
+    constructor(private _client: SqlToolsServiceClient) {
         this._client.onNotification(TaskCreatedNotification.type, taskInfo => this.handleTaskCreatedNotification(taskInfo));
         this._client.onNotification(TaskStatusChangedNotification.type, taskProgressInfo => this.handleTaskChangedNotification(taskProgressInfo));
     }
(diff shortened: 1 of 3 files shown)


src/services/sqlTasksService.ts
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * A simple service that hooks into the SQL Task Service feature provided by SQL Tools Service. This handles detecting when
+ * new tasks are started and displaying a progress notification for those tasks while they're running.
+ */
+export class SqlTasksService {
+
+    private _activeTasks = new Map<string, ActiveTaskInfo>();
+
+    constructor(
+        private _client: SqlToolsServiceClient) {
```
kisantia, line 69: move up to same line as the constructor( above

src/services/sqlTasksService.ts
```diff
@@ -0,0 +1,179 @@
...
+            // Task is completed, complete the progress notification and display a final toast informing the
+            // user of the final status.
+            this._activeTasks.delete(taskProgressInfo.taskId);
+            taskInfo.completionPromise.resolve();
+            // Only include the message if it isn't the same as the task status string we already have - some (but not all) task status
+            // notifications include this string as the message
+            const taskMessage = taskProgressInfo.message && taskProgressInfo.message.toLowerCase() !== taskStatusString.toLowerCase() ?
+                utils.formatString(localizedConstants.taskStatusWithMessage, taskInfo.taskInfo.name, taskStatusString, taskProgressInfo.message) :
+                utils.formatString(localizedConstants.taskStatusWithName, taskInfo.taskInfo.name, taskStatusString);
+            vscode.window.showInformationMessage(taskMessage);
```
kisantia, line 127: it would be nice if there was also showErrorMessage() if the TaskStatus was failed

kisantia (approved):

src/services/sqlTasksService.ts
Charles-Gagnon: Good idea!

PR #17032, opened 2021-08-05
Cleanup

Charles-Gagnon:
- Update tslint to latest (we should switch to eslint at some point, but out of scope for this)
- Remove 'use strict' from ts files - those are added automatically by compiler
- require imports -> newer import style (there were a few edge cases that I left as requires for various reasons, either because they needed to be or because the compiler was giving some weird errors when using import which I didn't have time to investigate)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,8401 @@
+{
+  "name": "mssql",
+  "version": "1.10.1",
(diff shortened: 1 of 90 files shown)


abist (approved):
Thanks!

PR #17035, opened 2021-08-05
Enable noUnusedLocals

Charles-Gagnon:
Enable the compiler option to not allow unused locals - code hygiene thing. None of these things were being used so removing them shouldn't affect functionality.


abist (approved):

Charles-Gagnon:
@cssuh There's one issue left : https://github.com/microsoft/vscode-mssql/commit/a5f772757783aec28527151109c150bff83a1a4d#diff-5eeace56def153787efa337fcd2f2e2771a202d038a1da07302cb41f6672a255R435

Here you're getting a "refreshedProfile" back but not doing anything with it. This seems like a bug...would you be able to look at this and figure out what's supposed to be happening here?

Charles-Gagnon:
Ok so I don't think this is actually a bug - the getTokens call is populating the passed profile with the new tokens and then just returning it. Can you confirm that that's the intended behavior here? In that case I'll just remove the var since it's not needed.

cssuh:
> Ok so I don't think this is actually a bug - the getTokens call is populating the passed profile with the new tokens and then just returning it. Can you confirm that that's the intended behavior here? In that case I'll just remove the var since it's not needed.

Yes this is the intended behavior.

PR #17037, opened 2021-08-12
Connection Profile Interface optional fields 

llali:
Making some connection properties accept undefined so they will be set to default value


Charles-Gagnon (commented):
Please also update https://github.com/microsoft/vscode-mssql/blob/main/src/models/connectionCredentials.ts

llali:
> Please also update https://github.com/microsoft/vscode-mssql/blob/main/src/models/connectionCredentials.ts

done

Charles-Gagnon (approved):

PR #17043, opened 2021-08-19
expose getAzureFunctions

kisantia:
Adds the getAzureFunctions api that was added to SqlToolsService in https://github.com/microsoft/sqltoolsservice/pull/1228. This will be used by the add sql bindings flow in sql database projects (https://github.com/microsoft/azuredatastudio/pull/16553)

### DIFF
--- a/typings/vscode-mssql.d.ts
+++ b/typings/vscode-mssql.d.ts
@@ -264,8 +264,21 @@ declare module 'vscode-mssql' {
     }
 
     export interface IAzureFunctionsService {
+        /**
+         * Adds a SQL Binding to a specified Azure function in a file
+         * @param bindingType Type of SQL Binding
+         * @param filePath Path of the file where the Azure Functions are
+         * @param functionName Name of the function where the SQL Binding is to be added
+         * @param objectName Name of Object for the SQL Query
+         * @param connectionStringSetting Setting for the connection string
+         */
         addSqlBinding(bindingType: BindingType, filePath: string, functionName: string, objectName: string, connectionStringSetting: string): Thenable<ResultStatus>;
-        getAzureFunctions(filePath: string): Thenable<getAzureFunctionsResult>;
+        /**
+         * Gets the names of the Azure functions in the file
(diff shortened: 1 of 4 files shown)


typings/vscode-mssql.d.ts
```diff
@@ -265,6 +265,7 @@ declare module 'vscode-mssql' {
 
     export interface IAzureFunctionsService {
         addSqlBinding(bindingType: BindingType, filePath: string, functionName: string, objectName: string, connectionStringSetting: string): Thenable<ResultStatus>;
+        getAzureFunctions(filePath: string): Thenable<getAzureFunctionsResult>;
```
Charles-Gagnon, line 268: Comments for things added to this file please

alanrenmsft (approved):

typings/vscode-mssql.d.ts
```diff
@@ -554,4 +555,12 @@ declare module 'vscode-mssql' {
         bindingType: BindingType;
         connectionStringSetting: string;
     }
+
+    export interface getAzureFunctionsParams {
```
Charles-Gagnon, line 559: Interfaces should be CapitalCased

typings/vscode-mssql.d.ts
```diff
@@ -554,4 +555,12 @@ declare module 'vscode-mssql' {
...
+        filePath: string;
+    }
+
+    export interface getAzureFunctionsResult extends ResultStatus {
+        azureFunctions: string[];
```
Charles-Gagnon, line 564: Something to consider for the future is making this return an object instead - that for right now can just contain a single name property. But it seems to me that it's pretty likely you'll want to get other data associated with a function eventually and this would make it a lot easier to work with later. 

But it's probably fine to just leave it like this for now anyways, since it also seems pretty unlikely that anyone else would be using this so updating later when that happens wouldn't be too hard.

Charles-Gagnon (approved):

abist (approved):

PR #17045, opened 2021-08-23
Add more comments

kisantia:
Adding more comments that got added in https://github.com/microsoft/azuredatastudio/pull/16838

### DIFF
--- a/typings/vscode-mssql.d.ts
+++ b/typings/vscode-mssql.d.ts
@@ -568,10 +568,29 @@ declare module 'vscode-mssql' {
      * Parameters for adding a SQL binding to an Azure function
      */
     export interface AddSqlBindingParams {
+        /**
+         * Aboslute file path of file to add SQL binding
+         */
         filePath: string;
+
+        /**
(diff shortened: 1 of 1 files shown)


abist (approved):

typings/vscode-mssql.d.ts
```diff
@@ -569,10 +575,16 @@ declare module 'vscode-mssql' {
...
 
+    /**
+     * Parameters for getting the names of the Azure functions in a file
+     */
     export interface GetAzureFunctionsParams {
         filePath: string;
     }
 
+    /**
+     * Result from a get Azure functions request
```
Charles-Gagnon, line 586: May want to consider adding some comments for the parameters as well. For example on 582 for filePath what types of "paths" does it support? Relative/absolute? What about file share?

And for this one - what are the string values? (this goes back to my earlier comment about either renaming this to azureFunctionNames or having it return an object that has a name property). But at least adding a comment explaining that these are the names of the functions is an ok start.

Charles-Gagnon (approved):

Issue #17044, opened 2021-08-20
Default isolation level is set to `read uncommitted`


sjp363:
- MSSQL Extension Version:  1.7+
- VSCode Version: 1.59.1
- OS Version: windows 10 21H1

Steps to Reproduce:

1. version mssql 1.6;   execute dbcc useroptions; isolation level = "read committed" 
2. version mssql 1.7+; execute dbcc useroptions; isolation level = "read UNcommitted" 
not able to permanently change or set default using statement "set transaction isolation level read committed"

sjp363:
@kburtram you are the man 💯 !! thanks so much, that fixed it. 😄

kburtram:
@sjp363 happy to hear that resolved the problem!  I'll leave this open to track adding the settings into the vscode settings UI and possibly changing the default isolation mode.  Thanks!

PR #17052, opened 2021-08-31
Add query execution settings

kburtram:
This fixes https://github.com/microsoft/vscode-mssql/issues/17044 by changing the default isolation level to `read committed` to match SSMS.  This also adds in several other settings into the package.json so those options are displayed in the User Settings UI.


package.nls.json
```diff
@@ -93,5 +93,33 @@
...
-"mssql.queryHistoryLimit":"Number of query history entries to show in the Query History view"
+"mssql.queryHistoryLimit":"Number of query history entries to show in the Query History view",
+"mssql.query.maxXmlCharsToStore":"Number of XML characters to store after running a query",
+"mssql.tracingLevel":"[Optional] Log level for backend services. Azure Data Studio generates a file name every time it starts and if the file already exists the logs entries are appended to that file. For cleanup of old log files see logRetentionMinutes and logFilesRemovalLimit settings. The default tracingLevel does not log much. Changing verbosity could lead to extensive logging and disk space requirements for the logs. Error includes Critical, Warning includes Error, Information includes Warning and Verbose includes Information",
+"mssql.logRetentionMinutes":"Number of minutes to retain log files for backend services. Default is 1 week.",
+"mssql.logFilesRemovalLimit":"Maximum number of old files to remove upon startup that have expired mssql.logRetentionMinutes. Files that do not get cleaned up due to this limitation get cleaned up next time Azure Data Studio starts up.",
```
abist, line 100: I think there might be an extra `mssql.logRetentionMinutes` here

abist (approved):
One small comment in the strings, but otherwise looks good

package.nls.json
kburtram: I don't see two `mssql.logRetentionMinutes` settings in this file.  Could you refer me to the lines with the duplicate entries?

abist: In this line 100, we have
`"Maximum number of old files to remove upon startup that have expired mssql.logRetentionMinutes. Files that do not get cleaned up due to this limitation get cleaned up next time Azure Data Studio starts up."` 
Was the ` mssql.logRetentionMinutes` in the first sentence intended?

kburtram: It was copy/paste from ADS source code https://github.com/microsoft/azuredatastudio/blob/main/extensions/mssql/package.nls.json#L55, which also has the reference to `mssql.logRetentionMinutes`.  I think this is saying that setting controls the time the files will expire.  Not sure why this setting is even needed.  I'm just exposing the same settings we have in ADS.  I'll investigate separately if this setting is actually useful.

PR #17056, opened 2021-09-15
SQL Binding for Azure Functions

abist:
Added bindings from generated Azure Functions and then format the function.


src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,175 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import * as os from 'os';
+import * as vscode from 'vscode';
+import * as af from '../../typings/vscode-azurefunctions.api';
+import * as cp from 'child_process';
+import * as fs from 'fs';
+import * as mssql from 'vscode-mssql';
+import { azureFunctionsExtensionName, defaultBindingResult, defaultSqlBindingTextLines, genericCollectionImport, sqlBindingResult, sqlConnectionString } from '../constants/constants';
+import LocalizedConstants = require('../constants/localizedConstants');
+import path = require('path');
```
kisantia, line 13: does this import need to use requires and not the format like the other imports above?

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,175 @@
...
+import { AzureFunctionsService } from '../services/azureFunctionsService';
+
+const sqlBindingNugetSource = 'https://www.myget.org/F/azure-appservice/api/v3/index.json';
+const sqlBindingPackageName = 'Microsoft.Azure.WebJobs.Extensions.Sql';
+const sqlBindingPackageVersion = '1.0.0-preview3';
```
kisantia, line 18: it would be nice if it pulls in the latest version so this doesn't need to be updated every time a new version is released

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,175 @@
...
+    private async isAzureFunctionProjectOpen(): Promise<boolean> {
+        if (vscode.workspace.workspaceFolders === undefined || vscode.workspace.workspaceFolders.length === 0) { return false; }
+        const projFile = await this.getProjectFile();
+        const hostFile = await this.getHostFile();
+        return projFile !== undefined && hostFile !== undefined;
+    }
+
+    private async getProjectFile(): Promise<string | undefined> {
+        const projFiles = await vscode.workspace.findFiles('**/*.csproj');
+        return projFiles.length > 0 ? projFiles[0].fsPath : undefined;
```
kisantia, line 117: what happens if there's more than one csproj in the workspace?

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,175 @@
...
+        if (currentSources.indexOf(sqlBindingNugetSource) === -1) {
+            await this.executeCommand(`dotnet nuget add source ${sqlBindingNugetSource}`);
+        }
+        const projFile = await this.getProjectFile();
+        await this.executeCommand(`dotnet add package ${sqlBindingPackageName} --version ${sqlBindingPackageVersion}`, path.dirname(projFile));
+    }
+
+    private async addConnectionStringToConfig(connectionString: string): Promise<void> {
+        const settingsFile = await this.getSettingsFile();
+        const content = await fs.promises.readFile(settingsFile);
```
kisantia, line 152: Does there need to be a check if settingsFile is undefined since `getSettingsFile()` can return undefined?

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,175 @@
...
+        const config = JSON.parse(content.toString());
+        if (!(sqlConnectionStringPropertyName in config.Values)) {
+            config.Values[sqlConnectionStringPropertyName] = connectionString;
+            await fs.promises.writeFile(settingsFile, JSON.stringify(config, undefined, 2));
```
kisantia, line 156: the azure functions extension has helper functions like `setLocalAppSetting()` that might be good to use since there's more error checking

abist:
@alanrenmsft Bump

alanrenmsft:
@chlafreniere @Charles-Gagnon , do you still need this feature?

Charles-Gagnon:
We'll likely want to implement this kind of behavior at some point, but for now I think it's fine to close the PR (but leave the branch for code reference)

chlafreniere:
@dzsquared and us are chatting about the best way to release these bits, but yes, ok to close for now (please keep the branch!!!)

alanrenmsft:
sounds good.

PR #17060, opened 2021-09-26
Include AzureMFA as valid value in settings

nic-paul:
Added support for Settings.json to recognise AzureMFA as a valid value for authenticationType.

### DIFF
--- a/package.json
+++ b/package.json
@@ -271,7 +271,8 @@
                 "default": "SqlLogin",
                 "enum": [
                   "Integrated",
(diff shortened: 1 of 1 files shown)


kburtram:
@nic-paul thanks for this contribution!

mgenereu:
@kburtram, why did this merge into dev instead of main?

Charles-Gagnon:
Thanks for catching that @mgenereu! I ported it to the correct branch

Issue #17004, opened 2021-06-25
Azure SQL database query execution fails second day


kcaswick:
- MSSQL Extension Version: v1.10.1
- VSCode Version: Version: 1.57.1 (system setup)
Commit: 507ce72a4466fbb27b715c3722558bb15afa9f48
Date: 2021-06-17T13:28:07.755Z
Electron: 12.0.7
Chrome: 89.0.4389.128
Node.js: 14.16.0
V8: 8.9.255.25-electron.0
OS: Windows_NT x64 10.0.18362
- OS Version: Windows 10 Version 1903

Steps to Reproduce:

1. Add Connection via sidebar
2. Connect
3. Run SQL file multiple times the first day, without exiting, works fine
4. Shutdown entire computer
5. Start up computer and VS Code the next day
6. Click connect on connection in SQL Server side bar
7. Click "Execute Query" toolbar button while in a SQL file
8. Choose the connection profile from the list
9. The errors below appear, and the list re-appears
10. Choose the same connection profile again
11. Execution tab opens with just the message "[1:41:11 PM]	Executing query..." and a loading spinner.

> mssql: Error 10054: A connection was successfully established with the server, but then an error occurred during the login process. (provider: TCP Provider, error: 0 - An existing connection was forcibly closed by the remote host.)

> mssql: Execution failed: Cannot connect to the database due to invalid OwnerUri (Parameter 'OwnerUri')

Connection information from settings.json is:

```
        {
            "server": "xxxxxx01.database.windows.net",
            "database": "Xxxxx999Xxxxxx_2021-06-23T18-45Z",
            "authenticationType": "AzureMFA",
            "profileName": "Xxxxx999Xxxxxx as admin_xxxx",
            "azureAccountToken": "",
            "email": "admin_xxxx@xxxxxxxxx.onmicrosoft.com",
            "accountId": "aaa99a99-9999-9999-99a9-9a999999a999",
            "password": ""
        }
```

[...]

PR #17066, opened 2021-10-04
Fix query retry bug

cssuh:
fixes https://github.com/microsoft/vscode-mssql/issues/17004

It was not able to connect due to a missing azure account token, so I fetch the token before calling `connect()`


cssuh:
I'm going to make a helper method for fetching the account token because we do it in multiple places, but I will put that in a separate PR.

src/controllers/connectionManager.ts
```diff
@@ -600,6 +601,35 @@ export default class ConnectionManager {
             // close active connection
             await this.disconnect(fileUri);
             // connect to the server/database
+            if (connectionCreds.authenticationType === Constants.azureMfa) {
+                if (!connectionCreds.azureAccountToken) {
+                    let account = this.accountStore.getAccount(connectionCreds.accountId);
+                    let profile = new ConnectionProfile();
```
Charles-Gagnon, line 607: Do we not have a clone function on connection profile? That would be nice to add if not.

src/controllers/connectionManager.ts
```diff
@@ -600,6 +601,35 @@ export default class ConnectionManager {
             // close active connection
             await this.disconnect(fileUri);
             // connect to the server/database
```
Charles-Gagnon, line 603: Move this comment down below your changes

src/controllers/connectionManager.ts
```diff
@@ -600,6 +601,35 @@ export default class ConnectionManager {
...
+                    profile.azureAccountToken = connectionCreds.azureAccountToken;
+                    profile.database = connectionCreds.database;
+                    profile.email = connectionCreds.email;
+                    profile.password = connectionCreds.password;
+                    profile.server = connectionCreds.server;
+                    let azureAccountToken = await this.azureController.refreshToken(account, this.accountStore, providerSettings.resources.databaseResource);
+                    if (!azureAccountToken) {
+                        let errorMessage = LocalizedConstants.msgAccountRefreshFailed;
+                        await this.vscodeWrapper.showErrorMessage(
+                            errorMessage, LocalizedConstants.refreshTokenLabel).then(async refreshResult => {
```
abist, line 619: I would stick to just using awaits here instead of using a Promise.then just to be consistent

package.json
```diff
@@ -122,6 +121,8 @@
     "find-remove": "1.2.1",
     "fs-extra-promise": "^0.3.1",
     "getmac": "1.2.1",
+    "gulp": "^4.0.0",
```
Charles-Gagnon, line 124: What are these changes for?

src/controllers/connectionManager.ts
```diff
@@ -600,6 +601,35 @@ export default class ConnectionManager {
...
+                            if (refreshResult === LocalizedConstants.refreshTokenLabel) {
+                                await this.azureController.getTokens(
+                                    profile, this.accountStore, providerSettings.resources.databaseResource);
+
+                            } else {
+                                return undefined;
```
abist, line 625: I think we should be rejecting the promise or throw an error instead of returning undefined connectionInfo

src/controllers/connectionManager.ts
```diff
@@ -600,6 +601,35 @@ export default class ConnectionManager {
             // close active connection
             await this.disconnect(fileUri);
             // connect to the server/database
+            if (connectionCreds.authenticationType === Constants.azureMfa) {
+                if (!connectionCreds.azureAccountToken) {
+                    let account = this.accountStore.getAccount(connectionCreds.accountId);
+                    let profile = new ConnectionProfile();
+                    profile.accountId = connectionCreds.accountId;
```
abist, line 608: It would be cleaner if we could just add a static method to create connection profiles in one constructor/helper function in the connection profile class

package.json
Charles-Gagnon: gulp shouldn't be a runtime dependency so not sure why you moved the stuff down here.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -421,13 +421,7 @@ export class ObjectExplorerService {
...
                     let account = this._connectionManager.accountStore.getAccount(connectionCredentials.accountId);
                     let profile = new ConnectionProfile();
-                    profile.accountId = connectionCredentials.accountId;
-                    profile.authenticationType = connectionCredentials.authenticationType;
-                    profile.azureAccountToken = connectionCredentials.azureAccountToken;
-                    profile.database = connectionCredentials.database;
-                    profile.email = connectionCredentials.email;
-                    profile.password = connectionCredentials.password;
-                    profile.server = connectionCredentials.server;
+                    profile.setProfile(connectionCredentials);
```
Charles-Gagnon, line 424: Why not just have this be a part of the constructor? Seems easier and I don't know of a scenario where we'd want to set it afterwards

src/models/connectionProfile.ts
```diff
@@ -152,4 +153,14 @@ export class ConnectionProfile extends ConnectionCredentials implements IConnect
         }
         return choices;
     }
+
+    public setProfile(connectionCredentials: IConnectionInfo): void {
```
abist, line 157: This is fine, but you could perhaps create a connection profile within this helper and then return that instead of just setting fields, unless you plan to use this helper elsewhere

src/controllers/connectionManager.ts
```diff
@@ -672,6 +673,27 @@ export default class ConnectionManager {
     // create a new connection with the connectionCreds provided
     public async connect(fileUri: string, connectionCreds: IConnectionInfo, promise?: Deferred<boolean>): Promise<boolean> {
         const self = this;
+        if (connectionCreds.authenticationType === Constants.azureMfa) {
```
Charles-Gagnon, line 676: Comment explaining what this code block is doing

src/controllers/connectionManager.ts
```diff
@@ -672,6 +673,27 @@ export default class ConnectionManager {
...
+            if (!connectionCreds.azureAccountToken) {
```
Charles-Gagnon, line 677: What if the token exists but is expired?

caohai: Yes, once the existing token gets expired, all later requests using that token will fail. What I did with Query Editor in ADS was to store the expiry timestamp along with the token itself. We could do something similar here and refresh token if the token doesn't exist  or token is expired.

kburtram: we definitely need to port of the renew token fix from ADS since that bug significantly impacts the AAD querying experience.  Though if that bug is unrelated to the bug being resolved with this PR it's probably fine to take that as a follow-up PR.

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -125,6 +125,9 @@
         <trans-unit id="azureAddAccount">
             <source xml:lang="en">Add an Account...</source>
         </trans-unit>
+        <trans-unit id="cannotConnect">
+            <source xml:lang="en">Cannot connect due to expired tokens</source>
```
Charles-Gagnon, line 129: Maybe something like `Cannot connect due to expired tokens. Please reauthenticate and try again.`? This just seems a bit abrupt.

PR #17069, opened 2021-10-05
Adding an option to connect API to save the connection

llali:
Adding an option to connect API to save the connection


src/extension.ts
```diff
@@ -50,6 +51,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<IExten
             const connectionSucceeded = await connectionPromise;
             if (!connectionSucceeded) {
                 throw new Error(`Connection for ${JSON.stringify(connectionInfo)} failed`);
+            } else if (saveConnection) {
+                let profile: IConnectionProfile = <IConnectionProfile>connectionInfo;
```
kisantia, line 55: const instead of let

src/extension.ts
```diff
@@ -37,7 +38,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<IExten
         promptForConnection: (ignoreFocusOut?: boolean) => {
             return controller.connectionManager.connectionUI.promptForConnection(ignoreFocusOut);
         },
-        connect: async (connectionInfo: IConnectionInfo) => {
+        connect: async (connectionInfo: IConnectionInfo, saveConnection?: boolean) => {
```
Charles-Gagnon, line 41: IConnectionProfile already has saveConnection on it...so I think instead you should be looking at utilizing that. Maybe by moving that interface into the API type declarations...although it's kind of weird to have these two types then that are basically the same thing.

Charles-Gagnon (changes requested):

src/extension.ts
llali: I was basically trying to be consistent with ADS. ADS has connect API with saveConnection in the method. Although it's also part of connection profile

llali: well that's what being consistent means :) if we decided that's how to same a connection in ADS then I assume we thought it over. but if that's not the right way to do then I can also add a separate API for saving connection and get the IConnectionProfile as input, this way I won't have problem with casting. Also I need to add an option to save in workspace or user setting, it would be easier to add that into a new API function

src/extension.ts
```diff
@@ -50,6 +51,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<IExten
             const connectionSucceeded = await connectionPromise;
             if (!connectionSucceeded) {
                 throw new Error(`Connection for ${JSON.stringify(connectionInfo)} failed`);
+            } else if (saveConnection) {
+                let profile: IConnectionProfile = <IConnectionProfile>connectionInfo;
+                await controller.connectionManager.connectionStore.saveProfile(profile);
```
Charles-Gagnon, line 56: Where is the profile normally saved now? Is that just done in the connection picker flow or something? We should make sure there isn't somewhere later on that we can leverage the logic for to save the connection (say by converting this into a profile and setting saveConnection to true here and the passing it on to connect)

llali: connection will only get saved when saveconnection is set to true. I'm not sure what you mean here

src/extension.ts
llali: right now we assume they are all optional but  I think the correct to handle it is the API accept IConnectionInfo or IConnectionProfile so that if anyone wants to save, have to pass IconnectionProfile

src/extension.ts
Charles-Gagnon: I'm asking how the extension saves profiles currently. Because if I go through the connection profile picker then those connections are saved already. So where is that done? We shouldn't be duplicating logic if there's somewhere else that it's already saving connections we can utilize instead of calling save directly here.

llali: sure I can look it up again but what happens is, if connect using the profile picker, then connect method is called and then save method. it always saves after. but in the API we don't want to save all the connections using the API so that's why we give an option to call save if the option is set to true.

src/extension.ts
llali: I'm guessing in future when we work with projects in a workspace, we might want to store the connections for the project in the workspace so it can be shared with others using the workspace. but I'm now doing anything on that right now since nothing is final

src/extension.ts
Charles-Gagnon: Can you link where that's done?

Also what about my earlier question regarding calling connect for duplicate profiles?

llali: I made the change in the code to do the same as connection is saved from picker.

src/controllers/mainController.ts
```diff
@@ -617,6 +617,20 @@ export default class MainController implements vscode.Disposable {
         return false;
     }
 
+    /**
+     * Makes a connection and save if saveConnection is set to true
+     */
+    public async onConnect(uri: string, connectionInfo: IConnectionInfo, connectionPromise: Deferred<boolean>, saveConnection?: boolean): Promise<boolean> {
```
Benjin, line 623: docstring for return, please!  Looks like it returns whether the connection is both successful and saved, which wouldn't have been my first guess based on the name of the function

typings/vscode-mssql.d.ts
```diff
@@ -54,9 +54,10 @@ declare module 'vscode-mssql' {
          * Attempts to create a new connection for the given connection info. An error is thrown and displayed
          * to the user if an error occurs while connecting.
          * @param connectionInfo The connection info
+         * @param saveConnection Save the connection profile if sets to true
          * @returns The URI associated with this connection
          */
```
llali, line 61: add a warning so users know connection gets saved every time

src/controllers/mainController.ts
```diff
@@ -617,6 +617,29 @@ export default class MainController implements vscode.Disposable {
...
+     * Makes a connection and save if saveConnection is set to true
+     * @param uri The URI of the connection to list the databases for.
+     * @param connectionInfo The connection info
+     * @param connectionPromise connection promise object
+     * @param saveConnection saves the connection profile if sets to true
+     * @returns if saveConnection is set to true, returns true for successful connection and saving the profile
+     * otherwise returns true for successful connection
+     *
+     */
+     public async onConnect(uri: string, connectionInfo: IConnectionInfo, connectionPromise: Deferred<boolean>, saveConnection?: boolean): Promise<boolean> {
```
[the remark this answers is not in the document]
llali: no reason. sure

Issue #17064, opened 2021-09-30
Repeated keychain prompts in v1.10.2 & 1.11.0


leighghunt:
- MSSQL Extension Version: v1.10.2
- VSCode Version: 1.60.2
- OS Version: BigSur 11.5.2

Steps to Reproduce:

1. Start VS Code
2. Open SQL file
3. Shift-Cmd-E to execute
4. Choose connection profile
5. Dialog appears - **MicrosoftSqlToolsServiceLayer wants to use your confidential information stored in "M" in your keychain**
6. Enter password and click **Allow** or **Always Allow**.
7. Query will run, and can then be run successfully for rest of session.
8. However, if you disconnect and then reconnect to database, our restart VSCode, the security dialogue reappears.
<img width="447" alt="Screen Shot 2021-10-01 at 10 44 59 AM" src="https://user-images.githubusercontent.com/349153/135534423-052a440e-3991-40bc-be13-4365ed337538.png">

Note - if this is simply because it's a preview release and is perhaps unsigned, or something like that kicking off MacOSX being fussy, happy to close ticket.

Thanks for the work on getting it working on M1.

kburtram:
@leighghunt thanks for reporting!  It shouldn't be related to this being a preview build so we'll investigate to ensure this doesn't also impact the upcoming 1.11.0 build.

kburtram:
@cssuh can you take a look if anything's changed here in 1.11.0?  I think when we push an update users will need to reenter their keychain password to access saved passwords using the new binary, but this should be a one-time (or at least one-time per password).  Not something that keep reoccurring.

kburtram:
A few follow-up questions/responses regarding previous comments...

@chigia001 this extension saves credentials in the keystore for saved connections.  You should only get this prompt when connecting to a server, so if it's happening outside that scenario then that sounds like an unintended design change.  When the extension is updated the binary that reads these keys is updated and needs to be granted permissions to the keychain again.

@sharumpe @boxofyellow it's strange that this is happening on startup if you aren't connecting to a server.  Chris will investigate what's going on there.  Are you granting SQL Tools Service access and then it continues to prompt subsequently, or are you denying STS permissions?

@benzdouglas you would need to grant STS permissions to the keychain to read saved credential information and establish a connection.  Though this should be a one-time thing after upgrade.  Downgrading may workaround this issue since the previous version of STS may already have permissions to the keychain.

Please let me know if we're missing any details with this issue.  Thanks!

[...]

PR #17095, opened 2021-11-08
Remove credential service call upon startup

cssuh:
Credential service is only called when a connection is created, this change will also improve extension startup times
fixes https://github.com/microsoft/vscode-mssql/issues/17064

### DIFF
--- a/src/azure/azureController.ts
+++ b/src/azure/azureController.ts
@@ -73,8 +73,10 @@ export class AzureController {
     private context: vscode.ExtensionContext;
     private logger: AzureLogger;
     private _vscodeWrapper: VscodeWrapper;
+    private initialize: boolean;
 
     constructor(context: vscode.ExtensionContext, logger?: AzureLogger) {
+        this.initialize = false;
         this.context = context;
         if (!this.logger) {
             this.logger = new AzureLogger();
@@ -86,11 +88,6 @@ export class AzureController {
     public async init(): Promise<void> {
         this.authRequest = new AzureAuthRequest(this.context, this.logger);
         await this.authRequest.startServer();
-        let storagePath = await findOrMakeStoragePath();
-        let credentialStore = new CredentialStore();
-        this.cacheService = new SimpleTokenCache('aad', storagePath, true, credentialStore);
-        await this.cacheService.init();
-        this.storageService = this.cacheService.db;
         this.azureStringLookup = new AzureStringLookup();
         this.azureUserInteraction = new AzureUserInteraction(this.authRequest.getState());
         this.azureErrorLookup = new AzureErrorLookup();
@@ -188,7 +185,7 @@ export class AzureController {
 
     private async createAuthCodeGrant(): Promise<AzureCodeGrant> {
         let azureLogger = new AzureLogger();
-        await this.init();
+        await this.initializeHelper();
         return new AzureCodeGrant(
             providerSettings, this.storageService, this.cacheService, azureLogger,
             this.azureMessageDisplayer, this.azureErrorLookup, this.azureUserInteraction,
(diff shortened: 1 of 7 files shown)


Charles-Gagnon:
Link the issue please

src/azure/azureController.ts
```diff
@@ -73,8 +73,10 @@ export class AzureController {
     private context: vscode.ExtensionContext;
     private logger: AzureLogger;
     private _vscodeWrapper: VscodeWrapper;
+    private initialize: boolean;
 
     constructor(context: vscode.ExtensionContext, logger?: AzureLogger) {
+        this.initialize = false;
```
Charles-Gagnon, line 79: Just set it to false where you declare it

src/azure/azureController.ts
```diff
@@ -211,4 +208,19 @@ export class AzureController {
         await azureAuth.deleteAccountCache(account.key);
         return;
     }
+
+    /**
+     * Checks if this.init() has already been called
+     */
+    private async initializeHelper(): Promise<void> {
```
Charles-Gagnon, line 215: Suggest just calling this `initializeCredentialStore` or something similar to make it clear what it's doing

src/azure/azureController.ts
```diff
@@ -73,8 +73,10 @@ export class AzureController {
     private context: vscode.ExtensionContext;
     private logger: AzureLogger;
     private _vscodeWrapper: VscodeWrapper;
+    private initialize: boolean;
```
Charles-Gagnon, line 76: Would suggest `credentialStoreInitialized` to be explicit

src/views/connectionUI.ts
```diff
@@ -490,7 +489,7 @@ export class ConnectionUI {
      * false otherwise
      */
     public async handleFirewallError(uri: string, profile: IConnectionProfile, ipAddress: string): Promise<boolean> {
-        let azureController = new AzureController(this._context);
+        let azureController = this.connectionManager.azureController;
```
Charles-Gagnon, line 492: There's a bunch of places where we're creating the AzureController it looks like - are those fine? Why did this one need to change?

cssuh: I think those should be changed too, when we declare new azureControllers the credentialStoreInitialized will once again be false which will lead to the same code running again unnecessarily.

Charles-Gagnon: Are you planning on doing that here? What's the effect if we don't? Is that why we're getting multiple prompts - every time we create a new instance of the azure controller it prompts the user again for access?

cssuh: I have removed all the new azureController creations except for where it is created in connectionManager

src/models/connectionProfile.ts
```diff
@@ -53,6 +53,7 @@ export class ConnectionProfile extends ConnectionCredentials implements IConnect
         connectionStore: ConnectionStore,
         context: vscode.ExtensionContext,
         accountStore?: AccountStore,
+        azureController?: AzureController,
```
Charles-Gagnon, line 56: This should never be undefined I would think (just put it before the azureController to avoid the compile error)

src/azure/azureController.ts
```diff
@@ -211,4 +207,19 @@ export class AzureController {
         await azureAuth.deleteAccountCache(account.key);
         return;
     }
+
+    /**
+     * Checks if this.init() has already been called
```
Charles-Gagnon, line 212: Would suggest actually saying what this is doing (like what does it "initialize") and mention that it will only do that stuff once.

Charles-Gagnon (approved):

kburtram (approved):

Charles-Gagnon:
@cssuh Looks like you have some compile errors in the tests, make sure you run `gulp build` to build everything to catch all those types of errors.

Charles-Gagnon:
Also I'd suggest using one of the keywords to auto-close the issue when this PR is merged in (e.g. `fixes #17064`)

PR #17097, opened 2021-11-10
Update yaml pipeline

Charles-Gagnon:
Currently the ADO pipeline is using the UI based pipeline - but that makes it difficult to review changes or test things out.

The first step to using the yaml pipeline is just updating the current yaml we have to use the same content as the ADO pipeline (there's a "View YAML" button to do this)

So the edits made here are simply just copying over the ADO pipeline yaml to this one. There's some things - such as using npm instead of yarn - that we should fix up. But that can be done as a separate task. 

https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=128060&view=results

### DIFF
--- a/build/build.yml
+++ b/build/build.yml
@@ -1,8 +1,5 @@
 pool:
   name: Hosted macOS
-  demands:
-  - node.js
-
 steps:
 - task: ArchiveFiles@1
   displayName: 'Archive source '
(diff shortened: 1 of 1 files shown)


Charles-Gagnon:
@kburtram @abist Look good to you?

abist (approved):

PR #17100, opened 2021-11-11
Cleanup/update pipeline

Charles-Gagnon:
- Don't bother archiving source (that's what commit history is for)
- Specify image to use for builds
- Switch to yarn instead of npm
- Lock dependencies
- Use gulp from npm and bump version


Charles-Gagnon:
Merging this in so I can make other fixes/improvements. If you have any comments leave them here and I'll follow up

Issue #17105, opened 2021-11-12
Unable to add azure account 


cssuh:
- MSSQL Extension Version: 1.11.1
- VSCode Version: 1.62.2
- OS Version: macOS Big Sur 11.6

Steps to Reproduce:

1. Click "Add Connection"
2. Go through steps to add a connection, selecting AAD as the authentication method
3. Try to add an azure account
4. The following screen appears:
![Screen Shot 2021-11-12 at 1 19 53 PM](https://user-images.githubusercontent.com/18150417/141536053-047696a8-f9b2-424d-b4ce-b9385c84947b.png)

cssuh:
not seeing this anymore, something must have been wrong with my machine

PR #17106, opened 2021-11-12
Revert "Remove credential service call upon startup"

cssuh:
Reverts microsoft/vscode-mssql#17095
This introduced a different bug: https://github.com/microsoft/vscode-mssql/issues/17105 so I will revert it for now and find a better solution


Charles-Gagnon (approved):
Since we haven't released the build yet I'm not sure we need a bug that will (shortly) cease to exist, but if you find that easier for tracking I guess that's ok 😄

kburtram (approved):

cssuh:
not needed anymore, since the add account feature works (has been confirmed by myself and @abist)

PR #17107, opened 2021-11-12
Add ads-adal-library to @microsoft scope

Charles-Gagnon:
All packages owned by our team should generally be published under the @microsoft scope

Added basic instructions - for now this will be a manual process since it isn't something I anticipate needing to be updated very often.


Charles-Gagnon:
Merging in quickly since this is a very minor change...

PR #17115, opened 2021-11-16
Remove default profiles when adding nodes

abist:
Remove default profiles when populating the OE nodes. Not sure why we need this template, we could perhaps get rid of it, but since it's been there forever, I don't want to assume it's use.


Charles-Gagnon:
Do we really need this with @cssuh's fix? We haven't had many reports of this so it seems to not be a widespread issue..

At the very least you should indicate in the description specifically why we're doing this check.

abist:
@Charles-Gagnon We discussed this in the Shiproom chat

kburtram:
@abist since we're going with the other fix for this bug, I think we can close this.  We can reopen if needed.  Thanks for putting this together so quickly!

Issue #17065, opened 2021-10-04
Connection placeholder content written to user settings file


af4jm:
- MSSQL Extension Version: 1.10.1
- VSCode Version: 1.60.2
- OS Version: Win10

Steps to Reproduce: install extension, restart VSCode, see nonsense added to user settings

```json
    "mssql.connections": [
        {
            "server": "{{put-server-name-here}}",
            "database": "{{put-database-name-here}}",
            "user": "{{put-username-here}}",
            "password": "{{put-password-here}}"
        }
    ],
```

Expected behavior: setting should be there if and only if it contains actual connection data, the template makes sense in the default settings pane, not live user settings

justin-dx:
I've just had the same thing happen after adding a new server. I got my password wrong on the first attempt and after getting it correct I end up with two connections.
![vscode mssql](https://user-images.githubusercontent.com/39646902/140836845-86ae3523-b64f-4500-9d54-aee27c98e169.JPG)

kburtram:
@justin-dx we've observed this before and will fix in the next release.  In our repo the placeholder went away when vscode was restarted.  Please let us know if that's not what you also observed in your environment.

kburtram:
This should be fixed in the 1.11.1 release.  Thanks!

PR #17116, opened 2021-11-16
Remove default placeholder values

cssuh:
delete the placeholder values in `package.json`
fixes https://github.com/microsoft/vscode-mssql/issues/17065

### DIFF
--- a/package.json
+++ b/package.json
@@ -578,12 +578,6 @@
         "mssql.connections": {
           "type": "array",
           "default": [
-            {
-              "server": "{{put-server-name-here}}",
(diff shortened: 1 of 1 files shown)


abist:
you can perhaps remove the default field altogether

kburtram (approved):
LGTM.  Let's merge and queue a build so we can test this change.

kburtram:
@abist it may make sense to remove default all together as you suggest.  Maybe we can consider that for Dec or if this change has unexpected side-effects.  I don't see a need to add any content to settings.json if there are no saved connections.

PR #17118, opened 2021-11-18
Handle connect button press with firewall error

cssuh:
Upon the firewall rule button firing, and selecting the Connect button, some undefined errors were thrown because it was handling the Connect button as a tree node. I added some checks to verify the instance of a node.  
Screen Shot 2021-11-18 at 6 07 21 AM

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -31,7 +31,7 @@ function getParentNode(node: TreeNodeType): TreeNodeInfo {
     node = node.parentNode;
     if (!(node instanceof TreeNodeInfo)) {
         vscode.window.showErrorMessage(LocalizedConstants.nodeErrorMessage);
-        throw new Error(`Something went wrong when trying to find the correct node.`);
+        throw new Error(`Parent node was not TreeNodeInfo.`);
     }
     return node;
 }
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -591,7 +591,6 @@ export class ObjectExplorerService {
         return;
     }
 
-
     /** Getters */
     public get currentNode(): TreeNodeInfo {
         return this._currentNode;
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -195,13 +195,16 @@ export class ObjectExplorerService {
         }
     }
 
-    public updateNode(node): void {
-        for (let rootTreeNode of this._rootTreeNodeArray) {
+    public updateNode(node: TreeNodeInfo | ConnectTreeNode): void {
+        if (node instanceof ConnectTreeNode) {
+            node = node.parentNode;
             if (!(node instanceof TreeNodeInfo)) {
-                node = node.parentNode;
+                throw new Error(`Something went wrong when trying to find the correct node.`)
             }
+        }
+        for (let rootTreeNode of this._rootTreeNodeArray) {
             if (Utils.isSameConnection(node.connectionInfo, rootTreeNode.connectionInfo) &&
-                    rootTreeNode.label === node.label) {
+                rootTreeNode.label === node.label) {
                     const index = this._rootTreeNodeArray.indexOf(rootTreeNode);
                     delete this._rootTreeNodeArray[index];
                     this._rootTreeNodeArray[index] = node;
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -27,6 +27,15 @@ import { ConnectionProfile } from '../models/connectionProfile';
 import providerSettings from '../azure/providerSettings';
 import { IConnectionInfo } from 'vscode-mssql';
 
+function getParentNode(node: TreeNodeType): TreeNodeInfo {
+    node = node.parentNode;
+    if (!(node instanceof TreeNodeInfo)) {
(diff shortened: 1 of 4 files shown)


src/objectExplorer/objectExplorerService.ts
```diff
@@ -191,10 +195,13 @@ export class ObjectExplorerService {
         }
     }
 
-    public updateNode(node: TreeNodeInfo): void {
+    public updateNode(node): void {
```
Charles-Gagnon, line 198: Please always have types (I would suggest making a type union of the possible types a node could be)

src/objectExplorer/objectExplorerUtils.ts
```diff
@@ -26,8 +26,13 @@ export class ObjectExplorerUtils {
         }
     }
 
-    public static getNodeUri(node: TreeNodeInfo): string {
-        const profile = <IConnectionProfile>node.connectionInfo;
+    public static getNodeUri(node): string {
```
Charles-Gagnon, line 29: Same here, add type

src/objectExplorer/objectExplorerUtils.ts
```diff
@@ -26,8 +27,13 @@ export class ObjectExplorerUtils {
         }
     }
 
-    public static getNodeUri(node: TreeNodeInfo): string {
-        const profile = <IConnectionProfile>node.connectionInfo;
+    public static getNodeUri(node: TreeNodeInfo | ConnectTreeNode): string {
```
Charles-Gagnon, line 30: I would still suggest combining these into a single type union so that if we add other types in the future (or modify existing ones) we have less places to update and can find the places these are used easier.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -191,7 +195,13 @@ export class ObjectExplorerService {
         }
     }
 
-    public updateNode(node: TreeNodeInfo): void {
+    public updateNode(node: TreeNodeInfo | ConnectTreeNode): void {
+        if (node instanceof ConnectTreeNode) {
+            node = node.parentNode;
+            if (!(node instanceof TreeNodeInfo)) {
```
Charles-Gagnon, line 201: 1. Is there a reason you added this check here but not anywhere else? Do you actually expect this to be able to happen ever? I don't have a huge issue with being careful, but maybe this would be better put in a helper function (`getParentTreeNode` or something)
2. If this error ends up being displayed to the user it should be localized

src/objectExplorer/objectExplorerUtils.ts
cssuh: ok, I will look into doing that

src/objectExplorer/objectExplorerService.ts
cssuh: yeah, I was mostly being careful

abist: The error is thrown here, but what happens in the UI? Would the user know there was an error in the one of the nodes?

cssuh: Do you think I should have a popup to the user in addition to the error message? @abist

abist: Either that, or put that error in the node

Charles-Gagnon: Having a popup seems to be the cleanest option, I think trying to fit text in the node would be hard to read and not immediately indicate an issue occurred.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -575,6 +582,14 @@ export class ObjectExplorerService {
         return;
     }
 
+    private getParentNode(node: TreeNodeType): TreeNodeInfo {
+        node = node.parentNode;
+        if (!(node instanceof TreeNodeInfo)) {
+            throw new Error(`${LocalizedConstants.nodeErrorMessage}`);
```
Charles-Gagnon, line 588: You don't need to put it in a string, just use the error message directly

src/objectExplorer/objectExplorerService.ts
```diff
@@ -575,6 +582,14 @@ export class ObjectExplorerService {
         return;
     }
 
+    private getParentNode(node: TreeNodeType): TreeNodeInfo {
```
Charles-Gagnon, line 585: This doesn't need to be a member of the service, just have it be a module-level function outside the class.

Charles-Gagnon (approved):

src/objectExplorer/objectExplorerService.ts
```diff
@@ -27,6 +27,15 @@ import { ConnectionProfile } from '../models/connectionProfile';
 import providerSettings from '../azure/providerSettings';
 import { IConnectionInfo } from 'vscode-mssql';
 
+function getParentNode(node: TreeNodeType): TreeNodeInfo {
+    node = node.parentNode;
+    if (!(node instanceof TreeNodeInfo)) {
+        vscode.window.showErrorMessage(LocalizedConstants.nodeErrorMessage);
+        throw new Error(`Something went wrong when trying to find the correct node.`);
```
Charles-Gagnon, line 34: You've verified that this isn't displayed to the user?

I'd also make it a little clearer what the actual problem is - something like `Parent node was not TreeNodeInfo`. (and if there's any general name or something you can print as well for the current node that'd definitely be useful)

Charles-Gagnon (approved):

src/objectExplorer/objectExplorerService.ts
cssuh: yes I've verified this - it shows up in the debug console

PR #17134, opened 2021-11-29
Use native secret store for credentials

abist:
Adding functionality to let users use the native secret store just like in ADS.

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -121,5 +121,6 @@
 "mssql.query.ansiWarnings":"Enable SET ANSI_WARNINGS",
 "mssql.query.ansiNulls":"Enable SET ANSI_NULLS",
 "mssql.query.alwaysEncryptedParameterization":"Enable Parameterization for Always Encrypted",
-"mssql.ignorePlatformWarning":"[Optional] Do not show unsupported platform warnings"
+"mssql.ignorePlatformWarning":"[Optional] Do not show unsupported platform warnings",
+"mssql.useNativeCredentials":"Migrate all passwords to native password storage"
 }
--- a/src/credentialstore/credentialstore.ts
+++ b/src/credentialstore/credentialstore.ts
@@ -9,6 +9,7 @@ import SqlToolsServerClient from '../languageservice/serviceclient';
 import * as Utils from '../models/utils';
 import * as Constants from '../constants/constants';
 import { ConnectionStore } from '../models/connectionStore';
+import { IConnectionProfile } from '../models/interfaces';
 
 /**
  * Implements a credential storage for Windows, Mac (darwin), or Linux.
@@ -28,13 +29,20 @@ export class CredentialStore implements ICredentialStore {
             this._client = SqlToolsServerClient.instance;
         }
         this._secretStorage = this._context.secrets;
+        this._context.globalState.update(Utils.configPasswordsMigrated, false);
         this._passwordsMigrated = this._context.globalState.get(Utils.configPasswordsMigrated);
         this._useNativeCredentials = Utils.useNativeCredentials();
+    }
+
+    /**
+     * Initializes the credential store by migrating any old passwords
+     * to the native secret store if the native credential setting is set
+     */
+    public async initialize(): Promise<void> {
         if (this._useNativeCredentials && !this._passwordsMigrated) {
-            this.migratePasswords().then(() => {
-                this._passwordsMigrated = true;
-                this._context.globalState.update(Utils.configPasswordsMigrated, this._passwordsMigrated);
-            });
+            await this.migratePasswords();
+            this._passwordsMigrated = true;
+            this._context.globalState.update(Utils.configPasswordsMigrated, this._passwordsMigrated);
         }
     }
 
@@ -115,14 +123,14 @@ export class CredentialStore implements ICredentialStore {
         const savedPasswordConnections = connections.filter(conn => conn.savePassword === true);
         for (let i = 0; i < savedPasswordConnections.length; i++) {
             let conn = savedPasswordConnections[i];
-            await this.cleanCredential(conn);
+            await this.migrateCredential(conn);
         }
         return Utils.removeCredentialFile();
     }
 
 
-    private async cleanCredential(conn): Promise<boolean> {
-        const credentialId = ConnectionStore.formatCredentialId(conn);
+    private async migrateCredential(conn: IConnectionProfile): Promise<boolean> {
+        const credentialId = ConnectionStore.formatCredentialIdForCred(conn);
         const credential = await this._client.sendRequest(Contracts.ReadCredentialRequest.type, { credentialId, password: undefined });
         if (credential.password) {
             const password = credential.password;
--- a/src/credentialstore/credentialstore.ts
+++ b/src/credentialstore/credentialstore.ts
@@ -2,10 +2,13 @@
 *  Copyright (c) Microsoft Corporation. All rights reserved.
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
-
+import * as vscode from 'vscode';
 import * as Contracts from '../models/contracts';
 import { ICredentialStore } from './icredentialstore';
 import SqlToolsServerClient from '../languageservice/serviceclient';
+import * as Utils from '../models/utils';
+import * as Constants from '../constants/constants';
+import { ConnectionStore } from '../models/connectionStore';
 
 /**
  * Implements a credential storage for Windows, Mac (darwin), or Linux.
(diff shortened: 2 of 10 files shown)


package.nls.json
```diff
@@ -121,5 +121,6 @@
 "mssql.query.ansiWarnings":"Enable SET ANSI_WARNINGS",
 "mssql.query.ansiNulls":"Enable SET ANSI_NULLS",
 "mssql.query.alwaysEncryptedParameterization":"Enable Parameterization for Always Encrypted",
-"mssql.ignorePlatformWarning":"[Optional] Do not show unsupported platform warnings"
+"mssql.ignorePlatformWarning":"[Optional] Do not show unsupported platform warnings",
+"mssql.useNativeCredentials":"Migrate all passwords to native password storage"
```
Charles-Gagnon, line 125: I'd just have this be something like `Use the built-in native credential storage for storing connection passwords`. The "migration" part is just a one-time thing and isn't really what this setting is primarily about.

src/credentialstore/credentialstore.ts
```diff
@@ -14,10 +17,25 @@ import SqlToolsServerClient from '../languageservice/serviceclient';
...
+        private _context: vscode.ExtensionContext,
+        private _client?: SqlToolsServerClient) {
         if (!this._client) {
             this._client = SqlToolsServerClient.instance;
         }
+        this._secretStorage = this._context.secrets;
+        this._passwordsMigrated = this._context.globalState.get(Utils.configPasswordsMigrated);
+        this._useNativeCredentials = Utils.useNativeCredentials();
+        if (this._useNativeCredentials && !this._passwordsMigrated) {
+            this.migratePasswords().then(() => {
```
Charles-Gagnon, line 34: If this takes a while then that means that users may see connections fail because it tries to them up in the native store but they haven't been inserted yet. I'd suggest having a deferred that the read credentials can await on to ensure that the passwords are all migrated.

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +44,141 @@ export class CredentialStore implements ICredentialStore {
...
+            let returnedCred: Contracts.Credential;
+
+            // Use native credential if the setting is on
+            if (this._useNativeCredentials) {
+                returnedCred = await this.nativeReadCredential(credentialId, cred);
+            } else {
+                returnedCred = await this._client.sendRequest(Contracts.ReadCredentialRequest.type, cred);
+            }
+            return returnedCred;
+        } catch (err) {
```
Charles-Gagnon, line 59: Just remove the try/catch if you aren't doing anything with the error (here and elsewhere below)

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +44,141 @@ export class CredentialStore implements ICredentialStore {
...
 
-    public saveCredential(credentialId: string, password: any): Promise<boolean> {
-        let self = this;
+    /**
+     * Saves a credential in the credential store
+     *
+     * @param credentialId the ID uniquely identifying this credential
+     * @param password the password that needs to be saved with the credential
+     */
+    public async saveCredential(credentialId: string, password: any): Promise<boolean> {
```
Charles-Gagnon, line 71: Does password actually have to be an any? It's not just a string?

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +44,141 @@ export class CredentialStore implements ICredentialStore {
...
-            }, err => reject(err));
-        });
+    public async readCredential(credentialId: string): Promise<Contracts.Credential> {
+        const cred: Contracts.Credential = new Contracts.Credential();
+        try {
+            let returnedCred: Contracts.Credential;
+
+            // Use native credential if the setting is on
+            if (this._useNativeCredentials) {
+                returnedCred = await this.nativeReadCredential(credentialId, cred);
```
Charles-Gagnon, line 54: I'd just return the calls directly instead of awaiting - that'll be doing the same thing and save having to unwrap and then re-wrap the promise. (here and in similar functions below)

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +44,141 @@ export class CredentialStore implements ICredentialStore {
...
+            return cred;
+        } catch (err) {
+            throw(err);
+        }
+    }
+
+    private async nativeSaveCredential(credentialId: string, password: any): Promise<boolean> {
+        try {
+            await this._context.secrets.store(credentialId, password);
+            const savedPassword = await this._context.secrets.get(credentialId);
```
Charles-Gagnon, line 125: Why do you do this check? Seems like trusting secrets.store to do the right thing should be fine.

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +44,141 @@ export class CredentialStore implements ICredentialStore {
...
+            return false;
+        } catch (err) {
+            throw(err);
+        }
+    }
+
+    private async nativeDeleteCredential(credentialId: string): Promise<boolean> {
+        try {
+            await this._context.secrets.delete(credentialId);
+            const savedPassword = await this._context.secrets.get(credentialId);
```
Charles-Gagnon, line 138: Why do you need to check this? We should generally trust their APIs to do what they say at runtime.

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +44,141 @@ export class CredentialStore implements ICredentialStore {
...
+        const savedPasswordConnections = connections.filter(conn => conn.savePassword === true);
+        for (let i = 0; i < savedPasswordConnections.length; i++) {
+            let conn = savedPasswordConnections[i];
+            await this.cleanCredential(conn);
+        }
+        await Utils.removeCredentialFile();
+    }
+
+
+    private async cleanCredential(conn): Promise<boolean> {
```
Charles-Gagnon, line 163: This seems like it should be named `migrateCredential` or something instead - cleaning it up is only part of what it's doing.

Also please add a type for conn

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +44,141 @@ export class CredentialStore implements ICredentialStore {
...
+        const credentialId = ConnectionStore.formatCredentialId(conn);
+        const credential = await this._client.sendRequest(Contracts.ReadCredentialRequest.type, { credentialId, password: undefined });
+        if (credential.password) {
+            const password = credential.password;
+            // save it in secret store
+            await this._secretStorage.store(credentialId, password);
+            // check if it's saved
```
Charles-Gagnon, line 170: Same as above, why do we need to check this? Once it's done being saved we should be fine to just delete it from STS.

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +44,141 @@ export class CredentialStore implements ICredentialStore {
...
+            const savedPassword = await this._secretStorage.get(credentialId);
+            if (savedPassword === password) {
+                // delete from tools service
+                const result = await this._client.sendRequest(Contracts.DeleteCredentialRequest.type,
+                    { credentialId, password: undefined });
+                return result;
+            } else {
+                return false;
```
Charles-Gagnon, line 178: Anywhere you're returning false should be an error thrown - if we can't successfully migrate all the passwords it will currently just silently skip them which doesn't seem ideal. Throwing an error and letting the user decide if they care seems like a better way to handle that.

src/models/utils.ts
```diff
@@ -494,3 +495,23 @@ export function generateQueryUri(scheme = 'vscode-mssql-adhoc'): vscode.Uri {
         authority: `Query${uriIndex++}`
     });
 }
+
+/**
+ * Returns whether the credentials should use the native credential
+ * service or not
+ */
+export function useNativeCredentials(): boolean {
+    const isLinux: boolean = os.platform() === 'linux';
```
Charles-Gagnon, line 504: I'd add an explanation of why you're only doing this on Linux.

Charles-Gagnon (changes requested):

src/credentialstore/credentialstore.ts
abist: The user wouldn't know the passwords are migrated, since it's always done. What suggestion do you have to ask users in case there was an error here?

src/credentialstore/credentialstore.ts
```diff
@@ -14,10 +18,32 @@ import SqlToolsServerClient from '../languageservice/serviceclient';
...
+
+    constructor(
+        private _context: vscode.ExtensionContext,
+        private _client?: SqlToolsServerClient) {
         if (!this._client) {
             this._client = SqlToolsServerClient.instance;
         }
+        this._secretStorage = this._context.secrets;
+        this._context.globalState.update(Utils.configPasswordsMigrated, false);
+        this._passwordsMigrated = this._context.globalState.get(Utils.configPasswordsMigrated);
```
Charles-Gagnon, line 33: You don't need to store this anymore - just read/set the stored var in initialize directly

kburtram:
@abist what's the plan with this PR?  do we still need it or should we close this now?

abist:
@kburtram This is supposed to go in the next release, but I'll create a new PR since there are conflicts now.

PR #17137, opened 2021-11-30
Refresh expired azure account tokens before running query

cssuh:
Porting [this](https://github.com/microsoft/azuredatastudio/pull/16936) change from ADS
Before, running queries with an expired azure account token (expires after 1 hour) would fail an error message.

Now, upon running a query it will check the expiredOn field and if a token is expired, it will automatically refresh the token and run the query.


src/controllers/connectionManager.ts
```diff
@@ -814,6 +818,52 @@ export default class ConnectionManager {
...
     }
 
+    public async refreshAzureAccountToken(uri: string): Promise<boolean> {
+        const profile = this.getConnectionInfo(uri);
+		if (!profile) {
+            this.vscodeWrapper.logToOutputChannel(`Connection not found for uri ${uri}`);
+			return false;
+		}
+
+        // Wait for the pending reconnction promise if any
```
Charles-Gagnon, line 828: Spacing seems off in this file, make sure you format so everything is consistent

src/controllers/connectionManager.ts
```diff
@@ -814,6 +819,52 @@ export default class ConnectionManager {
...
+        if (typeof expiry === 'number' && !Number.isNaN(expiry)) {
+            const currentTime = new Date().getTime() / 1000;
+            const maxTolerance = 2 * 60; // two minutes
+            if (expiry - currentTime < maxTolerance) {
+                this.vscodeWrapper.logToOutputChannel(`Access token expired for connection ${profile.connectionId} with uri ${uri}`);
+                try {
+                    let connectionResult = await this.connect(uri, profile.credentials);
+                    if (!connectionResult) {
+                        this.vscodeWrapper.showErrorMessage(`Failed to refresh connection ${profile.connectionId} with uri ${uri}, invalid connection result.`);
+                        throw new Error('Connection result is invalid');
```
Charles-Gagnon, line 855: Maybe just `Unable to refresh connection` - "invalid" doesn't really help here

src/controllers/connectionManager.ts
```diff
@@ -814,6 +819,52 @@ export default class ConnectionManager {
...
+                    }
+                    this.vscodeWrapper.logToOutputChannel(`Successfully refreshed token for connection ${profile.connectionId} with uri ${uri}, result: ${connectionResult}, isConnected: ${this.isConnected(uri)}, ${this.getConnectionInfo(uri)}`);
+                    return true;
+                } catch {
+                    this.vscodeWrapper.showInformationMessage('Error when refreshing token to execute query');
```
Charles-Gagnon, line 860: This is more generic than just queries

src/controllers/mainController.ts
```diff
@@ -734,6 +734,8 @@ export default class MainController implements vscode.Disposable {
             if (!self.connectionManager.isConnected(uri)) {
                 await self.onNewConnection();
             }
+            // check if current connection is still valid / active - if not, refresh azure account token
+            this._connectionMgr.refreshAzureAccountToken(uri);
```
Charles-Gagnon, line 738: await?

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -329,6 +329,36 @@
         <trans-unit id="msgConnecting">
             <source xml:lang="en">Connecting to server "{0}" on document "{1}".</source>
         </trans-unit>
+        <trans-unit id="msgConnectionNotFound">
```
Charles-Gagnon, line 332: I'm not sure about adding essentially debug messages into the output pane...but it seems fine to throw in there to make it easier to track down issues and if it gets to be too spammy later on we can change it I guess.

typings/vscode-mssql.d.ts
```diff
@@ -5,6 +5,7 @@
 
 declare module 'vscode-mssql' {
 
+    import { Token } from '@microsoft/ads-adal-library';
```
Charles-Gagnon, line 8: It doesn't look like you need this anymore

src/controllers/connectionManager.ts
```diff
@@ -814,6 +819,53 @@ export default class ConnectionManager {
         }
     }
 
+    public async refreshAzureAccountToken(uri: string): Promise<boolean> {
```
abist, line 822: doesn't seem like the boolean is used anywhere, so the return type could just be `Promise<void>`

src/models/connectionProfile.ts
```diff
@@ -27,6 +27,7 @@ export class ConnectionProfile extends ConnectionCredentials implements IConnect
     public emptyPasswordInput: boolean;
     public azureAuthType: AzureAuthType;
     public azureAccountToken: string;
+    public expiresOn: number;
```
abist, line 30: shouldn't this also be `number | undefined` just like how it is in `ConnectionCredentials` ?

Charles-Gagnon (approved):

abist (approved):

PR #17141, opened 2021-12-02
Added when clause for run and cancel

### LINKED ISSUE #1584: Cancel query button is always enabled
This should be disabled when a query isn't currently running

image

Clicking on it results in this error 

`mssql: Cannot cancel query as no query is running.`

smartguest:
Fixes #1584 by adding a when clause to indicate if a query is running or not.

When there is no query running, the run query button is visible, when there is a query running, the stop button is visible. This allows for a single start and stop button

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -35,7 +35,6 @@ import { IConnectionInfo } from 'vscode-mssql';
 import { SchemaCompareService } from '../services/schemaCompareService';
 import { SqlTasksService } from '../services/sqlTasksService';
 import { AzureFunctionsService } from '../services/azureFunctionsService';
-import { formArrayNameProvider } from '../../out/src/views/htmlcontent/src/js/lib/@angular/forms/src/directives/reactive_directives/form_group_name';
 
 /**
  * The main controller class that initializes the extension
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -751,9 +751,9 @@ export default class MainController implements vscode.Disposable {
             }
             vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', true);
             await self._outputContentProvider.runQuery(self._statusview, uri, querySelection, title);
+            vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', false);
         } catch (err) {
             console.warn(`Unexpected error running query : ${err}`);
-        } finally {
             vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', false);
         }
     }
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -535,10 +535,10 @@ export default class MainController implements vscode.Disposable {
         try {
             let uri = this._vscodeWrapper.activeTextEditorUri;
             this._outputContentProvider.cancelQuery(uri);
-            vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', false);
+            //vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', false);
         } catch (err) {
             console.warn(`Unexpected error cancelling query : ${err}`);
-            vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', true);
+            //vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', true);
         }
     }
 
@@ -749,12 +749,13 @@ export default class MainController implements vscode.Disposable {
             if (editor.document.getText(selectionToTrim).trim().length === 0) {
                 return;
             }
-            vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', true);
+            let filename = editor.document.fileName;
+            console.log('filename is ' + filename + ' uri is ' + uri);
+            //vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', true);
             await self._outputContentProvider.runQuery(self._statusview, uri, querySelection, title);
-            vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', false);
         } catch (err) {
             console.warn(`Unexpected error running query : ${err}`);
-            vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', false);
+            //vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', false);
         }
     }
 
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -535,6 +535,7 @@ export default class MainController implements vscode.Disposable {
         try {
             let uri = this._vscodeWrapper.activeTextEditorUri;
             this._outputContentProvider.cancelQuery(uri);
+            vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', false);
         } catch (err) {
             console.warn(`Unexpected error cancelling query : ${err}`);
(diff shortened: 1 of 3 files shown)


Charles-Gagnon:
Please include a gif of this in action

src/controllers/mainController.ts
```diff
@@ -747,9 +749,11 @@ export default class MainController implements vscode.Disposable {
             if (editor.document.getText(selectionToTrim).trim().length === 0) {
                 return;
             }
+            vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', true);
```
abist, line 752: where is it set to false again once the query is completed?

abist (changes requested):
I think there's more logic required to set the flag to false again once a query is completed

smartguest:
There is an issue where the button won't start running the query immediately after selecting a connection, however this is not a regression and happens when you debug with the regular code as well so this likely will not happen in the actual release.

https://user-images.githubusercontent.com/18018452/144521336-0a3a4771-25a6-4c32-8507-60563aae6f7a.mp4

src/controllers/mainController.ts
smartguest: You're right about that, it needs to be added after it.

abist:
> There is a debug only issue where the button won't start running the query immediately after selecting a connection, however this is not a regression and happens when you debug with the regular code as well so this likely will not happen in the actual release. @Charles-Gagnon
> 
>  StopStartButton.mp4

What happens if you let the query run and dont touch the button?

src/controllers/mainController.ts
smartguest: The button disappears after it's run, but it reappears once you click around the area.

smartguest:
@abist Nothing happens really aside from the connection being connected.

src/controllers/mainController.ts
```diff
@@ -747,9 +749,13 @@ export default class MainController implements vscode.Disposable {
             if (editor.document.getText(selectionToTrim).trim().length === 0) {
                 return;
             }
-            await self._outputContentProvider.runQuery(self._statusview, uri, querySelection, title);
+            vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', true);
+            await self._outputContentProvider.runQuery(self._statusview, uri, querySelection, title).then(() => {
+                vscode.commands.executeCommand('setContext', 'mssql.isQueryRunning', false);
```
Charles-Gagnon, line 754: You're already awaiting this, you don't need to put it in a .then

But regardless, it would be easier just to remove this and the one on 758 and put it inside a finally block instead.

Charles-Gagnon (changes requested):
Is this going to work correctly when running multiple queries? i.e try this : 

1. Start long running (infinite) query
2. Open a new query window - does the run button show correctly?

smartguest:
@Charles-Gagnon unfortunately no, it won't work properly with multiple queries, still need to investigate how to get individual sessions

src/controllers/mainController.ts
smartguest: Apparently, the code seems to ignore the await and the finally block executes before the query is complete (in the case of an infinite query that is). This means it is not working as expected.

Charles-Gagnon:
@smartguest I'm going to close this while you work on a fix to avoid sending a bunch of notifications, please reopen once you think you have it figured out!

smartguest:
Due to the design of when clauses in vscode, currently there is no way to check whether an object is NOT in a list (no parentheses support for negating sub-clauses), therefore the run button will always be enabled. However, the stop button will properly appear when a query is running for a file and only for that file (tested with different paths with the same filename), it will also disappear appropriately if an error occurs or when the query is finished/stopped.

Charles-Gagnon:
Please add screenshots/gif of the new functionality

smartguest:
startstoprun12-8-21

GIF demonstrating handling multiple windows and same named files and unsaved files (in different directories).

src/controllers/queryRunner.ts
```diff
@@ -195,19 +196,31 @@ export default class QueryRunner {
...
+            vscode.commands.executeCommand('setContext', 'mssql.activeQueries', QueryRunner._activeQueries);
+            console.log(this.getFilePath() + ' has been added to activeQueries');
             this.eventEmitter.emit('start', this.uri);
             this._notificationHandler.registerRunner(this, this._ownerUri);
         };
         let onError = (error) => {
             this._statusView.executedQuery(this.uri);
             this._isExecuting = false;
+            QueryRunner._activeQueries = QueryRunner._activeQueries.filter(fileName => fileName !== this.getFilePath());
+            vscode.commands.executeCommand('setContext', 'mssql.activeQueries', QueryRunner._activeQueries);
```
Charles-Gagnon, line 209: I'd suggest `mssql.runningQueries` to be even more explicit

src/controllers/queryRunner.ts
```diff
@@ -195,19 +196,31 @@ export default class QueryRunner {
...
+            console.log(this.getFilePath() + ' has been removed from activeQueries');
             // TODO: localize
             this._vscodeWrapper.showErrorMessage('Execution failed: ' + error.message);
         };
 
         await queryCallback(onSuccess, onError);
     }
 
+    // helper function for parsing the ownerUri
```
Charles-Gagnon, line 218: Doc comments should be in JS doc form (/** ... */)

But also I don't think a one-liner needs a whole function for it. You can just inline `vscode.Uri.parse(this._ownerUri).fsPath` where you're calling this instead.

src/controllers/queryRunner.ts
```diff
@@ -232,6 +245,9 @@ export default class QueryRunner {
         }
         this._statusView.executedQuery(result.ownerUri);
         let hasError = this._batchSets.some(batch => batch.hasError === true);
+        QueryRunner._activeQueries = QueryRunner._activeQueries.filter(fileName => fileName !== this.getFilePath());
```
Charles-Gagnon, line 248: Filtering out the completed/errored path though is duplicated so I'd suggest moving that into its own helper function.

Charles-Gagnon (commented):
What happens if : 

1. Someone starts a long running (infinite) query in an untitled editor (Untitled-1)
2. Closes that editor
3. Opens a new one (now also called Untitled-1 but without any content)

Do we properly clear out the editor from the list when it's closed?

abist:
> Due to the design of when clauses in vscode, currently there is no way to check whether an object is NOT in a list (no parentheses support for negating sub-clauses), therefore the run button will always be enabled. However, the stop button will properly appear when a query is running for a file and only for that file (tested with different paths with the same filename), it will also disappear appropriately if an error occurs or when the query is finished/stopped.

You could have another array called inactiveQueries that gets filled up after the query is completed and when you remove it from active queries, and then use that for the when clause

smartguest:
> What happens if :
> 
> 1. Someone starts a long running (infinite) query in an untitled editor (Untitled-1)
> 2. Closes that editor
> 3. Opens a new one (now also called Untitled-1 but without any content)
> 
> Do we properly clear out the editor from the list when it's closed?

That has been tested and the scenario works, closing the file triggers an error which clears out its entry

smartguest:
> 

While this approach does work initially, it breaks upon saving the file to a different name (one that has never been active), and this persists with every new file until you restart, so this is not ideal.

Charles-Gagnon (approved):

PR #17142, opened 2021-12-03
fix the connections setting issue

### LINKED ISSUE #16983: Adding connection to workspace settings, copies to settings.json
- MSSQL Extension Version: v1.10.1
- VSCode Version:
Version: 1.55.2 (user setup)
Commit: 3c4e3df9e89829dce27b7b5c24508306b151f30d
Date: 2021-04-13T09:35:57.887Z
Electron: 11.3.0
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Windows_NT x64 10.0.19042

- OS Version: Windows 10 - 20H2

Steps to Reproduce:

1. Add the following to the Workspace setting file:

```
		"mssql.connections": [
			{
				"connectionString": "Server=(localdb)\\LocalDevDB;Database=MainDB;Trusted_Connection=True;ConnectRetryCount=0",
				"profileName": "LocalLogs",
				"password": ""
			},
```

2. When you save, then load the extension in the left panel and look at Settings.Json. The settings for the SQL has been copied there. 

This is frustrating is it's not needed in my other workspaces, yet the connection appears as it's in the 
`C:\Users\<user>\AppData\Roaming\Code\User\settings.json`

alanrenmsft:
- this pr fixes #16983.
    when we sanitize the connections (remove aad token and password), we are directly saving the settings to global settings. the fix is to load connections from different settings then sanitize and save them to the original settings.
- when adding connections to workspace settings, the changes won't be reflected immediately, a reload is needed. 
    In the config change monitoring handler, we are reading the connections using the get configuration API which will not merge the connections in different settings. the fix is to use the same API the OE is using to get all the connections.
- when connection is added to settings directly, removal of AAD token is not performed, it is a startup time only operation.

there are a lot of indention changes, please hide whitespaces to make it easier to review.
image

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -67,8 +67,8 @@ export default class MainController implements vscode.Disposable {
      * @constructor
      */
     constructor(context: vscode.ExtensionContext,
-                connectionManager?: ConnectionManager,
-                vscodeWrapper?: VscodeWrapper) {
+        connectionManager?: ConnectionManager,
+        vscodeWrapper?: VscodeWrapper) {
         this._context = context;
         if (connectionManager) {
             this._connectionMgr = connectionManager;
@@ -114,7 +114,7 @@ export default class MainController implements vscode.Disposable {
     /**
      * Initializes the extension
      */
-    public async activate():  Promise<boolean> {
+    public async activate(): Promise<boolean> {
         // initialize the language client then register the commands
         const didInitialize = await this.initialize();
         if (didInitialize) {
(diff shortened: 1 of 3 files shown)


abist (approved):
LGTM

src/controllers/mainController.ts
```diff
@@ -235,24 +233,46 @@ export default class MainController implements vscode.Disposable {
         if (activeTextEditor && this._vscodeWrapper.isEditingSqlFile) {
             this.onDidOpenTextDocument(activeTextEditor.document);
         }
-        await this.sanitizeProfiles();
+        this.sanitizeConnectionProfiles();
```
Charles-Gagnon, line 236: Why are we not awaiting on this anymore? Won't we potentially run into race condition issues if the sanitization isn't finished but the extension is finished activating and the user tries to modify connections?

alanrenmsft: good catch. https://github.com/microsoft/vscode-mssql/pull/17147

Issue #17128, opened 2021-11-23
AAD auth method message is not localized


abist:
https://github.com/microsoft/vscode-mssql/blob/main/package.json#L559

PR #17146, opened 2021-12-03
Added localization

cssuh:
Fixes: https://github.com/microsoft/vscode-mssql/issues/17128

### DIFF
--- a/package.json
+++ b/package.json
@@ -566,6 +566,10 @@
             "AuthCodeGrant",
             "DeviceCode"
           ],
+          "enumDescriptions": [
+            "%mssql.authCodeGrant.description%",
+            "%mssql.deviceCode.description%"
+          ],
           "scope": "application"
         },
         "mssql.logDebugInfo": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -556,12 +556,12 @@
...
       "type": "object",
-      "title": "MSSQL configuration",
+      "title": "%mssql.Configuration%",
       "properties": {
         "mssql.azureActiveDirectory": {
           "type": "string",
           "default": "AuthCodeGrant",
-          "description": "Chooses which Authentication method to use",
+          "description": "%mssql.chooseAuthMethod%",
           "enum": [
```
Charles-Gagnon, line 565: While you're here could you add enum descriptions for these?

abist (approved):

package.json
cssuh: @Charles-Gagnon what do you mean by enum descriptions?

Charles-Gagnon: https://github.com/microsoft/azuredatastudio/blob/a2ded8bd3c5c53a371e344b9b7c729b8adf540ca/extensions/azcli/package.json#L49

Shows up as extra text describing the option 

image

PR #17152, opened 2021-12-05
credential for recent connections

### LINKED ISSUE #1698: KeyChain password not completely removed when Profile is removed
- MSSQL Extension Version: 1.9.0
- VSCode Version: 1.46.1
- OS Version: 10.15.5

Steps to Reproduce:

1. Open Keychain Access and sort by Date Modified to see the entry being created and deleted
2. In SQL extension, click + to add connection
3. Enter information, including password, finally save password
4. Notice in Keychain, 2 "application password" Kind was created with name "M"
5. Right click on connection and select "Remove"
6. Notice in Keychain, 1 entry was removed but another remains.

alanrenmsft:
this pr fixes #1698 and also fixes the issue when we do cleanup of recent connections we are not removing the credentials.


src/models/connectionStore.ts
```diff
@@ -294,8 +294,14 @@ export class ConnectionStore {
...
+            // Get all recent connection profiles and delete the associated credentials.
+            const mruList = this.getRecentlyUsedConnections();
+            for (const connection of mruList) {
+                const credentialId = ConnectionStore.formatCredentialId(connection.server, connection.database, connection.user, ConnectionStore.CRED_MRU_USER);
+                await this._credentialStore.deleteCredential(credentialId);
+            }
+            // Update the MRU list to be empty
             await this._context.globalState.update(Constants.configRecentConnections, []);
         } catch (error) {
             Promise.reject(error);
```
Charles-Gagnon, line 307: We should just remove this try/catch, not only is it wrong (it's not even returning the rejected promise....) but it's unnecessary since this is an async function anyways.

src/models/connectionStore.ts
```diff
@@ -294,8 +294,14 @@ export class ConnectionStore {
      * Clear all recently used connections from the MRU list.
      */
     public async clearRecentlyUsed(): Promise<void> {
-        // Update the MRU list to be empty
         try {
+            // Get all recent connection profiles and delete the associated credentials.
+            const mruList = this.getRecentlyUsedConnections();
+            for (const connection of mruList) {
+                const credentialId = ConnectionStore.formatCredentialId(connection.server, connection.database, connection.user, ConnectionStore.CRED_MRU_USER);
+                await this._credentialStore.deleteCredential(credentialId);
```
Charles-Gagnon, line 302: We should have each connection be wrapped in a try/catch so that if one fails to be removed we still delete the others. And for ones that error out we have a choice : 

1. Leave it in the MRU (since we couldn't delete the credential, this gives the person another chance to remove it from ADS in case it was some weird transient error)
2. Still clear the MRU at the end but maybe warn them we couldn't delete it so the user can go clean up the credential themself? Not sure how we expect them to find it easily though.

I'm slightly in favor of 2 since that seems like a better user experience if the user actually wants to clear out their MRU - and this is pretty unlikely to happen anyways.

src/models/connectionStore.ts
```diff
@@ -294,11 +294,23 @@ export class ConnectionStore {
...
+        const mruList = this.getRecentlyUsedConnections();
+        let errorOccured = false;
+        for (const connection of mruList) {
+            const credentialId = ConnectionStore.formatCredentialId(connection.server, connection.database, connection.user, ConnectionStore.CRED_MRU_USER);
+            try {
+                await this._credentialStore.deleteCredential(credentialId);
+            }
+            catch (err) {
+                errorOccured = true;
+                console.error(LocalizedConstants.deleteCredentialError, credentialId, err);
```
Charles-Gagnon, line 307: It would be better to log this to the output channel instead so users don't have to hunt around for how to open the developer tools (and where in the tools this message would be...) and then include an action in the error message to open the output channel.

alanrenmsft, line 307: pushed the changes too soon, i am still testing the changes

src/controllers/connectionManager.ts
```diff
@@ -314,6 +312,13 @@ export default class ConnectionManager {
             let mruConnection: IConnectionInfo = <any>{};
 
             if (Utils.isNotEmpty(result.connectionId)) {
+                // Use the original connection information to save the MRU connection.
+                // for connections that a database is not provided, the database information will be updated
+                // to the default database name, if we use the new information as the MRU connection,
+                // the connection information will be different from the saved connections (saved connection's database property is empty).
+                // When deleting the saved connection, we won't be able to find its corresponding recent connection,
+                // and the saved connection credentials will become orphaned.
+                mruConnection = Object.assign({}, connection.credentials);
```
Charles-Gagnon, line 321: I'd consider doing a deep clone here just to be safe...right now the credentials only holds base types I think but if we add more complex properties later on it would be good not to have to remember to update this as well.

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -554,6 +554,12 @@
         <trans-unit id="nodeErrorMessage">
             <source xml:lang="en">Parent node was not TreeNodeInfo.</source>
         </trans-unit>
+        <trans-unit id="deleteCredentialError">
+            <source xml:lang="en">Failed to delete credential with id: {0}, Error: {1}</source>
```
abist, line 558: does localization work with typescript variables like that?

alanrenmsft: yes, need to use Utils.formatString()

abist (approved):

Charles-Gagnon (approved):

src/models/connectionStore.ts
alanrenmsft: done. the method i am using will open the output channel directly.

PR #17165, opened 2021-12-09
Fixes connection issue with existing Azure SQL connections

lewis-sanchez:
This PR fixes #16964 

### Description
The change in this PR resolves errors that are encountered while reconnecting to a previously connected database connection. The reason for the connection errors was a result of passing a credential ID, stored in the connection string property, instead of the actual connection string to the SQL Tools Service. When the SQL Tools Service attempted to instantiate a `SqlConnectionStringBuilder` using the credential ID, a formatting exception would be thrown that stopped the connection attempt from completing.

### Test Steps
* Open VS Code
* Install the "SQL Server (MSSQL)" extension
* Open the command palette and enter `>MS SQL: Add Connection`
* Add a connection to an Azure SQL database by providing a connection string:

      Server={uri,port};Initial Catalog={connName};Persist Security Info=False;User ID={userName};Password={password};Connection Timeout=60;

* Close VS Code
* Open VS Code
* Open SQL Server tab and connect to the previously added connection
* Expand `Tables` folder > right click on any of the listed tables > click "Select Top 1000"
* Query results appear and no errors

PR #17167, opened 2021-12-09
fix connection not working issue

alanrenmsft:
I am not able to open the saved connections with a nullref error in the console, debugged and found it is caused by this PR: https://github.com/microsoft/vscode-mssql/pull/17165


abist (approved):

abist:
we're probably taking this for ask mode right?

Charles-Gagnon:
The breaking PR was checked in post-ask mode so we should be fine

Charles-Gagnon (approved):
One of these days it would be amazing to enable strict mode so we can avoid issues like this ☹️

PR #17186, opened 2022-01-06
Expose getConnectionString API 

VasuBhog:
To be used by the sql-database-projects extension to be used for the Add SQL Binding scenarios. 

In order to get the connectionString for SQL Server, we must set up a way to utilize the STS request for [GetConnectionStringRequest](https://github.com/Microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs#L953). I utilized the connect() API in sql-db extension to first get the connectionUri and then pass that into the request param for it to then return the connection string.

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -260,9 +260,9 @@ export default class ConnectionManager {
     }
 
     /**
-     * Get the connection string for the provided connection ID
-     * @param connectionId The connection ID for the connection.
-     * @param includePassword if password should be included in connection string.
+     * Get the connection string for the provided connection Uri
+     * @param connectionUri The connection Uri for the connection.
+     * @param includePassword (optional) if password should be included in connection string.
      * @returns connection string for the connection
      */
     public async getConnectionString(connectionUri: string, includePassword: boolean = false): Promise<string> {
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -265,7 +265,7 @@ export default class ConnectionManager {
      * @param includePassword if password should be included in connection string.
      * @returns connection string for the connection
      */
-    public async getConnectionString(connectionId: string, includePassword: boolean = false): Promise<string> {
+    public async getConnectionString(connectionInfo: IConnectionInfo, includePassword: boolean = false): Promise<string> {
         const listParams = new ConnectionContracts.GetConnectionStringParams();
         const result = await this.client.sendRequest(ConnectionContracts.GetConnectionStringRequest.type, listParams);
         return result.connectionString;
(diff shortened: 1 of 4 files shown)


src/models/contracts/connection.ts
```diff
@@ -254,3 +254,26 @@ export class ListDatabasesResult {
...
+export namespace GetConnectionStringRequest {
+    export const type = new RequestType<GetConnectionStringParams, GetConnectionStringResult, void, void>('connection/getconnectionstring');
+}
+
+// Get Connection String request format
+export class GetConnectionStringParams {
+    // Connection key to lookup connection string for the provided connection Uri
+    public ownerUri: string;
+
+    // Indicates whether to include the password in the connection string
```
Charles-Gagnon, line 270: If you're adding comments like this they should be in js doc format (/** ... */)

src/controllers/connectionManager.ts
```diff
@@ -259,6 +259,19 @@ export default class ConnectionManager {
         return (fileUri in this._connections && this._connections[fileUri].connecting);
     }
 
+    /**
+     * Get the connection string for the provided connection ID
+     * @param connectionId The connection ID for the connection.
+     * @param includePassword if password should be included in connection string.
+     * @returns connection string for the connection
+     */
+    public async getConnectionString(connectionUri: string, includePassword: boolean = false): Promise<string> {
```
Charles-Gagnon, line 268: You take in "includePassword" but never actually set it

typings/vscode-mssql.d.ts
```diff
@@ -76,6 +76,10 @@ declare module 'vscode-mssql' {
          */
         getDatabaseNameFromTreeNode(node: ITreeNodeInfo): string;
 
+        /**
+         * Get the connection string for the provided connection Uri
```
Charles-Gagnon, line 80: Missing parameters in js doc

Charles-Gagnon (changes requested):

src/models/contracts/connection.ts
VasuBhog: Sure thing will update this request section. It seems other requests in this file that I was following are not like that.

typings/vscode-mssql.d.ts
```diff
@@ -76,6 +76,13 @@ declare module 'vscode-mssql' {
...
+         * @param connectionUri The URI of the connection to get the connection string for.
+         * @param includePassword to include password with connection string
+         * @returns connection string
+         */
+        getConnectionString(connectionUri: String, includePassword: boolean): Promise<string>;
```
Charles-Gagnon, line 85: includePassword should be optional per the params typing (and update the comment to mention default is false)

Charles-Gagnon (approved):

kisantia (approved):

src/controllers/connectionManager.ts
```diff
@@ -259,6 +259,20 @@ export default class ConnectionManager {
         return (fileUri in this._connections && this._connections[fileUri].connecting);
     }
 
+    /**
+     * Get the connection string for the provided connection ID
+     * @param connectionId The connection ID for the connection.
```
chlafreniere, line 264: connectionUri

PR #17187, opened 2022-01-06
Connection String request edit

VasuBhog:
After debugging found an issue with the STS value that gets returned as undefined if we do .connectionString previously, as such we need to ensure to make it a string to be used in sql-database-project. 

Original PR: https://github.com/microsoft/vscode-mssql/pull/17186

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -270,7 +270,7 @@ export default class ConnectionManager {
         listParams.ownerUri = connectionUri;
         listParams.includePassword = includePassword;
         const result = await this.client.sendRequest(ConnectionContracts.GetConnectionStringRequest.type, listParams);
-        return result.toString();
+        return result;
     }
 
     /**
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -269,8 +269,7 @@ export default class ConnectionManager {
         const listParams = new ConnectionContracts.GetConnectionStringParams();
         listParams.ownerUri = connectionUri;
         listParams.includePassword = includePassword;
-        const result = await this.client.sendRequest(ConnectionContracts.GetConnectionStringRequest.type, listParams);
-        return result;
+        return this.client.sendRequest(ConnectionContracts.GetConnectionStringRequest.type, listParams);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/controllers/connectionManager.ts
```diff
@@ -270,7 +270,7 @@ export default class ConnectionManager {
         listParams.ownerUri = connectionUri;
         listParams.includePassword = includePassword;
         const result = await this.client.sendRequest(ConnectionContracts.GetConnectionStringRequest.type, listParams);
-        return result.connectionString;
+        return result.toString();
```
Charles-Gagnon, line 273: According to https://github.com/Microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Connection/Contracts/GetConnectionStringRequest.cs the issue here is that you have the response be an object. 

So the correct fix is to remove the result object and just have the type be a string in https://github.com/microsoft/vscode-mssql/blob/main/src/models/contracts/connection.ts#L263

Charles-Gagnon (changes requested):

src/controllers/connectionManager.ts
Charles-Gagnon: Out of curiosity where did you get that original result object definition from? I'm surprised this ever worked given what I'm seeing.

VasuBhog: I was following a similar function listDatabases and I did not clearly see the STS result until I started debugging it. Not sure why it worked at some point and didn't. But it's fixed now.

kisantia (approved):

src/controllers/connectionManager.ts
```diff
@@ -270,7 +270,7 @@ export default class ConnectionManager {
         listParams.ownerUri = connectionUri;
         listParams.includePassword = includePassword;
         const result = await this.client.sendRequest(ConnectionContracts.GetConnectionStringRequest.type, listParams);
-        return result.connectionString;
+        return result;
```
Charles-Gagnon, line 273: You can just return the sendRequest call, that way you aren't having to unnecessarily re-wrap the result in a promise.

Charles-Gagnon (approved):

chlafreniere (approved):

PR #17190, opened 2022-01-08
add loc strings

alanrenmsft:
Noticed this while debugging the extension.

image

the issue was introduced by: https://github.com/microsoft/vscode-mssql/pull/17146

### DIFF
--- a/package.json
+++ b/package.json
@@ -561,14 +561,14 @@
         "mssql.azureActiveDirectory": {
           "type": "string",
           "default": "AuthCodeGrant",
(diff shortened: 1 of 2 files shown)


Charles-Gagnon (approved):
Were these just never added or something?

abist (approved):

kburtram (approved):
Thanks for fixing!

alanrenmsft:
> Were these just never added or something?

they were added a month back and never had actual strings in the localized file.

PR #17191, opened 2022-01-08
more loc strings

alanrenmsft:
committed my previous PR too soon https://github.com/microsoft/vscode-mssql/pull/17190, another warning pop up after fixing those three.

also I noticed in the original PR https://github.com/microsoft/vscode-mssql/pull/17146 that these strings were added directly to the xlf file, which I think is not correct for the strings in package.json file. xlf files are meant to be used by the strings in the code and should not be edited by us directly. @smartguest , please confirm.


Charles-Gagnon (changes requested):
@smartguest Can verify, but I'm pretty sure that the xlf stuff should be kept. That's what the loc team uses as the baseline for the translations similar to https://github.com/microsoft/azuredatastudio/blob/aed561c7efd560c3cc3d335ce9b7fdde022179c8/resources/xlf/en/arc.xlf

Although I'm looking at this and it seems like the extension is mostly not localized. @kburtram @smartguest is that something you're aware of?

alanrenmsft:
I saw that we have loc files for package.json independent of 
image

but yes, @smartguest please confirm how does localization for different resources.

smartguest:
@Charles-Gagnon  @alanrenmsft  The package NLS json file is where you update the strings normally and there usually is a process to update the XLFs with the changes from the json file. But yes, the loc team uses the XLF files as a base to push their changes in. I'm looking into the vscode-mssql gulp/build files right now to see if there's such a process in it.

smartguest:
@Charles-Gagnon  @alanrenmsft It appears that in the case of vscode-mssql, you are supposed to add the strings to the XLF file then run gulp build (which runs the localization task), however the gulp task doesn't appear to work at the moment, so for the time being, you need to add the strings to package.nls.json manually for now, I'll post a bug for the gulp task not working in issues.

Charles-Gagnon:
It seems like the process should be to check in changes to both nls and enu.xlf files at the same time (and then loc team picks up xlf changes and pushes update to other xlf files). Is there any reason we would want to split up that process?

This goes back to my comments about how it's done in ADS - having to do a separate PR/merge just to update the ENU XLF files after updating the loc stuff seems like something that should be unnecessary.

smartguest (changes requested):
The localizedPackage.json.enu.xlf source file still needs to have the strings as that is where the translations are received by the localization team. The package.nls.json file should be updated when we run gulp build, but since the localization process isn't working right now, it's fine to add it to the package.nls.json file for now.

smartguest:
@Charles-Gagnon there is a task in the gulp file that is supposed to generate the package.nls.json and other localized files from the XLFs, but that gulp task is currently not working at the moment, so the current thing right now is to just update the XLF and package.nls.json files then push them in and get translations back from the loc team (only issue from that point is to figure out why the localization task is not working).

Charles-Gagnon:
@smartguest Could you please document the process somewhere (as it works currently, and then when we fix it we can update it as needed then) for what people need to update when adding strings (both in the package.json and in the code)

Charles-Gagnon:
Please add that to the wiki 😄 Adding a new section to https://github.com/microsoft/vscode-mssql/wiki/contributing seems fine for now

alanrenmsft:
@smartguest also when looking at the ADS repo, the resource id for the strings in package.json is not the directly using the resource id in packages.json, please also document how to come up with the new id in xlf file.

smartguest:
@Charles-Gagnon  @alanrenmsft right now, it appears you have to edit the XLF files directly to get any translations back as they're the source files read by the localization process and the localization tasks in the gulp file are designed around using the XLF files. Currently the tasks are:

1. Export json files from *.xlf files (generate i18n files in the localization i18n folder)
2. Generate a localizedConstants.ts file in src/constants (that is based on whatever is put into the localizedConstants.enu.xlf file),
3. Create localized nls files based on the English package.nls.json file and values found in localizedPackage.json.*.xlf (there is a problem with the path that I'm fixing currently)

It seems for now you will need to add constants values to BOTH the  constants.ts/package.nls.json files and the respective English xlf files in order to generate localized values properly with this build task.

Charles-Gagnon:
We shouldn't need to be modifying two files just to get the correct localized strings. Can you fix it so that they're generated automatically like we have for ADS?

smartguest:
I can attempt to reproduce the XLF generation from ADS, but it will take some time to write a number of gulp tasks that works with the structure of vscode-mssql. For now, the XLF files, package.nls and constants file must share the same strings for translation and the building of localized resources to work properly.

smartguest:
@alanrenmsft I merged in the changes from main into your branch, and also added back in the mssql.Configuration string, also those mssql.auth strings were used by package.json so they were added back in. Right now, when I run the build, I no longer get a pop up message.

smartguest (approved):

Charles-Gagnon (approved):
Excellent, thank you for taking care of this @smartguest !

PR #17210, opened 2022-01-12
Add SBOM build task

kburtram:


kburtram:
Push as-is to get the initial manifests generated.  Will add package version field later.

PR #17213, opened 2022-01-13
Fix spacing in build.yml

kburtram:
A example space got added into build.yaml when merging spaces->tabs change from main.


kburtram:
fixing build break, so merging.

PR #17216, opened 2022-01-19
Removed new position in text editor for accessibility reasons

### LINKED ISSUE #17203: A11y_mssql for VSCode_Select top 1000 _Keyboard:When pressing enter in the "select top 1000" tab item, the first keyboard focus is moving to the second line in the editor section instead of first line.
“[Check out Accessibility Insights!](https://accessibilityinsights.io/) - Identify accessibility bugs before check-in and make bug fixing faster and easier.”

**GitHubTags:**#A11yMAS;#A11yTCS;#BM_mssql for VSCode_Jan2022;#mssql for VSCode;#Win32;#FTP;#DesktopApp;#A11ySev2;#Keyboard;#AINotInScope;#WCAG2.4.3;#FocusOrder;

## Environment Details:
Application Name: mssql for VSCode
Windows Version: Win11

## Repro Steps:

1. Open Visual studio code 
2. Tab till extension button and search for "mssql"
3. Tab till install and move to the SQL Server  button
4. Tab till Add Connection Button and insert the valid credentials
5. After adding the server move to database expand it -->System database expand-->Master expand-->Tables expand -->System table expand and right click on it and select top 1000 option
6. Select top 1000 window will open and check if the focus in the editor is moving to second line.

## Actual:
When pressing enter in the "select top 1000" tab item, the first keyboard focus is moving to the second line in the editor section instead of first line.

#### Similar issue is also observed in following screens.

1. Script as create_first focus moving to second line.
2. Script as Drop_first focus moving to second line.
3. Script as Alter_first focus moving to second line.

## Expected:
When pressing enter in the "select top 1000" tab item, the first keyboard focus should move to the first line in the editor section.

## User Impact:
Keyboard users will face difficulty in navigating the page.

smartguest:
Fixes #17203

@abist Will this affect any feature or bug fix you previously implemented earlier? It seems to work fine when I used it but do let me know if this will break anything.

### DIFF
--- a/src/controllers/untitledSqlDocumentService.ts
+++ b/src/controllers/untitledSqlDocumentService.ts
@@ -27,10 +27,6 @@ export default class UntitledSqlDocumentService {
 				preserveFocus: false,
 				preview: false
 			});
-		const position = editor.selection.active;
-		let newPosition = position.with(position.line + 1, 0);
-		let newSelection = new vscode.Selection(newPosition, newPosition);
-		editor.selection = newSelection;
 		return editor;
 	}
 }


Charles-Gagnon (approved):
Looks like this was changed in https://github.com/microsoft/vscode-mssql/pull/1290 but I can't tell which bug that was supposedly fixing at first glance. Let's wait on @abist before merging in, but I'm not really sure what the point of that was supposed to be so seems fine to me.

abist:
I don't quite remember why I did this, but if you've played around selections and running selected queries (and not the whole file) and tried running multiple queries together by opening new tabs and it works fine, then it should be good.

PR #17217, opened 2022-01-19
Create Azure Function from Table

VasuBhog:
Previous PR: https://github.com/microsoft/vscode-mssql/pull/17056
Work that will be done in the future:
- convert the Connection String to be received via STS (already planned enhancement
- Work to be moved to STS if discussed to be better 

Overall Goal: 
- Allow the user to be able to create a SQL binding directly from a table (that is from a connected server)

Assumptions – User has a server already set up a SQL Server connection with a table

1. User has an Azure function already established and does not need to create a trigger 
2. Once done with creating the project. User clicks onto SQL Server connections tab in VS Code and right clicks on the table after connecting to specific SQL Server and chooses “Create Azure Function with SQL Binding” from. 
image
3. Create Azure Function From Table steps – Function Name (HTTPTrigger1), namespace (Company.Function), AccessRight (Anonymous) 
4. Once completed the HttpTrigger will be refactored in real-time to add the Sql binding that includes the table name and the SQL Connection String that is saved in local.settings.json 

Video demo:
https://user-images.githubusercontent.com/23587151/151084615-ffd3ea4e-3b2b-453e-9157-e226c0aff74d.mov


.vscode/launch.json
```diff
@@ -29,8 +29,7 @@
             "args": ["--extensionDevelopmentPath=${workspaceRoot}" ],
             "stopOnEntry": false,
             "sourceMaps": true,
-            "outFiles": ["${workspaceRoot}/out/src/**/*.js"],
-            "preLaunchTask": "cover:disable"
```
Charles-Gagnon, line 33: Why did you remove this?

package-lock.json
```diff
@@ -0,0 +1,8442 @@
+{
```
Charles-Gagnon: We use yarn, not npm. Remove this and run yarn install instead.

package.nls.json
```diff
@@ -94,6 +94,7 @@
 "mssql.intelliSense.lowerCaseSuggestions":"Should IntelliSense suggestions be lowercase",
 "mssql.persistQueryResultTabs":"Should query result selections and scroll positions be saved when switching tabs (may impact performance)",
 "mssql.queryHistoryLimit":"Number of query history entries to show in the Query History view",
+"mssql.createAzureFunction":"Create Azure Function…",
```
Charles-Gagnon, line 97: `Create Azure Function with SQL Binding...`?

@dzsquared

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,156 @@
...
+		await this.addNugetReferenceToProjectFile();
+		await this.addConnectionStringToConfig(connectionString);
+		const functionFile = await newFilePromise;
+
+
+		await this.azureFunctionsService.addSqlBinding(
+			mssql.BindingType.input,
+			functionFile,
+			functionName,
+			schema + '.' + table,
```
Charles-Gagnon, line 62: This needs to be properly escaped

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,156 @@
...
+			return;
+		}
+
+		// because of an AF extension API issue, we have to get the newly created file by adding
+		// a watcher: https://github.com/microsoft/vscode-azurefunctions/issues/2908
+		const newFilePromise = this.getNewFunctionFile();
+
+		// get function name from user
+		const functionName = await vscode.window.showInputBox({
+			title: 'Function Name',
```
Charles-Gagnon, line 43: Localize

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,156 @@
...
+			value: 'HttpTrigger1'
```
Charles-Gagnon, line 44: Seems reasonable to have this default to the table name

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,156 @@
...
+		);
+
+		this.refactorAzureFunction(functionFile);
+	}
+
+	private async getAzureFunctionsExtensionApi(): Promise<af.AzureFunctionsExtensionApi> {
+		const afExtension = vscode.extensions.getExtension(azureFunctionsExtensionName);
+		if (afExtension) {
+			let azureFunctionApi;
+			if (!afExtension.isActive) {
```
Charles-Gagnon, line 73: This is all pretty unnecessary, you can just call activate and if it's already activated it'll just immediately return the promise. (there may be some very slight overhead, but in general it's not worth the extra code here to do that ourselves)

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,156 @@
...
+				azureFunctionApi = await afExtension.activate();
+			} else {
+				azureFunctionApi = afExtension.exports;
+			}
+			return azureFunctionApi.getApi('*') as af.AzureFunctionsExtensionApi;
```
Charles-Gagnon, line 78: For this I would suggest pulling down the typings from https://github.com/microsoft/vscode-azuretools/blob/main/ui/api.d.ts so we have proper typings

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,156 @@
...
+			functionFile,
+			functionName,
+			schema + '.' + table,
+			sqlConnectionString
+		);
+
+		this.refactorAzureFunction(functionFile);
+	}
+
+	private async getAzureFunctionsExtensionApi(): Promise<af.AzureFunctionsExtensionApi> {
```
Charles-Gagnon, line 69: `Promise<af.AzureFunctionsExtensionApi | undefined>`

Also I would suggest putting this in the utils module. There's no reason this needs to be a part of any class.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,156 @@
...
+				azureFunctionApi = afExtension.exports;
+			}
+			return azureFunctionApi.getApi('*') as af.AzureFunctionsExtensionApi;
+		} else {
+			vscode.window.showErrorMessage(LocalizedConstants.azureFunctionsExtensionNotInstalled);
+			return undefined;
+		}
+	}
+
+	private refactorAzureFunction(filePath: string): void {
```
Charles-Gagnon, line 85: Comments for these functions please - what exactly is being "refactored"? Examples in the doc would be nice too

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,156 @@
...
+				// Normal lines to be included
+				replacedValueLines.push(defaultLine);
+			}
+		}
+		newValue = replacedValueLines.join(os.EOL);
+		fs.writeFileSync(filePath, newValue, 'utf-8');
+	}
+
+	private async isAzureFunctionProjectOpen(): Promise<boolean> {
+		if (vscode.workspace.workspaceFolders === undefined || vscode.workspace.workspaceFolders.length === 0) { return false; }
```
Charles-Gagnon, line 109: Format so it's not all on one line

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,156 @@
...
+			} else {
+				// Normal lines to be included
+				replacedValueLines.push(defaultLine);
+			}
+		}
+		newValue = replacedValueLines.join(os.EOL);
+		fs.writeFileSync(filePath, newValue, 'utf-8');
+	}
+
+	private async isAzureFunctionProjectOpen(): Promise<boolean> {
```
Charles-Gagnon, line 108: All these helper functions can just be at the module level, no reason to have them in the class.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,156 @@
...
+		const projFile = await this.getProjectFile();
+		azureFunctionUtils.addPackageToAFProjectContainingFile(vscode.Uri.file(projFile), constants.sqlExtensionPackageName);
+	}
+
+	private async addConnectionStringToConfig(connectionString: string): Promise<void> {
+		const settingsFile = await this.getSettingsFile();
+		const content = await fs.promises.readFile(settingsFile);
+		const config = JSON.parse(content.toString());
+
+		if (!(sqlConnectionStringPropertyName in config.Values)) {
```
Charles-Gagnon, line 152: We should still be setting the connection string even if the setting already exists (or at least warning/prompting the user)

It actually looks like setLocalAppSetting already does this so you can probably just remove this check

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+/**
+ * Adds specified package to Azure Functions project the specified file is a part of
+ * @param filePath uri of file to find the containing AF project of to add package reference to
+ * @param packageName package to add reference to
+ * @param packageVersion optional version of package. If none, latest will be pulled in
+ */
+export async function addPackageToAFProjectContainingFile(
+	fileUri: vscode.Uri, packageName: string,
```
Charles-Gagnon, line 82: If you're going to have these on separate lines please format them so they're each on their own line for readability

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,154 @@
...
+			await this.addPackage(project, packageName, packageVersion);
+		} else {
+			void vscode.window.showInformationMessage(constants.addPackageReferenceMessage, constants.moreInformation).then((result) => {
+				if (result === constants.moreInformation) {
+					void vscode.env.openExternal(vscode.Uri.parse(constants.sqlBindingsHelpLink));
+				}
+			});
+		}
+	} catch (e) {
+		const result = await vscode.window.showErrorMessage(constants.addSqlBindingPackageError, constants.checkoutOutputMessage);
```
Charles-Gagnon, line 100: This shouldn't be blocking here - make it a .then like in the else above

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,154 @@
...
+ * @returns uri of project or undefined if project couldn't be found
+ */
+export async function getAFProjectContainingFile(fileUri: vscode.Uri): Promise<vscode.Uri | undefined> {
+	// get functions csprojs in the workspace
+	const projectPromises = vscode.workspace.workspaceFolders?.map(f => getAllProjectsInFolder(f.uri, '.csproj')) ?? [];
+	const functionsProjects = (await Promise.all(projectPromises)).reduce((prev, curr) =>
+		prev.concat(curr), []).filter(p => isFunctionProject(path.dirname(p.fsPath)));
+
+	// look for project folder containing file if there's more than one
+	if (functionsProjects.length > 1) {
```
Charles-Gagnon, line 120: Are we going to warn the user when this happens?

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,154 @@
...
+		console.error('need to find which project contains the file ' + fileUri.fsPath);
+		return undefined;
+	} else if (functionsProjects.length === 0) {
+		throw new Error(constants.noAzureFunctionsProjectsInWorkspace);
+	} else {
+		return functionsProjects[0];
+	}
+}
+
+// Use 'host.json' as an indicator that this is a functions project
```
Charles-Gagnon, line 133: Proper js doc here (and actual link to source code would be ideal)

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,154 @@
...
+	return fse.pathExists(path.join(folderPath, constants.hostFileName));
+}
+
+/**
+ * Gets all the projects of the specified extension in the folder
+ * @param folder
+ * @param projectExtension project extension to filter on
+ * @returns array of project uris
+ */
+export async function getAllProjectsInFolder(folder: vscode.Uri, projectExtension: string): Promise<vscode.Uri[]> {
```
Charles-Gagnon, line 145: Can probably even just call this `getAllFilesInFolder`

src/constants/constants.ts
```diff
@@ -151,3 +154,47 @@ export const tenantDisplayName = 'Microsoft';
...
+		'string requestBody = await new StreamReader(req.Body).ReadToEndAsync();',
+		'dynamic data = JsonConvert.DeserializeObject(requestBody);',
+		'name = name ?? data?.name;',
+		'string responseMessage = string.IsNullOrEmpty(name) ? \"This HTTP triggered function executed successfully. Pass a name in the query string or in the request body for a personalized response.\" : $\"Hello, {name}. This HTTP triggered function executed successfully.\";'
+	];
+export const defaultBindingResult = 'return new OkObjectResult(responseMessage);';
+export const sqlBindingResult = `return new OkObjectResult(result);`;
+export const azureFunctionLocalSettingsFileName = 'local.settings.json';
+export const sqlExtensionPackageName = 'Microsoft.Azure.WebJobs.Extensions.Sql';
+// tslint:disable-next-line:typedef
```
Charles-Gagnon, line 175: Generally we shouldn't be disabling tslint rules. Why did you need to do this?

src/utils/connectionStringUtil.ts
```diff
@@ -0,0 +1,20 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AuthenticationTypes } from '../models/interfaces';
+
+export function getConnectionString(server: string, authType: string, database: string, userName: string, password: string): string {
```
Charles-Gagnon, line 8: Why not use the STS functionality instead? This isn't very future-proof and will miss out on any custom settings the user might have specified for their connection.

Charles-Gagnon (changes requested):

.vscode/launch.json
VasuBhog: some of these changes I did not make. @abist / @alanrenmsft made, and as such, I refactored and fixed the parts that needed to be changed

src/azureFunction/azureFunctionProjectService.ts
VasuBhog: the entire string or just the period?

src/azureFunction/azureFunctionProjectService.ts
VasuBhog: Following how the Create Azure Function via Azure Functions is set up. HttpTrigger# is typically set up as the default value. @dzsquared for thoughts.

src/constants/constants.ts
VasuBhog: Originally was getting errors with the linting so disabled, however, realized it needed to have a defined return type.

src/azureFunction/azureFunctionUtils.ts
VasuBhog: As of this moment it seems sql-database-projects also just console.error out at the moment. But I can put something like this for now - 

> 		vscode.window.showWarningMessage('Unable to find which project contains the file: ' + fileUri.fsPath);

.vscode/launch.json
Charles-Gagnon: Please remove it for now unless there's a reason to keep it, we can do a follow up PR if there's a reason it needs to be removed.

Charles-Gagnon:
Is there a reason this is still a draft?

src/azureFunction/azureFunctionProjectService.ts
Charles-Gagnon: That would make sense if this was an empty HttpTrigger function, but we're specifically making a function with a SQL binding to a table so leaving it as the default name is unlikely to be what the user wants.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,144 @@
...
+		await addNugetReferenceToProjectFile();
+		await addConnectionStringToConfig(connectionString);
+		const functionFile = await newFilePromise;
+
+
+		await this.azureFunctionsService.addSqlBinding(
+			mssql.BindingType.input,
+			functionFile,
+			functionName,
+			schema + escape('.') + table,
```
Charles-Gagnon, line 59: This isn't what I meant by escape the string. You need to escape the object names to prevent possible SQL injection attacks (and allow names that have things like periods in them).

So escaping in this case is by bracket quoting the schema and table name (separately). So `dbo.myTable` becomes `[dbo].[myTable]`

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,176 @@
...
+ */
+export async function addPackageToAFProjectContainingFile(fileUri: vscode.Uri, packageName: string, packageVersion?: string): Promise<void> {
+	try {
+		const project = await getAFProjectContainingFile(fileUri);
+
+		// if no AF projects were found, an error gets thrown from getAFProjectContainingFile(). This check is temporary until
+		// multiple AF projects in the workspace is handled. That scenario returns undefined and shows an info message telling the
+		// user to make sure their project has the package reference
+		if (project) {
+			await this.addPackage(project, packageName, packageVersion);
```
Charles-Gagnon, line 92: Is this correct?

1. I don't see the `addPackage` defined
2. You're using `this` but this is at the module level so there's no "this" to bind to

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,144 @@
...
+		// create C# HttpTrigger
+		await azureFunctionApi.createFunction({
+			language: 'C#',
+			templateId: 'HttpTrigger',
+			functionName: functionName
+		});
+
+		// once steps are addressed we add SQL Bindings Nuget and Connection String to local.settings.json
+		// and then create a new functions file
+		await addNugetReferenceToProjectFile();
```
Charles-Gagnon, line 50: I think we need to revisit how this all works. Currently it's kind of odd that the logic is split up into two places : 

1. The extension handles stuff like adding the nuget reference, adding the connection string config and this whole "refactoring" stuff
2. STS handles actually adding the sql binding

There doesn't seem to be any reason why we would split that logic up. Having STS handle everything seems to make more sense to me - it has a lot more power (such as access to the rosyln parser) to make code modifications and keeping things in one place (especially a common place like STS) means that if anything else wants to use it we don't have to re-implement this logic elsewhere (say if we wanted to add this functionality into ADS for example).

@alanrenmsft @abist Do you know why this was done like this originally?

I'm doing some work that is going to need to use this functionality which I would greatly prefer to have all in STS if possible. Now since we've done the work here already I guess we can get it in like this for the first step, but if there's nothing blocking it then I'd like to see the next step after merging this in be to move the logic over to STS (the logic itself can mostly stay the same so that shouldn't be a huge task).

@chlafreniere Thoughts?

package.nls.json
dzsquared: yes, including the SQL binding mention is important context for the user
"Create Azure Function with SQL Binding"
or
"Generate Azure Function with SQL Binding"

src/azureFunction/azureFunctionUtils.ts
VasuBhog: removed this found a simpler solution for the current functionality.

src/utils/connectionStringUtil.ts
VasuBhog: as said in the PR description. This would require the same functionality that we discussed in STS of accepting connectionInfo otherwise would have to reconnect to the Server in order to get the proper uri to send to the connection string request. Will address this again since most work seems to be going to STS in near future.

src/azureFunction/azureFunctionProjectService.ts
VasuBhog: I can setup a meeting tomorrow / async discussion on this in order to get this resolved sooner rather than later.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,165 @@
...
+		await addConnectionStringToConfig(connectionString);
+		const functionFile = await newFilePromise;
+
+
+		await this.azureFunctionsService.addSqlBinding(
+			mssql.BindingType.input,
+			functionFile,
+			functionName,
+			'[' + schema + ']' + '.' + '[' + table + ']',
```
Charles-Gagnon, line 62: You need to properly escape ] (replace ] with ]] in the original string). I would suggest a simple helper function to do this

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// adds the required nuget package to the project
+async function addNugetReferenceToProjectFile(): Promise<void> {
+	// Make sure the nuget source is added
+	const currentSources = await executeCommand('dotnet nuget list source');
+	if (currentSources.indexOf(sqlBindingNugetSource) === -1) {
+		await executeCommand(`dotnet nuget add source ${sqlBindingNugetSource}`);
+	}
+	const projFile = await getProjectFile();
```
Charles-Gagnon, line 141: Right now you're just grabbing the first project file that is returned from the findFiles call. If the user has multiple projects open and the first one isn't a azure functions project then this is going to be getting the incorrect one.

I believe STS already has logic to get the project based on the existence of the host.json at the same level - you should likely be doing something similar. (first check whether host.json is guaranteed to be at the same level. But that would be a better place to start for now since I imagine other locations aren't as common)

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,165 @@
...
+	await executeCommand(`dotnet add package ${sqlBindingPackageName} --prerelease`, path.dirname(projFile));
+}
+
+// adds the Sql Connection String to the local.settings.json
+async function addConnectionStringToConfig(connectionString: string): Promise<void> {
+	const settingsFile = await getSettingsFile();
+	await azureFunctionUtils.setLocalAppSetting(path.dirname(settingsFile), constants.sqlConnectionString, connectionString);
+}
+
+async function executeCommand(command: string, cwd?: string): Promise<string> {
```
Charles-Gagnon, line 151: This should go in a generic utils file ideally - it's not really specific to AF stuff

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+export interface ILocalSettingsJson {
+	IsEncrypted?: boolean;
+	Values?: { [key: string]: string };
+	Host?: { [key: string]: string };
+	ConnectionStrings?: { [key: string]: string };
+}
+
+/**
+ * copied and modified from vscode-azurefunctions extension
```
Charles-Gagnon, line 26: A link to where this was copied from would be great

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+	settings.Values[key] = value;
+	await fse.writeJson(localSettingsPath, settings, { spaces: 2 });
+
+	return true;
+}
+
+export async function getAzureFunctionsExtensionApi(): Promise<af.AzureFunctionsExtensionApi | undefined> {
+	const afExtension = vscode.extensions.getExtension(constants.azureFunctionsExtensionName);
+	if (afExtension) {
+		let azureFunctionApi = await afExtension.activate();
```
Charles-Gagnon, line 80: Please include the typings for this extension API as well (https://github.com/microsoft/vscode-azuretools/blob/main/ui/api.d.ts)

See https://github.com/Azure/Azure-Data-API/blob/main/src/providers/azureFunctionApiLayerProvider.ts#L13 for an example

(and adding links to the original sources is a helpful thing too)

src/constants/constants.ts
```diff
@@ -151,3 +154,44 @@ export const tenantDisplayName = 'Microsoft';
...
+	return localize('SettingAlreadyExists', 'Local app setting \'{0}\' already exists. Overwrite?', settingName);
+}
+export const yesString = localize('yesString', 'Yes');
+export const noAzureFunctionsProjectsInWorkspace = localize('noAzureFunctionsProjectsInWorkspace', 'No Azure functions projects found in the workspace');
+export const hostFileName = 'host.json';
+export const sqlBindingsHelpLink = 'https://github.com/Azure/azure-functions-sql-extension/blob/main/README.md';
+export const addPackageReferenceMessage = localize('addPackageReferenceMessage', 'To use SQL bindings, ensure your Azure Functions project has a reference to {0}', sqlExtensionPackageName);
+export const moreInformation = localize('moreInformation', 'More Information');
+export const addSqlBindingPackageError = localize('addSqlBindingPackageError', 'Error adding Sql Binding extension package to project');
+export const checkoutOutputMessage = localize('checkoutOutputMessage', 'Check output pane for more details');
```
Charles-Gagnon, line 192: It's generally helpful to make `output pane` a link to actually show the output channel with messages like this since otherwise the user has no idea what channel to look in.

src/constants/parseJson.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Parses and returns JSON
+ * Has extra logic to remove a BOM character if it exists and handle comments
+ */
+export function parseJson<T extends object>(data: string): T {
+	if (data.charCodeAt(0) === 0xFEFF) {
+		data = data.slice(1);
+	}
+
+	const errors: jsonc.ParseError[] = [];
+	const result: T = <T>jsonc.parse(data, errors, { allowTrailingComma: true });
```
Charles-Gagnon, line 21: Is this from something? Is there a reason you can't just use JSON.parse instead of relying on a 3rd party package?

Charles-Gagnon (changes requested):

src/azureFunction/azureFunctionProjectService.ts
VasuBhog: The original string we get back does not include brackets (current dbo.table), so the escape that I do returns this [dbo].[table].

Charles-Gagnon: Yes, that's going to be the case for most tables I imagine....but this is user input. The table/schema names can be anything that's a valid object name and so can include ] in the name. Then if you quote it like you do currently you'll end up with something like `[My]Table]` which is invalid (it'll see the first ] and think that's the end of the name). 

Try making a function from a table with a ] in its name to see what happens.

src/constants/parseJson.ts
VasuBhog: This is used in getLocalSettingsJson that was reused from the sql-database-projects repo.

src/constants/constants.ts
VasuBhog: removed this with other constants that are not needed anymore

src/azureFunction/azureFunctionProjectService.ts
VasuBhog: In order to address this I added a new prompt: 
image

src/constants/constants.ts
VasuBhog: did not need this as the function that was using it was deleted

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+				// Normal lines to be included
+				replacedValueLines.push(defaultLine);
+			}
+		}
+		defaultBindedFunctionText = replacedValueLines.join(os.EOL);
+		fs.writeFileSync(filePath, defaultBindedFunctionText, 'utf-8');
+	}
+}
+
+// A C# Azure Functions project must be present in order to create a new Azure Function for the table
```
Charles-Gagnon, line 93: Property xml doc comments please

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+async function isAzureFunctionProjectOpen(): Promise<boolean> {
+	if (vscode.workspace.workspaceFolders === undefined || vscode.workspace.workspaceFolders.length === 0) {
+		return false;
+	} else {
+		const hostFile = await getHostFile();
+		const projectFiles = await getProjectFile();
+		if (projectFiles !== undefined && hostFile !== undefined) {
+			// select project to add azure function to
+			selectedProjectFile = (await vscode.window.showQuickPick(projectFiles, {
```
Charles-Gagnon, line 102: We should only be showing function project files. How I would suggest handling this : 

1. Look for all host.json files
2. For each of those files then filter it down to just ones that also have a csproj next to it
3. If there's more than 1 project file left then prompt the user to select at that point

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// gets the local.settings.json file path
+async function getSettingsFile(): Promise<string | undefined> {
+	const settingsFiles = await vscode.workspace.findFiles('**/local.settings.json');
+	return settingsFiles.length > 0 ? settingsFiles[0].fsPath : undefined;
+}
+
+// retrieves the new function file once the file is created
+function getNewFunctionFile(): Promise<string> {
```
Charles-Gagnon, line 136: You know which folder the project is in at this point right? You should only be watching for files under that folder - no reason to watch the entire workspace.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+	return new Promise((resolve) => {
+		const watcher = vscode.workspace.createFileSystemWatcher(new vscode.RelativePattern(
```
Charles-Gagnon, line 138: We should really have a timeout here too, otherwise we might hang forever if something breaks creating the file.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// adds the required nuget package to the project
+async function addNugetReferenceToProjectFile(): Promise<void> {
+	// Make sure the nuget source is added
+	const currentSources = await executeCommand('dotnet nuget list source');
+	if (currentSources.indexOf(sqlBindingNugetSource) === -1) {
+		await executeCommand(`dotnet nuget add source ${sqlBindingNugetSource}`);
+	}
+	await executeCommand(`dotnet add package ${constants.sqlExtensionPackageName} --prerelease`, path.dirname(selectedProjectFile));
```
Charles-Gagnon, line 154: You should pass in the file path as a parameter, not use this global var.

Also do you need to do the dirname? dotnet add should take a direct reference to the project file just fine

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// adds the Sql Connection String to the local.settings.json
```
Charles-Gagnon, line 157: proper xml doc

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+async function addConnectionStringToConfig(connectionString: string): Promise<void> {
+	const settingsFile = await getSettingsFile();
+	await azureFunctionUtils.setLocalAppSetting(path.dirname(settingsFile), constants.sqlConnectionString, connectionString);
+}
+
+// escapes objectName for add SQL Binding
```
Charles-Gagnon, line 163: 1. proper xml doc
2. The doc shouldn't reference sql bindings. This is nothing specific to sql bindings - something like `Returns the quoted and escaped full name of the specified schema and object`.

And then make the signature `generateQuotedFullName(schema: string, objectName: string): string` since it can be used for any object (not just tables).
3. I would also suggest putting this in a more common utils file - it doesn't need to live here.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+function escapeObjectName(schema: string, table: string): string {
+	table = table.replace(/\]/g, ']]');
```
Charles-Gagnon, line 165: You need to do this for the schema too...

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+	return '[' + schema + ']' + '.' + '[' + table + ']';
```
Charles-Gagnon, line 166: It'd be more readable to just do 

```
return `[${schema}].[${table}]`;
```

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,90 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the Source EULA. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import * as fse from 'fs-extra';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import * as constants from '../constants/constants';
+import * as LocalizedConstants from '../constants/localizedConstants';
+// https://github.com/microsoft/vscode-azuretools/blob/main/ui/api.d.ts
```
Charles-Gagnon, line 10: This isn't the correct link - this typings file is from https://github.com/microsoft/vscode-azurefunctions/blob/main/src/vscode-azurefunctions.api.d.ts

src/constants/constants.ts
```diff
@@ -151,3 +154,37 @@ export const tenantDisplayName = 'Microsoft';
...
+export function failedToParse(errorMessage: string): string {
+	return localize('failedToParse', 'Failed to parse "{0}": {1}.',
+		azureFunctionLocalSettingsFileName, errorMessage);
+}
+export function settingAlreadyExists(settingName: string): string {
+	return localize('SettingAlreadyExists', 'Local app setting \'{0}\' already exists. Overwrite?', settingName);
+}
+export const yesString = localize('yesString', 'Yes');
+export const functionNameTitle = localize('functionNameTitle', 'Function Name');
+export const selectProject = localize('selectProject', 'Select an project to add a Azure function with SQL binding to');
```
Charles-Gagnon, line 187: `Select the project to create the Azure Function with SQL binding in`

src/constants/parseJson.ts
```diff
@@ -0,0 +1,45 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the Source EULA. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// copied from vscode-azurefunctions extension
```
Charles-Gagnon, line 6: Why is this up here? And why is it in the constants folder? Having a separate module for this seems fine, but I'd just put it in the utils folder instead.

Also please provide an actual link to where this is from

Charles-Gagnon (changes requested):

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -521,6 +521,11 @@
         <trans-unit id="azureSignInToAzureCloudDescription">
             <source xml:lang="en">Sign in to your Azure subscription in one of the sovereign clouds.</source>
         </trans-unit>
+        <trans-unit id="azureFunctionsExtensionNotInstalled">
+            <source xml:lang="en">Azure Functions extension must be installed in order to use this feature.</source>
+        </trans-unit>
+        <trans-unit id="azureFunctionsProjectMustBeOpened">
```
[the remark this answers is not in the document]
VasuBhog: All I have been doing is yarn install and gulp build & gulp lint. Did not manually create this.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+import * as mssql from 'vscode-mssql';
+import { defaultBindingResult, defaultSqlBindingTextLines, sqlBindingResult } from '../constants/constants';
+import * as LocalizedConstants from '../constants/localizedConstants';
+import { AzureFunctionsService } from '../services/azureFunctionsService';
+import * as azureFunctionUtils from '../azureFunction/azureFunctionUtils';
+import * as constants from '../constants/constants';
+import * as path from 'path';
+import { executeCommand } from '../utils/utils';
+const sqlBindingNugetSource = 'https://www.myget.org/F/azure-appservice/api/v3/index.json';
+let selectedProjectFile: string | undefined = '';
```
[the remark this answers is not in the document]
VasuBhog: This is set by isAzureFunctionProjectOpen and then used in addNugetReferenceToProjectFile. If it was local to that where it is set then we would not know what the user chose.

localization/xliff/enu/constants/localizedConstants.enu.xlf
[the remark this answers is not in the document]
Charles-Gagnon: Right, the MSSQL extension has a different process for this. Ok nevermind, this is fine.

package.json
```diff
@@ -116,14 +118,16 @@
...
   },
   "dependencies": {
-    "@types/node": "^14.14.16",
     "@microsoft/ads-adal-library": "1.0.13",
+    "@types/node": "^14.14.16",
     "core-js": "^2.4.1",
     "decompress-zip": "^0.2.2",
     "ejs": "^3.1.6",
     "error-ex": "^1.3.0",
+    "fast-glob": "^3.2.7",
```
chlafreniere, line 127: Why is fast-glob necessary to add?

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+		let objectName = escapeObjectName(schema, table);
+		await this.azureFunctionsService.addSqlBinding(
+			mssql.BindingType.input,
+			functionFile,
+			functionName,
+			objectName,
+			constants.sqlConnectionString
+		);
+
+		this.refactorAzureFunction(functionFile);
```
chlafreniere, line 65: Just from a naming perspective, I don't quite understand what's going on here.

Above, the sql binding is "added", and now it's being refactor to "include" the sql binding (see comment below).

Please clean up the strings here.

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,90 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the Source EULA. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import * as fse from 'fs-extra';
```
chlafreniere, line 5: Same comment about the other dependency. Do we really need fs-extra, or can we use fs?

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,167 @@
...
+
+		// once steps are addressed we add SQL Bindings Nuget and Connection String to local.settings.json
+		// and then create a new functions file
+		await addNugetReferenceToProjectFile();
+		await addConnectionStringToConfig(connectionString);
+		const functionFile = await newFilePromise;
+
+		let objectName = escapeObjectName(schema, table);
+		await this.azureFunctionsService.addSqlBinding(
+			mssql.BindingType.input,
```
chlafreniere, line 58: Since this is hardcoded to input binding, we need to improve the strings for the experience.

@dzsquared can you help with the string for the context menu? Should it say "input binding"?

src/constants/parseJson.ts
VasuBhog: Many of these files were re-used from sql-database-projects and as such, they were checked in like this. Will move and add link.

chlafreniere (changes requested):

src/azureFunction/azureFunctionProjectService.ts
VasuBhog: Yes it needs to be the dirname

src/constants/constants.ts
```diff
@@ -2,6 +2,8 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
+import * as nls from 'vscode-nls';
+const localize = nls.loadMessageBundle();
```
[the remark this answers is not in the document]
Charles-Gagnon: For VS Code MSSQL these should actually just go into the XLF and then you run the build which will generate a localizedConstants.ts file (with the new localized constants added).

I have no idea why this is the process but it seems like that was how it was set up...

typings/vscode-azurefunctions.api.d.ts
```diff
@@ -0,0 +1,86 @@
+/*---------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
VasuBhog: Please see this comment here: https://github.com/microsoft/vscode-mssql/pull/17217#discussion_r794948986.

src/azureFunction/azureFunctionProjectService.ts
Charles-Gagnon: Oh, didn't realize you were passing this in as the cwd. You can just specify the project file as an arg

`dotnet add <PATH TO PROJECT> package Microsoft.Azure.WebJobs.Extensions.Sql --prerelease`

src/constants/constants.ts
```diff
@@ -151,3 +154,37 @@ export const tenantDisplayName = 'Microsoft';
 export const firewallErrorMessage = 'To enable access, use the Windows Azure Management Portal or run sp_set_firewall_rule on the master database to create a firewall rule for this IP address or address range.';
 export const windowsResourceClientPath = 'SqlToolsResourceProviderService.exe';
 export const unixResourceClientPath = 'SqlToolsResourceProviderService';
+
+// Azure Functions
+export const azureFunctionsExtensionName = 'ms-azuretools.vscode-azurefunctions';
+export const sqlConnectionString = 'SqlConnectionString';
+export const genericCollectionImport = 'using System.Collections.Generic;';
+export const defaultSqlBindingTextLines =
```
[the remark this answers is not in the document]
VasuBhog: As far as I see this is what matches the generated azure function file.

src/utils/connectionStringUtil.ts
```diff
@@ -0,0 +1,20 @@
...
+	const serverDatabasePart = `Server=${server};Database=${database};`;
```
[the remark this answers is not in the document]
VasuBhog: Since we click on the specific table to create the azure function we have all the necessary information. As a table would have access to the connection info.

package.json
VasuBhog: not necessary - was used previously

src/azureFunction/azureFunctionUtils.ts
VasuBhog: Unable to use fs as it uses writeFileSync which would need the contents to be JSON, and since we are putting a map of contents instead fse allows us to do that instead directly.

package.json
```diff
@@ -72,6 +72,7 @@
     "@angular/router": "~3.1.2",
     "@angular/upgrade": "~2.1.2",
     "@types/ejs": "^3.1.0",
+    "@types/fs-extra": "^9.0.13",
```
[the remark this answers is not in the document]
VasuBhog: The latest version for this was 9.0.13

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+
+		// create C# HttpTrigger
+		await azureFunctionApi.createFunction({
+			language: 'C#',
+			templateId: 'HttpTrigger',
+			functionName: functionName,
+			folderPath: projectFile
+		});
+
+		// once steps are addressed we add SQL Bindings Nuget and Connection String to local.settings.json
```
Charles-Gagnon, line 55: What's the "once steps are addressed" for? Typically comments describe what's happening below so that wouldn't be necessary here.

(although honestly this whole comment doesn't super necessary - the function names already indicate what's being done pretty clearly)

src/azureFunction/azureFunctionProjectService.ts
Charles-Gagnon: We definitely should - it'd be nice to support creating an output binding as well but until that's done we should be explicit as possible

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+
+	/**
+	 * Overwrites the Azure function file to include the new binding in the method body
+	 * @param filePath is the path for the function file (.cs for C# functions)
+	 */
+	private overwriteAzureFunctionMethodBody(filePath: string): void {
+		let defaultBindedFunctionText = fs.readFileSync(filePath, 'utf-8');
+		// Replace default binding text
+		let newValueLines = defaultBindedFunctionText.split(os.EOL);
+		const defaultLineSet = new Set(defaultSqlBindingTextLines);
```
Charles-Gagnon, line 81: Why not just make this a set initially so you don't have to create it every time?

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+		let replacedValueLines = [];
+		for (let defaultLine of newValueLines) {
+			// Skipped lines
+			if (defaultLineSet.has(defaultLine.trimStart())) {
```
Charles-Gagnon, line 85: Maybe name this array/set something like `defaultFunctionTextToSkip` - that's a lot clearer about what it's going to be used for.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+			functionName,
+			objectName,
+			constants.sqlConnectionString
+		);
+
+		this.overwriteAzureFunctionMethodBody(functionFile);
+	}
+
+	/**
+	 * Overwrites the Azure function file to include the new binding in the method body
```
Charles-Gagnon, line 74: This isn't really accurate - the binding is added in the method signature. This is just updating the method body to work with the binding and return the correct result.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+				replacedValueLines.push(defaultLine);
+			}
+		}
+		defaultBindedFunctionText = replacedValueLines.join(os.EOL);
+		fs.writeFileSync(filePath, defaultBindedFunctionText, 'utf-8');
+	}
+}
+
+/**
+ * A C# Azure Functions project must be present in order to create a new Azure Function for the table
```
Charles-Gagnon, line 100: Descriptions should say what a function is doing - this is more a general statement

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+					title: constants.selectProject,
+					ignoreFocusOut: true
+				}));
+				return selectedProjectFile;
+			} else if (count === 1) {
+				// only one azure function project found
+				return functionProjects[0];
+			}
+		}
+		return;
```
Charles-Gagnon, line 132: You're returning here but the function signature says it only returns a string, which isn't valid. You should either update the return type to be `string | undefined` (and have this be `return undefined`) or do something else like throw an error.

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+		defaultBindedFunctionText = replacedValueLines.join(os.EOL);
+		fs.writeFileSync(filePath, defaultBindedFunctionText, 'utf-8');
+	}
+}
+
+/**
+ * A C# Azure Functions project must be present in order to create a new Azure Function for the table
+ * @returns the selected project file path or indicates to user there is no C# Azure Functions project
+ */
+async function isAzureFunctionProjectOpen(): Promise<string> {
```
Charles-Gagnon, line 103: The function name implies it returns a boolean (`is...Open`  is typically for questions) 

Something like `getAzureFunctionProject` seems clearer (and then use the description to describe what happens if one doesn't exist or the user exits out of choosing)

This also seems like it should be in azureFunctionUtils

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+			} else if (count === 1) {
+				// only one azure function project found
+				return functionProjects[0];
+			}
+		}
+		return;
+	}
+}
+
+// gets the azure functions project file path
```
Charles-Gagnon, line 136: xml doc

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+async function getProjectFile(): Promise<string[] | undefined> {
```
Charles-Gagnon, line 137: `getProjectFiles`

But also i don't think we need this - we only care about getting project files that are next to a host.json so why not get all those first and then check just those folders instead of iterating over the entire workspace again?

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+	const projUris = await vscode.workspace.findFiles('**/*.csproj');
+	let projFiles: string[] = [];
+	for (let projUri of projUris) {
+		projFiles.push(projUri.fsPath);
+	}
+	return projFiles.length > 0 ? projFiles : undefined;
+}
+
+// gets the host file config file path
```
Charles-Gagnon, line 146: xml doc

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+async function getHostFile(): Promise<string[] | undefined> {
+	const hostUris = await vscode.workspace.findFiles('**/host.json');
+	let hostFiles = [];
+	for (let hostUri of hostUris) {
+		hostFiles.push(hostUri.fsPath);
+	}
+	return hostFiles.length > 0 ? hostFiles : undefined;
+}
+
+// gets the local.settings.json file path
```
Charles-Gagnon, line 156: xml doc

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+		);
+
+		this.overwriteAzureFunctionMethodBody(functionFile);
+	}
+
+	/**
+	 * Overwrites the Azure function file to include the new binding in the method body
+	 * @param filePath is the path for the function file (.cs for C# functions)
+	 */
+	private overwriteAzureFunctionMethodBody(filePath: string): void {
```
Charles-Gagnon, line 77: This doesn't appear to need to be in the class, suggest moving it to azureFunctionUtils

src/azureFunction/azureFunctionProjectService.ts
Charles-Gagnon: (and really - all these module level functions here should probably just be in `azureFunctionUtils`

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+		return selectedProjectFile;
+	} else {
+		const hostFiles = await getHostFile();
+		const projectFiles = await getProjectFile();
+		if (projectFiles !== undefined && hostFiles !== undefined) {
+			// Check to see if its only a function project using host.json and project file
+			let functionProjects = [];
+			let count = 0;
+			for (let host of hostFiles) {
+				for (let project of projectFiles) {
```
Charles-Gagnon, line 115: Easier and more readable to just do 

`const functionsProjects = projectFiles.filter(p => ...)`

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+	const projUris = await vscode.workspace.findFiles('**/*.csproj');
+	let projFiles: string[] = [];
+	for (let projUri of projUris) {
+		projFiles.push(projUri.fsPath);
+	}
+	return projFiles.length > 0 ? projFiles : undefined;
+}
+
+// gets the host file config file path
+async function getHostFile(): Promise<string[] | undefined> {
```
Charles-Gagnon, line 147: `getHostFiles`

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+	for (let hostUri of hostUris) {
+		hostFiles.push(hostUri.fsPath);
+	}
+	return hostFiles.length > 0 ? hostFiles : undefined;
+}
+
+// gets the local.settings.json file path
+async function getSettingsFile(): Promise<string | undefined> {
+	const settingsFiles = await vscode.workspace.findFiles('**/local.settings.json');
+	return settingsFiles.length > 0 ? settingsFiles[0].fsPath : undefined;
```
Charles-Gagnon, line 159: This is hardcoded as well to just use the first local.settings.json that it finds which isn't correct (especially now that we're picking the project from the list).

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// retrieves the new function file once the file is created
```
Charles-Gagnon, line 162: xml doc

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+function getNewFunctionFile(projectFile: string): Promise<string> {
```
Charles-Gagnon, line 163: Name would be better as `waitForNewFunctionFile` I think

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,186 @@
...
+		const watcher = vscode.workspace.createFileSystemWatcher((
+			path.dirname(projectFile), '**/*.cs'), false, true, true);
+		watcher.onDidCreate((e) => {
+			resolve(e.fsPath);
+			watcher.dispose();
+		});
+	});
+}
+
+// adds the required nuget package to the project
```
Charles-Gagnon, line 174: xml doc

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,87 @@
...
+ * @returns settings in local.settings.json. If no settings are found, returns default "empty" settings
+ */
+export async function getLocalSettingsJson(localSettingsPath: string): Promise<ILocalSettingsJson> {
+	if (await fse.pathExists(localSettingsPath)) {
+		const data: string = (await fse.readFile(localSettingsPath)).toString();
+		if (/[^\s]/.test(data)) {
+			try {
+				return JSON.parse(data);
+			} catch (error) {
+				throw new Error(constants.failedToParse(error.message));
```
Charles-Gagnon, line 38: Should log the original error to console in case it has some other information (such as the stack trace) that could be useful

src/controllers/mainController.ts
```diff
@@ -439,6 +442,19 @@ export default class MainController implements vscode.Disposable {
 					}
 				}
 			}));
+
+		const afService = new AzureFunctionProjectService(this.azureFunctionsService);
```
Charles-Gagnon, line 446: Would suggest making this up with the rest of the services....

(but also as mentioned earlier, why is there not just a single AzureFunctionsService? Two seems unnecessary

src/utils/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/**
+ * Generates a quoted full name for the object
+ * @param schema of the object
+ * @param objectName object chosen by the user
+ * @returns the quoted and escaped full name of the specified schema and object
+ */
+export function generateQuotedFullName(schema: string, objectName: string): string {
+	schema = schema.replace(/\]/g, ']]');
```
Charles-Gagnon, line 31: Utils already has a function to do the escaping (escapeClosingBrackets) so I'd just use that

```
```

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * copied and modified from vscode-azurefunctions extension
+ * https://github.com/microsoft/vscode-azurefunctions/blob/main/src/funcConfig/local.settings.ts
+ * @param localSettingsPath full path to local.settings.json
+ * @returns settings in local.settings.json. If no settings are found, returns default "empty" settings
+ */
+export async function getLocalSettingsJson(localSettingsPath: string): Promise<ILocalSettingsJson> {
+	if (await fse.pathExists(localSettingsPath)) {
+		const data: string = (await fse.readFile(localSettingsPath)).toString();
+		if (/[^\s]/.test(data)) {
```
Charles-Gagnon, line 34: Is this check actually necessary? JSON.parse doesn't seem to care if the content starts with a whitespace character.

src/azureFunction/azureFunctionUtils.ts
Charles-Gagnon: I'm not really sure what you mean by "putting a map of contents" @VasuBhog , but unless I'm missing something the reason is because we're writing the JSON and want to pretty-print it. But `JSON.stringify` takes a parameter for that so is there any reason we could just use fs.promises.writeFile(...) with the JSON.stringify content of settings?

(I realize that this is just copied code from AF but I agree with Chris that we should avoid extra dependencies unless we actually need them)

src/controllers/mainController.ts
```diff
@@ -146,14 +148,15 @@ export default class MainController implements vscode.Disposable {
 			this.registerCommand(Constants.cmdAadRemoveAccount);
 			this._event.on(Constants.cmdAadRemoveAccount, () => this.removeAadAccount(this._prompter));
 
+			this.azureFunctionsService = new AzureFunctionsService(SqlToolsServerClient.instance);
```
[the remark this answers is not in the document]
VasuBhog: Tested after moving back. This would cause an error when trying to add the SQL Binding - 
error: cannot addSqlBinding of undefined

package.json
```diff
@@ -116,14 +118,14 @@
...
   "dependencies": {
-    "@types/node": "^14.14.16",
     "@microsoft/ads-adal-library": "1.0.13",
     "core-js": "^2.4.1",
     "decompress-zip": "^0.2.2",
     "ejs": "^3.1.6",
     "error-ex": "^1.3.0",
     "figures": "^1.4.0",
     "find-remove": "1.2.1",
+    "fs-extra": "^10.0.0",
```
Charles-Gagnon, line 128: Do we need this now with the refactoring you did?

src/azureFunction/azureFunctionProjectService.ts
```diff
@@ -0,0 +1,66 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import * as vscode from 'vscode';
+import * as mssql from 'vscode-mssql';
+import * as LocalizedConstants from '../constants/localizedConstants';
+import { AzureFunctionsService } from '../services/azureFunctionsService';
+import * as azureFunctionUtils from '../azureFunction/azureFunctionUtils';
+import * as constants from '../constants/constants';
+import { generateQuotedFullName } from '../utils/utils';
+
+export class AzureFunctionProjectService {
```
Charles-Gagnon, line 13: The stuff here should really just be a part of the `AzureFunctionsService`

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,214 @@
...
+	return true;
+}
+
+export async function getAzureFunctionsExtensionApi(): Promise<AzureFunctionsExtensionApi | undefined> {
+	const apiProvider = await vscode.extensions.getExtension(constants.azureFunctionsExtensionName)?.activate() as AzureExtensionApiProvider;
+	const azureFunctionApi = apiProvider.getApi<AzureFunctionsExtensionApi>('*');
+	if (azureFunctionApi) {
+		return azureFunctionApi;
+	} else {
+		vscode.window.showErrorMessage(LocalizedConstants.azureFunctionsExtensionNotInstalled);
```
Charles-Gagnon, line 85: We should provide buttons that will either install the extension directly or bring them to the marketplace page (using commands)

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/**
+ * Gets the local.settings.json file path
+ * @param projectFile path of the azure function project
+ * @returns the local.settings.json file path
+ */
+export async function getSettingsFile(projectFile: string): Promise<string | undefined> {
+	let settingsFile = path.join(path.dirname(projectFile), 'local.settings.json');
+	return settingsFile.length > 0 ? settingsFile : undefined;
```
Charles-Gagnon, line 179: This check doesn't seem necessary - it shouldn't be possible to have an empty string here

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,214 @@
...
+	return projFiles.length > 0 ? projFiles : undefined;
+}
+
+/**
+ * Gets the host files from the workspace
+ * @returns the host file paths
+ */
+export async function getHostFiles(): Promise<string[] | undefined> {
+	const hostUris = await vscode.workspace.findFiles('**/host.json');
+	let hostFiles = [];
```
Charles-Gagnon, line 165: This can be simplified to 

`const hostFiles = hostUris.map(uri => uri.fsPath);`

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/**
+ * Retrieves the new function file once the file is created
+ * @param projectFile is the path to the project file
+ * @returns the function file path once created
+ */
+export function waitForNewFunctionFile(projectFile: string): Promise<string> {
+	return new Promise((resolve) => {
+		const watcher = vscode.workspace.createFileSystemWatcher((
```
Charles-Gagnon, line 189: Add a timeout reject here

src/constants/constants.ts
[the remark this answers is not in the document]
Charles-Gagnon: We should revisit this logic - we don't want this to break if the AF template changes what code it generates. That'll be easier to do if we move this to STS so probably just wait for it there.

src/constants/constants.ts
```diff
@@ -151,3 +154,33 @@ export const tenantDisplayName = 'Microsoft';
...
+		'string requestBody = await new StreamReader(req.Body).ReadToEndAsync();',
+		'dynamic data = JsonConvert.DeserializeObject(requestBody);',
+		'name = name ?? data?.name;',
+		'string responseMessage = string.IsNullOrEmpty(name) ? \"This HTTP triggered function executed successfully. Pass a name in the query string or in the request body for a personalized response.\" : $\"Hello, {name}. This HTTP triggered function executed successfully.\";'
+	];
+export const defaultBindingResult = 'return new OkObjectResult(responseMessage);';
+export const sqlBindingResult = `return new OkObjectResult(result);`;
+export const azureFunctionLocalSettingsFileName = 'local.settings.json';
+export const sqlExtensionPackageName = 'Microsoft.Azure.WebJobs.Extensions.Sql';
+export function jsonParseError(error: string, line: number, column: number): string {
```
Charles-Gagnon, line 174: There's some strings here that aren't used anymore, go through the new ones added and clean them up

src/controllers/mainController.ts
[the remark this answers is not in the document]
Charles-Gagnon: You need to make sure that things are done in order, it sounds like that wasn't being done correctly.

If you combine the two AF services as suggested then it should be easy to move this back to where it was.

src/utils/connectionStringUtil.ts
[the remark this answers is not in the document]
Charles-Gagnon: If we already have an active connection (because we're coming from OE) then why aren't you using the connection string API from STS?

Charles-Gagnon (commented):
This is pretty close to being ready to get in as a first draft. I think the main thing here is removing the fs-extra dependency and a few cleanup items. Here's the list of stuff that I think could be done as follow up PRs (I left comments for all of them too)

- Improve context menu string (specify input binding)
- AzureFunctionProjectService combine
- AF ext not installed buttons
- Wait for new file timeout
- Revisit method body replacement logic
- Move AF service init back to original location
- Use connection string API instead of hardcoding

package.json
VasuBhog: fs-extra is still used in serviceDownloaderProvider. If I remove that then I would also have to go and change this as well to use fs. (can be done in follow-up if necessary.

src/azureFunction/azureFunctionProjectService.ts
VasuBhog: Can look to move the stuff in follow-up PR - as there have been a ton of changes thus far.

src/azureFunction/azureFunctionUtils.ts
VasuBhog: Follow-up PR.

src/azureFunction/azureFunctionProjectService.ts
dzsquared: "Create Azure Function with SQL input binding"

src/utils/connectionStringUtil.ts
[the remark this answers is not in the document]
VasuBhog: Will do this in separate PR. I did not find a connection URI to then pass that into GetConnectiongString Request for STS.

VasuBhog:
@Charles-Gagnon @chlafreniere 

Follow PR(s) to address enhancing this PR:
- Fix breaking change in this PR: `AF service init back to the original location`
   - AzureFunctionProjectService combine 
- AF ext not installed buttons
- Revisit method body replacement logic
- Use connection string API instead of hardcoding

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -0,0 +1,215 @@
...
+ * Retrieves the new function file once the file is created
+ * @param projectFile is the path to the project file
+ * @returns the function file path once created
+ */
+export function waitForNewFunctionFile(projectFile: string): Promise<string> {
+	return new Promise((resolve, reject) => {
+		const watcher = vscode.workspace.createFileSystemWatcher((
+			path.dirname(projectFile), '**/*.cs'), false, true, true);
+		const timeout = setTimeout(async () => {
+			reject('Timed out waiting for azure function file creation');
```
Charles-Gagnon, line 188: This should actually be `reject(new Error('Timed out waiting for Azure Function file creation'));` (we should always throw errors, not strings directly)

And if the rejection is displayed to the user I would suggest localizing the string as well.

Charles-Gagnon (approved):

chlafreniere (approved):

PR #17219, opened 2022-01-21
Credentials with same behavior as ADS

abist:
Putting in the same behavior we have in ADS for Feb release as discussed. The other open PR is to be merged in a future release.

### DIFF
--- a/src/credentialstore/credentialstore.ts
+++ b/src/credentialstore/credentialstore.ts
@@ -37,12 +37,8 @@ export class CredentialStore implements ICredentialStore {
 		let self = this;
 		let cred: Contracts.Credential = new Contracts.Credential();
 		cred.credentialId = credentialId;
-		try {
-			const returnedCred = await self._client.sendRequest(Contracts.ReadCredentialRequest.type, cred)
-			return returnedCred;
-		} catch (error) {
-			throw error;
-		}
+		const returnedCred = await self._client.sendRequest(Contracts.ReadCredentialRequest.type, cred);
+		return returnedCred;
 	}
 
 	public async saveCredential(credentialId: string, password: any): Promise<boolean> {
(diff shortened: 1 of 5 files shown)


src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +33,50 @@ export class CredentialStore implements ICredentialStore {
...
-			self._client
-				.sendRequest(Contracts.ReadCredentialRequest.type, cred)
-				.then(returnedCred => {
-					resolve(returnedCred);
-				}, err => reject(err));
-		});
+		try {
+			const returnedCred = await self._client.sendRequest(Contracts.ReadCredentialRequest.type, cred)
+			return returnedCred;
+		} catch (error) {
```
Charles-Gagnon, line 43: You can just get rid of the try/catch if you're immediately rethrowing the error. This is doing nothing right now.

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +33,50 @@ export class CredentialStore implements ICredentialStore {
...
 		cred.password = password;
-		return new Promise<boolean>((resolve, reject) => {
-			self._client
-				.sendRequest(Contracts.SaveCredentialRequest.type, cred)
-				.then(status => {
-					resolve(status);
-				}, err => reject(err));
-		});
+		/* This is only done for linux because this is going to be
+		* the default credential system for linux in the next release
```
Charles-Gagnon, line 54: I would avoid using terms like "next release" in comments since that relies on the person having context of when "next" is.

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +33,50 @@ export class CredentialStore implements ICredentialStore {
...
+		*/
+		if (Utils.isLinux) {
+			await this._secretStorage.store(credentialId, password);
+		}
+		try {
+			const success = await self._client.sendRequest(Contracts.SaveCredentialRequest.type, cred);
+			return success;
+		} catch (error) {
```
Charles-Gagnon, line 62: Same, get rid of try/catch

src/credentialstore/credentialstore.ts
```diff
@@ -26,44 +33,50 @@ export class CredentialStore implements ICredentialStore {
...
-					resolve(status);
-				}, err => reject(err));
-		});
+		if (Utils.isLinux) {
+			await this._secretStorage.delete(credentialId);
+		}
+		try {
+			const success = await self._client.sendRequest(Contracts.DeleteCredentialRequest.type, cred);
+			return success;
+		} catch (error) {
```
Charles-Gagnon, line 77: Remove try/catch

src/models/utils.ts
```diff
@@ -516,3 +516,9 @@ export function deepClone<T>(obj: T): T {
 	});
 	return result;
 }
+
+/**
+ * Returns whether the credentials should use the native credential
```
Charles-Gagnon, line 521: This comment doesn't make sense for something as generic as this - just because it's used for that purpose in one place doesn't mean it's the only way it's going to be used. I'd just remove the comment (the name alone is pretty clear what it means)

Charles-Gagnon:
Run `gulp lint` and fix the linting errors

Charles-Gagnon (approved):

alanrenmsft (approved):

PR #17226, opened 2022-01-31
Remove fs-extra-promise and fs-extra

Charles-Gagnon:
Noticed in another PR that we had this package - but it's really not necessary and so removing it and replacing it with the built-in fs module instead.


alanrenmsft (approved):

Charles-Gagnon:
@alanrenmsft Resetting since I missed some more that involved some rewriting functions to be async

alanrenmsft (approved):

abist (approved):

PR #17229, opened 2022-02-01
Improve connection string and empty workspace folder case

VasuBhog:
Fixes https://github.com/microsoft/vscode-mssql/pull/17217 - connection string to use STS request instead

Also found a case where if there is a workspace and no azure function project is found it would give an incorrect error of unable to iterate over hostFiles. Now it indicates to the user to create a C# function first.


src/controllers/mainController.ts
```diff
@@ -443,14 +442,9 @@ export default class MainController implements vscode.Disposable {
...
-			const database = ObjectExplorerUtils.getDatabaseName(node);
-			const connStr = getConnectionString(
-				node.connectionInfo.server,
-				node.connectionInfo.authenticationType,
-				database,
-				node.connectionInfo.user,
-				node.connectionInfo.password);
-			await this.azureFunctionsService.createAzureFunction(connStr, node.metadata.schema, node.metadata.name);
+			const connectionUri = this._connectionMgr.getUriForConnection(node.connectionInfo);
+			let connectionString = await this._connectionMgr.getConnectionString(connectionUri);
```
chlafreniere, line 446: nit: can this be const?

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -149,11 +149,14 @@ export async function getAzureFunctionProject(): Promise<string | undefined> {
 export async function getAzureFunctionProjectFiles(): Promise<string[] | undefined> {
 	let projFiles: string[] = [];
 	const hostFiles = await getHostFiles();
-	for (let host of hostFiles) {
-		let projectFile = await vscode.workspace.findFiles('*.csproj', path.dirname(host));
-		projectFile.filter(file => path.dirname(file.fsPath) === path.dirname(host) ? projFiles.push(file?.fsPath) : projFiles);
+	if (hostFiles !== undefined) {
```
chlafreniere, line 152: This is honestly super minor, but alternatively we could do:

```
if (!hostFiles) {
    return undefined;
}
```

...and then leave the rest the same. Just would touch fewer lines and involve less scrolling.

Honestly, it's more of a style thing than anything. Feel completely empowered to ignore me, just throwing it out there 😄

chlafreniere (approved):

src/azureFunction/azureFunctionUtils.ts
VasuBhog: All good, this is a better solution. I changed it to this way originally just due to debugging.

PR #17230, opened 2022-02-01
Prompt if Azure Function Extension not installed 

VasuBhog:
Follow up from https://github.com/microsoft/vscode-mssql/pull/17217.

This creates a prompt to allow users to install the azure function via a link. Wording can be updated
Screen Shot 2022-01-31 at 11 07 40 PM


src/azureFunction/azureFunctionUtils.ts
```diff
@@ -78,6 +78,14 @@ export async function setLocalAppSetting(projectFolder: string, key: string, val
 
 export async function getAzureFunctionsExtensionApi(): Promise<AzureFunctionsExtensionApi | undefined> {
 	const apiProvider = await vscode.extensions.getExtension(constants.azureFunctionsExtensionName)?.activate() as AzureExtensionApiProvider;
+	if (!apiProvider) {
+		const response = await vscode.window.showInformationMessage(constants.azureFunctionsExtensionNotFound,
+			constants.installAzureFunction, constants.doNotInstall);
+		if (response === constants.installAzureFunction) {
+			await vscode.env.openExternal(vscode.Uri.parse(constants.linkToAzureFunctionExtension));
```
Charles-Gagnon, line 85: I think it would be better to have 3 buttons

1. `Install` - directly installs extension by calling the `workbench.extensions.installExtension` and passing it the extension ID. I don't believe that a progress dialog is shown while it's installing so it'd be good to wrap that in a vscode.window.withProgress call as well
2. `View Gallery` - opens gallery, ideally to the actual page for the extension. Look at the `workbench.extensions.search` command to see if that'll work, there might be another command you can use too if that doesn't fit our needs
3. `Do not install` - same as what you have

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -78,6 +78,14 @@ export async function setLocalAppSetting(projectFolder: string, key: string, val
...
+		}
+		return undefined;
```
Charles-Gagnon, line 87: If we install the extension you should await the command to install it and then be able to call activate again to get the API.

Although you should check to see if the AF extension requires a vs code restart. If it does then you should indicate that in the dialog and call the command to reload the window after it's done installing.

src/constants/constants.ts
```diff
@@ -182,3 +182,8 @@ export const yesString = localize('yesString', 'Yes');
 export const functionNameTitle = localize('functionNameTitle', 'Function Name');
 export const selectProject = localize('selectProject', 'Select the Azure Function project for the SQL Binding');
 export const timeoutError = localize('timeoutError', 'Timed out waiting for azure function file creation');
+export const azureFunctionsExtensionNotFound = localize('azureFunctionsExtensionNotFound', 'Azure Functions extension is not installed. Install the extension to use the Azure Functions binding.');
+export const installAzureFunction = localize('install', 'Link to install');
+export const linkToAzureFunctionExtension = localize('linkToAzureFunction', 'https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azurefunctions');
```
Charles-Gagnon, line 187: You should never localize links

Charles-Gagnon (changes requested):

VasuBhog:
Updated the Prompt:
image

Install directly installs az fn and then begins to prompt the user the steps of "Create Azure function with SQL input binding" or will tell them they need a C# azure function project first.
Learn More directs to the link: https://docs.microsoft.com/en-us/azure/azure-functions/functions-develop-vs-code
Do not install: Does what it says and just closes prompt

cc @Charles-Gagnon and @dzsquared

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -76,8 +76,48 @@ export async function setLocalAppSetting(projectFolder: string, key: string, val
...
+				{
+					location: vscode.ProgressLocation.Notification,
+					title: constants.azureFunctionsExtensionName,
+					cancellable: false
+				}, async (_progress, _token) => {
+					await vscode.commands.executeCommand('workbench.extensions.installExtension', constants.azureFunctionsExtensionName);
+				}
+			);
+			// the extension has not been notified that the extension is installed so wait till it is to then activate it
+			await new Promise<void>((resolve, reject) => {
```
Charles-Gagnon, line 100: You should create (but not await) this promise before the `withProgress` call so that the event is hooked up in case there's a race condition here and we end up having the extension host updated before the withProgress returns. 

(and then await it down here)

src/constants/constants.ts
```diff
@@ -182,3 +182,9 @@ export const yesString = localize('yesString', 'Yes');
 export const functionNameTitle = localize('functionNameTitle', 'Function Name');
 export const selectProject = localize('selectProject', 'Select the Azure Function project for the SQL Binding');
 export const timeoutError = localize('timeoutError', 'Timed out waiting for azure function file creation');
+export const azureFunctionsExtensionNotFound = localize('azureFunctionsExtensionNotFound', 'The Azure Functions extension is required to create a new Azure Function with SQL binding but is not installed, install it now?');
+export const installAzureFunction = localize('install', 'Install');
+export const learnMore = localize('learnMore', 'Learn more');
+export const linkToAzureFunctionExtension = 'https://docs.microsoft.com/azure/azure-functions/functions-develop-vs-code';
```
chlafreniere, line 188: nit: please organize such that unlocalized constants aren't in the middle of localized ones

Charles-Gagnon (approved):

PR #17231, opened 2022-02-01
vBump STS to 1.9.5 & add optional application name parameter for connection string

### LINKED ISSUE #17221: mssql: Execution failed: Could not find the resource "Microsoft.SqlTools.ManagedBatchParser.Localization.SR.resources" among the resources "Microsoft.SqlTools.ManagedBatchParser.SR.resources" embedded in the assembly "Microsoft.SqlTools.ManagedBatchParser", nor among the resources in any satellite assemblies for the specified culture. Perhaps the resources were embedded with an incorrect name.
- MSSQL Extension Version: 1.12.0
- VSCode Version: 1.63
- OS Version: 20H2

Steps to Reproduce:

1.open a .tsql extension file
2.use "execute query" on a query

VasuBhog:
STS PR: https://github.com/microsoft/sqltoolsservice/pull/1380

The goal is to make the application name in the connection string be optional to include. 

Only when is it set to false will it not be set.

[Edit by @smartguest ] Also fixes #17221

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -265,10 +265,11 @@ export default class ConnectionManager {
 	 * @param includePassword (optional) if password should be included in connection string.
 	 * @returns connection string for the connection
 	 */
-	public async getConnectionString(connectionUri: string, includePassword: boolean = false): Promise<string> {
+	public async getConnectionString(connectionUri: string, includePassword: boolean = false, includeApplicationName: boolean = true): Promise<string> {
 		const listParams = new ConnectionContracts.GetConnectionStringParams();
 		listParams.ownerUri = connectionUri;
 		listParams.includePassword = includePassword;
+		listParams.includeApplicationName = includeApplicationName;
 		return this.client.sendRequest(ConnectionContracts.GetConnectionStringRequest.type, listParams);
 	}
 
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -263,6 +263,7 @@ export default class ConnectionManager {
 	 * Get the connection string for the provided connection Uri
 	 * @param connectionUri The connection Uri for the connection.
 	 * @param includePassword (optional) if password should be included in connection string.
+	 * @param includeApplicationName (optional) if application name should be included in connection string.
 	 * @returns connection string for the connection
 	 */
 	public async getConnectionString(connectionUri: string, includePassword: boolean = false, includeApplicationName: boolean = true): Promise<string> {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -62,7 +62,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<IExten
 		dacFx: controller.dacFxService,
 		schemaCompare: controller.schemaCompareService,
 		azureFunctions: controller.azureFunctionsService,
-		getConnectionString: (connectionUri: string, includePassword: boolean, includeApplicationName: boolean) => {
+		getConnectionString: (connectionUri: string, includePassword?: boolean, includeApplicationName?: boolean) => {
 			return controller.connectionManager.getConnectionString(connectionUri, includePassword, includeApplicationName);
 		}
 	};
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -62,8 +62,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<IExten
 		dacFx: controller.dacFxService,
 		schemaCompare: controller.schemaCompareService,
 		azureFunctions: controller.azureFunctionsService,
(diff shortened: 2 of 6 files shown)


src/controllers/connectionManager.ts
```diff
@@ -265,10 +265,11 @@ export default class ConnectionManager {
 	 * @param includePassword (optional) if password should be included in connection string.
 	 * @returns connection string for the connection
 	 */
-	public async getConnectionString(connectionUri: string, includePassword: boolean = false): Promise<string> {
+	public async getConnectionString(connectionUri: string, includePassword: boolean = false, includeApplicationName: boolean = true): Promise<string> {
```
chlafreniere, line 268: What's the plan for actually updating the version of STS?

VasuBhog: I would like to try and get my other connectionString improvements in as well prior to updating the version of STS.

typings/vscode-mssql.d.ts
```diff
@@ -82,7 +82,7 @@ declare module 'vscode-mssql' {
 		 * @param includePassword to include password with connection string
 		 * @returns connection string
 		 */
-		getConnectionString(connectionUri: String, includePassword?: boolean): Promise<string>;
+		getConnectionString(connectionUri: String, includePassword?: boolean, includeApplicationName?: boolean): Promise<string>;
```
Charles-Gagnon, line 85: Update JS doc params as well

src/models/contracts/connection.ts
```diff
@@ -276,6 +276,11 @@ export class GetConnectionStringParams {
 	 * Indicates whether to include the password in the connection string
 	 */
 	public includePassword?: boolean;
+
+	/**
+	 * Indicates whether to include the password in the connection string
```
Charles-Gagnon, line 281: Specify what the default value is

src/extension.ts
```diff
@@ -62,8 +62,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<IExten
 		dacFx: controller.dacFxService,
 		schemaCompare: controller.schemaCompareService,
 		azureFunctions: controller.azureFunctionsService,
-		getConnectionString: (connectionUri: string, includePassword: boolean) => {
-			return controller.connectionManager.getConnectionString(connectionUri, includePassword);
+		getConnectionString: (connectionUri: string, includePassword: boolean, includeApplicationName: boolean) => {
```
Charles-Gagnon, line 65: includeApplicationName should be optional here

typings/vscode-mssql.d.ts
```diff
@@ -80,9 +80,10 @@ declare module 'vscode-mssql' {
 		 * Get the connection string for the provided connection Uri
 		 * @param connectionUri The URI of the connection to get the connection string for.
 		 * @param includePassword to include password with connection string
+		 * @param includeApplicationName to include application name with connection string
```
Charles-Gagnon, line 83: `Whether the Application Name is included in the connection string. Default is true`

src/controllers/connectionManager.ts
Charles-Gagnon: Are you implying you're going to merge this in before updating STS? Generally we shouldn't do that - once there's a build of STS with your changes you should include it in this PR and then merge it in.

Charles-Gagnon (changes requested):

src/controllers/connectionManager.ts
VasuBhog: @Charles-Gagnon I'll follow that approach so that we can get these changes into the current insiders.

Charles-Gagnon: What approach? If you merge it in now the STS changes aren't going to be there so this won't actually do anything until STS is updated to a version that has your changes.

VasuBhog: The approach you stated in your previous comment. Merge STS, update version of STS, and then use the latest STS version. I don't see any docs on how to go about that. Is there a standard process for it?

Charles-Gagnon: https://github.com/microsoft/azuredatastudio/pull/18212

VasuBhog:
Updating to latest STS 
to 1.95 from 1.74: 
image
image

cc @alanrenmsft, @smartguest, @kburtram, @aasimkhan30, and @Charles-Gagnon

src/controllers/connectionManager.ts
Charles-Gagnon, line 268: Update doc params

Charles-Gagnon (approved):

alanrenmsft (approved):

smartguest (approved):

smartguest:
This will also fix #17221 as well with the bump

PR #17232, opened 2022-02-08
Optional condition to use the Connection Info for GetConnectionString

VasuBhog:
PR from STS: https://github.com/microsoft/sqltoolsservice/pull/1394

This PR addresses the disconnected connections. 

This will enable SQL database projects to not have to connect to servers to retrieve the connection string.

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -260,18 +260,18 @@ export default class ConnectionManager {
 	}
 
 	/**
-	 * Get the connection string for the provided connection Uri
-	 * @param connectionUri The connection Uri for the connection.
-	 * @param connectionInfo The connection info for the connection.
+	 * Get the connection string for the provided connection Uri or ConnectionDetails.
+	 * @param connectionUri The connection Uri for the connection. If undefined then a ConnectionDetails should be specified.
+	 * @param connectionDetails The connection details for the connection. If undefined then a connection URI should be specified.
 	 * @param includePassword (optional) if password should be included in connection string.
 	 * @param includeApplicationName (optional) if application name should be included in connection string.
 	 * @returns connection string for the connection
 	 */
-	public async getConnectionString(connectionUri?: string, connectionInfo?: IConnectionInfo,
+	public async getConnectionString(connectionUri?: string, connectionDetails?: ConnectionContracts.ConnectionDetails,
 		includePassword: boolean = false, includeApplicationName: boolean = true): Promise<string> {
 		const listParams = new ConnectionContracts.GetConnectionStringParams();
 		listParams.ownerUri = connectionUri;
-		listParams.connectionInfo = connectionInfo;
+		listParams.connectionDetails = connectionDetails;
 		listParams.includePassword = includePassword;
 		listParams.includeApplicationName = includeApplicationName;
 		return this.client.sendRequest(ConnectionContracts.GetConnectionStringRequest.type, listParams);
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -262,13 +262,16 @@ export default class ConnectionManager {
 	/**
 	 * Get the connection string for the provided connection Uri
 	 * @param connectionUri The connection Uri for the connection.
+	 * @param connectionDetails The connection info for the connection.
 	 * @param includePassword (optional) if password should be included in connection string.
 	 * @param includeApplicationName (optional) if application name should be included in connection string.
 	 * @returns connection string for the connection
 	 */
-	public async getConnectionString(connectionUri: string, includePassword: boolean = false, includeApplicationName: boolean = true): Promise<string> {
+	public async getConnectionString(connectionUri?: string, connectionDetails?: IConnectionInfo,
+		includePassword: boolean = false, includeApplicationName: boolean = true): Promise<string> {
 		const listParams = new ConnectionContracts.GetConnectionStringParams();
 		listParams.ownerUri = connectionUri;
+		listParams.connectionDetails = connectionDetails;
 		listParams.includePassword = includePassword;
 		listParams.includeApplicationName = includeApplicationName;
 		return this.client.sendRequest(ConnectionContracts.GetConnectionStringRequest.type, listParams);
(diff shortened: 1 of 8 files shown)


src/controllers/connectionManager.ts
```diff
@@ -262,13 +262,16 @@ export default class ConnectionManager {
 	/**
 	 * Get the connection string for the provided connection Uri
 	 * @param connectionUri The connection Uri for the connection.
+	 * @param connectionDetails The connection info for the connection.
 	 * @param includePassword (optional) if password should be included in connection string.
 	 * @param includeApplicationName (optional) if application name should be included in connection string.
 	 * @returns connection string for the connection
 	 */
-	public async getConnectionString(connectionUri: string, includePassword: boolean = false, includeApplicationName: boolean = true): Promise<string> {
+	public async getConnectionString(connectionUri?: string, connectionDetails?: IConnectionInfo,
```
Charles-Gagnon, line 270: You should have the first parameter be

`connectionContext: string | IConnectionInfo`

where you can pass it either a URI string or info object. Easier to use on the typescript side.

Charles-Gagnon (changes requested):

src/controllers/connectionManager.ts
```diff
@@ -260,15 +260,18 @@ export default class ConnectionManager {
...
-	 * @param connectionUri The connection Uri for the connection.
+	 * Get the connection string for the provided connection Uri or ConnectionDetails.
+	 * @param connectionUri The connection Uri for the connection. If undefined then a ConnectionDetails should be specified.
+	 * @param connectionDetails The connection details for the connection. If undefined then a connection URI should be specified.
 	 * @param includePassword (optional) if password should be included in connection string.
 	 * @param includeApplicationName (optional) if application name should be included in connection string.
 	 * @returns connection string for the connection
 	 */
-	public async getConnectionString(connectionUri: string, includePassword: boolean = false, includeApplicationName: boolean = true): Promise<string> {
+	public async getConnectionString(connectionUri?: string, connectionDetails?: ConnectionContracts.ConnectionDetails,
```
Charles-Gagnon, line 270: It'd be better to make it `connectionUriOrDetails: string | ConnectionContracts.ConnectionDetails` - less chance of someone calling this incorrectly by not passing in either a URI or details

src/extension.ts
```diff
@@ -62,8 +63,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<IExten
 		dacFx: controller.dacFxService,
 		schemaCompare: controller.schemaCompareService,
 		azureFunctions: controller.azureFunctionsService,
-		getConnectionString: (connectionUri: string, includePassword?: boolean, includeApplicationName?: boolean) => {
-			return controller.connectionManager.getConnectionString(connectionUri, includePassword, includeApplicationName);
+		getConnectionString: (connectionUri?: string,
```
Charles-Gagnon, line 66: This should also be `connectionUriOrDetails: string | ConnectionDetails`

src/models/contracts/connection.ts
```diff
@@ -270,7 +270,12 @@ export class GetConnectionStringParams {
 	/**
 	 * Connection key to lookup connection string for the provided connection Uri
 	 */
-	public ownerUri: string;
+	public ownerUri?: string;
```
Charles-Gagnon, line 273: Add note here and in the details about requiring one or the other

typings/vscode-mssql.d.ts
```diff
@@ -77,13 +77,13 @@ declare module 'vscode-mssql' {
...
-		 * @param includeApplicationName Whether the Application Name is included in the connection string. Default is true
-		 * @returns connection string
+		 * Get the connection string for the provided connection Uri or connection details.
+		 * @param connectionUriOrDetails Either the connection Uri for the connection or the connection details for the connection is required.
+		 * @param includePassword (optional) if password should be included in connection string.
+		 * @param includeApplicationName (optional) if application name should be included in connection string.
+		 * @returns connection string for the connection
 		 */
-		getConnectionString(connectionUri: String, includePassword?: boolean, includeApplicationName?: boolean): Promise<string>;
+		getConnectionString(connectionUriOrDetails: string | { [name: string]: any }, includePassword?: boolean, includeApplicationName?: boolean): Promise<string>;
```
Charles-Gagnon, line 86: Would suggest moving the type definition into here if we're going to do this.

(we could also just not add it to the extension API for now. It isn't needed AFAIK)

Charles-Gagnon (changes requested):
This needs the STS update as well right?

typings/vscode-mssql.d.ts
VasuBhog: I tried adding ConnectionContracts.ConnectionDetails but was having trouble importing that into this file. Also, this is needed to be added so sql-database-projects can utilize it for disconnected scenarios.

VasuBhog:
> This needs the STS update as well right?

Should I add this to this PR or open a new one?

Charles-Gagnon:
This one is fine. Just go through the same process of verifying that changes in the update are OK to get in.

typings/vscode-mssql.d.ts
Charles-Gagnon: What "problems"? Feel free to reach out if you're having issues with stuff like this

VasuBhog:
Updating to latest STS to 2.05 from 1.95

Screen Shot 2022-02-16 at 2 52 54 PM

cc @smartguest, @Charles-Gagnon, @alanrenmsft

Charles-Gagnon (approved):
This connection details interface is pretty odd, I don't know why we aren't just using ConnectionInfo or any of the other usual types of connection objects. But this is fine for now...

alanrenmsft (approved):

VasuBhog:
> This connection details interface is pretty odd, I don't know why we aren't just using ConnectionInfo or any of the other usual types of connection objects. But this is fine for now...

I remember trying to send the getConnectionString with the IConnectionInfo and it would not as you mentioned in your previous commit/discussion. I can look at this again why its unable to parse the IConnectionInfo that we sent, but for now this works and helps unblock disconnected scenarios.

alanrenmsft:
@VasuBhog is this for Feb release? if so you have to go through the process.

VasuBhog:
> @VasuBhog is this for Feb release? if so you have to go through the process.

this is not for the Feb release.

PR #17233, opened 2022-02-08
Add task to generate package.nls.json from English package xlf

smartguest:
This change adds in a gulp task that generates package.nls.json from the English XLF files, and also fixes the file paths of the localized package.nls.json files. This makes it so we are able to generate the package.nls.json files from the localized package XLF file (already the localizedConstants.ts file is generated this way). It is a separate task from the package.nls.*.json generation task because that one relies on using package.nls.json to fill in strings that are missing from the localized XLF file.

However, you will need to run build twice due to a small bug where if you make a change to the English XLF, it will only show up in the package.nls.json file and not in the localized one unless you run the gulp task again, but this is probably due to limitations with gulp/file writing and is a minor thing. 

I've tested both package.nls.json and the localized constants file (which was already in the code) and it seems to work for both. 

This will require adding strings that are currently in package.nls.json and not in the XLF to the XLF itself afterwards.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -236,7 +236,7 @@ gulp.task('ext:copy-js', () => {
 // Copy the files which aren't used in compilation
 gulp.task('ext:copy', gulp.series('ext:copy-tests', 'ext:copy-js', 'ext:copy-config', 'ext:copy-systemjs-config', 'ext:copy-dependencies', 'ext:copy-html', 'ext:copy-css', 'ext:copy-images'));
 
-gulp.task('ext:localization', gulp.series('ext:localization:xliff-to-ts', 'ext:localization:xliff-to-json', 'ext:localization:xliff-to-package.nls'));
+gulp.task('ext:localization', gulp.series('ext:localization:generate-eng-package.nls', 'ext:localization:xliff-to-ts', 'ext:localization:xliff-to-json', 'ext:localization:xliff-to-package.nls'));
 
 gulp.task('ext:build', gulp.series('ext:localization', 'ext:copy', 'ext:clean-library-ts-files', 'ext:compile', 'ext:compile-view')); // removed lint before copy
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -236,6 +236,7 @@ gulp.task('ext:copy-js', () => {
 // Copy the files which aren't used in compilation
 gulp.task('ext:copy', gulp.series('ext:copy-tests', 'ext:copy-js', 'ext:copy-config', 'ext:copy-systemjs-config', 'ext:copy-dependencies', 'ext:copy-html', 'ext:copy-css', 'ext:copy-images'));
 
+// Localization (or build) task must be run twice separately to update localized package.nls file after the main English one is updated.
 gulp.task('ext:localization', gulp.series('ext:localization:generate-eng-package.nls', 'ext:localization:xliff-to-ts', 'ext:localization:xliff-to-json', 'ext:localization:xliff-to-package.nls'));
 
 gulp.task('ext:build', gulp.series('ext:localization', 'ext:copy', 'ext:clean-library-ts-files', 'ext:compile', 'ext:compile-view')); // removed lint before copy
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -236,7 +236,6 @@ gulp.task('ext:copy-js', () => {
 // Copy the files which aren't used in compilation
 gulp.task('ext:copy', gulp.series('ext:copy-tests', 'ext:copy-js', 'ext:copy-config', 'ext:copy-systemjs-config', 'ext:copy-dependencies', 'ext:copy-html', 'ext:copy-css', 'ext:copy-images'));
 
-// Localization (or build) task must be run twice separately to update localized package.nls file after the main English one is updated.
 gulp.task('ext:localization', gulp.series('ext:localization:generate-eng-package.nls', 'ext:localization:xliff-to-ts', 'ext:localization:xliff-to-json', 'ext:localization:xliff-to-package.nls'));
 
 gulp.task('ext:build', gulp.series('ext:localization', 'ext:copy', 'ext:clean-library-ts-files', 'ext:compile', 'ext:compile-view')); // removed lint before copy
--- a/tasks/localizationtasks.js
+++ b/tasks/localizationtasks.js
@@ -244,8 +244,9 @@ gulp.task('ext:localization:xliff-to-package.nls', function () {
 				file.basename = 'package.nls.' + iso639_3_to_2[language] + '.json';
 			}
 
-			// Make the new file create on root
-			file.dirname = file.dirname.replace(language, '').replace('xliff', '').replace('localization', '');
+			// Make the new file create on root.
+			// the original directory path was 'vscode-mssql\localization\xliff\<lang>'
+			file.dirname = file.dirname.replace(language, '').replace('localization', '').replace('xliff', '');
 
 			// callback to notify we have completed the current file
 			callback(null, file);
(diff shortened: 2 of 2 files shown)


Charles-Gagnon:
This is an improvement, but I'm not sure I understand why we'd need to run the task twice. If you're doing this work it seems worth it to spend a bit more time to make it so that it all gets done the first time.

smartguest:
@Charles-Gagnon basically the first time you run the task, it updates the package.nls.json file first, but it never updates the localized package.nls.*.json files for some reason (not sure why the file writing is delayed here). The task for the localized package.nls.*.json files relies on the main package.nls.json file for any missing strings, so it seems to not find it there. Any tips on how I can delay the next gulp task?

Charles-Gagnon:
Can you point to where you're seeing this problem?

smartguest:
Updated, it turns out that the localized package.nls task was using outdated resources before the package.nls.json generation task was run, now it uses the resources received when the task itself is run (after the package.nls.json generation task)

tasks/localizationtasks.js
```diff
@@ -149,12 +148,62 @@ gulp.task('ext:localization:xliff-to-ts', function () {
...
+
 // Generates a localized package.nls.*.json
 gulp.task('ext:localization:xliff-to-package.nls', function () {
-	return gulp.src([config.paths.project.localization + '/xliff/**/localizedPackage.json.*.xlf', '!' + config.paths.project.localization + '/xliff/enu/localizedPackage.json.enu.xlf'], { base: '' })
+	return gulp.src([config.paths.project.localization + '/xliff/**/localizedPackage.json.*.xlf', '!' + config.paths.project.localization + '/xliff/enu/localizedPackage.json.enu.xlf'], { base: '.' })
 		.pipe(through.obj(function (file, enc, callback) {
 			// convert xliff into json document
 			let dict = convertXmlToDictionary(String(file.contents), false);
+			// Get the latest changes from package.nls.json after being updated.
+			var packageAllKeys = require('./../package.nls.json')
```
Charles-Gagnon, line 206: Yeah, was going to point this out as the likely culprit here. I assume it's working better now? Everything else looks fine for what it's doing.

gulpfile.js
```diff
@@ -236,7 +236,8 @@ gulp.task('ext:copy-js', () => {
 // Copy the files which aren't used in compilation
 gulp.task('ext:copy', gulp.series('ext:copy-tests', 'ext:copy-js', 'ext:copy-config', 'ext:copy-systemjs-config', 'ext:copy-dependencies', 'ext:copy-html', 'ext:copy-css', 'ext:copy-images'));
 
-gulp.task('ext:localization', gulp.series('ext:localization:xliff-to-ts', 'ext:localization:xliff-to-json', 'ext:localization:xliff-to-package.nls'));
+// Localization (or build) task must be run twice separately to update localized package.nls file after the main English one is updated.
```
Charles-Gagnon, line 239: Remove this comment now

tasks/localizationtasks.js
```diff
@@ -149,12 +148,62 @@ gulp.task('ext:localization:xliff-to-ts', function () {
...
+				}
+				let instantiation = '"' + key + '":"' + value + '"';
+				contents.push(instantiation);
+			});
+
+			// end the function
+			contents.push('}');
+
+			// Join with new lines in between
+			let fullFileContents = contents.join('\r\n') + '\r\n';
```
Charles-Gagnon, line 181: You should use os.EOL here

tasks/localizationtasks.js
```diff
@@ -149,12 +148,62 @@ gulp.task('ext:localization:xliff-to-ts', function () {
...
 
+// Generates a package.nls.json file from localizedPackage.json.enu.xlf
+gulp.task('ext:localization:generate-eng-package.nls', function () {
+	return gulp.src([config.paths.project.localization + '/xliff/enu/localizedPackage.json.enu.xlf'], { base: '.' })
+		.pipe(through.obj(function (file, enc, callback) {
+			// convert xliff into json document
+			let dict = convertXmlToDictionary(String(file.contents), false);
+
+			var contents = ['{'];
+			var regxForReplacingQuots = new RegExp('"', 'g');
```
Charles-Gagnon, line 159: Quotes*

tasks/localizationtasks.js
```diff
@@ -149,12 +148,62 @@ gulp.task('ext:localization:xliff-to-ts', function () {
...
+					value = dict[key]['source'];
+				}
+				if (value && value.indexOf('"') >= 0) {
+					value = value.replace(regxForReplacingQuots, '\'');
+				}
+				let instantiation = '"' + key + '":"' + value + '"';
+				contents.push(instantiation);
+			});
+
+			// end the function
```
Charles-Gagnon, line 177: What function? This is just a dictionary (or object)

Not sure this needs a comment regardless

tasks/localizationtasks.js
smartguest: Yep, it's working better

tasks/localizationtasks.js
```diff
@@ -202,7 +245,7 @@ gulp.task('ext:localization:xliff-to-package.nls', function () {
 			}
 
 			// Make the new file create on root
-			file.dirname = file.dirname.replace(language, '');
+			file.dirname = file.dirname.replace(language, '').replace('xliff', '').replace('localization', '');
```
alanrenmsft, line 248: please add a comment with the original directory name format

tasks/localizationtasks.js
```diff
@@ -149,15 +144,64 @@ gulp.task('ext:localization:xliff-to-ts', function () {
...
+	return gulp.src([config.paths.project.localization + '/xliff/enu/localizedPackage.json.enu.xlf'], { base: '.' })
+		.pipe(through.obj(function (file, enc, callback) {
+			// convert xliff into json document
+			let dict = convertXmlToDictionary(String(file.contents), false);
+
+			var contents = ['{'];
+			var regxForReplacingQuotes = new RegExp('"', 'g');
+
+			// Get all keys and values from localizedPackage.json.enu.xlf
+			Object.keys(dict).forEach(key => {
```
Charles-Gagnon, line 158: This dictionary stuff appears to be duplicated below - would suggest looking into making a helper function for this as well

tasks/localizationtasks.js
```diff
@@ -149,15 +144,64 @@ gulp.task('ext:localization:xliff-to-ts', function () {
 		.pipe(gulp.dest(config.paths.project.root + '/src/constants/'));
 });
 
+// Generates a package.nls.json file from localizedPackage.json.enu.xlf
```
Charles-Gagnon, line 147: It would be nice to give a brief description of how it accomplishes that. i.e. what are the input files, and what does it do to them to get the output files? (that can be done as more thorough comments in the function itself too)

Charles-Gagnon (approved):
Now that you've done this - could you do a similar thing for ADS so that when string updates and made the original PR will update the xlf as well? Save you having to go and manually do it for every file updated.

smartguest:
@alanrenmsft That is correct, with the current localization structure in this extension, you need to make the changes in the XLF files first.

alanrenmsft (approved):

tasks/localizationtasks.js
smartguest: There are subtle differences between the two dictionary loops, but I'll try to implement something to consolidate it

smartguest:
> Now that you've done this - could you do a similar thing for ADS so that when string updates and made the original PR will update the xlf as well? Save you having to go and manually do it for every file updated.

The ADS code already grabs any changes made recently to the strings and update the English XLFs accordingly when the localization task is run, it's pretty much a single step for me.

Charles-Gagnon:
Yes, but it's a task that you constantly have to be doing pretty much every single day. Having it instead be something that's generated as people are making changes seems like a much better way of handling that and less error prone (what happens if you leave/forget to run the task for example)

PR #17235, opened 2022-02-10
Allows for both azure function with input or output binding 

VasuBhog:
Prompt the user for which type of binding a user would like to do instead of only allowing input binding.

Not sure why the package.nls.json was entirely updated when I only changed 
`"mssql.createAzureFunction":"Create Azure Function with SQL input binding",` to
`"mssql.createAzureFunction":"Create Azure Function with SQL binding",`.
Seems to be due to this PR: https://github.com/microsoft/vscode-mssql/pull/17233

Input binding resulting file:
image

Output binding resulting file: 
image


src/services/azureFunctionsService.ts
```diff
@@ -94,13 +94,35 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
+				label: constants.input,
+				type: mssql.BindingType.input
+			},
+			{
+				label: constants.output,
+				type: mssql.BindingType.output
+			}
+		];
+
+		const selectedBinding = (await vscode.window.showQuickPick(inputOutputItems, {
```
Charles-Gagnon, line 109: extra ()'s

Charles-Gagnon (changes requested):
Spacing changes are indeed due to the NLS task changes - this is fine to go in as is though and a fix will be made separately. 

@smartguest I thought this new task was going to also update the XLF but that doesn't seem to be the case. Does that still need to be updated manually or something?

lucyzhang929 (approved):

Charles-Gagnon (commented):
You need to update the command display name as well right? Currently it's hardcoded to input binding

VasuBhog:
> You need to update the command display name as well right? Currently it's hardcoded to input binding

In the package.nls.json I ensure that I put this to reflect the new options:
`"mssql.createAzureFunction":"Create Azure Function with SQL binding",` so this should fix it.

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -551,6 +551,15 @@
         <trans-unit id="doNotInstall">
             <source xml:lang="en">Do not install</source>
         </trans-unit>
+        <trans-unit id="input">
```
Charles-Gagnon, line 554: Shouldn't these be updated in the package.nls.json after re-running build?

VasuBhog: No, these live in the LocalizedConstants file. package.nls.json only has extension commands with `mssql.`. Thats why I update the location where they come using from `import * as LocalizedConstants from '../constants/localizedConstants';` 

I also did re-run build and lint this is what I pushed.

Charles-Gagnon: Pretty sure that isn't how it works for the MSSQL extension - vscode-nls doesn't understand xlf files as far as I know. It needs the strings to be in an nls.json file to be picked up. The task that @smartguest added is supposed to be doing that (reading in xlf and then writing the updated package.nls.json) I thought.

@smartguest am I missing something here?

VasuBhog: These are localized strings, not commands. I discussed this with Alex 
the localizedPackage.json.enu.xlf updates package.nls.json 
and
the localizedConstants.enu.xlf updates the localizedConstants.ts 

This is how it is set up.

Charles-Gagnon: Ok, I have no idea how the localization is supposed to work for stuff inside the extension. We generate the package.nls.json from the localizedPackage.json.enu.xlf but don't seem to do anything for any other xlf files (such as localizedConstants.enu.xlf)

So I guess this is expected but I'm curious how these strings are even being localized for the actual release then. It seems like that's just broken at this time.

Let's see what @smartguest has to say about this...

smartguest: The localizedConstants.enu.xlf file is to update the localizedConstants.ts file essentially yes @Charles-Gagnon they have a localize call which is processed by the compile-src gulp task, (this is how it was before I made any changes to the localization, my change just made it so that package.nls.json was generated in a similar way).

Charles-Gagnon: Ok, thanks. I'll follow up separately since this thing seems to be broken for these strings currently.

Charles-Gagnon (approved):

PR #17236, opened 2022-02-10
Increment azure function name when using create azure func from table

VasuBhog:
This will ensure that if the user would like to create an azure function with SQL binding from the same table again, then we don't use the azure function name (.cs). As such, I made an increment where each one is set up with a count starting at 1. The same way Azure Function: Create Function does it.


src/utils/utils.ts
```diff
@@ -31,3 +33,26 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
...
+ * @param functionName objectName that was chosen by the user
+ * @returns the function name that will always add at least `1` to the function name
+ */
+export async function getUniqueFsPath(folderPath: string, functionName: string): Promise<string | undefined> {
+	let count: number = 1;
+	const maxCount: number = 1024;
+
+	while (count < maxCount) {
+		const fileName: string = functionName + count.toString();
+		if (!(fs.existsSync(path.join(folderPath, '.cs' ? fileName + '.cs' : fileName)))) {
```
Charles-Gagnon, line 51: You can remove the ternary check here - `'.cs'` is always truthy

src/utils/utils.ts
```diff
@@ -31,3 +33,26 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
 export function generateQuotedFullName(schema: string, objectName: string): string {
 	return `[${escapeClosingBrackets(schema)}].[${escapeClosingBrackets(objectName)}]`;
 }
+
+
+/**
+ * Copied from Azure function: Create Function
```
Charles-Gagnon, line 39: For something as basic as this I don't think it matters to link the source. Instead it would be more helpful to have the description actually say what this is doing

src/services/azureFunctionsService.ts
```diff
@@ -77,9 +78,10 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 		const newFilePromise = azureFunctionUtils.waitForNewFunctionFile(projectFile);
 
 		// get function name from user
+		let uniqueFunctionName = await getUniqueFsPath(path.dirname(projectFile), table);
 		const functionName = await vscode.window.showInputBox({
 			title: constants.functionNameTitle,
-			value: table,
+			value: uniqueFunctionName,
```
Charles-Gagnon, line 84: While you're here - it would be good to add a validation check for the name to be non-empty

src/utils/utils.ts
```diff
@@ -31,3 +33,26 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
...
+ * https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/createFunction/FunctionNameStepBase.ts
+ * @param folderPath selected azure project folder path
+ * @param functionName objectName that was chosen by the user
+ * @returns the function name that will always add at least `1` to the function name
+ */
+export async function getUniqueFsPath(folderPath: string, functionName: string): Promise<string | undefined> {
```
Charles-Gagnon, line 45: This function name isn't accurate - it's not getting the path but rather just the unique filename (without the extension)

src/utils/utils.ts
```diff
@@ -31,3 +33,26 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
...
 	return `[${escapeClosingBrackets(schema)}].[${escapeClosingBrackets(objectName)}]`;
 }
+
+
+/**
+ * Copied from Azure function: Create Function
+ * https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/createFunction/FunctionNameStepBase.ts
+ * @param folderPath selected azure project folder path
+ * @param functionName objectName that was chosen by the user
+ * @returns the function name that will always add at least `1` to the function name
```
Charles-Gagnon, line 43: I think it would be better to first try the table name without a number, and then if that already exists start adding numbers.

Charles-Gagnon (changes requested):

src/utils/utils.ts
```diff
@@ -31,3 +33,25 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
...
+/**
+ * Gets a unique file name
+ * Increment the file name by adding 1 to function name if the file already exists
+ * @param folderPath selected azure project folder path
+ * @param fileName objectName that was chosen by the user
+ * @returns the function name that will always add at least `1` to the function name
+ */
+export async function getUniqueFileName(folderPath: string, fileName: string): Promise<string | undefined> {
+	let count: number = 0;
+	const maxCount: number = 1024;
```
lucyzhang929, line 47: Why is 1024 chosen as the maxCount?

src/utils/utils.ts
```diff
@@ -31,3 +33,25 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
 export function generateQuotedFullName(schema: string, objectName: string): string {
 	return `[${escapeClosingBrackets(schema)}].[${escapeClosingBrackets(objectName)}]`;
 }
+
+
+/**
+ * Gets a unique file name
+ * Increment the file name by adding 1 to function name if the file already exists
+ * @param folderPath selected azure project folder path
```
lucyzhang929, line 41: "selected azure function project folder path"

src/utils/utils.ts
VasuBhog: This was just the default number. Azure Functions does the same thing with their Create Function.

MaddyDev (approved):

src/utils/utils.ts
```diff
@@ -31,3 +33,26 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
...
 	return `[${escapeClosingBrackets(schema)}].[${escapeClosingBrackets(objectName)}]`;
 }
+
+
+/**
+ * Gets a unique file name
+ * Increment the file name by adding 1 to function name if the file already exists
+ * @param folderPath selected azure function project folder path
+ * @param fileName objectName that was chosen by the user
+ * @returns the unique file name that will always add at least `1` to the function name
```
chlafreniere, line 43: or undefined (and specify what case returns undefined)

Charles-Gagnon: The returns comment needs to be updated too - it no longer always add a string to the end (same with the overall description)

src/utils/utils.ts
```diff
@@ -31,3 +33,26 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
 export function generateQuotedFullName(schema: string, objectName: string): string {
 	return `[${escapeClosingBrackets(schema)}].[${escapeClosingBrackets(objectName)}]`;
 }
+
+
+/**
+ * Gets a unique file name
+ * Increment the file name by adding 1 to function name if the file already exists
+ * @param folderPath selected azure function project folder path
```
Charles-Gagnon, line 41: Not azure function specific

src/utils/utils.ts
```diff
@@ -31,3 +33,26 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
...
+ * @param fileName objectName that was chosen by the user
```
Charles-Gagnon, line 42: Make this more generic - `base filename to use` or something

src/utils/utils.ts
chlafreniere: the word "promise" should probably exist here too

Charles-Gagnon (approved):

PR #17239, opened 2022-02-10
Fixes for localization task

smartguest:
changes the dictionaryMapping function to use the correct source original string when generating the English package.nls.json file, and also replaces EOL with '\r\n'


smartguest:
nls screenshot

PR #17243, opened 2022-02-10
Only show "Create Azure Function with SQL Input Binding" in object explorer

### LINKED ISSUE #17242: "Create Azure Function with SQL Input Binding" is showing in the command palette
Selecting it gives an error since there's no table connection.

image

We should either prompt the user to select a connection - or not show it at all.

### LINKED ISSUE #17244: "Create Azure Function with SQL binding" menu item should be in a separate group
Currently it's listed inline with all the others

image

For separate functionality like this it's preferable to split it from the others by having it in a separate group.

VasuBhog:
Fixes #17242
Fixes #17244

Screen Shot 2022-02-11 at 2 07 30 PM


package.json
```diff
@@ -529,7 +529,8 @@
       {
         "command": "mssql.createAzureFunction",
         "title": "%mssql.createAzureFunction%",
-        "category": "MS SQL"
+        "category": "MS SQL",
+        "when": "view == objectExplorer && viewItem == database"
```
Charles-Gagnon, line 533: Why database? Don't you want this on tables instead so it has the correct table context to insert?

(also - side note. Is there a reason we're not supporting views for this?)

Charles-Gagnon (changes requested):

package.nls.json
```diff
@@ -1,129 +1,78 @@
...
-"mssql.query.ansiPadding":"Enable SET ANSI_PADDING",
-"mssql.query.ansiWarnings":"Enable SET ANSI_WARNINGS",
-"mssql.query.ansiNulls":"Enable SET ANSI_NULLS",
-"mssql.query.alwaysEncryptedParameterization":"Enable Parameterization for Always Encrypted",
-"mssql.ignorePlatformWarning":"[Optional] Do not show unsupported platform warnings",
-"mssql.chooseAuthMethod":"Chooses which Authentication method to use",
-"mssql.authCodeGrant.description":"Prompts users to sign in using their browser.",
-"mssql.deviceCode.description":"Allows users to sign in to input-constrained devices."
-}
+	"extension.runQuery": "Execute Query",
```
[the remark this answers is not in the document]
Charles-Gagnon: I'm not sure I understand. You haven't made any string changes here so this file shouldn't have been changed. And especially not the line endings change that this PR has a lot of - that was a bug that was fixed. So if you merge in latest from main and re-run the build task and commit the changes to your branch I'd expect these changes to go away.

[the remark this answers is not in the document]
smartguest: You need to update your main branch first before merging into your branch, otherwise, the resources provided will be outdated.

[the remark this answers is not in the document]
VasuBhog: image

VasuBhog:
> Please include a picture of the context menu

image

package.json
```diff
@@ -303,7 +303,7 @@
         {
           "command": "mssql.createAzureFunction",
           "when": "view == objectExplorer && viewItem == Table",
-          "group": "MS_SQL@6"
+          "group": "Azure_Function@1"
```
Charles-Gagnon, line 306: This should go below the other group - you can just name it `zAzure_Function@1` to do that

VasuBhog: updated!

Charles-Gagnon (approved):

package.json
```diff
@@ -303,7 +303,7 @@
         {
           "command": "mssql.createAzureFunction",
           "when": "view == objectExplorer && viewItem == Table",
-          "group": "MS_SQL@6"
+          "group": "zAzure_Function@1"
```
chlafreniere, line 306: what's the `z`? Is that part of the group name, or does that mean something else?

Charles-Gagnon: Yes. VS Code orders groups alphabetically so putting z on the group name makes it so it's at the end (which this should be).

AFAIK there isn't any other way to force the ordering, not sure why they don't have an "order" field or something 🤷

chlafreniere: hah ok 😄

chlafreniere (approved):

package.json
Charles-Gagnon: Another common pattern is to put numbers, e.g. 

1_MS_SQL
2_Azure_Function

Charles-Gagnon: https://code.visualstudio.com/api/references/contribution-points#Sorting-of-groups

VasuBhog: The group order is based on the alphabetical order of the group, so this ensures that it's below the second group.

PR #17245, opened 2022-02-10
provide more info or create project

### LINKED ISSUE #17240: Create new SQL Binding from Table with no project should provide more info

VasuBhog:
Fixes #17240

Screen Shot 2022-02-10 at 2 11 14 PM
Learn More: https://aka.ms/sqlbindings;
Create Azure Function Project: Asks users which folder they would like to create the project for and then starts the create function prompts

### DIFF
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -187,3 +187,6 @@ export const azureFunctionsExtensionNotFound = localize('azureFunctionsExtension
 export const installAzureFunction = localize('install', 'Install');
 export const learnMore = localize('learnMore', 'Learn more');
 export const doNotInstall = localize('doNotInstall', 'Do not install');
+export const createProject = localize('createProject', 'Create Azure Function Project');
+export const selectFolder = localize('selectFolder', 'Select Folder for Azure Function Project');
+export const select = localize('select', 'Select');
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -159,6 +159,7 @@ export const unixResourceClientPath = 'SqlToolsResourceProviderService';
 export const azureFunctionsExtensionName = 'ms-azuretools.vscode-azurefunctions';
 export const sqlConnectionString = 'SqlConnectionString';
 export const linkToAzureFunctionExtension = 'https://docs.microsoft.com/azure/azure-functions/functions-develop-vs-code';
+export const sqlBindingsDoc = 'https://aka.ms/sqlbindings';
 export const defaultSqlBindingTextLines =
 	[
 		'log.LogInformation(\"C# HTTP trigger function processed a request.\");',
--- a/src/services/azureFunctionsService.ts
+++ b/src/services/azureFunctionsService.ts
@@ -70,7 +70,7 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 		if (!projectFile) {
 			let projectCreate = await vscode.window.showErrorMessage(LocalizedConstants.azureFunctionsProjectMustBeOpened, constants.learnMore, constants.createProject);
 			if (projectCreate === constants.learnMore) {
-				vscode.commands.executeCommand('vscode.open', vscode.Uri.parse('https://aka.ms/sqlbindings'));
+				vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(constants.sqlBindingsDoc));
 			} else if (projectCreate === constants.createProject) {
 				// ask what folder the project should be created in
 				const options: vscode.OpenDialogOptions = {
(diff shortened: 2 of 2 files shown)


src/constants/constants.ts
```diff
@@ -187,3 +188,6 @@ export const azureFunctionsExtensionNotFound = localize('azureFunctionsExtension
 export const installAzureFunction = localize('install', 'Install');
 export const learnMore = localize('learnMore', 'Learn more');
 export const doNotInstall = localize('doNotInstall', 'Do not install');
+export const createProject = localize('createProject', 'Create Azure Function Project');
```
lucyzhang929, line 191: Should it always be Azure Functions with the 's'? Just want to make sure we are consistent

src/constants/constants.ts
```diff
@@ -187,3 +188,6 @@ export const azureFunctionsExtensionNotFound = localize('azureFunctionsExtension
...
+export const selectFolder = localize('selectFolder', 'Select Folder for Azure Function Project');
```
Charles-Gagnon, line 192: `Select folder for the Azure Function project`

I'd also suggest having the var be `selectAzureFunctionProjFolder` so it's clearer what the string is when reading the code where it's used

src/services/azureFunctionsService.ts
```diff
@@ -68,7 +68,25 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 		}
 		let projectFile = await azureFunctionUtils.getAzureFunctionProject();
 		if (!projectFile) {
-			vscode.window.showErrorMessage(LocalizedConstants.azureFunctionsProjectMustBeOpened);
+			let projectCreate = await vscode.window.showErrorMessage(LocalizedConstants.azureFunctionsProjectMustBeOpened, constants.learnMore, constants.createProject);
```
Charles-Gagnon, line 71: I'd suggest switching the order of the options, we want to encourage people to create the project and usually the ones on the left are the preferred options

src/services/azureFunctionsService.ts
```diff
@@ -68,7 +68,25 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
+			if (projectCreate === constants.learnMore) {
+				vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(constants.sqlBindingsDoc));
+			} else if (projectCreate === constants.createProject) {
+				// ask what folder the project should be created in
+				const options: vscode.OpenDialogOptions = {
```
Charles-Gagnon, line 76: I'm not sure we need to prompt here - the create flow will already do that so might as well let them handle the whole thing.

src/services/azureFunctionsService.ts
```diff
@@ -68,7 +68,25 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
+					title: constants.selectFolder,
+					canSelectMany: false,
+					openLabel: 'Select',
+					canSelectFiles: false,
+					canSelectFolders: true
+				};
+				// create the project based on the folder selected
+				await vscode.window.showOpenDialog(options).then(fileUri => {
```
Charles-Gagnon, line 84: If you're already awaiting this here then there isn't any reason to use .then - just put that code after the await.

src/services/azureFunctionsService.ts
```diff
@@ -68,7 +68,25 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
+					if (fileUri && fileUri[0]) {
+						azureFunctionApi.createFunction({ folderPath: fileUri[0].fsPath });
+					}
+				});
+			}
 			return;
```
Charles-Gagnon, line 90: If they create a new project then we should be able to go loop back and call getAzureFunctionProject again to save them having to re-run the command.

Charles-Gagnon (changes requested):

src/services/azureFunctionsService.ts
VasuBhog: not sure if this is feasible for most cases since there are instances where the create function needs to reload the window and therefore we lose the context of the connection string and such.

Charles-Gagnon: Yeah, there's very little we can do there. But no reason not to try - since the window doesn't need to reload unless this is the first folder being added to the workspace. 

Anyways I'll file an improvement for it. This isn't something we need off the bat.

Charles-Gagnon (approved):

src/services/azureFunctionsService.ts
VasuBhog: Yup agreed can look into this further as well.

PR #17246, opened 2022-02-10
Add create azure function xlf

VasuBhog:
This adds the XLF correctly due to the new process that is set up via this PR: https://github.com/microsoft/vscode-mssql/pull/17233


smartguest (changes requested):
in package.nls.json, you need to remove the changes to package.nls.json besides this line:

"mssql.createAzureFunction":"Create Azure Function with SQL binding"

The other changes are because the XLF source file is outdated and needs significant updates. Only add your changes to the package.nls.json and xlf files.

I am planning to update the XLF file later to better reflect package.nls.json as it is, but for now, just add your change to it.

VasuBhog:
> in package.nls.json, you need to remove the changes to package.nls.json besides this line:
> 
> "mssql.createAzureFunction":"Create Azure Function with SQL binding"
> 
> The other changes are because the XLF source file is outdated and needs significant updates. Only add your changes to the package.nls.json and xlf files.
> 
> I am planning to update the XLF file later to better reflect package.nls.json as it is, but for now, just add your change to it.

@smartguest 
This is due to your changes and if I revert that means I will always see this error when building (gulp build) which to me does not seem realistic for development. I can wait for you to fix the xlf / package.nls.json before pushing this in since all my PRs would have this issue of reverting the changes each time.

smartguest:
@VasuBhog That is the current state of what we have unfortunately, but sure, I'll update them as soon as I can.

Charles-Gagnon:
@VasuBhog Why are you doing this as a separate PR instead of in https://github.com/microsoft/vscode-mssql/pull/17235 where you're actually making the change to allow both input and output bindings?

VasuBhog:
> @VasuBhog Why are you doing this as a separate PR instead of in #17235 where you're actually making the change to allow both input and output bindings?

I could do it in that but I wanted to get this in so that I can pull it down to that other PR and merge main. That way I know that all strings and localizations should be fine in main now.

Charles-Gagnon:
But the string shouldn't be changed until that PR...and the current state of main is already correct with how things currently are there (it has the `mssql.createAzureFunction` string as this PR clearly shows)

VasuBhog:
To make this easier on everyone I just combined this with other PR. https://github.com/microsoft/vscode-mssql/pull/17235

PR #17248, opened 2022-02-11
Fix Timeout Issue

### LINKED ISSUE #17241: Timeout error after creating Azure Function
After creating an Azure Function from a table I'm seeing a popup about the azure function file not being created, although the file was created and displayed in the editor almost instantly

image

Other smaller issues : 

- [ ] We should capitalize `Azure Function` in the text
- [ ] Make sure we only show a single error instead of the two we currently show

VasuBhog:
Fixes #17241

So the issue here is that we do `azureFunctionApi.createFunction` which creates the generic azure function boilerplate. 

Due to https://github.com/microsoft/vscode-mssql/blob/main/src/services/azureFunctionsService.ts#L77, we have to watch for the file to be created and as such if it times out we get the first error (`Error running command`) from the mssql extension. 

The main solution that I found was to ensure we send the proper error message back.

image


Charles-Gagnon (commented):
I'm not sure I understand how this fixes the issue. The file was created just fine so why did I see a timeout error in the first place?

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -229,7 +229,7 @@ export function waitForNewFunctionFile(projectFile: string): Promise<string> {
 		const watcher = vscode.workspace.createFileSystemWatcher((
 			path.dirname(projectFile), '**/*.cs'), false, true, true);
 		const timeout = setTimeout(async () => {
-			reject(new Error(constants.timeoutError));
+			reject(constants.timeoutSqlBindingError);
```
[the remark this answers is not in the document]
VasuBhog: If I do `new Error...` then this will output two error messages as you see currently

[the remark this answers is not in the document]
Charles-Gagnon: Well that's likely a bug in our error handling then. Did you look at where the second error dialog is coming from? I suppose it could be something from VS Code but that'd be pretty odd behavior to me.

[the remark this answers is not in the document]
VasuBhog: This does come from VS Code as far as I see since I searched for the strings and did not find them in mssql repo.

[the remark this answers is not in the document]
Charles-Gagnon: Not the "Error running command" one - that comes from VS Code yes. I'm saying the other one which just has the error message that you were rejecting with.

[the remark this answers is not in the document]
VasuBhog: As far as I know. `reject(new Error...)` would be the same as `throw new Error...`. So that's why we get one from VS Code as mentioned and then another one (which we do not need anymore).
https://javascript.info/promise-error-handling#implicit-try-catch

VasuBhog:
> I'm not sure I understand how this fixes the issue. The file was created just fine so why did I see a timeout error in the first place?

So what I am seeing is that the vscode.workspace.createFileSystemWatcher was waiting for adding the Nuget Package reference, addConnectionStringToConfig as well to finish. So it will still output the error. I moved the check to see the promise sooner now and that should alleviate the issue of timeout being affected by other things.

src/azureFunction/azureFunctionUtils.ts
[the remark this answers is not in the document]
Charles-Gagnon: So throwing directly vs rejecting is the same - but you should be throwing an Error regardless. Not throwing a string.

What isn't clear to me is why this fixes the duplicate error messages. Because unless VS Code has something built in I don't see why throwing an Error object vs a string would cause an extra dialog to show like this - I've never seen that before and we throw errors everywhere else. Are you sure we aren't catching the error and displaying the additional message ourselves? That seems much more likely.

Charles-Gagnon:
So looking at the code I think the real issue is where we start the timer. Currently we start it BEFORE we prompt the user to go through the flow to create the file : 

https://github.com/microsoft/vscode-mssql/blob/main/src/services/azureFunctionsService.ts#L77

Which means we then immediately start the timer, but the user still needs to input everything to create the file which could easily take a number of seconds.

So instead, we should move that down to line 96 and start waiting right after we finish with the flow from the AF extension. 

This would be in addition to your change to move the await to line 97 - since I agree that seems like a better place to put it.

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -229,8 +229,9 @@ export function waitForNewFunctionFile(projectFile: string): Promise<string> {
...
 			watcher.dispose();
+			throw new Error();
```
Charles-Gagnon, line 234: You should be rejecting with an error or throwing an error - not both.

src/services/azureFunctionsService.ts
```diff
@@ -94,9 +90,13 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 			folderPath: projectFile
 		});
 
+		// because of an AF extension API issue, we have to get the newly created file by adding
+		// a watcher: https://github.com/microsoft/vscode-azurefunctions/issues/2908
+		const newFilePromise = azureFunctionUtils.waitForNewFunctionFile(projectFile);
```
Charles-Gagnon, line 95: Oh right, I remember why we set this up now. We have to start the file watcher before we finish the create flow because otherwise if the file is created before we set up the file watcher then we'll run into a race condition where we might never get the notification the file was created.

So what we'll have to do is set up the file watcher promise before the create flow, and then as soon as that's done set up another promise with the timeout and use Promise.race to catch the first one.

src/services/azureFunctionsService.ts
```diff
@@ -94,9 +95,17 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 			folderPath: projectFile
 		});
 
+		// check for the new function file to be created
+		const functionFile = await newFilePromise;
+		const timeoutFunctionFile = await checkFunctionFile;
+		try {
+			await Promise.race([functionFile, timeoutFunctionFile]);
+		} catch (err) {
+			console.log(err);
```
Charles-Gagnon, line 104: You shouldn't catch and just log here - we want this to fail if we hit the timeout since that implies the file wasn't detected and so trying to continue on will just result in an error (since we need to know the file path to update it...)

src/utils/utils.ts
```diff
@@ -31,3 +32,15 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
 export function generateQuotedFullName(schema: string, objectName: string): string {
 	return `[${escapeClosingBrackets(schema)}].[${escapeClosingBrackets(objectName)}]`;
 }
+
+/**
+ * Promise with timeout to check for another promise fulfillment
+ * @returns a promise that resolves after the specified timeout
```
Charles-Gagnon, line 38: This isn't true - it never resolves. It _rejects_ after the specified timeout.

And you should either say here that the timeout is 10sec or have the timeout be a parameter (with a default of 10sec)

src/utils/utils.ts
```diff
@@ -31,3 +32,15 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
...
+ */
+export function waitForPromise(): Promise<void> {
+	return new Promise((_, reject) => {
+		setTimeout(() => {
+			reject(new Error(constants.timeoutSqlBindingError));
```
Charles-Gagnon, line 43: This message is pretty specific to sql bindings. I'd suggest just having this take in the message as a parameter

src/utils/utils.ts
```diff
@@ -31,3 +32,15 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
 export function generateQuotedFullName(schema: string, objectName: string): string {
 	return `[${escapeClosingBrackets(schema)}].[${escapeClosingBrackets(objectName)}]`;
 }
+
+/**
+ * Promise with timeout to check for another promise fulfillment
+ * @returns a promise that resolves after the specified timeout
+ */
+export function waitForPromise(): Promise<void> {
```
Charles-Gagnon, line 40: I'd suggest just `timeoutPromise` or something similar - this function isn't actually doing any "waiting" so naming it `waitFor...` isn't really accurate

src/utils/utils.ts
```diff
@@ -31,3 +32,15 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
 export function generateQuotedFullName(schema: string, objectName: string): string {
 	return `[${escapeClosingBrackets(schema)}].[${escapeClosingBrackets(objectName)}]`;
 }
+
+/**
+ * Promise with timeout to check for another promise fulfillment
```
Charles-Gagnon, line 37: `Returns a promise that will reject after the specified timeout`

src/services/azureFunctionsService.ts
```diff
@@ -75,6 +75,7 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 		// because of an AF extension API issue, we have to get the newly created file by adding
 		// a watcher: https://github.com/microsoft/vscode-azurefunctions/issues/2908
 		const newFilePromise = azureFunctionUtils.waitForNewFunctionFile(projectFile);
+		const checkFunctionFile = waitForPromise();
```
Charles-Gagnon, line 78: As discussed earlier we shouldn't start the timeout until AFTER we've gotten the function name from the user and called createFunction.

src/services/azureFunctionsService.ts
```diff
@@ -94,9 +95,17 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 			folderPath: projectFile
 		});
 
+		// check for the new function file to be created
+		const functionFile = await newFilePromise;
+		const timeoutFunctionFile = await checkFunctionFile;
+		try {
+			await Promise.race([functionFile, timeoutFunctionFile]);
```
Charles-Gagnon, line 102: This isn't how you use Promise.race - you shouldn't be awaiting the promises above. It should look like this:

```
const timeoutPromise = waitForPromise();
const functionFile = await Promise.race([newFilePromise, timeoutPromise]);
```

But now there's the slight problem since the type of functionFile is going to be unknown (since you have mixed promise types). There's a couple solutions : 

1. Just cast the return type to string since you know that if it resolves it'll always be a string
2. Have the timeout promise be generic with the same type as the other promise (string in this case)
3. Use a different way of doing the timeout, for example [here](https://github.com/microsoft/azuredatastudio/blob/2bc6a0cd012de69504061a876070cfa43533fedc/src/vs/base/common/async.ts#L79) they're just having it return undefined if it times out. So you could use that functionality and then check if it's undefined after the await and throw in that case.

Charles-Gagnon (changes requested):

src/azureFunction/azureFunctionUtils.ts
```diff
@@ -229,17 +229,12 @@ export async function getSettingsFile(projectFile: string): Promise<string | und
...
 export function waitForNewFunctionFile(projectFile: string): Promise<string> {
-	return new Promise((resolve, reject) => {
+	return new Promise((resolve, _) => {
 		const watcher = vscode.workspace.createFileSystemWatcher((
 			path.dirname(projectFile), '**/*.cs'), false, true, true);
-		const timeout = setTimeout(async () => {
-			reject(new Error(LocalizedConstants.timeoutError));
-			watcher.dispose();
-		}, 10000);
 		watcher.onDidCreate((e) => {
```
Charles-Gagnon, line 235: You need to still clean up the watcher if the timeout happens as well

src/services/azureFunctionsService.ts
```diff
@@ -101,6 +101,13 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 			folderPath: projectFile
 		});
 
+		// check for the new function file to be created
+		const timeout = timeoutPromise(10000, LocalizedConstants.timeoutSqlBindingError);
+		const functionFile = await Promise.race([newFilePromise, timeout]);
+		if (!functionFile) {
```
Charles-Gagnon, line 107: This should never be possible - it'll either return a file path or throw

src/utils/utils.ts
```diff
@@ -31,3 +31,17 @@ export async function executeCommand(command: string, cwd?: string): Promise<str
...
 	return `[${escapeClosingBrackets(schema)}].[${escapeClosingBrackets(objectName)}]`;
 }
+
+/**
+ * Returns a promise that will reject after the specified timeout
+ * @param ms timeout in milliseconds. Default is 10 seconds
+ * @param errorMessage error message to be returned in the rejection
+ * @returns a promise that rejects after the specified timeout
+ */
+export function timeoutPromise(ms: number = 10000, errorMessage: string): Promise<string> {
```
Charles-Gagnon, line 41: Optional params should always be last

Charles-Gagnon (changes requested):

Charles-Gagnon (approved):

src/services/azureFunctionsService.ts
```diff
@@ -104,6 +104,10 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 			folderPath: projectFile
 		});
 
+		// check for the new function file to be created and dispose of the file system watcher
+		const timeout = timeoutPromise(LocalizedConstants.timeoutAzureFunctionFileError);
+		const functionFile = await Promise.race([newFunctionFileObject.filePromise, timeout]).finally(newFunctionFileObject.watcherDisposable.dispose());
```
Charles-Gagnon, line 109: Actually - let's wrap lines 87 - 109 in a try/finally to ensure that we clean the watcher up even if something else happens like the user exits the flow.

Charles-Gagnon (changes requested):

Charles-Gagnon (approved):

Issue #17247, opened 2022-02-10
Strings are not localized properly


VasuBhog:
Dev Main building does not include the proper xlf for strings that used to be in package.nls.json

PR that changed package.nls.json: https://github.com/microsoft/vscode-mssql/pull/17233

![image](https://user-images.githubusercontent.com/23587151/153516868-4afdf302-ba7c-4503-b12d-df96dc459e9d.png)

PR #17249, opened 2022-02-11
Update to XLFs and constants.ts

smartguest:
Fixes: #17247

This PR adds in updated XLFs that removes unused strings from the localizedConstants xlf and makes the localizedPackage xlf reflect the current package.nls.json. 

Additionally, the constants.ts file had localized calls that weren't being recognized at all or translated, I decided to move them to the localizedConstants xlf file where they should belong. 

Finally, the original file attribute of the xlfs has been changed to match ADS and better reflect the actual original file (it doesn't affect the functionality but it's just for consistency)


VasuBhog (approved):
Tested. Looks good to me.

PR #17176, opened 2021-12-15
Bump STS to 3.0.0.174 to pick up data type fix

kburtram:
Mitigation for https://github.com/microsoft/azuredatastudio/issues/17926 for Dec release.

PR #17265, opened 2022-02-18
Remove dev config

Charles-Gagnon:
Some cleanup - we switched a while back to only using the "dev" config so there isn't any point in keeping multiple versions around anymore. It's just confusing about which config to update. 

(we could remove the copy-config task at some point completely if we wanted by moving config.json into the root or something, but the logic to read the config is pretty messy and I didn't want to spend a bunch of time fixing all that up so just sticking with the simple change for now)


abist:
@Charles-Gagnon We update the production config when releasing (e.g -https://github.com/microsoft/vscode-mssql/pull/17176/files). Do releases also target dev now?

Charles-Gagnon:
The current version of the extension we ship is using the dev config though - take a look at the config file in the installed extension folder.

I vaguely recall that this was changed a while back and we discussed reverting it back to use the blob storage but I can't remember the details there. @kburtram do you remember?

Charles-Gagnon:
(I don't know exactly why this is happening - the pipeline is set to "release" which should mean it uses the "production" config but it isn't. I think it may be because of us having the var like this in the build yaml : 

```
variables:
  - name: release
    value: false
```

PR #17266, opened 2022-02-18
Add tenant selection during adding connection using AAD auth

### LINKED ISSUE #17042: Cannot select tenant with Azure AD auth
- MSSQL Extension Version: 1.10.1
- VSCode Version: 1.59.0
- OS Version: macOS 11.4

Steps to Reproduce:

1. Add connection
1. Enter Azure server hostname
1. Enter database name
1. Use Azure AD authentication to connect to a database
1. Select the account
1. Leave profile name empty
1. The extension tries to connect with your default tenant instead of asking which tenant to use
1. If the database is in another tenant, an error occurs: "Unable to connect using the connection information provided. Retry profile creation?"

Same connection works fine in Azure Data Studio because it asks you which tenant to use.

caohai:
Fixes #17042

Root cause: The current AAD implementation always chooses default tenant to use and doesn't handle the case where the user has multiple home tenants. This causes failure when the user tries to access resources that are not in the default tenant.

https://github.com/microsoft/vscode-mssql/blob/e36ef1f2c4bb3d4887b90232af2b5241abacd487/src/azure/azureController.ts#L159-L169
https://github.com/microsoft/vscode-mssql/blob/593a01c345170d349f3475d6adadccd3135a7536/lib/ads-adal-library/src/engine/AzureAuth.ts#L62-L65

Eventually the endpoint url for getToken calls contains tenant id.
https://github.com/microsoft/vscode-mssql/blob/593a01c345170d349f3475d6adadccd3135a7536/lib/ads-adal-library/src/engine/AzureAuth.ts#L190-L191
Fix: prompt the user for a tenant selection if there are multiple tenants available on the account.
1) If the user is adding a connection using an added account, prompt a tenant selection after account selection.
2) If the user is adding a new account during adding a connection, prompt a tenant selection after the account is successfully added.

image

To test this change locally, you'll need to add another tenant to your Azure account.

### DIFF
--- a/src/azure/azureController.ts
+++ b/src/azure/azureController.ts
@@ -100,7 +100,7 @@ export class AzureController {
 
 	private async promptForTenantChoice(account: AzureAccount, profile: ConnectionProfile): Promise<void> {
 		let tenantChoices: INameValueChoice[] = account.properties.tenants?.map(t => ({ name: t.displayName, value: t }));
-		if (tenantChoices.length === 1) {
+		if (tenantChoices && tenantChoices.length === 1) {
 			profile.tenantId = tenantChoices[0].value.id;
 			return;
 		}
@@ -124,11 +124,12 @@ export class AzureController {
 			let azureCodeGrant = await this.createAuthCodeGrant();
 			account = await azureCodeGrant.startLogin();
 			await accountStore.addAccount(account);
-			if (profile.tenantId === undefined) {
+			if (!profile.tenantId) {
 				await this.promptForTenantChoice(account, profile);
 			}
+			let tid = profile.tenantId ? profile.tenantId : azureCodeGrant.getHomeTenant(account).id;
 			const token = await azureCodeGrant.getAccountSecurityToken(
-				account, profile.tenantId, settings
+				account, tid, settings
 			);
 			if (!token) {
 				let errorMessage = LocalizedConstants.msgGetTokenFail;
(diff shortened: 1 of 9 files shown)


abist (approved):

src/azure/azureController.ts
```diff
@@ -93,15 +98,37 @@ export class AzureController {
 		this.azureMessageDisplayer = new AzureMessageDisplayer();
 	}
 
+	private async promptForTenantChoice(account: AzureAccount, profile: ConnectionProfile): Promise<void> {
+		let tenantChoices: INameValueChoice[] = account.properties.tenants?.map(t => ({ name: t.displayName, value: t }));
+		if (tenantChoices.length === 1) {
```
kburtram, line 103: `tenantChoices` can never be `undefined` or length 0 here?

src/azure/azureController.ts
```diff
@@ -93,15 +98,37 @@ export class AzureController {
...
+			profile.tenantId = tenantChoices[0].value.id;
+			return;
+		}
+		let tenantQuestion: IQuestion = {
+			type: QuestionTypes.expand,
+			name: LocalizedConstants.tenant,
+			message: LocalizedConstants.azureChooseTenant,
+			choices: tenantChoices,
+			shouldPrompt: (answers) => profile.isAzureActiveDirectory() && tenantChoices.length > 1,
```
kburtram, line 112: nit: don't think you need `tenantChoices.length > 1` since you return in the if above, but it doesn't hurt to make preconditions clear either.

caohai: I'll keep it here to make it clear. Plus if tenantChoices.length becomes zero, this question should also not prompt.

kburtram (approved):

PR #17270, opened 2022-02-22
Start the Add SQL Binding flow once azure project is created

### LINKED ISSUE #17250: Create new binding from table should pick up new project if user chose to make one
See https://github.com/microsoft/vscode-mssql/pull/17245 for details

If the user is prompted to create a new project when creating a new function with sql binding from a table then we should wait until the create is finished and continue the flow using the project they just created instead of requiring that they run the command again. 

This won't always be possible since the extension host may restart if the new project is the first one in the workspace, but it's not uncommon for people to have a workspace already opened that the project can be added to which won't require a restart.

VasuBhog:
Fixes #17250

This will ensure we wait for createFunction to be completed if the user chooses to do so and start the Add SQL binding flow to the newly created function project.

### DIFF
--- a/src/services/azureFunctionsService.ts
+++ b/src/services/azureFunctionsService.ts
@@ -77,7 +77,14 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 				// start the create azure function project flow
 				await azureFunctionApi.createFunction({});
 			}
-			return;
+			if (await azureFunctionUtils.getHostFiles()) {
+				// once project is created then start sql binding flow
+				projectFile = await azureFunctionUtils.getAzureFunctionProject();
+			} else {
+				// if user cancels prompt show warning
+				vscode.window.showErrorMessage(LocalizedConstants.azureFunctionsProjectMustBeOpened);
+				return;
+			}
 		}
 
 		// because of an AF extension API issue, we have to get the newly created file by adding
--- a/src/services/azureFunctionsService.ts
+++ b/src/services/azureFunctionsService.ts
@@ -73,6 +73,7 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 				LocalizedConstants.createProject, LocalizedConstants.learnMore);
 			if (projectCreate === LocalizedConstants.learnMore) {
 				vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(constants.sqlBindingsDoc));
+				return;
 			} else if (projectCreate === LocalizedConstants.createProject) {
 				// start the create azure function project flow
 				await azureFunctionApi.createFunction({});
--- a/src/services/azureFunctionsService.ts
+++ b/src/services/azureFunctionsService.ts
@@ -91,7 +91,7 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 						projectFile = await azureFunctionUtils.getAzureFunctionProject();
 					}
 				} catch (error) {
-					vscode.window.showErrorMessage(utils.formatString(LocalizedConstants.errorNewAzureFunction, error.message) ?? error);
+					vscode.window.showErrorMessage(utils.formatString(LocalizedConstants.errorNewAzureFunction, error.message ?? error));
 					return;
 				} finally {
 					newHostProjectFile.watcherDisposable.dispose();
--- a/src/services/azureFunctionsService.ts
+++ b/src/services/azureFunctionsService.ts
@@ -82,15 +82,15 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 					// start the create azure function project flow
 					// create a watcher to see if the host file is created after the flow is complete
 					newHostProjectFile = await azureFunctionUtils.waitForNewHostFile();
-					const timeoutForHostFile = timeoutPromise(LocalizedConstants.azureFunctionsProjectMustBeOpened);
 					await azureFunctionApi.createFunction({});
+					const timeoutForHostFile = timeoutPromise(LocalizedConstants.azureFunctionsProjectMustBeOpened);
 					hostFile = await Promise.race([newHostProjectFile.filePromise, timeoutForHostFile]);
 					if (hostFile) {
 						// start the add sql binding flow
 						projectFile = await azureFunctionUtils.getAzureFunctionProject();
 					}
 				} catch (error) {
-					vscode.window.showErrorMessage(error.message);
+					vscode.window.showErrorMessage(LocalizedConstants.errorNewAzureFunction + error.message ?? error);
 					return;
 				} finally {
 					newHostProjectFile.watcherDisposable.dispose();
--- a/src/services/azureFunctionsService.ts
+++ b/src/services/azureFunctionsService.ts
@@ -68,6 +68,9 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 			return;
 		}
 		let projectFile = await azureFunctionUtils.getAzureFunctionProject();
+		let newHostProjectFile: azureFunctionUtils.IFileFunctionObject;
+		let hostFile: string;
+
 		if (!projectFile) {
 			let projectCreate = await vscode.window.showErrorMessage(LocalizedConstants.azureFunctionsProjectMustBeOpened,
 				LocalizedConstants.createProject, LocalizedConstants.learnMore);
@@ -75,16 +78,23 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 				vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(constants.sqlBindingsDoc));
 				return;
 			} else if (projectCreate === LocalizedConstants.createProject) {
-				// start the create azure function project flow
-				await azureFunctionApi.createFunction({});
-			}
-			if (await azureFunctionUtils.getHostFiles()) {
-				// once project is created then start sql binding flow
-				projectFile = await azureFunctionUtils.getAzureFunctionProject();
-			} else {
-				// if user cancels prompt show warning
-				vscode.window.showErrorMessage(LocalizedConstants.azureFunctionsProjectMustBeOpened);
-				return;
+				try {
+					// start the create azure function project flow
+					// create a watcher to see if the host file is created after the flow is complete
+					newHostProjectFile = await azureFunctionUtils.waitForNewHostFile();
+					const timeoutForHostFile = timeoutPromise(LocalizedConstants.azureFunctionsProjectMustBeOpened);
+					await azureFunctionApi.createFunction({});
+					hostFile = await Promise.race([newHostProjectFile.filePromise, timeoutForHostFile]);
+					if (hostFile) {
+						// start the add sql binding flow
+						projectFile = await azureFunctionUtils.getAzureFunctionProject();
+					}
+				} catch (error) {
+					vscode.window.showErrorMessage(error.message);
+					return;
+				} finally {
+					newHostProjectFile.watcherDisposable.dispose();
+				}
 			}
 		}
 
@@ -116,8 +126,8 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 			});
 
 			// check for the new function file to be created and dispose of the file system watcher
-			const timeout = timeoutPromise(LocalizedConstants.timeoutAzureFunctionFileError);
-			functionFile = await Promise.race([newFunctionFileObject.filePromise, timeout]);
+			const timeoutForFunctionFile = timeoutPromise(LocalizedConstants.timeoutAzureFunctionFileError);
+			functionFile = await Promise.race([newFunctionFileObject.filePromise, timeoutForFunctionFile]);
 		} finally {
 			newFunctionFileObject.watcherDisposable.dispose();
 		}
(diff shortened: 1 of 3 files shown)


src/services/azureFunctionsService.ts
```diff
@@ -77,7 +77,14 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
 				// start the create azure function project flow
 				await azureFunctionApi.createFunction({});
 			}
-			return;
+			if (await azureFunctionUtils.getHostFiles()) {
+				// once project is created then start sql binding flow
+				projectFile = await azureFunctionUtils.getAzureFunctionProject();
+			} else {
+				// if user cancels prompt show warning
+				vscode.window.showErrorMessage(LocalizedConstants.azureFunctionsProjectMustBeOpened);
```
Charles-Gagnon, line 85: So if the user clicks the learn more button doesn't that mean that they're going to get the same notification twice in a row? That doesn't seem like a great experience, if they choose learn more we should either

1. End the flow there
2. Show the error message with the buttons again (for the case where they want to learn more but also do want to make a new project)

(I think 2 would be my preference, @dzsquared thoughts?)

Charles-Gagnon (changes requested):

src/services/azureFunctionsService.ts
VasuBhog: I personally like the first one as showing the same message would possibly create confusion, and they should be able to understand that in order to run is to create an azure function stated in the docs and in the error message they already saw. So they would have to re-run the flow or create an azure function project separately.

Charles-Gagnon: I don't think it'd be that confusing - it's pretty standard to have docs links not take you away from the page/dialog you're on. But it's a minor detail so I'm fine leaving it like this too.

src/services/azureFunctionsService.ts
```diff
@@ -73,11 +73,19 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
 			} else if (projectCreate === LocalizedConstants.createProject) {
 				// start the create azure function project flow
 				await azureFunctionApi.createFunction({});
 			}
-			return;
+			if (await azureFunctionUtils.getHostFiles()) {
+				// once project is created then start sql binding flow
+				projectFile = await azureFunctionUtils.getAzureFunctionProject();
+			} else {
+				// if user cancels prompt show warning
```
Charles-Gagnon, line 85: Is this comment correct? The if has nothing to do about prompting the user for anything - it's just looking for the host files. And why do you check the host files first? That's already done as part of the `getAzureFunctionProject` flow so doing it here too seems unnecessary

Charles-Gagnon (changes requested):

src/services/azureFunctionsService.ts
Charles-Gagnon: Oh I see - if they cancel out of the dialog on line 83 you're saying?

I'd suggest just calling `getAzureFunctionProject` instead - that already checks for host files and if none exist returns undefined.

VasuBhog: Line 79 specifically. `await azureFunctionApi.createFunction({});` - if user cancels that prompt early then we show the error. 

If the complete it then we do the getAzureFunctionProject call. But I cant just do getAzureFunctionProject call as I found if a user completes the createFunction promise then that is completed before the project is even made. Similar to what happened below in the logic with the `waitForNewFunctionFile`

Charles-Gagnon: I don't think this is the right solution then - the check you added isn't waiting for anything to be created. It just immediately checks if the host.json file isn't there. But if the project isn't done being created we can't guarantee the host.json file is there.

So if you have seen that happening then I would suggest you look into adding a file watcher here for the host.json file being created. That's going to be a much more robust way of doing that. 

Although the tricky thing here is that we have no idea where the project is created, and IIRC the createFunction flow asks the user if they want to add it to their workspace so we aren't even guaranteed that the project will be visible to us after creation. I don't have a good answer for those right now, but we definitely should be doing something more than just hoping the host.json file is created by the time we call that if the createFunction call is returning before the project is completely created.

src/services/azureFunctionsService.ts
```diff
@@ -68,16 +68,34 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
 				vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(constants.sqlBindingsDoc));
+				return;
 			} else if (projectCreate === LocalizedConstants.createProject) {
-				// start the create azure function project flow
-				await azureFunctionApi.createFunction({});
+				try {
+					// start the create azure function project flow
+					// create a watcher to see if the host file is created after the flow is complete
+					newHostProjectFile = await azureFunctionUtils.waitForNewHostFile();
+					const timeoutForHostFile = timeoutPromise(LocalizedConstants.azureFunctionsProjectMustBeOpened);
```
Charles-Gagnon, line 85: You should create this after createFunction, otherwise you'll have the same bug we fixed before if the user takes a while to go through the createFunction flow

src/services/azureFunctionsService.ts
```diff
@@ -68,16 +68,34 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
+					await azureFunctionApi.createFunction({});
+					hostFile = await Promise.race([newHostProjectFile.filePromise, timeoutForHostFile]);
+					if (hostFile) {
+						// start the add sql binding flow
+						projectFile = await azureFunctionUtils.getAzureFunctionProject();
+					}
+				} catch (error) {
+					vscode.window.showErrorMessage(error.message);
```
Charles-Gagnon, line 93: Generally we should include header text saying where the error is from `Error creating new Azure Function : ...`

Also it's good practice to not assume that a caught error is an error object. Most of the time it should be but sometimes other libraries throw strings instead. So you should do `error.message ?? error`

Charles-Gagnon (changes requested):

src/services/azureFunctionsService.ts
```diff
@@ -68,16 +68,34 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
+					newHostProjectFile = await azureFunctionUtils.waitForNewHostFile();
+					await azureFunctionApi.createFunction({});
+					const timeoutForHostFile = timeoutPromise(LocalizedConstants.azureFunctionsProjectMustBeOpened);
+					hostFile = await Promise.race([newHostProjectFile.filePromise, timeoutForHostFile]);
+					if (hostFile) {
+						// start the add sql binding flow
+						projectFile = await azureFunctionUtils.getAzureFunctionProject();
+					}
+				} catch (error) {
+					vscode.window.showErrorMessage(LocalizedConstants.errorNewAzureFunction + error.message ?? error);
```
Charles-Gagnon, line 93: You can't do localized strings like this - it doesn't work for non-LTR languages (among other problems). Use a format string instead (there's plenty of examples in this repo)

src/services/azureFunctionsService.ts
```diff
@@ -68,16 +68,34 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
+				return;
 			} else if (projectCreate === LocalizedConstants.createProject) {
-				// start the create azure function project flow
-				await azureFunctionApi.createFunction({});
+				try {
+					// start the create azure function project flow
+					// create a watcher to see if the host file is created after the flow is complete
+					newHostProjectFile = await azureFunctionUtils.waitForNewHostFile();
+					await azureFunctionApi.createFunction({});
+					const timeoutForHostFile = timeoutPromise(LocalizedConstants.azureFunctionsProjectMustBeOpened);
```
Charles-Gagnon, line 86: Is this the error string we want to display? If we time out it means the user tried to create a project but something else happened, so we should probably say something like we did for the other ones ("Timed out waiting for project to be created" or something...)

Charles-Gagnon (changes requested):

src/services/azureFunctionsService.ts
```diff
@@ -68,16 +69,34 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
+					newHostProjectFile = await azureFunctionUtils.waitForNewHostFile();
+					await azureFunctionApi.createFunction({});
+					const timeoutForHostFile = timeoutPromise(LocalizedConstants.timeoutProjectError);
+					hostFile = await Promise.race([newHostProjectFile.filePromise, timeoutForHostFile]);
+					if (hostFile) {
+						// start the add sql binding flow
+						projectFile = await azureFunctionUtils.getAzureFunctionProject();
+					}
+				} catch (error) {
+					vscode.window.showErrorMessage(utils.formatString(LocalizedConstants.errorNewAzureFunction, error.message) ?? error);
```
Charles-Gagnon, line 94: I don't see the format index in the string - pretty sure that's required for this to work correctly.

VasuBhog: I used it the same way we use it in utils file. https://github.com/microsoft/vscode-mssql/blob/main/src/models/utils.ts#L186

Charles-Gagnon: What do you mean "used it in the utils file". It doesn't get used there. It's defined there...

Look at the logic - it's going through and searching for instances of `{\d+}` (e..g {0}, {1}, etc.) and replacing them with the values you pass in.

Just find other uses of utils.formatString and look at how those strings are defined.

VasuBhog: This is how it is defined. You pass in multiple string to format together. This is also in azurefunctionutils in the same way.

VasuBhog: https://github.com/microsoft/vscode-mssql/blob/d628907b30b49760edc87fd90276dd121af27572/src/azureFunction/azureFunctionUtils.ts#L46

Charles-Gagnon: I'm saying that the string itself is wrong - it needs to have the index specifier `{0}` for the replacement to work.

e.g. for failedToParse https://github.com/microsoft/vscode-mssql/blob/main/localization/xliff/enu/constants/localizedConstants.enu.xlf#L531 it looks like this

`Failed to parse "{0}": {1}.`

You currently have the string defined as just 

`Error creating new Azure Function:`

in [localizedConstants.enu.xlf](https://github.com/microsoft/vscode-mssql/blob/vabhog/newProj/localization/xliff/enu/constants/localizedConstants.enu.xlf#L579). which means the function will never do the replacement to add the error text. It needs to be 

`Error creating new Azure Function: {0}`

VasuBhog: Ah, I did miss that. Thanks for clarifying, I will be sure to make a note of this when addressing future work. Updated the localizedConstant.

Charles-Gagnon, line 94: You need to move the `?? error` bit now to inside the formatString call

src/services/azureFunctionsService.ts
```diff
@@ -68,16 +69,34 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
 				vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(constants.sqlBindingsDoc));
+				return;
 			} else if (projectCreate === LocalizedConstants.createProject) {
-				// start the create azure function project flow
-				await azureFunctionApi.createFunction({});
+				try {
+					// start the create azure function project flow
+					// create a watcher to see if the host file is created after the flow is complete
+					newHostProjectFile = await azureFunctionUtils.waitForNewHostFile();
+					await azureFunctionApi.createFunction({});
```
Charles-Gagnon, line 86: So this is going to start the create project flow from the start, right? That could potentially be a problem, as we have some specific requirements such as not supporting the .NET 5 runtime currently. Are you planning on following up to ensure that this create experience isn't going to lead users down the wrong path?

src/services/azureFunctionsService.ts
```diff
@@ -68,16 +69,34 @@ export class AzureFunctionsService implements mssql.IAzureFunctionsService {
...
 				LocalizedConstants.createProject, LocalizedConstants.learnMore);
 			if (projectCreate === LocalizedConstants.learnMore) {
 				vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(constants.sqlBindingsDoc));
+				return;
 			} else if (projectCreate === LocalizedConstants.createProject) {
-				// start the create azure function project flow
-				await azureFunctionApi.createFunction({});
+				try {
+					// start the create azure function project flow
+					// create a watcher to see if the host file is created after the flow is complete
```
Charles-Gagnon, line 84: You should add to this comment explaining why this is necessary like we do in the other place (with the link to the AF extension issue filed)

Charles-Gagnon (approved):
It seems fine as a start, but we may want to follow up with trying to tighten up the edge case logic. Or at least make it so when something happens the user is able to determine why this fails (for example - if they choose to make a project and not add it to the workspace it'll time out but the user might be confused why)

VasuBhog:
Also have an issue tracking the watchers' issue we have set up: https://github.com/microsoft/vscode-azurefunctions/issues/3052 as it seems to be a core issue with how they write the json file.

src/services/azureFunctionsService.ts
VasuBhog: Is there documentation of what runtimes / sdks are supported in regards to SQL binding?

Charles-Gagnon: I thought we did but I'm not seeing that specifically listed anywhere. https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-azure-sql just mentions that we only support AF v2

@dzsquared Do you know if we list the requirements anywhere?

PR #17271, opened 2022-02-22
Fix to find child function folder

### LINKED ISSUE #17269: add SQL binding when functions project is within a folder asks to create a new project
not a ship blocker

- MSSQL Extension Version: RC 1.13
Version: 1.64.2 (Universal)
Commit: f80445acd5a3dadef24aa209168452a3d97cc326
Date: 2022-02-09T22:00:58.347Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Darwin arm64 21.3.0

Steps to Reproduce:

1. open an empty folder
2. add a folder, we'll call it "functions"
3. create a functions project within "functions"
4. open object explorer and try to add a sql binding to a table

VasuBhog:
Fixes #17269

This will ensure we check to get all function files even if it is a subfolder of another folder.

### DIFF
--- a/src/azureFunction/azureFunctionUtils.ts
+++ b/src/azureFunction/azureFunctionUtils.ts
@@ -203,7 +203,7 @@ export async function getAzureFunctionProjectFiles(): Promise<string[] | undefin
 		return undefined;
 	}
 	for (let host of hostFiles) {
-		let projectFile = await vscode.workspace.findFiles('**/*.csproj', path.dirname(host));
+		let projectFile = await vscode.workspace.findFiles(new vscode.RelativePattern(path.dirname(host), '*.csproj'));
 		projectFile.filter(file => path.dirname(file.fsPath) === path.dirname(host) ? projFiles.push(file?.fsPath) : projFiles);
 	}
 	return projFiles.length > 0 ? projFiles : undefined;
--- a/src/azureFunction/azureFunctionUtils.ts
+++ b/src/azureFunction/azureFunctionUtils.ts
@@ -203,7 +203,7 @@ export async function getAzureFunctionProjectFiles(): Promise<string[] | undefin
 		return undefined;
 	}
 	for (let host of hostFiles) {
-		let projectFile = await vscode.workspace.findFiles('*.csproj', path.dirname(host));
+		let projectFile = await vscode.workspace.findFiles('**/*.csproj', path.dirname(host));
 		projectFile.filter(file => path.dirname(file.fsPath) === path.dirname(host) ? projFiles.push(file?.fsPath) : projFiles);
 	}
 	return projFiles.length > 0 ? projFiles : undefined;


src/azureFunction/azureFunctionUtils.ts
```diff
@@ -203,7 +203,7 @@ export async function getAzureFunctionProjectFiles(): Promise<string[] | undefin
 		return undefined;
 	}
 	for (let host of hostFiles) {
-		let projectFile = await vscode.workspace.findFiles('*.csproj', path.dirname(host));
+		let projectFile = await vscode.workspace.findFiles('**/*.csproj', path.dirname(host));
```
Charles-Gagnon, line 206: So this is supposed to be finding the csproj that are in the folder containing the host.json file, which in the issue filed seems to be the case (both are contained in the subfolder). And 

Was that not working or something?

VasuBhog: So I found that the issue is that even though we get the proper host file location, we do not get the proper csproj file associated with it based on the host location. This is an issue as now the find files properly search the host file directory.

I agree that the other issue in regards to both host.json and csproj being on the same level may not always be guaranteed. But as of now, I am not sure what else we could utilize that could strictly indicate that the folder is an azure function project.

Charles-Gagnon: Oh, so the issue here is that the second parameter isn't the root folder to search from. 

https://github.com/microsoft/azuredatastudio/blob/main/src/vs/vscode.d.ts#L10652

`		export function findFiles(include: GlobPattern, exclude?: GlobPattern | null, maxResults?: number, token?: CancellationToken): Thenable<Uri[]>;`

```
		 * @param exclude  A {@link GlobPattern glob pattern} that defines files and folders to exclude. The glob pattern
		 * will be matched against the file paths of resulting matches relative to their workspace. When `undefined`, default excludes and the user's
		 * configured excludes will apply. When `null`, no excludes will apply.
```

So what we have there currently is incorrect and should be removed.

Instead you should look at using a [RelativePattern](https://github.com/microsoft/azuredatastudio/blob/main/src/vs/vscode.d.ts#L1932) and specify the root folder there.

VasuBhog: If it is at the root folder then it could possibly retrieve the wrong project file since right now we are specifying that a host.json and project file needs to be on the same level to ensure that it is an azure function. Right now the second parameter searches the given host.json file parent directory to see if there is also a csproj and if not then we show the user a warning.

VasuBhog: Hmm, this is a bit odd on how it's currently retrieving it correctly. But I will be sure to update, I see that the second parameter is supposed to be exclude. So should I be using the host.json parent folder (instead of the root folder) in the relative pattern instead to see if there is csproj located on the same level?

Charles-Gagnon (approved):

Charles-Gagnon:
Can you do a quick check through the other repos to make sure we're using this correctly there as well?

VasuBhog:
> Can you do a quick check through the other repos to make sure we're using this correctly there as well?

Did not see any issues in ADS.

barbaravaldez (approved):

PR #17281, opened 2022-03-01
Cleanup CHANGELOG and remove change log from README

Charles-Gagnon:
Fixed indentation in CHANGELOG (version h2 with "what's new" and "contributions" at h3 for each version)

Removed change log entries from README - it just adds clutter and isn't necessary with CHANGELOG existing


erpett (approved):
Thank you for saving me from future duplicity

abist:
@Charles-Gagnon We've always had change logs, is it fine to remove @kburtram ?

Charles-Gagnon:
Not sure what you're saying - we still have the CHANGELOG file. It's just not in the README. It's definitely not standard to have the changelog in the README - especially when you have a separate file for that already. 

(VS Code extension gallery even has a separate tab for the changelog which it gets from the CHANGELOG.md)

image

PR #17282, opened 2022-03-02
Include database in connection string when creating az func from table

### LINKED ISSUE #17267: add SQL binding from connection to master uses master in connection string added to local.settings.json
- MSSQL Extension Version: RC for 1.13
Version: 1.64.2 (Universal)
Commit: f80445acd5a3dadef24aa209168452a3d97cc326
Date: 2022-02-09T22:00:58.347Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Darwin arm64 21.3.0

Steps to Reproduce:

1. connect to a SQL server instance at master
2. navigate to an app db in the object explorer
3. create an AF input binding for a table
4. connection string is accessed and added to local.settings.json for functions project with Initial Catalog=master, which may be missed by a user expecting it to be the database the table is under (required for the binding to work)

VasuBhog:
Fixes #17267

After looking at the connection node that we get back from right-clicking a table and doing `Create Azure function with SQL binding" I found that the database property in the node.connectionInfo was undefined. 

As such, I found that we can use the label from the two-parent nodes above (first parent node label being table name) to get the database name. Once we set that, and pass it to STS we will get back the connection string to include Initial Catalog='database' (master or user's db).

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -443,7 +443,11 @@ export default class MainController implements vscode.Disposable {
 
 		// Generate Azure Function command
 		this._context.subscriptions.push(vscode.commands.registerCommand(Constants.cmdCreateAzureFunction, async (node: TreeNodeInfo) => {
-			const connectionDetails = ConnectionCredentials.createConnectionDetails(node.connectionInfo);
+			let connectionInfo = node.connectionInfo;
+			// set the database containing the selected table so it can be used
+			// for the initial catalog property of the connection string
+			connectionInfo.database = node.parentNode.parentNode.label as string;
+			const connectionDetails = ConnectionCredentials.createConnectionDetails(connectionInfo);
 			const connectionString = await this._connectionMgr.getConnectionString(connectionDetails, false, false);
 			await this.azureFunctionsService.createAzureFunction(connectionString, node.metadata.schema, node.metadata.name);
 		}));
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -446,7 +446,7 @@ export default class MainController implements vscode.Disposable {
 			let connectionInfo = node.connectionInfo;
 			// set the database containing the selected table so it can be used
 			// for the initial catalog property of the connection string
-			connectionInfo.database = node.parentNode.parentNode.label as string;
+			connectionInfo.database = node.parentNode.parentNode.metadata.name;
 			const connectionDetails = ConnectionCredentials.createConnectionDetails(connectionInfo);
 			const connectionString = await this._connectionMgr.getConnectionString(connectionDetails, false, false);
 			await this.azureFunctionsService.createAzureFunction(connectionString, node.metadata.schema, node.metadata.name);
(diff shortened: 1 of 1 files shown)


lucyzhang929 (approved):

src/controllers/mainController.ts
```diff
@@ -443,7 +443,11 @@ export default class MainController implements vscode.Disposable {
 
 		// Generate Azure Function command
 		this._context.subscriptions.push(vscode.commands.registerCommand(Constants.cmdCreateAzureFunction, async (node: TreeNodeInfo) => {
-			const connectionDetails = ConnectionCredentials.createConnectionDetails(node.connectionInfo);
+			let connectionInfo = node.connectionInfo;
+			// set the database containing the selected table so it can be used
+			// for the initial catalog property of the connection string
+			connectionInfo.database = node.parentNode.parentNode.label as string;
```
Charles-Gagnon, line 449: You shouldn't use the label - that may have additional text added to it. e.g. for external tables : 

image

Use node.metadata.name instead.

Charles-Gagnon (changes requested):

barbaravaldez (approved):

src/controllers/mainController.ts
```diff
@@ -443,7 +443,11 @@ export default class MainController implements vscode.Disposable {
 
 		// Generate Azure Function command
 		this._context.subscriptions.push(vscode.commands.registerCommand(Constants.cmdCreateAzureFunction, async (node: TreeNodeInfo) => {
-			const connectionDetails = ConnectionCredentials.createConnectionDetails(node.connectionInfo);
+			let connectionInfo = node.connectionInfo;
+			// set the database containing the selected table so it can be used
+			// for the initial catalog property of the connection string
+			connectionInfo.database = node.parentNode.parentNode.metadata.name;
```
Charles-Gagnon, line 449: You should also be walking up the tree until you find a node with a nodeType === 'Database' and then grabing the name from that node. Tables aren't always guaranteed to be 2 levels below the database node, e.g. for system tables : 

image

src/controllers/mainController.ts
```diff
@@ -443,7 +443,19 @@ export default class MainController implements vscode.Disposable {
 
 		// Generate Azure Function command
 		this._context.subscriptions.push(vscode.commands.registerCommand(Constants.cmdCreateAzureFunction, async (node: TreeNodeInfo) => {
-			const connectionDetails = ConnectionCredentials.createConnectionDetails(node.connectionInfo);
+			let connectionInfo = node.connectionInfo;
+			// set the database containing the selected table so it can be used
+			// for the initial catalog property of the connection string
+			let newNode: TreeNodeInfo = node;
+			while (true) {
```
Charles-Gagnon, line 450: `while(newNode)` - if for some reason there isn't a db node then we should still continue on and use the default connection db.

Charles-Gagnon (approved):

PR #17287, opened 2022-03-05
Added Extension APIs to get azure account, access token and firewall rule prompt 

llali:
Added APIs to mssql extension to use in SQL project extension

- New APIs to prompt for firewall rule if connection fails with firewall rule error
- New APIs to return azure account and token


src/azure/azureController.ts
```diff
@@ -117,48 +117,61 @@ export class AzureController {
...
+			let azureDeviceCode = await this.createDeviceCode();
+			tenantId = tenantId ? tenantId : azureDeviceCode.getHomeTenant(account).id;
+			token = await azureDeviceCode.getAccountSecurityToken(
+				account, tenantId, settings
+			);
+		}
+		return token;
+	}
+
+	public async getTokens(profile: ConnectionProfile, accountStore: AccountStore, settings: AADResource): Promise<ConnectionProfile> {
```
Charles-Gagnon, line 155: `populateAccountProperties` or something seems like a better name. Please also add doc comments for these new helper functions you're adding.

src/extension.ts
```diff
@@ -64,7 +65,12 @@ export async function activate(context: vscode.ExtensionContext): Promise<IExten
 		azureFunctions: controller.azureFunctionsService,
 		getConnectionString: (connectionUriOrDetails: string | ConnectionDetails, includePassword?: boolean, includeApplicationName?: boolean) => {
 			return controller.connectionManager.getConnectionString(connectionUriOrDetails, includePassword, includeApplicationName);
-		}
+		},
+		promptForFirewallRule: async (connectionUri: string, connectionInfo: IConnectionInfo) => {
+			const connectionProfile = new ConnectionProfile(connectionInfo);
+			return await this._connectionMgr.connectionUI.addFirewallRule(connectionUri, connectionProfile);
```
Charles-Gagnon, line 71: You shouldn't need to await here - you can just return directly

src/views/connectionUI.ts
```diff
@@ -488,6 +482,22 @@ export class ConnectionUI {
 		});
 	}
 
+	public async addFirewallRule(uri: string, profile: IConnectionProfile): Promise<boolean> {
+		const self = this;
```
Charles-Gagnon, line 486: This shouldn't be necessary (especially since you're using `this` in the code below already anyways...)

typings/vscode-mssql.d.ts
```diff
@@ -61,6 +66,14 @@ declare module 'vscode-mssql' {
 		 */
 		connect(connectionInfo: IConnectionInfo, saveConnection?: boolean): Promise<string>;
 
+		/**
+		 * Prompts the user to add firewall rule if connection failed with firerule error.
+		 * @param connectionUri The URI of the connection to list the databases for.
```
Charles-Gagnon, line 71: This param description doesn't seem correct

typings/vscode-mssql.d.ts
```diff
@@ -284,6 +297,107 @@ declare module 'vscode-mssql' {
 		validateStreamingJob(packageFilePath: string, createStreamingJobTsql: string): Thenable<ValidateStreamingJobResult>;
 	}
 
+	/**
+ * Represents a key that identifies an account.
```
Charles-Gagnon, line 301: Fix indentation

typings/vscode-mssql.d.ts
```diff
@@ -284,6 +297,107 @@ declare module 'vscode-mssql' {
...
+	export interface Token extends AccessToken {
+		/**
+		 * TokenType
+		 */
+		tokenType: string;
+	}
+
+	export interface IAzureAccountService {
+		getAccount(): Promise<IAccount>;
+		getAccountSecurityToken(account: IAccount, tenantId: string | undefined): Thenable<Token>;
```
Charles-Gagnon, line 398: We should be consistent and make this return a Promise

yarn.lock
```diff
@@ -43,23 +43,23 @@
   integrity sha1-/JQj+H/K5BigrjuNVsYV2TVj1E8=
 
 "@babel/code-frame@^7.0.0":
-  version "7.14.5"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.14.5.tgz#23b08d740e83f49c5e59945fbf1b43e80bbf4edb"
-  integrity sha512-9pzDqyc6OLDaqe+zbACgFkb6fKMNG6CObKpnYXChRsvYGyEdc7CA2BaqeOM+vOtCS5ndmJicPJhKAwYRI6UfFw==
+  version "7.16.7"
```
Charles-Gagnon, line 46: Please revert these changes if they aren't relevant to your PR

Charles-Gagnon (commented):
@llali Did you forget to add reviewers to this?

alanrenmsft:
@cssuh these are mostly AAD related changes, please take a look.

src/azure/azureController.ts
```diff
@@ -117,48 +117,64 @@ export class AzureController {
 		await this.prompter.promptSingle(tenantQuestion, true);
 	}
 
-	public async getTokens(profile: ConnectionProfile, accountStore: AccountStore, settings: AADResource): Promise<ConnectionProfile> {
+	public async getAccount(accountStore: AccountStore): Promise<IAccount> {
```
alanrenmsft, line 120: 1. The method is actually doing add account operation, right? if so we should rename the method.
2. it isweird that I have to pass in the `accountStore` object, add account to it and then return the account. the AccountStore should be globally accessible.

typings/vscode-mssql.d.ts
```diff
@@ -61,6 +66,14 @@ declare module 'vscode-mssql' {
 		 */
 		connect(connectionInfo: IConnectionInfo, saveConnection?: boolean): Promise<string>;
 
+		/**
+		 * Prompts the user to add firewall rule if connection failed with firerule error.
+		 * @param connectionUri The URI of the connection to add firewall rule to.
+		 * @param connectionInfo The connection info
+		 * @returns True if firewall rulle added
```
Charles-Gagnon, line 73: rule

src/azure/azureController.ts
llali: this method was already in the code I just divided to two methods. The previous one would add account and generate token. Now there are methods one add Account and other generates token

typings/vscode-mssql.d.ts
```diff
@@ -61,6 +66,14 @@ declare module 'vscode-mssql' {
 		 */
 		connect(connectionInfo: IConnectionInfo, saveConnection?: boolean): Promise<string>;
 
+		/**
+		 * Prompts the user to add firewall rule if connection failed with firerule error.
```
Charles-Gagnon, line 70: `failed with a firewall error`

typings/vscode-mssql.d.ts
```diff
@@ -292,6 +305,136 @@ declare module 'vscode-mssql' {
...
+		 */
+		key: IAccountKey;
+		/**
+		 * Display information for the account
+		 */
+		displayInfo: IAccountDisplayInfo;
+		/**
+		 * Custom properties stored with the account
+		 */
+		properties: any;
```
Charles-Gagnon, line 387: Can this be given a better type - even just an object with string key/value pairs?

src/extension.ts
```diff
@@ -14,9 +14,17 @@ import { Deferred } from './protocol';
 import * as utils from './models/utils';
 import { ObjectExplorerUtils } from './objectExplorer/objectExplorerUtils';
 import SqlToolsServerClient from './languageservice/serviceclient';
+import { ConnectionProfile } from './models/connectionProfile';
 
 let controller: MainController = undefined;
 
+class FirewallRuleError extends Error {
```
Charles-Gagnon, line 21: This should go somewhere else - it's own file or generic utils file or something

src/azure/azureController.ts
```diff
@@ -117,48 +117,64 @@ export class AzureController {
...
-			let tid = profile.tenantId ? profile.tenantId : azureDeviceCode.getHomeTenant(account).id;
-			const token = await azureDeviceCode.getAccountSecurityToken(
-				account, tid, settings
+		}
+
+		return account;
+	}
+
+	public async getAccountSecurityToken(account: IAccount, tenantId: string | undefined, settings: AADResource): Promise<Token> {
+		let token;
```
Charles-Gagnon, line 137: Give this a type

Charles-Gagnon (approved):

Charles-Gagnon:
I assume you've done plenty of testing of AAD connections with these changes?

llali:
> I assume you've done plenty of testing of AAD connections with these changes?

these changes won't affect anything in AAD connection. There are just APIs so other extensions can get azure account and generate token

src/extension.ts
Charles-Gagnon: And actually - it would be good to add an interface for this in the extension typings file so that others using this can catch and handle these errors more easily.

src/azure/azureController.ts
```diff
@@ -117,48 +117,64 @@ export class AzureController {
...
+			await this.promptForTenantChoice(account, profile);
+		}
+
+		const token = await this.getAccountSecurityToken(
+			account, profile.tenantId, settings
+		);
+
+		if (!token) {
+			let errorMessage = LocalizedConstants.msgGetTokenFail;
+			this._vscodeWrapper.showErrorMessage(errorMessage);
```
alanrenmsft, line 171: we need to throw exception and stop the execution here, right?

cssuh:
I agree that we should probably put this through a full testing run before merging

llali:
> I agree that we should probably put this through a full testing run before merging

I did test connecting with AAD auth but if you send me other scenarios, I'll do a full testing. I assume the automated tests won't cover the scenarios?

cssuh:
Here are the 5 things to test:
- Add azure account
- Remove azure account
- Connect to azure DB (new connection)
- Connect to azure DB (existing connection)
- Add firewall rule

I can also help with the testing if needed

Charles-Gagnon:
@cssuh Are these added to the release test plan? It would be great to use that (and add anything that's missing if it's not completed) so that we have an easy reference for these in the future.

cssuh:
> @cssuh Are these added to the release test plan? It would be great to use that (and add anything that's missing if it's not completed) so that we have an easy reference for these in the future.

Yes I got these from the vscode-mssql ADO test plan

src/services/azureAccountServie.ts
```diff
@@ -0,0 +1,30 @@
+/* --------------------------------------------------------------------------------------------
```
abist, line 1: File name typo

typings/vscode-mssql.d.ts
llali: I'm worried it breaks some scenarios for AAD.

src/extension.ts
llali: I already added interface. will add the "implements"

src/azure/azureController.ts
```diff
@@ -117,48 +117,64 @@ export class AzureController {
...
+		return token;
+	}
+
+	/**
+	 * Gets the token for given account and updates the connection profile with token infromation needed for AAD authetnication
+	 */
+	public async populateAccountProperties(profile: ConnectionProfile, accountStore: AccountStore, settings: AADResource): Promise<ConnectionProfile> {
+		let account = await this.getAccount(accountStore);
+
+		if (!profile.tenantId) {
```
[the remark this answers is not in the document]
llali: this is again existing logic but I don't mind adding a check and error message. should we prompt for tenant in that case. @cssuh please let me know what you think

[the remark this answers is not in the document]
Charles-Gagnon: Let's do any changes in a separate PR please so it can be cleanly reviewed.

abist (approved):

alanrenmsft (approved):

src/azure/azureController.ts
```diff
@@ -117,48 +117,65 @@ export class AzureController {
...
-			);
-			if (!token) {
-				let errorMessage = LocalizedConstants.msgGetTokenFail;
-				this._vscodeWrapper.showErrorMessage(errorMessage);
-			}
-			profile.azureAccountToken = token.token;
-			profile.expiresOn = token.expiresOn;
-			profile.email = account.displayInfo.email;
-			profile.accountId = account.key.id;
 		} else if (config === utils.azureAuthTypeToString(AzureAuthType.DeviceCode)) {
```
llali, line 127: @cssuh before my change when config was AzureAuthType.DeviceCode, the getToken method didn't set the account and token in the profile. Should I keep it that way or was it a bug?

cssuh: hmm, that seems like a bug.  I think it should be set.

llali: ok I keep the change then

llali:
> Here are the 5 things to test:
> 
> * Add azure account
> * Remove azure account
> * Connect to azure DB (new connection)
> * Connect to azure DB (existing connection)
> * Add firewall rule
> 
> I can also help with the testing if needed

@cssuh I verified all these scenarios still work as expected. The only difference I see if, when I login to my AAD account, I see a prompt for selecting tenant. I think it's expected since I have two tenants. but in existing version of mssql, I don't see any prompt using the same account. I don't know if it's a bug in the current version and fixed in my change or if I broke anything. Can you please take look at the code and let me know if you see anything wrong? Thanks

Charles-Gagnon:
@llali That's expected - that was a feature added since the last release : https://github.com/microsoft/vscode-mssql/pull/17266

So great to hear that we have additional validation that it's working!

PR #17288, opened 2022-03-07
Remove code coverage badge

Charles-Gagnon:
Right now it just displays "unknown". Seems better to remove it for now and then add it back in when we get around to fixing the coverage numbers.


abist (approved):
It's unfortunate we lost this, it used to work earlier. But agreed, better to remove than have an unknown badge.

PR #17295, opened 2022-03-14
Extend MSSQL api to prompt user for binding type

barbaravaldez:
Moving the addSqlBindingQuickpick logic here.

 https://github.com/microsoft/azuredatastudio/blob/main/extensions/sql-bindings/src/dialogs/addSqlBindingQuickpick.ts#L63


src/constants/constants.ts
```diff
@@ -171,3 +171,6 @@ export const defaultBindingResult = 'return new OkObjectResult(responseMessage);
 export const sqlBindingResult = `return new OkObjectResult(result);`;
 export const azureFunctionLocalSettingsFileName = 'local.settings.json';
 export const sqlExtensionPackageName = 'Microsoft.Azure.WebJobs.Extensions.Sql';
+export const input = 'Input';
```
Charles-Gagnon, line 174: These should be localized

Charles-Gagnon (changes requested):
What is this for? And is this temporary or something? The code for this should be living in the sql-bindings extension eventually, not here.

PR #17296, opened 2022-03-15
generic way to send STS requests using mssql API

VasuBhog:
Add a way to send STS requests and get back the response. Will be used in SQL bindings extension to send addSqlBinding send request.

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -295,8 +295,8 @@ export default class ConnectionManager {
 	 * @param params The params to pass with the request
 	 * @returns A thenable object for when the request receives a response
 	 */
-	public sendRequest<P, R, E, R0>(requestType: RequestType<P, R, E, R0>, params?: P): Thenable<R> {
-		return this.client.sendRequest(requestType, params);
+	public async sendRequest<P, R, E, R0>(requestType: RequestType<P, R, E, R0>, params?: P): Promise<R> {
+		return await this.client.sendRequest(requestType, params);
 	}
 
 	/**
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -16,7 +16,7 @@ import StatusView from '../views/statusView';
 import SqlToolsServerClient from '../languageservice/serviceclient';
 import { IPrompter } from '../prompts/question';
 import VscodeWrapper from './vscodeWrapper';
-import { NotificationHandler } from 'vscode-languageclient';
+import { NotificationHandler, RequestType } from 'vscode-languageclient';
 import { Runtime, PlatformInformation } from '../models/platform';
 import { Deferred } from '../protocol';
 import { AccountService } from '../azure/accountService';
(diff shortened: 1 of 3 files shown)


src/controllers/connectionManager.ts
```diff
@@ -289,6 +289,18 @@ export default class ConnectionManager {
...
 	}
 
+	/**
+	 * Send a request to the SQL Tools Server client
+	 * @param type The of the request to make
+	 * @param params The params to pass with the request
+	 * @returns A thenable object for when the request receives a response
+	 */
+	public sendRequest<P, R, E, R0>(requestType: RequestType<P, R, E, R0>, params?: P): Thenable<R> {
+		if (this.client !== undefined) {
```
Charles-Gagnon, line 299: `this.client?.sendRequest(...)`

But also - why is this necessary? I wouldn't expect client to ever be undefined. It's passed in or set in the constructor.

src/controllers/connectionManager.ts
```diff
@@ -289,6 +289,18 @@ export default class ConnectionManager {
 		return ConnectionCredentials.createConnectionDetails(connectionInfo);
 	}
 
+	/**
+	 * Send a request to the SQL Tools Server client
+	 * @param type The of the request to make
+	 * @param params The params to pass with the request
+	 * @returns A thenable object for when the request receives a response
+	 */
+	public sendRequest<P, R, E, R0>(requestType: RequestType<P, R, E, R0>, params?: P): Thenable<R> {
```
Charles-Gagnon, line 298: We generally shouldn't be using Thenables in our code - that's only something we do for extension APIs. This should be returning `Promise<R>`

src/controllers/connectionManager.ts
```diff
@@ -289,6 +289,18 @@ export default class ConnectionManager {
 		return ConnectionCredentials.createConnectionDetails(connectionInfo);
 	}
 
+	/**
+	 * Send a request to the SQL Tools Server client
+	 * @param type The of the request to make
```
Charles-Gagnon, line 294: requestType

src/controllers/connectionManager.ts
VasuBhog: The issue here is that the returned sendRequest of type Thenable ([serverclient.ts](https://github.com/microsoft/vscode-mssql/blob/main/src/languageservice/serviceclient.ts#L380)) which then also calls the sendRequest from vscode-languageclient. They are all of type Thenable.

Charles-Gagnon: You can solve this by making this function async, awaiting the call to sendRequest and then returning the result

typings/vscode-mssql.d.ts
```diff
@@ -3,9 +3,11 @@
  *  Licensed under the Source EULA. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+
```
chlafreniere, line 6: nit: extra space

src/controllers/connectionManager.ts
```diff
@@ -289,6 +289,16 @@ export default class ConnectionManager {
 		return ConnectionCredentials.createConnectionDetails(connectionInfo);
 	}
 
+	/**
+	 * Send a request to the SQL Tools Server client
+	 * @param requestType The of the request to make
```
chlafreniere, line 294: nit: "The type of the request"

Can we make this a bit clearer, by the way?

src/controllers/connectionManager.ts
```diff
@@ -289,6 +289,16 @@ export default class ConnectionManager {
...
+	 * @param params The params to pass with the request
+	 * @returns A thenable object for when the request receives a response
```
chlafreniere, line 296: promise not thenable

src/controllers/connectionManager.ts
VasuBhog: Not sure how much clearer since this is supposed to be generic to be used by many different contract request types

chlafreniere (approved):

PR #17297, opened 2022-03-15
Removing all logic of sql bindings

VasuBhog:
Remove all sql-bindings from vscode-mssql as it is now in the sql-bindings repo. 
This will be the first step to remove and then bundle the sql-bindings extension in the extension pack.


typings/sql-bindings.d.ts
```diff
@@ -0,0 +1,102 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the Source EULA. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+
+declare module 'sql-bindings' {
```
Charles-Gagnon, line 7: Why is this here? The extension shouldn't need to have this.

VasuBhog: Yeah, I can delete the typing files as well now.

Charles-Gagnon (approved):
Make sure you go through the package dependencies and see if there's any of those we can remove as well

VasuBhog:
> Make sure you go through the package dependencies and see if there's any of those we can remove as well

Checked and found there are no package dependencies that we can remove.

PR #17305, opened 2022-04-04
Downgrading a package

aasimkhan30:


yarn.lock
```diff
@@ -1714,7 +1714,7 @@ error-ex@^1.2.0, error-ex@^1.3.0:
   dependencies:
     is-arrayish "^0.2.1"
 
-es5-ext@^0.10.12, es5-ext@^0.10.35, es5-ext@^0.10.46, es5-ext@^0.10.50:
+es5-ext@0.10.53, es5-ext@^0.10.12, es5-ext@^0.10.35, es5-ext@^0.10.46, es5-ext@^0.10.50:
```
Charles-Gagnon, line 1717: As mentioned offline, we're locked to the .53 version so this doesn't actually affect us.

Charles-Gagnon (changes requested):

PR #17312, opened 2022-04-14
Fix tests and add to pipeline

abist:
This PR fixes tests, adds the tests to the pipeline and adds code coverage to the ADO pipelines

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -260,7 +260,7 @@ gulp.task('ext:test', async (done) => {
 		console.error(`exec error: ${error}`);
 	}
 	done();
-	process.exit(1);
+	process.exit(0);
 });
 
 gulp.task('test', gulp.series('ext:test'));
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -237,7 +237,7 @@ gulp.task('ext:localization', gulp.series('ext:localization:generate-eng-package
 
 gulp.task('ext:build', gulp.series('ext:localization', 'ext:copy', 'ext:clean-library-ts-files', 'ext:compile', 'ext:compile-view')); // removed lint before copy
 
-gulp.task('ext:test', (done) => {
+gulp.task('ext:test', async (done) => {
 	let workspace = process.env['WORKSPACE'];
 	if (!workspace) {
 		workspace = process.cwd();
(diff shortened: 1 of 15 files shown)


gulpfile.js
```diff
@@ -246,21 +246,21 @@ gulp.task('ext:test', (done) => {
...
+		await vscodeTest.runTests({
+			vscodeExecutablePath: vscodePath,
+			extensionDevelopmentPath: workspace,
+			extensionTestsPath: extensionTestsPath,
+			launchArgs: args
+		});
+	} catch (error) {
+		console.log(`stdout: ${process.stdout}`);
+		console.log(`stderr: ${process.stderr}`);
+		console.error(`exec error: ${error}`);
```
Charles-Gagnon, line 260: Shouldn't we fail this task if the tests fail?

gulpfile.js
```diff
@@ -246,21 +246,21 @@ gulp.task('ext:test', (done) => {
...
+	}
+	done();
+	process.exit(0);
```
Charles-Gagnon, line 263: What's the exit for here? I would expect that the gulp task would exit after returning from this

abist: It doesn't since the vscode starts a process and the gulp task is left hanging after finishing.

gulpfile.js
abist: The task does fail because of inner assertion errors etc

Charles-Gagnon: What do you mean "inner assertion errors"? How is the gulp task being notified of these errors if we're catching everything here?

gulpfile.js
Charles-Gagnon: What process? The extension host process it runs tests in?

gulpfile.js
```diff
@@ -246,21 +246,21 @@ gulp.task('ext:test', (done) => {
...
-				launchArgs: args
-			}).then(() => done()).catch((error) => {
-				console.log(`stdout: ${process.stdout}`);
-				console.log(`stderr: ${process.stderr}`);
-				console.error(`exec error: ${error}`);
-				done(error);
-			});
-		}
-	})
+	let vscodePath = await vscodeTest.downloadAndUnzipVSCode(vscodeVersion);
```
Charles-Gagnon, line 249: We should probably always be downloading the latest stable to verify we're working against that.

gulpfile.js
abist: it starts a vscode instance to run these tests so it's probably the extension host's child

gulpfile.js
abist: It's handled by the vscode-test module. The original idea was to capture all the errors and not just fail after one error, I just refactored it.

Charles-Gagnon: I assume the "not fail after one error" was you removing the original `done(error)` call? But right now I'm not seeing you ever call done with an error which seems to me to imply that this task will never fail. Did you test this with purposely making a test fail to ensure that it fails the task if a test fails?

gulpfile.js
abist: We download the same version as in the package.json. We should probably update the vscode engine in a separate PR.

Charles-Gagnon: No, because that's the minimum version we support. That doesn't (and shouldn't) need to be the latest version so that users who don't want to/can't update VS Code for some reason can still use our extension if they want.

Charles-Gagnon: But for testing we should just focus on the latest release - we could run tests against older versions if we wanted but generally that's not worth the extra time and effort.

gulpfile.js
abist: yes, I tested it and it ends with an error and prints out the test with the error (exit code is 1 in that case)

gulpfile.js
abist: @Charles-Gagnon I tried updating it to latest and some tests are failing (non functional ones with no side-effects). I didn't get much time to investigate since I'm working on other tasks, but this can be added as another task for later. I'd like to get the tests running in the pipelines initially. Thoughts @kburtram ?

Charles-Gagnon: This seems like a good task for the DRI to take up

Charles-Gagnon (approved):

PR #17321, opened 2022-04-25
Adding extension APIs for managing Azure Resources (subscriptions, resource groups, locations and SQL server)

llali:
Adding extension APIs for managing Azure Resources (subscriptions, resource groups, locations and SQL server). The APIs only include functions to get the list of subscriptions, resource groups, locations and a function to create a SQL server in Azure. These APIs are used in SQL projects (https://github.com/microsoft/azuredatastudio/pull/18833)


src/azure/azureController.ts
```diff
@@ -279,4 +318,15 @@ export class AzureController {
 		}
 	}
 
+	public set SubscriptionClient(v: SubscriptionClient) {
```
Charles-Gagnon, line 321: What is this for? We don't want to use the same client for the entire azureController - many different things can call into that

src/azure/azureController.ts
```diff
@@ -239,6 +249,35 @@ export class AzureController {
...
+						token: token
+					};
+				});
+				subscriptions = subscriptions.concat(array);
+			}
+
+			return subscriptions.sort((a, b) => (a.subscription.displayName || '').localeCompare(b.subscription.displayName || ''));
+		} catch (error) {
+
+			return [];
```
Charles-Gagnon, line 277: Why are you returning an empty array here? Seems reasonable to let the exception bubble up so the caller knows something went wrong.

src/azure/azureController.ts
```diff
@@ -239,6 +249,35 @@ export class AzureController {
 		}
 	}
 
+	/**
+	 * Returns Azure subscriptions with tenant and token for each given account
+	 */
+	public async getAccountSessions(account: IAccount): Promise<mssql.IAzureAccountSession[]> {
```
Charles-Gagnon, line 255: Why is this here? You already have a subscription client in the azureResourceController so it seems better to just use that one

src/azure/azureResourceController.ts
```diff
@@ -0,0 +1,96 @@
...
+import * as mssql from 'vscode-mssql';
+import * as azureUtils from './utils';
+import { TokenCredentialWrapper } from "./TokenCredentialWrapper";
+
+export class AzureResourceController {
+	private _subscriptionClient: SubscriptionClient | undefined;
+	private _resourceManagementClient: ResourceManagementClient | undefined;
+	private _sqlManagementClient: SqlManagementClient | undefined;
+
+	public set SubscriptionClient(v: SubscriptionClient) {
```
Charles-Gagnon, line 18: Same here - we shouldn't ever be setting the client on the entire controller like this.

src/azure/azureResourceController.ts
```diff
@@ -0,0 +1,96 @@
...
+	public set ResourceManagementClient(v: ResourceManagementClient) {
+		this._resourceManagementClient = v;
+	}
+
+	public set SqlManagementClient(v: SqlManagementClient) {
+		this._sqlManagementClient = v;
+	}
+
+	/**
+	 * Returns Azure locations for given subscription
```
Charles-Gagnon, line 31: `for the given session`?

What is a session anyways in this context?

src/azure/azureResourceController.ts
```diff
@@ -0,0 +1,96 @@
...
+			}
+		}
+		return undefined;
+	}
+
+	/**
+	 * Returns Azure resource groups for given subscription
+	 */
+	public async getResourceGroups(session: mssql.IAzureAccountSession): Promise<ResourceGroup[]> {
+		if (session?.subscription?.subscriptionId) {
```
Charles-Gagnon, line 63: 1. session isn't marked as optional in the signature, which it should be if you actually intend on that being a valid case
2. Is it even a valid case for any of these to be undefined? It seems like you should be throwing an error instead of returning an empty array in that case so the user knows something was wrong

src/services/azureAccountService.ts
```diff
@@ -21,10 +21,19 @@ export class AzureAccountService implements mssql.IAzureAccountService {
...
 	public async getAccounts(): Promise<mssql.IAccount[]> {
 		return await this._accountStore.getAccounts();
 	}
+
 	public async getAccountSecurityToken(account: mssql.IAccount, tenantId: string | undefined): Promise<mssql.Token> {
 		return await this._azureController.getAccountSecurityToken(account, tenantId, providerSettings.resources.azureManagementResource);
 	}
+
+	/**
+	 * Returns Azure subscriptions with tenant and token for each given account
```
Charles-Gagnon, line 34: Shouldn't this be `Returns Azure account sessions...`?

src/services/azureResourceService.ts
```diff
@@ -0,0 +1,62 @@
...
+import providerSettings from '../azure/providerSettings';
+
+export class AzureResourceService implements mssql.IAzureResourceService {
+
+	private _accountStore: AccountStore;
+	constructor(
+		private _azureController: AzureController,
+		private _azureResourceController: AzureResourceController,
+		private _context: vscode.ExtensionContext) {
+		this._accountStore = new AccountStore(this._context);
```
Charles-Gagnon, line 23: We shouldn't be making our own account store. We should be using the one from the azure account service.

src/services/azureResourceService.ts
```diff
@@ -0,0 +1,62 @@
...
+	 public async createOrUpdateServer(session: mssql.IAzureAccountSession, resourceGroupName: string, serverName: string, parameters: Server): Promise<string | undefined> {
+		await this.checkAndRefreshToken(session);
+		return await this._azureResourceController.createOrUpdateServer(session.subscription.subscriptionId, resourceGroupName, serverName, parameters, session.token);
+	}
+
+	/**
+	 * Verifies if the token still valid, refreshes the token for given account
+	 * @param session
+	 */
+	private async checkAndRefreshToken(session: mssql.IAzureAccountSession) {
```
Charles-Gagnon, line 54: This should be a helper method in the azure controller

src/azure/azureController.ts
Charles-Gagnon: It seems like this is for testing purposes. It's much better to use the DI pattern where you inject a factory method into the constructor - something like this : 

```
export type SubscriptionClientFactory = (token: string) => string;
function defaultSubscriptionClientFactory(token: string) {
	return new SubscriptionClient(new TokenCredentialWrapper(token));
}
...
constructor(private _subscriptionClientFactory: SubscriptionClientFactory = defaultSubscriptionClientFactory) {
```

And then whenever you need to get the client you just call `this._subscriptionClientFactory(token)`

This removes a lot of the "extra" code and is a lot easier to understand since you're using the factory in the same way regardless of how it was implemented.

Charles-Gagnon (changes requested):

src/azure/azureController.ts
llali: if you are asking about why the function is here, I thought it's related to account and that's why I put it here. I was debating about having this function in this API or resources API and I was not sure at the end. as you see, in extension APIs, I put this in account service not resource service. a=if you think it make more sense to put it in resources, I can move it

src/azure/azureResourceController.ts
```diff
@@ -0,0 +1,68 @@
...
+			const locationsPages = await subClient.subscriptions.listLocations(session.subscription.subscriptionId);
+			let locations = await azureUtils.getAllValues(locationsPages, (v) => v);
+			return locations.sort((a, b) => (a.name || '').localeCompare(b.name || ''));
+		} else {
+			throw new Error('Invalid session');
+		}
+	}
+
+	/**
+	 * Creates or updates a Azure SQL server for given subscription, resource group and location
```
Charles-Gagnon, line 35: Please add doc comments for params and return statement

typings/vscode-mssql.d.ts
```diff
@@ -437,6 +452,29 @@ declare module 'vscode-mssql' {
...
+		/**
+		 * Returns Azure subscriptions with tenant and token for each given account
+		 */
+		getAccountSessions(account: IAccount): Promise<IAzureAccountSession[]>;
+	}
+
+	export interface IAzureResourceService {
+
+		/**
+		 * Returns Azure resource groups for given subscription
```
Charles-Gagnon, line 465: Please add doc comments for all these for the params and return values

Charles-Gagnon (approved):
Few minor things but otherwise looks good! Thanks for taking the time to split this out.

src/azure/azureResourceController.ts
```diff
@@ -0,0 +1,68 @@
...
+	 * Returns Azure locations for given session
+	 */
+	public async getLocations(session: mssql.IAzureAccountSession): Promise<Location[]> {
+		const subClient = this._subscriptionClientFactory(session.token);
+		if (session.subscription?.subscriptionId) {
+			const locationsPages = await subClient.subscriptions.listLocations(session.subscription.subscriptionId);
+			let locations = await azureUtils.getAllValues(locationsPages, (v) => v);
+			return locations.sort((a, b) => (a.name || '').localeCompare(b.name || ''));
+		} else {
+			throw new Error('Invalid session');
```
kisantia, line 30: localize

kisantia (approved):

src/azure/azureController.ts
```diff
@@ -239,6 +251,30 @@ export class AzureController {
 		}
 	}
 
+	/**
+	 * Returns Azure subscriptions with tenant and token for each given account
```
alanrenmsft, line 255: comment doesn't match the method name, also in the method body, you used subscriptions but the type is session, Subscription == Session?

src/azure/azureResourceController.ts
```diff
@@ -0,0 +1,68 @@
...
+
+	/**
+	 * Creates or updates a Azure SQL server for given subscription, resource group and location
+	 */
+	public async createOrUpdateServer(
+		subscriptionId: string,
+		resourceGroupName: string,
+		serverName: string,
+		parameters: Server,
+		token: mssql.Token): Promise<string | undefined> {
```
abist, line 42: this should be `Promise<string> | undefined`

src/azure/azureController.ts
```diff
@@ -279,4 +315,20 @@ export class AzureController {
...
 
+
+	/**
+	 * Verifies if the token still valid, refreshes the token for given account
+	 * @param session
+	 */
+	public async checkAndRefreshToken(
+		session: mssql.IAzureAccountSession,
+		accountStore: AccountStore): Promise<void> {
+		const currentTime = new Date().getTime() / 1000;
```
alanrenmsft, line 326: what does / 1000 mean?

src/azure/azureResourceController.ts
```diff
@@ -0,0 +1,68 @@
...
+		if (subscriptionId && resourceGroupName) {
+			const sqlClient = this._sqlManagementClientFactory(token, subscriptionId);
+			if (sqlClient) {
+				const result = await sqlClient.servers.beginCreateOrUpdateAndWait(resourceGroupName,
+					serverName, parameters);
+
+				return result.fullyQualifiedDomainName;
```
alanrenmsft, line 49: what if the result is undefined?

src/azure/azureResourceController.ts
Charles-Gagnon: That's not correct. It always returns a promise (because it's async). The wrapped type is either `string | undefined`. So what she has is correct.

src/azure/azureResourceController.ts
llali: I noticed that we didn't localize the error coming from extension API. I thought it might be intentional, so I did the same

src/azure/azureController.ts
llali: copy the code somewhere else in mssql extension which we refresh token when it's expired. getTime is in milliseconds and to compare to expire it needs to get converted

src/azure/azureResourceController.ts
llali: I don't think that happens. Azure APIs throw error if fails to create server

alanrenmsft (approved):

PR #17322, opened 2022-04-26
Fix Prompt for Connection when user moves focus away

VasuBhog:
This was found from this issue: https://github.com/microsoft/azuredatastudio/issues/19108

This will respect the parameter that would be passed in via the API and keep the prompt open

### DIFF
--- a/src/views/connectionUI.ts
+++ b/src/views/connectionUI.ts
@@ -136,7 +136,7 @@ export class ConnectionUI {
 			matchOptions: options,
 			choices: choices
 		};
-		return this._prompter.promptSingle(question);
+		return this._prompter.promptSingle(question, question.matchOptions.ignoreFocusOut);
 	}
 
 	/**
--- a/src/views/connectionUI.ts
+++ b/src/views/connectionUI.ts
@@ -90,7 +90,7 @@ export class ConnectionUI {
 		const selection = await this.promptItemChoice({
 			placeHolder: LocalizedConstants.recentConnectionsPlaceholder,
 			matchOnDescription: true,
-			ignoreFocusOut
+			ignoreFocusOut: ignoreFocusOut
 		}, picklist);
 		if (selection) {
 			return this.handleSelectedConnection(selection);


src/views/connectionUI.ts
```diff
@@ -90,7 +90,7 @@ export class ConnectionUI {
 		const selection = await this.promptItemChoice({
 			placeHolder: LocalizedConstants.recentConnectionsPlaceholder,
 			matchOnDescription: true,
-			ignoreFocusOut
+			ignoreFocusOut: ignoreFocusOut
```
Charles-Gagnon, line 93: Hm, this should be fine like this. Are you sure this was the problem? Looking at the code it seems like the actual issue is that we aren't passing in `ignoreFocusOut` to `promptSingle` in `promptItemChoice`

https://github.com/microsoft/vscode-mssql/blob/81710de6b4b580d1e8734502fa965d6ac085205a/src/views/connectionUI.ts#L131

Notice that `promptSingle` has a parameter for ignoreFocusOut https://github.com/microsoft/vscode-mssql/blob/9b8f4644ad676d277613ffc2adc78a927d103617/src/prompts/adapter.ts#L70

Charles-Gagnon (changes requested):

src/views/connectionUI.ts
VasuBhog: This was the problem since I tested this a few times. We never passed in ignoreFocusOut to be other than false. If the parameter is never set in promptItemChoice then it will never be respected. 
I see that promptSingle has a separate parameter but the IQuestion takes in ignoreFocusOut so I believe that it is still set via the first parameter.

Charles-Gagnon: What do you mean we never "passed in ignoreFocusOut to be other than false"? Your change here shouldn't be changing anything - we still pass in an object that has a property on it called ignoreFocusOut with the boolean value from the parameter...which sure the default is false but that's up to the caller to set if they want. 

And what do you mean IQuestion "takes in ignoreFocusOut". IQuestion is a type so I'm not sure how it could "take in" anything. 

I'm looking at the code and I'm never seeing ignoreFocusOut coming from anywhere except the parameter to `promptSingle`. Look at https://github.com/microsoft/vscode-mssql/blob/9b8f4644ad676d277613ffc2adc78a927d103617/src/prompts/prompt.ts - it specifically sets the option there which is then used in the implementation in https://github.com/microsoft/vscode-mssql/blob/9b8f4644ad676d277613ffc2adc78a927d103617/src/prompts/input.ts

Can you point me to where you see ignoreFocusOut being set in the code with how you currently have it?

VasuBhog: Again this never seemed to work previously. `ignoreFocusOut` property in `promptItemChoice` was never set to the parameter passed into `promptForConnection`. 

An example of this is using the API for `promptForConnection(true)`. With my current change, we now set the property correctly to the passed-in parameter of `promptForConnection`. 

Shown below is where ignoreFocusOut is within the context of promptSingle. As you can see it is in matchOptions within the IQuestion object property. Therefore, even though the promptSingle parameter for IgnoreFocusOut is not set the option is still there and therefore should be respected since it is then passed into createPrompt
Screen Shot 2022-04-26 at 11 59 26 AM

If you are saying to change the way we set ignoreFocusOut in the chain from PromptForConnection to creating the prompt, I can do that. It just seems like a much larger change.

Charles-Gagnon: Did you actually follow this through though? What I'm saying is that while the option is indeed in the "options" for the prompt code it's never actually used to set the ignoreFocusOut on the quick pick itself. 

Put a breakpoint on https://github.com/microsoft/vscode-mssql/blob/9b8f4644ad676d277613ffc2adc78a927d103617/src/prompts/input.ts#L42 and see what the options there are - those are the actual ones that we pass to the quickpick.

VasuBhog: I dont see this ever getting hit.

Charles-Gagnon: Did you step into PromptFactory.createPrompt? I realized I got the question type mixed up - it's actually "expand" it looks like. So you should be setting the breakpoint on https://github.com/microsoft/vscode-mssql/blob/9b8f4644ad676d277613ffc2adc78a927d103617/src/prompts/expand.ts instead.

But the logic is the same there so I expect the same result.

VasuBhog: I had some issues with debugging. `Reload Window` does not refresh any code changes I did. IgnoreFocusOut comes from here when building the expandPrompt - https://github.com/microsoft/vscode-mssql/blob/main/src/prompts/prompt.ts#L27

Now that we pass in ignoreFocusOut to promptSingle it will set this appropriately.

Charles-Gagnon (approved):

PR #17342, opened 2022-05-19
DeploymentOptions model updates according to STS model changes

ssreerama:
These changes are in sync to STS deploymentOptions model changes which have been added new options and changed the model to support options values, description and displayNames. Thanks.

Pending work: Need to update STS version after STS changes are checked in.


typings/vscode-mssql.d.ts
```diff
@@ -489,85 +489,112 @@ declare module 'vscode-mssql' {
 		executeAndScript = 2
 	}
 
+	export interface DacDeployOptionPropertyBoolean {
```
kisantia, line 492: please add doc comments for describing what these interfaces are for

kisantia (approved):
Please make sure to also update the STS version in this PR before merging.

typings/vscode-mssql.d.ts
ssreerama: done

ssreerama:
Verified the publish database with load profile (created from Visual studio) and publish database is succeeded from the mssql extension from VSCode.

PR #17350, opened 2022-06-02
Prompt For Connection Improvements 

### LINKED ISSUE #17325: Improve promptForConnection prompt to allow more options
PromptForConnection is set up so that it is restricted on the options we can provide to create a quickpick. 
https://github.com/microsoft/vscode-mssql/blob/main/src/views/connectionUI.ts#L90

We would like more control and in order to do this, we would need to create our own quickpick and provide options there.

### LINKED ISSUE #17335: SQL Bindings: No progress notification when creating new connection

VasuBhog:
This PR is not intended to go out with June Release (due to potential risk).

Initial Improvements to add more flexibility to promptForConnection: fixes #17325 

Fixes #17335: [Line here](https://github.com/microsoft/vscode-mssql/compare/vabhog/promptForConnectionImprovements?expand=1#diff-a6dd0c29a881dd2fd108e3f0fdcbeea035a85a1b43f37065a8b7fd6ba092e3d3R724)
Add Progress Notification when user is using `Create a Connection` we will show progress notification after they fill in the connection info:
Screen Shot 2022-06-02 at 4 42 42 PM. 

Also in order to fix this issue: https://github.com/microsoft/vscode-mssql/issues/17348 need to throw an error so that I would be able to re-prompt to the connection profile prompt for SQL bindings experiences. 
https://github.com/microsoft/azuredatastudio/blob/main/extensions/sql-bindings/src/services/azureFunctionsService.ts#L127

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -720,7 +720,6 @@ export default class ConnectionManager {
 
 	// create a new connection with the connectionCreds provided
 	public async connect(fileUri: string, connectionCreds: IConnectionInfo, promise?: Deferred<boolean>): Promise<boolean> {
-		const self = this;
 		return await vscode.window.withProgress({
 			location: vscode.ProgressLocation.Notification,
 			title: LocalizedConstants.connectProgressNoticationTitle,
@@ -762,12 +761,12 @@ export default class ConnectionManager {
 				this._connections[fileUri] = connectionInfo;
 
 				// Note: must call flavor changed before connecting, or the timer showing an animation doesn't occur
-				if (self.statusView) {
-					self.statusView.languageFlavorChanged(fileUri, Constants.mssqlProviderName);
-					self.statusView.connecting(fileUri, connectionCreds);
-					self.statusView.languageFlavorChanged(fileUri, Constants.mssqlProviderName);
+				if (this.statusView) {
+					this.statusView.languageFlavorChanged(fileUri, Constants.mssqlProviderName);
+					this.statusView.connecting(fileUri, connectionCreds);
+					this.statusView.languageFlavorChanged(fileUri, Constants.mssqlProviderName);
 				}
-				self.vscodeWrapper.logToOutputChannel(
+				this.vscodeWrapper.logToOutputChannel(
 					Utils.formatString(LocalizedConstants.msgConnecting, connectionCreds.server, fileUri)
 				);
 
@@ -789,7 +788,7 @@ export default class ConnectionManager {
 				// send connection request message to service host
 				this._uriToConnectionPromiseMap.set(connectParams.ownerUri, promise);
 				try {
-					const result = await self.client.sendRequest(ConnectionContracts.ConnectionRequest.type, connectParams);
+					const result = await this.client.sendRequest(ConnectionContracts.ConnectionRequest.type, connectParams);
 					if (!result) {
 						// Failed to process connect request
 						resolve(false);
@@ -798,8 +797,7 @@ export default class ConnectionManager {
 					reject(error);
 				}
 			});
-			let connectionResult = await connectionPromise;
-			return connectionResult;
+			return await connectionPromise;
 		});
 	}
 
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -739,7 +739,7 @@ export default class ConnectionManager {
 								profile, this.accountStore, providerSettings.resources.databaseResource);
 
 						} else {
-							throw new Error(`${LocalizedConstants.cannotConnect}`);
+							throw new Error(LocalizedConstants.cannotConnect);
 						}
 					} else {
 						connectionCreds.azureAccountToken = azureAccountToken.token;
@@ -797,7 +797,7 @@ export default class ConnectionManager {
 					reject(error);
 				}
 			});
-			return await connectionPromise;
+			return connectionPromise;
 		});
 	}
 
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -721,79 +721,86 @@ export default class ConnectionManager {
 	// create a new connection with the connectionCreds provided
 	public async connect(fileUri: string, connectionCreds: IConnectionInfo, promise?: Deferred<boolean>): Promise<boolean> {
 		const self = this;
-		// Check if the azure account token is present before sending connect request
-		if (connectionCreds.authenticationType === Constants.azureMfa) {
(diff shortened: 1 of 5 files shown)


src/models/connectionProfile.ts
```diff
@@ -127,7 +127,9 @@ export class ConnectionProfile extends ConnectionCredentials implements IConnect
 			});
 
 		return prompter.prompt(questions, true).then(async answers => {
-			if (answers.authenticationType === 'AzureMFA') {
+			if(!answers){
+				return undefined;
+			} else if (answers?.authenticationType === 'AzureMFA') {
```
VasuBhog, line 132: Saw an error in debug console when a user would exit out of `Create Connection Profile` while filling out the info. Therefore the `answers` would be undefined. We should handle this properly to result in an undefined instead, possibly could throw an error as well that user cancelled.

src/views/connectionUI.ts
```diff
@@ -553,7 +567,10 @@ export class ConnectionUI {
 				return ConnectionProfile.createProfile(this._prompter, this._connectionStore, this._context,
 					this.connectionManager.azureController, this._accountStore, profile);
 			} else {
-				return undefined;
+				// user cancelled the prompt - throw error so that we know user cancelled
+				return new Promise((_, reject) => {
+					reject(new CancelError('User cancelled retry connection warning'));
```
abist, line 572: localize

src/views/connectionUI.ts
```diff
@@ -553,7 +567,10 @@ export class ConnectionUI {
 				return ConnectionProfile.createProfile(this._prompter, this._connectionStore, this._context,
 					this.connectionManager.azureController, this._accountStore, profile);
 			} else {
-				return undefined;
+				// user cancelled the prompt - throw error so that we know user cancelled
+				return new Promise((_, reject) => {
```
Charles-Gagnon, line 571: Just refactor this function to be async, it'll be much easier to read that way

src/views/connectionUI.ts
```diff
@@ -84,19 +85,32 @@ export class ConnectionUI {
...
+		connectionProfileQuickPick.ignoreFocusOut = ignoreFocusOut;
+		connectionProfileQuickPick.canSelectMany = false;
+		connectionProfileQuickPick.busy = false;
+		connectionProfileQuickPick.show();
+		connectionProfileQuickPick.onDidChangeSelection(selection => {
+			if (selection[0]) {
+				// add progress notification and hide quickpick after user chooses an item from the quickpick
+				connectionProfileQuickPick.busy = true;
+				connectionProfileQuickPick.hide();
+				return this.handleSelectedConnection(selection[0]);
```
Charles-Gagnon, line 104: I'm assuming this is meant to be returned from `promptForConnection`? That isn't going to work currently, the return is for the `onDidChangeSelection` callback. 

Right now this function is always going to return undefined. You need to use a deferred promise (or wrap the whole thing in a promise) to be able to handle this correctly.

src/models/connectionProfile.ts
Charles-Gagnon: Can you please do this in a separate PR? That way it can be reviewed much quicker and won't clutter up this PR

src/controllers/connectionManager.ts
```diff
@@ -721,79 +721,86 @@ export default class ConnectionManager {
...
+						resolve(false);
+					}
+				} catch (error) {
+					reject(error);
 				}
-			} catch (error) {
-				reject(error);
-			}
+			});
+			let connectionResult = await connectionPromise;
```
Charles-Gagnon, line 801: `return connectionPromise`

src/views/connectionUI.ts
VasuBhog, line 104: @Charles-Gagnon one of the things you mentioned in a previous PR was that we should utilize the busy setting for when a user would like to test a connection. Right now the `this.handleSelectedConnection` returns back the connection if a user picks from a connection profile that is already saved shown below:
Screen Shot 2022-06-02 at 5 42 23 PM

Was wondering how we would bring in the `Test Connection` part that would utilize the busy property of the quickpick. 
@chlafreniere & @dzsquared

src/controllers/connectionManager.ts
```diff
@@ -721,79 +721,86 @@ export default class ConnectionManager {
 	// create a new connection with the connectionCreds provided
 	public async connect(fileUri: string, connectionCreds: IConnectionInfo, promise?: Deferred<boolean>): Promise<boolean> {
 		const self = this;
```
Charles-Gagnon, line 723: You should be able to get rid of this line and just use this since everything here is using lambdas

Charles-Gagnon (changes requested):

src/views/connectionUI.ts
Charles-Gagnon: I'm not sure I understand what you're saying. Could you give an example of what you're talking about?

src/views/connectionUI.ts
```diff
@@ -548,14 +562,16 @@ export class ConnectionUI {
...
-				return undefined;
-			}
-		});
+		let result = await this._vscodeWrapper.showErrorMessage(errorMessage, LocalizedConstants.retryLabel);
+		if (result === LocalizedConstants.retryLabel) {
+			return ConnectionProfile.createProfile(this._prompter, this._connectionStore, this._context,
+				this.connectionManager.azureController, this._accountStore, profile);
+		} else {
+			// user cancelled the prompt - throw error so that we know user cancelled
+			return new Promise((_, reject) => {
```
Charles-Gagnon, line 571: `throw new CancelError()`

(do you really need a message here? If there's a use for it I guess, but typically the messages aren't something we use except to display to the user)

VasuBhog:
So one of the scenarios that I will fix post-code-complete will be to remove the Test SQL Connection progress notification from sql-bindings since we have refactored the promptForConnection here: 
image

src/views/connectionUI.ts
VasuBhog: I can discuss this more in a separate PR since it would be a further enhancement.
Context of the previous PR: https://github.com/microsoft/azuredatastudio/pull/19211#discussion_r858907784

src/controllers/connectionManager.ts
```diff
@@ -720,80 +720,85 @@ export default class ConnectionManager {
...
+						resolve(false);
+					}
+				} catch (error) {
+					reject(error);
 				}
-			} catch (error) {
-				reject(error);
-			}
+			});
+			return await connectionPromise;
```
Charles-Gagnon, line 800: Just return the promise directly

Charles-Gagnon (approved):
Looks good, but hold off until after June release to get this in

PR #17368, opened 2022-06-17
Adding missing system function names to fix the coloring issue

llali:
Part one of the fix for colloring issues in ADS 
https://github.com/microsoft/azuredatastudio/issues/19756 and https://github.com/microsoft/azuredatastudio/issues/19737

### DIFF
--- a/syntaxes/SQL.plist
+++ b/syntaxes/SQL.plist
@@ -321,7 +321,7 @@
 		</dict>
 		<dict>
 			<key>match</key>
-			<string>(?i)\b(avg|checksum_agg|count|count_big|grouping|grouping_id|max|min|sum|stdev|stdevp|var|varp)\b</string>
+			<string>(?i)\b(aggregate|approx_count_distinct|avg|checksum_agg|count|count_big|grouping|grouping_id|max|min|sum|stdev|stdevp|var|varp)\b</string>
 			<key>name</key>
 			<string>support.function.aggregate.sql</string>
 		</dict>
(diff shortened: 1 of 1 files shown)


Charles-Gagnon:
@llali I linked an issue with a list of missing keywords, could you go through and make sure those are all included in this update?

syntaxes/SQL.plist
```diff
@@ -363,7 +381,13 @@
...
 		<dict>
 			<key>match</key>
-			<string>(?i)\b(choose|iif)\b</string>
+			<string>(?i)\b(json|isjson|json_object|json_array|json_value|json_query|json_modify|json_path_exists)\b</string>
+			<key>name</key>
+			<string>support.function.json.sql</string>
+		</dict>
+		<dict>
+			<key>match</key>
+			<string>(?i)\b(choose|iif|greatest|least)\b</string>
```
Charles-Gagnon, line 390: Out of curiosity how did you determine the "sections" each of these were in? Is there some core list you used for that?

llali:
> @llali I linked an issue with a list of missing keywords, could you go through and make sure those are all included in this update?

I did although some of the keywords were already there in the file so I'm not sure why the coloring didn't work for the. Example "Lag". Do you know what I'm missing?

Charles-Gagnon:
Oops, guess you were in the process of updating 😄 

But please remove the fixes tag - this doesn't actually fix the issue until we update the language file in ADS itself. So keep the issue open until that's done.

syntaxes/SQL.plist
llali: I grouped based on functions group. For instance all the date time functions are under function_datetime https://docs.microsoft.com/en-us/sql/t-sql/functions/date-bucket-transact-sql?view=sql-server-ver16

I'm actually not sure how to verify them or if my grouping is correct. I just followed the existing pattern

Charles-Gagnon:
@llali I assume you're looking at ADS? Because the language file there hasn't been updated in a very long time (nearly a year). It shows up correctly in VS Code : 

image

So first step here would be to grab the latest file from VS Code and update ADS so at least we match that.

syntaxes/SQL.plist
Charles-Gagnon: Cool, makes sense. Make sure you include that in your documentation!

As for verifying you should look through the VS Code docs and/or talk to Alex to see what we can do - we definitely should verify this all works correctly as early as we can.

Charles-Gagnon (approved):
Looks good but let's make sure that these new groups you added are being colored correctly, I'm not sure offhand what controls the colorization so you should make sure that the new names don't need to be added to some list somewhere

https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#tokenization
https://macromates.com/manual/en/language_grammars

llali:
> @llali I assume you're looking at ADS? Because the language file there hasn't been updated in a very long time (nearly a year). It shows up correctly in VS Code :
> 
> image
> 
> So first step here would be to grab the latest file from VS Code and update ADS so at least we match that.

so I guess that's one way to verify the new keywords 
https://github.com/microsoft/azuredatastudio/pull/19767
Screen Shot 2022-06-17 at 3 02 02 PM

syntaxes/SQL.plist
llali: I verified one of the function name and it showed correct coloring in ADS so I guess the way I did is correct

Charles-Gagnon:
Yeah but we still need to figure out how to get from plist -> sql.tmLanguage.json so we can verify the latest set of changes you're making

syntaxes/SQL.plist
Charles-Gagnon: What do you mean? How did you get your changes from the plist file into ADS?

llali: format in this file is xml, the one in vscode and ADS is json, that basically the only difference, I can add all these functions there manually and test.

Charles-Gagnon: Oh cool, that's handy. Would still be worth it to check with Alex to see if there's a way we can generate the full file automatically but at least this is something. 

Did you test with one of the new sections you added as well to make sure they don't need to be added to some other list somewhere?

llali: we can write a tool ourself :) convert xml to json. yes I added this 
		{
			"match": "(?i)\\b(asymkey_id|asymkeyproperty|certproperty|cert_id|crypt_gen_random|decryptbyasymkey|decryptbycert|decryptbykey|decryptbykeyautoasymkey|decryptbykeyautocert|decryptbypassphrase|encryptbyasymkey|encryptbycert|encryptbykey|encryptbypassphrase|hashbytes|is_objectsigned|key_guid|key_id|key_name|signbyasymkey|signbycert|symkeyproperty|verifysignedbycert|verifysignedbyasymkey)\\b",
			"name": "support.function.cryptographic.sql"
		},
and tested

alanrenmsft:
@llali @Charles-Gagnon do we have any doc explaining what goes here and what goes to the language service in STS? 

I am just about to start work on this one: https://github.com/microsoft/azuredatastudio/issues/18391, need some help.

Charles-Gagnon:
@llali is working on documentation for all this, although currently I believe her focus is on the colorization.

If you want actual auto-completion support you will need to get that updated in SqlParser (if it isn't already) - @shueybubbles or @Matteo-T can help there

Issue #17376, opened 2022-06-28
Test Failures Do Not Fail Builds


chlafreniere:
During the gulp test:cover step, there are currently 5 failing tests, but the build step shows as green.

<img width="759" alt="image" src="https://user-images.githubusercontent.com/40371649/176068172-6490c44c-a08d-44c7-b033-33f8dfb3ff55.png">

PR #17377, opened 2022-06-28
Fix test task to fail on test failures

Charles-Gagnon:
Fixes https://github.com/microsoft/vscode-mssql/issues/17376

We weren't failing the task when test failures occurred so the builds would happily complete.

I also did some cleanup by removing some unnecessary logging statements (the output and error messages are already printed out) and removing some disabled tasks from the pipeline definitions. 

This does mean we'll have failing tests, @kisantia @Benjin I'll leave that to your team as the owners to decide how you want to handle them. (with a preference towards getting them fixed of course 😄)

### DIFF
--- a/build/build.yml
+++ b/build/build.yml
@@ -104,22 +104,6 @@ steps:
     inputs:
       targets: "ext:install-service"
 
-  - task: gulp@0
-    displayName: "gulp cover:jenkins"
-    inputs:
-      targets: "cover:jenkins"
-      publishJUnitResults: true
-      testResultsFiles: "$(Build.SourcesDirectory)/test-reports/*.xml"
-    enabled: false
-    continueOnError: true
-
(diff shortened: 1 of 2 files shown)


kisantia (approved):

chlafreniere:
This fixes the uber issue around builds not failing when test failures occur (which is 100x more important). Any idea how the Tests tab in ADO works? that still thinks that all of the tests passed, strangely enough:

image

Charles-Gagnon:
Looks like this is a limitation of the default test result inference they do : https://docs.microsoft.com/en-us/azure/devops/pipelines/test/review-continuous-test-results-after-build?view=azure-devops

Trying manually uploading test results...

Charles-Gagnon:
Much better 😄 

image

PR #17380, opened 2022-06-29
Fix connection UI tests

kisantia:
This gets the 3 failing Connection UI tests passing as of commit b7cbc19d5453c2e0ec41e786bee38b2969060e41 - I ran the vscode-mssql adhoc pipeline with this commit on a branch off of that previous commit and verified these tests were passing. They were initially failing for the same reason as the test fixed in #17379. 

However, these 3 tests were then broken by the changes in https://github.com/microsoft/vscode-mssql/pull/17350, which significantly changed the code in `promptForConnection()`. @VasuBhog please fix these tests when you have a chance. 

Since the tests no longer are accurately testing the code in `promptForConnection()`, I'm skipping them until they get fixed.

### DIFF
--- a/test/connectionUI.test.ts
+++ b/test/connectionUI.test.ts
@@ -60,7 +60,7 @@ suite('Connection UI tests', () => {
 		outputChannel.verify(c => c.show(true), TypeMoq.Times.once());
 	});
 
-	test('showConnections with recent and new connection', () => {
+	test.skip('showConnections with recent and new connection', () => {
 		let item: IConnectionCredentialsQuickPickItem = {
 			connectionCreds: undefined,
 			quickPickItemType: CredentialsQuickPickItemType.NewConnection,
@@ -75,7 +75,7 @@ suite('Connection UI tests', () => {
 		});
 	});
 
-	test('showConnections with recent and edit connection', () => {
+	test.skip('showConnections with recent and edit connection', () => {
 		let testCreds = new ConnectionCredentials();
 		testCreds.connectionString = 'test';
 		let item: IConnectionCredentialsQuickPickItem = {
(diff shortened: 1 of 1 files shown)


VasuBhog:
These tests were broken prior to my changes, so I am bit confused as how my changes affected these. However, if you believe that I should be the one to fix them (due to my changes that were required)  then I can discuss with Chris. Since the only test that failed due to my PR was this [test](https://github.com/microsoft/vscode-mssql/blob/b51351e750db7e083618553b1134c07418239a06/test/connectionProfile.test.ts#L269)

Charles-Gagnon:
I agree we shouldn't skip these unless we've determined that we can't fix them by the next release cycle since they aren't causing any problems other than the build failing (local dev isn't impacted)

kisantia:
These 3 tests were already failing when the tests got reenabled in b51351e750db7e083618553b1134c07418239a06. They were failing with the same error as the test fixed in https://github.com/microsoft/vscode-mssql/pull/17379 which was broken by e66bb301e549b1137759da67415e4cd2c4304068 that was checked in the day before the tests got reenabled. I'm guessing these tests weren't fixed before the reenabling because Aditya didn't merge main and re-run the tests before merging the PR.

So at this point, these 3 tests were failing with the error `MockException - invocation count verification failed (expected invocation of Function.promptSingle(It.isAny()) exactly 1 times, invoked 0 times`. This is because these tests check `prompter.verify(p => p.promptSingle(TypeMoq.It.isAny()), TypeMoq.Times.once());` but it should actually be `p.promptSingle(TypeMoq.It.isAny(), TypeMoq.It.isAny()) ` after the `ignoreFocusOut` option started getting passed in with e66bb301e549b1137759da67415e4cd2c4304068.

After the prompt for connection improvements in 20750346a745890c93637ddfc2b4093c13904337 got checked in, the tests started failing with a different error: `TypeError: n.slice is not a function`. These 3 tests call `promptForConnection()`, which was changed in 20750346a745890c93637ddfc2b4093c13904337:

image


Does this make more sense on what broke these tests? 

I'm ok not skipping the tests if they can be fixed before the next release, but I'd prefer for the PR builds to not be red for an extended period of time since people might not notice new test failures caused by their PR.

VasuBhog:
Thanks for the clarification @kisantia. I believe that we should not skip the tests, I can prioritize this week but not sure if I will get it done (due to other work).

kisantia:
Ok sounds good. I'll go ahead and close this PR then. 

@Benjin @chlafreniere just fyi that the vscode-mssql pipelines won't be green until these tests are fixed.

PR #17385, opened 2022-07-01
Update minimatch to 3.0.2

cssuh:


package.json
```diff
@@ -150,6 +150,9 @@
     "vscode-nls": "^2.0.2",
     "zone.js": "^0.6.26"
   },
+  "resolutions": {
+    "minimatch": "^3.0.2"
```
Charles-Gagnon, line 154: Does the new version of remap-instanbul still not fix this? Or is this for some other package?

But in general it looks like this is again only for these instanbul dependencies, right? Are there any runtime packages using this?

cssuh: it's for another package - `gulp-istanbul-report`, and yes it's only for istanbul dependencies

cssuh:
discussed in dri channel

PR #17389, opened 2022-07-06
Changing to new lib for xmldom

aasimkhan30:


Charles-Gagnon (approved):

Charles-Gagnon:
@aasimkhan30 The builds are green so you're free to finish these up

PR #17393, opened 2022-07-12
Fix Connection Tests 

VasuBhog:
These tests were failing due to many changes to the promptForConnection improvements. 

The main PR that changed the functionality of promptForConnection is [here](https://github.com/microsoft/vscode-mssql/pull/17350). Instead of using prompter we now use createQuickPick to add flexibility as well as extensibility further down the road. 

The tests that were updated to setup the proper mocking aspects for the test to be verified. We verify the same assertions that we're originally being tested for these tests. 

Follow-up item: https://github.com/microsoft/vscode-mssql/issues/17394

### DIFF
--- a/test/connectionProfile.test.ts
+++ b/test/connectionProfile.test.ts
@@ -300,16 +300,15 @@ suite('Connection Profile tests', () => {
 			connectionStoreMock.object, mockAccountStore, prompter.object, vscodeWrapperMock.object);
 
 		// create a new connection profile
-		connectionUI.createAndSaveProfile().then(profile => {
+		connectionUI.createAndSaveProfile()
+		// CancelError will be thrown (expected)
+		.catch(() => {
 			// connection is attempted
-			connectionManagerMock.verify(x => x.connect(TypeMoq.It.isAny(), TypeMoq.It.isAny()), TypeMoq.Times.once());
+			connectionManagerMock.verify(x => x.connect(TypeMoq.It.isAny(), TypeMoq.It.isAny(), TypeMoq.It.isAny()), TypeMoq.Times.once());
 
 			// profile is not saved
-			connectionStoreMock.verify(x => x.saveProfile(TypeMoq.It.isAny()), TypeMoq.Times.never());
-
+			connectionStoreMock.verify(x => x.saveProfile(TypeMoq.It.isAny(), TypeMoq.It.isAny()), TypeMoq.Times.never());
 			done();
-		}).catch(err => {
-			done(err);
 		});
 	});
 
--- a/test/connectionProfile.test.ts
+++ b/test/connectionProfile.test.ts
@@ -294,7 +294,7 @@ suite('Connection Profile tests', () => {
 
 		let vscodeWrapperMock = TypeMoq.Mock.ofType(VscodeWrapper);
 		vscodeWrapperMock.setup(x => x.activeTextEditorUri).returns(() => 'test.sql');
-		// user cancels out of error message
+		// user cancels out of retry prompt
 		vscodeWrapperMock.setup(x => x.showErrorMessage(TypeMoq.It.isAny(), TypeMoq.It.isAny())).returns(() => Promise.resolve(undefined));
 
 		let connectionUI = new ConnectionUI(connectionManagerMock.object, contextMock.object,
--- a/test/connectionProfile.test.ts
+++ b/test/connectionProfile.test.ts
@@ -294,6 +294,7 @@ suite('Connection Profile tests', () => {
 
 		let vscodeWrapperMock = TypeMoq.Mock.ofType(VscodeWrapper);
 		vscodeWrapperMock.setup(x => x.activeTextEditorUri).returns(() => 'test.sql');
+		// user cancels out of error message
 		vscodeWrapperMock.setup(x => x.showErrorMessage(TypeMoq.It.isAny(), TypeMoq.It.isAny())).returns(() => Promise.resolve(undefined));
 
 		let connectionUI = new ConnectionUI(connectionManagerMock.object, contextMock.object,
--- a/test/connectionUI.test.ts
+++ b/test/connectionUI.test.ts
@@ -41,14 +41,15 @@ suite('Connection UI tests', () => {
 		vscodeWrapper.setup(v => v.createOutputChannel(TypeMoq.It.isAny())).returns(() => outputChannel.object);
 		vscodeWrapper.setup(v => v.showErrorMessage(TypeMoq.It.isAny()));
 		vscodeWrapper.setup(v => v.executeCommand(TypeMoq.It.isAnyString()));
-		quickPickMock = TypeMoq.Mock.ofType<vscode.QuickPick<IConnectionCredentialsQuickPickItem>>();
 		prompter = TypeMoq.Mock.ofType<IPrompter>();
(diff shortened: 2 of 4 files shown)


test/connectionUI.test.ts
```diff
@@ -83,21 +89,42 @@ suite('Connection UI tests', () => {
...
+			label: undefined
+		};
+		let quickPickMock = TypeMoq.Mock.ofType<vscode.QuickPick<IConnectionCredentialsQuickPickItem>>();
+		quickPickMock.setup(q => q.items);
+		quickPickMock.object.items = [item];
+		quickPickMock.setup(q => q.show());
+		quickPickMock.setup(q => q.onDidChangeSelection((e) => [] = e));
+		quickPickMock.object.onDidChangeSelection(undefined);
+		quickPickMock.setup(q => q.onDidHide(() => false));
+		vscodeWrapper.setup(v => v.createQuickPick()).returns(() => quickPickMock.object);
```
kisantia, line 122: these tests look like they have a lot of the same setup code. Can there be a helper function or beforeEach that does the common setup?

test/connectionProfile.test.ts
```diff
@@ -294,22 +294,22 @@ suite('Connection Profile tests', () => {
...
 		vscodeWrapperMock.setup(x => x.showErrorMessage(TypeMoq.It.isAny(), TypeMoq.It.isAny())).returns(() => Promise.resolve(undefined));
 
 		let connectionUI = new ConnectionUI(connectionManagerMock.object, contextMock.object,
 			connectionStoreMock.object, mockAccountStore, prompter.object, vscodeWrapperMock.object);
 
 		// create a new connection profile
-		connectionUI.createAndSaveProfile().then(profile => {
+		connectionUI.createAndSaveProfile()
+		// CancelError will be thrown (expected)
+		.catch(() => {
```
Charles-Gagnon, line 306: Why are you moving this into the catch? Isn't that changing the behavior of the test? Before if an exception was thrown it would automatically fail the test.

test/connectionUI.test.ts
```diff
@@ -83,21 +89,42 @@ suite('Connection UI tests', () => {
...
+		let quickPickMock = TypeMoq.Mock.ofType<vscode.QuickPick<IConnectionCredentialsQuickPickItem>>();
+		quickPickMock.setup(q => q.items);
+		quickPickMock.object.items = [item];
+		quickPickMock.setup(q => q.show());
+		quickPickMock.setup(q => q.onDidChangeSelection((e) => [item] = e));
+		quickPickMock.object.onDidChangeSelection((e) => [item] = e);
+		quickPickMock.setup(q => q.onDidHide(() => false));
+		vscodeWrapper.setup(v => v.createQuickPick()).returns(() => quickPickMock.object);
+
+		connectionUI.promptForConnection();
```
Charles-Gagnon, line 102: This returns a promise, you should be awaiting this call

test/connectionUI.test.ts
```diff
@@ -83,21 +89,42 @@ suite('Connection UI tests', () => {
...
+		quickPickMock.setup(q => q.items);
+		quickPickMock.object.items = [item];
+		quickPickMock.setup(q => q.show());
+		quickPickMock.setup(q => q.onDidChangeSelection((e) => [] = e));
+		quickPickMock.object.onDidChangeSelection(undefined);
+		quickPickMock.setup(q => q.onDidHide(() => false));
+		vscodeWrapper.setup(v => v.createQuickPick()).returns(() => quickPickMock.object);
+
+
+		connectionUI.promptForConnection();
```
Charles-Gagnon, line 125: await promise

test/connectionProfile.test.ts
VasuBhog: The logic where this is run was changed in the linked PR, but here is the specific line where instead of undefined we throw the CancelError: https://github.com/microsoft/vscode-mssql/pull/17350/files#diff-ddbe2fb5d84649cfa54b0ed716fb17faf46f59b3d6a152115f32ee2c3b1cd227R571

test/connectionUI.test.ts
chlafreniere: Do you have any ideas for how `promptForConnection` can be easier to test than it is at the moment? It this is awaited, then the function doesn't return. How can this pattern be made better?

https://github.com/microsoft/vscode-mssql/blob/228564b2d9c370539f1364a79cde661460e7a1d1/src/views/connectionUI.ts#L87

test/connectionProfile.test.ts
```diff
@@ -294,22 +294,22 @@ suite('Connection Profile tests', () => {
 
 		let vscodeWrapperMock = TypeMoq.Mock.ofType(VscodeWrapper);
 		vscodeWrapperMock.setup(x => x.activeTextEditorUri).returns(() => 'test.sql');
+		// user cancels out of error message
```
Charles-Gagnon, line 297: `user cancels out of retry prompt`

test/connectionUI.test.ts
```diff
@@ -40,11 +41,12 @@ suite('Connection UI tests', () => {
 		vscodeWrapper.setup(v => v.createOutputChannel(TypeMoq.It.isAny())).returns(() => outputChannel.object);
 		vscodeWrapper.setup(v => v.showErrorMessage(TypeMoq.It.isAny()));
 		vscodeWrapper.setup(v => v.executeCommand(TypeMoq.It.isAnyString()));
+		quickPickMock = TypeMoq.Mock.ofType<vscode.QuickPick<IConnectionCredentialsQuickPickItem>>();
 		prompter = TypeMoq.Mock.ofType<IPrompter>();
 		mockContext = TypeMoq.Mock.ofType<vscode.ExtensionContext>();
 		mockContext.setup(c => c.globalState).returns(() => globalstate.object);
 		connectionStore = TypeMoq.Mock.ofType(ConnectionStore, TypeMoq.MockBehavior.Loose, mockContext.object);
-		connectionStore.setup(c => c.getPickListItems()).returns(() => TypeMoq.It.isAny());
+		connectionStore.setup(c => c.getPickListItems()).returns(() => [TypeMoq.It.isAny()]);
```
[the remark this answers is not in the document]
VasuBhog: Removed this entirely since now we directly set items to the quickPickObject for the necessary tests.

test/connectionUI.test.ts
```diff
@@ -60,44 +64,62 @@ suite('Connection UI tests', () => {
...
 	});
 
-	test('showConnections with recent and new connection', () => {
+	test('showConnections with recent and new connection', async () => {
 		let item: IConnectionCredentialsQuickPickItem = {
 			connectionCreds: undefined,
 			quickPickItemType: CredentialsQuickPickItemType.NewConnection,
 			label: undefined
 		};
+		quickPickMock.object.items = [item];
```
Charles-Gagnon, line 73: Do you actually need to set this? The prompt just gets the selected item via the callback, the items property is only used to set the items (which this test doesn't currently care about).

But also you should consider capturing the values passed to the items setter and then use those items in the onDidChangeSelect callback. Because it's the product code that actually sets the properties - by overriding that here you're skipping part of the actual functionality. 

(same for others below)

Charles-Gagnon (approved):

test/connectionUI.test.ts
VasuBhog: Removed this since as you stated we send it via the emitter/callback.

kisantia (approved):

PR #17400, opened 2022-07-13
Add typings for test runner

Charles-Gagnon:
We should probably switch to vscodetestcover at some point to consolidate the runner logic, but for now at least making this stuff easier to work with and validate.


kisantia (approved):

Charles-Gagnon:
There are test failures that are actually due to updating the VS Code version the tests are running against. I'll investigate and follow up with an issue/fix separately

PR #17401, opened 2022-07-13
Fix service installer util tests

Charles-Gagnon:
In newer versions of VS Code they changed the extension host implementation of console.log which made it unusable for testing - we can't even stub/spy it. 

I looked at this code and realized that it's not even being used in the extension itself - it's only for installing the service from the gulp tasks. I didn't feel it was worth the effort to move this somewhere else or otherwise refactor it so it's not needed, so I just went with the simple route of replacing the direct console.log calls with an injected function that we can replace ourselves. 

I also removed the indent/outdent calls to not do anything - having them print out the words seemed weird and pointless. 

I considered removing the tests completely as they're pretty useless overall (testing that when we call a method that logs something that something is logged...woohoo...) but in the interest of not having to deal with going over this again later for code coverage decided just to leave it like this for now.


kisantia (approved):
yay now we can use sinon in tests in this extension :D

Issue #17375, opened 2022-06-27
Establishing connection to SQL Server from VSCode


Katzmann1983:
- MSSQL Extension Version: 1.15.0
- VSCode Version: 1.68.1
- OS Version: Windows 10 Enterprise 19042.1706

Steps to Reproduce:

1. Connect successfully with Azure Data Studio
2. Try to connect with Visual Studio Code. Use Azure Active Directory with MFA
3. Browser opens http://localhost:60681/signin?nonce=nIRgAEFuefgKQt9oFUNKEw%253D%253D which cannot be opened.
4. Uninstall Azure Data Studio.
5. Reconnect (using Oath2 with the browser twice) and receive the error:
"Token retrieval failed with an error. Open developer tools to view the error"
6. Delete the folder C:\Users{UserNameHere}\AppData\Roaming\azuredatastudio\Azure Accounts (following https://stackoverflow.com/questions/69481416/establishing-connection-to-azure-sql-server-error-token-retrieval-failed-with)
7. Receive error:
"mssql: Credential Error: Account credentials have expired. Please re-authenticate."
8. Reauthenticate and receive "mssql: Token retrieval failed with an error. Open developer tools to view the error"
9. Reauthenticate and receive error:
"mssql: Connection error : A connection was successfully established with the server, but then an error occurred during the login process. (provider: SSL Provider, error: 0 - The target principal name is incorrect.)"

`[2022-06-28 10:00:18.278] [exthost] [error] Error: Token retrieval failed with an error. Open developer tools to view the error
	at AzureCodeGrant.getToken (c:\Users\jens.humrich\.vscode\extensions\ms-mssql.mssql-1.15.0\node_modules\@microsoft\ads-adal-library\dist\engine\AzureAuth.js:193:19)
	at processTicksAndRejections (node:internal/process/task_queues:96:5)
	at async AzureCodeGrant.getAccountSecurityToken (c:\Users\jens.humrich\.vscode\extensions\ms-mssql.mssql-1.15.0\node_modules\@microsoft\ads-adal-library\dist\engine\AzureAuth.js:137:32) mssql.connectObjectExplorerNode {"value":"ms-mssql.mssql","_lower":"ms-mssql.mssql"}
[2022-06-28 10:00:18.278] [exthost] [error] Error: Token retrieval failed with an error. Open developer tools to view the error
[...]

Issue #17409, opened 2022-07-28
When changing the password, it still tries to use the old cached tokens


Katzmann1983:
MSSQL Extension Version: 1.15.0
    VSCode Version: 1.68.1
    OS Version: Windows 10 Enterprise 19042.1706

Steps to Reproduce:

1. Change password
2. The cached token will still be used

cheenamalhotra:
Do you mean Change Password or Refresh access token?

kburtram:
Closing based on non-response to previous clarification question.  Please let us know if you're still impacted by this issue.  Thanks!

PR #17410, opened 2022-07-28
Fixing Issue #17375

Katzmann1983:
Fixing an issue when changing the password

### DIFF
--- a/lib/ads-adal-library/src/engine/AzureAuth.ts
+++ b/lib/ads-adal-library/src/engine/AzureAuth.ts
@@ -126,7 +126,7 @@ export abstract class AzureAuth {
 			const remainingTime = expiry - currentTime;
 			const maxTolerance = 2 * 60; // two minutes
 
-			if (remainingTime < maxTolerance) {
+			if ((remainingTime < maxTolerance) && (remainingTime > 0)) {
 				const result = await this.refreshToken(tenant, azureResource, cachedTokens.refreshToken);
 				if (!result) {
 					return undefined;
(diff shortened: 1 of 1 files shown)


lib/ads-adal-library/src/engine/AzureAuth.ts
```diff
@@ -126,7 +126,7 @@ export abstract class AzureAuth {
 			const remainingTime = expiry - currentTime;
 			const maxTolerance = 2 * 60; // two minutes
 
-			if (remainingTime < maxTolerance) {
+			if ((remainingTime < maxTolerance) && (remainingTime > 0)) {
```
Charles-Gagnon, line 129: Why only when remainingTime > 0? If the expiration time has passed we need to refresh as well.

Charles-Gagnon (changes requested):
Please update the description with: 

1. A description of what issue you're fixing (ideally linking to an actual Github issue)
2. How this change is fixing the problem

lib/ads-adal-library/src/engine/AzureAuth.ts
```diff
@@ -135,14 +135,17 @@ export abstract class AzureAuth {
 				expiresOn = Number(result.expiresOn);
 			}
 			// Let's just return here.
-			if (accessToken) {
-				return {
-					...accessToken,
-					expiresOn: expiresOn,
-					tokenType: 'Bearer'
-				};
+			if (accessToken.expiresOn) {
```
Charles-Gagnon, line 138: What scenario would we have an accessToken but not an expiration time?

Katzmann1983: This is to check that the object exists (otherwise vscode was complaining):
imagen

lib/ads-adal-library/src/engine/AzureAuth.ts
Katzmann1983: 1. Yes, actually it is fixing #17409 and #17375
2. The issue is that sometime the refreshToken is outdated as well (e.g. Password change  #17409 or VSCode finds some old cached token #17375). In these cases it cannot refresh the token and returns an invalid token.

Katzmann1983: > If the expiration time has passed we need to refresh as well right? Yes, the issue is that we might try to refresh with an invalid token, due to it only being valid for 24h hours. Or due to a password change the refresh token is invalid, or it has been revoked. In these cases we end up without a valid token and return undefined;

Charles-Gagnon: Refreshing the token can fail for a number of reasons - including some like token revocation or password change that can happen at any time. So what seems like a better overall fix is to always try to refresh the token but then be better about handling errors (which it seems like we're not doing right now). Which would include ensuring that the old token is cleared out if it's invalid. 

@cssuh @kburtram Are these scenarios something you've had a chance to look into?

cheenamalhotra:
Hi @Katzmann1983 

I'm a bit confused about the intent of the PR with regards to 'Change Password'.. FYI I just opened a PR #17550 to fix the refresh access token issue, can you test with this change if you observe any remaining concerns?

Issue #17057, opened 2021-09-21
Incorrect highlighting when escaping quotes in SQL


alexr00:
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.60.0
- OS Version: Ubuntu 20.04 (Linux x64 5.11.0-27-generic)

Steps to Reproduce:

1. When escaping quotes, single our double inside the same quotes, the highlighting stops/starts on that quote. Causing incorrect highlighting

```sql
SELECT * FROM db WHERE foo = '\"bar\"'
SELECT * FROM db WHERE foo = '\'bar\''

SELECT * FROM db WHERE foo = 'You\'re cool'
SELECT * FROM db WHERE foo = 'bcd'
SELECT * FROM db WHERE foo = 'You\'re cool'

SELECT * FROM db WHERE foo = "\'bar\'"
SELECT * FROM db WHERE foo = "\"bar\""

SELECT * FROM db WHERE foo = "A quote symbol looks like this: \" "
SELECT * FROM db WHERE foo = "bar"

```

![image](https://user-images.githubusercontent.com/5701732/132988290-24dc2953-21a3-4833-8fd8-5002ec45d3d6.png)

kburtram:
This looks like a duplicate of https://github.com/microsoft/vscode-mssql/issues/1014.  I think the regular expression we're using for our colorization syntax aren't correctly handling escape characters.

RedCMD:
I'm sure if this comment is true
As both the single `match` and `begin`/`end` rules will start and finish at single quotes
https://github.com/microsoft/vscode-mssql/blob/22961ac77f84f6aeb5e0ef3066e008c572c09018/syntaxes/SQL.plist#L685

Removing everything from line 670 to 690 will fix the issue with single quotes
likewise line 722 to 742 for backticks
and 774 to 794 for double quotes
Some cleanup would need to happen as well.
like readding `(N)?` to single quote `begin` rule

![image](https://user-images.githubusercontent.com/33529441/176996012-4892516e-aae8-4e38-9de0-620a481548be.png)

ewauq:
Was there a fix done for this issue? I still encounter the problem today.

VSCode is updated and I tested without any extension enabled.

![image](https://user-images.githubusercontent.com/1007507/185578361-af7cf98b-7664-4c92-bf6d-44acdbaf3269.png)

RedCMD:
@kburtram 
Fix issue should not be closed
[...]

PR #17427, opened 2022-08-21
Fix escape characters inside strings

RedCMD:
Fixes https://github.com/microsoft/vscode-mssql/issues/17057

Before:
image

After:
image

I am very confused as what this comment means: 
`"comment": "this is faster than the next begin/end rule since sub-pattern will match till end-of-line and SQL files tend to have very long lines."`
I did some testing and found `begin/end` rules to be no slower than `match`
if anything, `match` was slightly slower due to https://github.com/microsoft/vscode-textmate/issues/167

For some reason double quote strings did not support escape characters what so ever
Does SQL support escape characters inside double quote strings??
```sql
`backtick \` string`
'single quote \' string'
"double quote \" string"
```

should also support the double char escape sequence?
```sql
`backtick `` string`
'single quote '' string'
"double quote "" string"
```

### DIFF
--- a/syntaxes/SQL.plist
+++ b/syntaxes/SQL.plist
@@ -668,32 +668,11 @@
 			<key>patterns</key>
 			<array>
 				<dict>
-					<key>captures</key>
-					<dict>
-						<key>2</key>
-						<dict>
-							<key>name</key>
-							<string>punctuation.definition.string.begin.sql</string>
-						</dict>
-						<key>3</key>
-						<dict>
-							<key>name</key>
-							<string>punctuation.definition.string.end.sql</string>
-						</dict>
-					</dict>
-					<key>comment</key>
-					<string>this is faster than the next begin/end rule since sub-pattern will match till end-of-line and SQL files tend to have very long lines.</string>
-					<key>match</key>
-					<string>(N)?(')[^']*(')</string>
-					<key>name</key>
-					<string>string.quoted.single.sql</string>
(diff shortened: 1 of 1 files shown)


Charles-Gagnon:
Thanks for looking into this!

T-SQL escapes quotes (both ' and ") by doubling them up. The backslash escape sequence isn't valid. 

So these are valid strings

`SELECT 'foo''bar'` -> `foo'bar` (escaped single quote)
`SELECT 'foo"bar'` -> `foo"bar` (double quote in single-quoted string)
`SELECT ''''` -> `'` (escaped single quote)
`SELECT "foo""bar"` -> `foo"bar` (escaped double quote)
`SELECT "foo'bar"` -> `foo'bar` (single quote in double-quoted string)
`SELECT """"` -> `"` (escaped double quote)

Not valid

`SELECT 'foo\'bar'`
`SELECT "foo\"bar"`

(Note that there's some trickiness around double quotes where you have to use `SET QUOTED_IDENTIFIER OFF` to use them, but for now just assuming both types of quotes are valid is fine)

cheenamalhotra:
Closing PR due to inactivity, please feel free to re-open when you have time to address feedback.

deemoney-rupendra:
The issue still persists. Can you please re-open this?

Charles-Gagnon:
See https://github.com/microsoft/vscode-mssql/issues/17057 for more discussion on this

PR #17443, opened 2022-09-14
Update README.md

akkadaya:
Version hasn't been updated.


Charles-Gagnon:
Oops, good catch. I'm just going to remove the version in a separate PR - there's no reason I can think of to have it embedded in the README like this.

PR #17445, opened 2022-09-14
Expose getServerInfo API

### LINKED ISSUE #17438: Expose server information
To complete this https://github.com/microsoft/azuredatastudio/issues/20576, server information needs to be exposed from vscode-mssql.

SakshiS-harma:
This PR fixes #17438 .
By exposing server information, target platform information can be set automatically for the sql database projects that are created from databases.

### DIFF
--- a/src/models/contracts/connection.ts
+++ b/src/models/contracts/connection.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { NotificationType, RequestType } from 'vscode-languageclient';
-import { ConnectionDetails } from 'vscode-mssql';
+import { ConnectionDetails, IServerInfo } from 'vscode-mssql';
 
 // ------------------------------- < Connect Request > ----------------------------------------------
 
@@ -42,7 +42,7 @@ export namespace ConnectionCompleteNotification {
 /**
  * Information about a SQL Server instance.
  */
-export class ServerInfo {
+export class ServerInfo implements IServerInfo {
 	/**
 	 * The major version of the SQL Server instance.
 	 */
--- a/src/models/contracts/connection.ts
+++ b/src/models/contracts/connection.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { NotificationType, RequestType } from 'vscode-languageclient';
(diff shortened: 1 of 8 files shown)


typings/vscode-mssql.d.ts
```diff
@@ -114,6 +114,67 @@ declare module 'vscode-mssql' {
 		 * @returns A promise object for when the request receives a response
 		 */
 		sendRequest<P, R, E, R0>(requestType: RequestType<P, R, E, R0>, params?: P): Promise<R>;
+
+		/**
+		 * Get the server info for a connection
+		 * @param connectionInfo connection info of the connection
```
kisantia, line 120: also add `@returns` doc comment

kisantia (approved):

src/models/contracts/connection.ts
```diff
@@ -42,7 +42,7 @@ export namespace ConnectionCompleteNotification {
 /**
  * Information about a SQL Server instance.
  */
-export class ServerInfo {
+export class ServerInfo implements IServerInfo {
```
Charles-Gagnon, line 45: I'd just remove this class and replace it with a `ServerInfo` interface in the typings. There isn't any reason this needs to be a class anyways.

typings/vscode-mssql.d.ts
```diff
@@ -114,6 +114,68 @@ declare module 'vscode-mssql' {
...
+		 * @returns server information
+	 	*/
+		getServerInfo(connectionInfo: IConnectionInfo): IServerInfo
+	}
+
+	/**
+	 * Information about a SQL Server instance.
+	 */
+	export interface IServerInfo {
```
alanrenmsft, line 129: we now have two places defining the ServerInfo, should we remove the old one and use this instead?

src/models/contracts/connection.ts
SakshiS-harma: Makes sense. Removed it.

typings/vscode-mssql.d.ts
SakshiS-harma: Yes, removed the old one.

Charles-Gagnon (approved):

Charles-Gagnon:
(you need to fix the linting errors first though)

PR #17449, opened 2022-09-16
Add Database engine edition information

### LINKED ISSUE #17448: Add DatabaseEngineEdition enum

SakshiS-harma:
This PR fixes #17448 .

This enum is needed to set target platform appropriately for Sql database projects.

### DIFF
--- a/typings/vscode-mssql.d.ts
+++ b/typings/vscode-mssql.d.ts
@@ -119,7 +119,7 @@ declare module 'vscode-mssql' {
 		 * Get the server info for a connection
 		 * @param connectionInfo connection info of the connection
 		 * @returns server information
-	 	*/
+		  */
 		getServerInfo(connectionInfo: IConnectionInfo): ServerInfo
 	}
 
@@ -176,6 +176,22 @@ declare module 'vscode-mssql' {
 		 * The Operating System version string of the machine running the SQL Server instance.
 		 */
 		osVersion: string;
+	}
+
+	/**
+	 * The possible values of the server engine edition
+	 */
+	export const enum DatabaseEngineEdition {
+		Unknown = 0,
+		Personal = 1,
+		Standard = 2,
+		Enterprise = 3,
+		Express = 4,
+		SqlDatabase = 5,
+		SqlDataWarehouse = 6,
+		SqlStretchDatabase = 7,
+		SqlManagedInstance = 8,
+		SqlOnDemand = 11
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


typings/vscode-mssql.d.ts
```diff
@@ -178,6 +178,22 @@ declare module 'vscode-mssql' {
 		osVersion: string;
 	}
 
+	/**
+	 * The possible values of the server engine edition
```
Charles-Gagnon, line 182: I'd also add a link to https://docs.microsoft.com/sql/t-sql/functions/serverproperty-transact-sql and direct them to look at the `EngineEdition` property

typings/vscode-mssql.d.ts
```diff
@@ -178,6 +178,22 @@ declare module 'vscode-mssql' {
...
+	 */
+	export const enum DatabaseEngineEdition {
+		Unknown = 0,
+		Personal = 1,
+		Standard = 2,
+		Enterprise = 3,
+		Express = 4,
+		SqlDatabase = 5,
+		SqlDataWarehouse = 6,
```
Charles-Gagnon, line 191: Should probably also update this to `SynapseAnalytics`

@shueybubbles Has there been any discussion over on the SMO side of updating the enums? I imagine not for compat reasons - maybe we should keep these the same here to match that.

typings/vscode-mssql.d.ts
```diff
@@ -178,6 +178,22 @@ declare module 'vscode-mssql' {
...
+		SqlStretchDatabase = 7,
+		SqlManagedInstance = 8,
+		SqlOnDemand = 11
```
Charles-Gagnon, line 194: And this to `SynapseServerlessSqlPool`

typings/vscode-mssql.d.ts
Charles-Gagnon: Actually let's just do that as a follow up task, you can go ahead and merge this in now and we can have a discussion about whether it makes sense to update later.

Charles-Gagnon (approved):

typings/vscode-mssql.d.ts
shueybubbles: I think if SMO were to do anything it would just add new symbols as synonyms for the current ones

typings/vscode-mssql.d.ts
SakshiS-harma: Raised a new issue https://github.com/microsoft/azuredatastudio/issues/20617 to keep track for discussion.

PR #17462, opened 2022-10-20
Fix function syntax highlighting including non-function keywords

Charles-Gagnon:
Currently the function captures just look for the keyword itself, which means if it's used in any other context then it's incorrectly colorized.

Example of current incorrect behavior - notice how the DAY is highlighted green in both the function AND the parameter to DATE_BUCKET

image

Fix is to update the captures to look for an open paren after the word itself, and then match only that word (so we don't match the paren since that shouldn't be colorized)

After fix : 

image

(Note that in this case DAY the parameter should be highlighted blue as a keyword. I'll be following up with fixing that in a separate PR to keep changes simple)

### DIFF
--- a/syntaxes/SQL.plist
+++ b/syntaxes/SQL.plist
@@ -351,18 +351,6 @@
 				<key>1</key>
 				<dict>
 					<key>name</key>
-					<string>support.function.bitmanipulation.sql</string>
-				</dict>
-			</dict>
-		</dict>
-		<dict>
-			<key>match</key>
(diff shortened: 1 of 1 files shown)


alanrenmsft (approved):

alanrenmsft:
@Charles-Gagnon does the same issue exists in ADS?

Charles-Gagnon:
Yes, but the ADS grammar is generated from this file. I'll be doing more fixes here and then doing one large update in ADS once I get those in (in addition to other fixes such as updating sqlparser and autocomplete stuff)

PR #17465, opened 2022-10-24
Syntax update

Charles-Gagnon:
Adding new syntax, mostly from SQL 2022 but also a few others I noticed were missing along the way. 

Used this list : https://learn.microsoft.com/en-us/sql/sql-server/what-s-new-in-sql-server-2022?view=sql-server-ver16&viewFallbackFrom=azuresqldb-current#language

General keywords : 

- ARRAY
- SCALAR
- WITHIN
- GROUP
- BLOCKERS
- DAY
- MONTH
- YEAR
- WEEK
- QUARTER
- HOUR
- MINUTE
- SECOND
- LEADING
- TRAILING
- BOTH
- RESPECT
- IGNORE
- NULLS
- WAIT_AT_LOW_PRIORITY
- AUTO_DROP

Functions : 

- APPROX_PERCENTILE_CONT
- APPROX_PERCENTILE_DISC
- DATETRUNC
- BIT_COUNT
- GET_BIT
- LEFT_SHIFT
- RIGHT_SHIFT
- SET_BIT

I also removed "Aggregate" from the list of functions since it isn't an actual function.


Charles-Gagnon:
Failure is due to CG alert, not something added in this change.

alanrenmsft:
Is the list version specific? for example, I have a query editor connected to a 2017 SQL Server,  will it show the new keywords/syntax introduced in SQL Server 2022 as valid for me?

alanrenmsft (approved):

Charles-Gagnon:
This list is just for colorization - which means it knows nothing about the connection and isn't used to display errors. The parser is what handles determining if syntax is valid (and applying the red squigglies if not). 

VS Code does have the ability to provide colorization from a language server, which is something I plan on looking into here with my other general improvements I plan on making. But I don't know if that'll end up being worth it - the downside to that approach is you add the overhead of having to start up the language server and wait for it to parse the document text whenever there's changes.

Charles-Gagnon:
(I'll be writing up a wiki how-to on all this language support stuff once I finish going through and piecing everything together to help others understand everything as well)

PR #17466, opened 2022-10-24
bump minimatch to fix CG alert

caohai:


package.json
```diff
@@ -151,6 +151,13 @@
     "vscode-nls": "^2.0.2",
     "zone.js": "^0.6.26"
   },
+  "resolutions": {
+    "pm-mocha-jenkins-reporter/mocha/minimatch": "^3.0.5",
```
Charles-Gagnon, line 155: So there are no runtime dependencies using minimatch? If that's the case we can just skip the update for now - resolutions isn't something we should be doing unless we absolutely need to.

lib/ads-adal-library/package.json
```diff
@@ -25,5 +25,8 @@
   "dependencies": {
     "axios": "^0.27.2",
     "qs": "^6.9.4"
+  },
+  "resolutions": {
+    "eslint/minimatch": "^3.0.5"
```
Charles-Gagnon, line 30: Did you try updating eslint directly? There's a ton of newer versions which may fix this - and something like eslint is generally going to be straightforward to update.

package.json
caohai: `ejs` is the only runtime dep using minimatch, all other ones are dev dep. And we only use `ejs` for rendering (in one line of code), no feature related to regex is used.

Charles-Gagnon: Alright. If you aren't able to update the root packages for these then just revert the changes and we can dismiss the alert.

Charles-Gagnon:
@caohai Do we need this anymore?

caohai:
> @caohai Do we need this anymore?

Nope, forgot to close this one.

PR #17473, opened 2022-11-07
update STS to align with ADS

erpett:
Updating STS to align with ADS


cheenamalhotra:
This update will enable encryption by default with MDS 5 as it comes with [4.3.0.42](https://github.com/microsoft/sqltoolsservice/releases/tag/4.3.0.42)
Should this be held off for December? @kburtram

erpett:
Canceling

PR #17484, opened 2022-11-26
Enable Encryption by default + introduce HostNameInCertificate

cheenamalhotra:
Bringing in Microsoft.Data.SqlClient 5.0.1 dependency with STS 4.4.0.7
- Sets `Encrypt` to `Mandatory` by default.
- Adds support for `HostNameInCertificate` SqlConnection property.
- Adds welcome note for encryption behavior change (Updated)
   
   _More information_ redirects to: https://aka.ms/vscodemssql-connection (placeholder link for blogpost)

- Enables setting `Trust Server Certificate` to true with prompt for cert validation failure.
   
   Read more redirects to: https://aka.ms/vscodemssql-connection (placeholder link for blogpost)

### Demo

VSCode-mssql
_Captures double failure scenario as well_


localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -2,6 +2,18 @@
...
   <file datatype="xml" original="src/constants/localizedConstants" source-language="en">
     <body>
+        <trans-unit id="viewMore">
+            <source xml:lang="en">View more</source>
+        </trans-unit>
+        <trans-unit id="releaseNotesPromptDescription">
+            <source xml:lang="en">View mssql for Visual Studio Code release notes?</source>
+        </trans-unit>
+        <trans-unit id="encryptionChangePromptDescription">
+            <source xml:lang="en">Azure Data Studio now has encryption enabled by default for all SQL Server connections. This may result in your existing connections no longer working unless certain Encryption related connection properties are changed. We recommend you visit the link below for more information.</source>
```
dzsquared, line 12: wouldn't we say "mssql extension for VS Code" instead of "Azure Data Studio"

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -267,7 +306,10 @@
             <source xml:lang="en">Firewall rule successfully added. Retry profile creation? </source>
         </trans-unit>
         <trans-unit id="msgAccountRefreshFailed">
-            <source xml:lang="en">Credential Error: Account credentials have expired.  Please re-authenticate.</source>
+            <source xml:lang="en">Credential Error: Account credentials have expired. Please re-authenticate.</source>
+        </trans-unit>
+        <trans-unit id="msgPromptProfileUpdateFailed">
+            <source xml:lang="en">Connection Profle could not be updated. Please modify the connection details manually in settings.json and try again.</source>
```
Charles-Gagnon, line 312:
```suggestion
            <source xml:lang="en">Connection Profile could not be updated. Please modify the connection details manually in settings.json and try again.</source>
```

dzsquared:
I'd prefer both the dialogs link to our aka.ms link where we will have specific documentation on VS Code connectivity.  From there, we can link to more information specifically on configuring SQL Server certificates but folks will also see info on the settings file and encryption options.

erinstellato-ms (approved):
I agree with Drew's comment - everything else looks great.

src/controllers/connectionManager.ts
```diff
@@ -462,6 +477,27 @@ export default class ConnectionManager {
...
+			.then(async (selection) => {
+				if (selection && selection === LocalizedConstants.enableTrustServerCertificate) {
+					profile.encrypt = 'Mandatory';
+					profile.trustServerCertificate = true;
+					await reconnectAction(profile);
+				} else if (selection && selection === LocalizedConstants.readMore) {
+					this.vscodeWrapper.openExternal(Constants.encryptionBlogLink);
+					this.showInstructionTextAsWarning(profile, reconnectAction);
+				}
+			});
```
Charles-Gagnon, line 498: nit: cleanup (use async/await style whenever possible)
```suggestion
		const selection = await this.vscodeWrapper.showWarningMessageAdvanced(instructionText,
			{ modal: false },
			[
				LocalizedConstants.enableTrustServerCertificate,
				LocalizedConstants.readMore,
				LocalizedConstants.cancel
			]);
		if (selection === LocalizedConstants.enableTrustServerCertificate) {
			profile.encrypt = 'Mandatory';
			profile.trustServerCertificate = true;
			await reconnectAction(profile);
		} else if (selection === LocalizedConstants.readMore) {
			this.vscodeWrapper.openExternal(Constants.encryptionBlogLink);
			this.showInstructionTextAsWarning(profile, reconnectAction);
		}
```

src/controllers/connectionManager.ts
```diff
@@ -462,6 +477,27 @@ export default class ConnectionManager {
...
+		await this.vscodeWrapper.showWarningMessageAdvanced(instructionText,
+			{ modal: false },
+			[
+				LocalizedConstants.enableTrustServerCertificate,
+				LocalizedConstants.readMore,
+				LocalizedConstants.cancel
+			])
+			.then(async (selection) => {
+				if (selection && selection === LocalizedConstants.enableTrustServerCertificate) {
+					profile.encrypt = 'Mandatory';
```
Charles-Gagnon, line 491: Would suggest making an enum for the possible (string) encrypt values

src/controllers/mainController.ts
```diff
@@ -151,6 +151,15 @@ export default class MainController implements vscode.Disposable {
 
 			this.initializeObjectExplorer();
 
+			this.registerCommandWithArgs(Constants.cmdConnectObjectExplorerProfile);
+			this._event.on(Constants.cmdConnectObjectExplorerProfile, async (profile: IConnectionProfile) => {
+				await this.createObjectExplorerSession(profile).then(async (result) => {
+					if (result === true) {
+						await this._connectionMgr.connectionUI.saveProfile(profile);
+					}
+				});
```
Charles-Gagnon, line 160: more async/await cleanup
```suggestion
				const result = await this.createObjectExplorerSession(profile);
				if (result === true) {
					await this._connectionMgr.connectionUI.saveProfile(profile);
				}
```

src/controllers/mainController.ts
```diff
@@ -282,15 +291,17 @@ export default class MainController implements vscode.Disposable {
 	/**
 	 * Creates a new Object Explorer session
 	 */
-	private async createObjectExplorerSession(connectionCredentials?: IConnectionInfo): Promise<void> {
+	private async createObjectExplorerSession(connectionCredentials?: IConnectionInfo): Promise<boolean> {
```
Charles-Gagnon, line 294: Expand doc comment to explain what the return value means

src/controllers/mainController.ts
```diff
@@ -308,9 +319,8 @@ export default class MainController implements vscode.Disposable {
 
 		// Sets the correct current node on any node selection
 		this._context.subscriptions.push(treeView.onDidChangeSelection((e: vscode.TreeViewSelectionChangeEvent<TreeNodeInfo>) => {
-			const selections: TreeNodeInfo[] = e.selection;
-			if (selections && selections.length > 0) {
-				self._objectExplorerProvider.currentNode = selections[0];
+			if (e.selection && e.selection.length > 0) {
```
Charles-Gagnon, line 322: nit - e.selection.length > 0 will already be false if selection is undefined/0
```suggestion
			if (e.selection.length > 0) {
```

src/controllers/mainController.ts
```diff
@@ -870,16 +880,31 @@ export default class MainController implements vscode.Disposable {
 	/**
 	 * Prompt the user to view release notes if this is new extension install
 	 */
-	private async showReleaseNotesPrompt(): Promise<void> {
+	private async showFirstLaunchPrompts(): Promise<void> {
```
Charles-Gagnon, line 883: Update doc comments

src/controllers/mainController.ts
```diff
@@ -870,16 +880,31 @@ export default class MainController implements vscode.Disposable {
...
+				LocalizedConstants.releaseNotesPromptDescription, confirmText)
+				.then(async (result) => {
+					if (result === confirmText) {
+						await self.launchReleaseNotesPage();
+					}
+				});
+
+
+			// ask the user to view encryption changes document
+			let confirmTextEncrypt = LocalizedConstants.moreInformation;
```
Charles-Gagnon, line 898: Don't we want to show this even for users who already had the extension installed?

src/controllers/mainController.ts
Charles-Gagnon, line 883:
```suggestion
	private async showFirstLaunchNotifications(): Promise<void> {
```

Charles-Gagnon, line 883: Given that we're not even awaiting this call anyways (and we shouldn't be) I think it'd make more sense to just make this non-async and then leave a comment in the below calls that we're purposely not blocking on the toasts being handled so as not to block extension startup

src/models/connectionStore.ts
```diff
@@ -202,9 +202,9 @@ export class ConnectionStore {
 	 * @param {forceWritePlaintextPassword} whether the plaintext password should be written to the settings file
 	 * @returns {Promise<IConnectionProfile>} a Promise that returns the original profile, for help in chaining calls
 	 */
-	public saveProfile(profile: IConnectionProfile, forceWritePlaintextPassword?: boolean): Promise<IConnectionProfile> {
+	public async saveProfile(profile: IConnectionProfile, forceWritePlaintextPassword?: boolean): Promise<IConnectionProfile> {
 		const self = this;
-		return new Promise<IConnectionProfile>((resolve, reject) => {
+		return new Promise<IConnectionProfile>(async (resolve, reject) => {
```
Charles-Gagnon, line 207: Good opportunity here to apply the async/await style again - makes it much cleaner and easier to read since you can get rid of the whole wrapped promise stuff. (or good task to do separately as a code cleanup thing if you wanted)

src/models/connectionStore.ts
```diff
@@ -341,28 +340,28 @@ export class ConnectionStore {
...
-		return this.doSaveCredential(profile, CredentialsQuickPickItemType.Profile, true);
+		return await this.doSaveCredential(profile, CredentialsQuickPickItemType.Profile, true);
 	}
 
-	private doSaveCredential(conn: IConnectionInfo, type: CredentialsQuickPickItemType, isConnectionString: boolean = false): Promise<boolean> {
+	private async doSaveCredential(conn: IConnectionInfo, type: CredentialsQuickPickItemType, isConnectionString: boolean = false): Promise<boolean> {
 		let self = this;
 		let password = isConnectionString ? conn.connectionString : conn.password;
-		return new Promise<boolean>((resolve, reject) => {
+		return new Promise<boolean>(async (resolve, reject) => {
```
Charles-Gagnon, line 360: Same here with converting this to async/await

src/models/connectionStore.ts
```diff
@@ -459,6 +457,14 @@ export class ConnectionStore {
 		// Connections defined in workspace scope are unioned with the Connections defined in user scope
 		let profilesInConfiguration = this._connectionConfig.getConnections(true);
 
+		// Update encrypt property value for each profile
+		profilesInConfiguration.forEach(async profile => {
```
Charles-Gagnon, line 461: FYI - this currently is going to continue on without awaiting for all the profiles to be updated, which probably isn't what you want. You should be doing `await Promise.all(profilesInConfiguration.map(async profile => ...` to do that.

src/views/connectionUI.ts
```diff
@@ -471,13 +471,13 @@ export class ConnectionUI {
...
 	 */
-	public validateAndSaveProfile(profile: IConnectionProfile): Promise<IConnectionProfile> {
+	public async validateAndSaveProfile(profile: IConnectionProfile): Promise<IConnectionProfile> {
 		const self = this;
 		let uri = self.vscodeWrapper.activeTextEditorUri;
 		if (!uri || !self.vscodeWrapper.isEditingSqlFile) {
 			uri = ObjectExplorerUtils.getNodeUriFromProfile(profile);
 		}
-		return self.connectionManager.connect(uri, profile).then(async (result) => {
+		return await self.connectionManager.connect(uri, profile).then(async (result) => {
```
Charles-Gagnon, line 480: Opportunity to cleanup with async/await here too

src/views/connectionUI.ts
```diff
@@ -674,15 +674,15 @@ export class ConnectionUI {
...
 	// Prompts the user to pick a profile for removal, then removes from the global saved state
-	public removeProfile(): Promise<boolean> {
+	public async removeProfile(): Promise<boolean> {
 		let self = this;
 
 		// Flow: Select profile to remove, confirm removal, remove, notify
 		let profiles = self._connectionStore.getProfilePickListItems(false);
 		return self.selectProfileForRemoval(profiles)
-			.then(profile => {
+			.then(async profile => {
```
Charles-Gagnon, line 683: Opportunity to cleanup with async/await here too

Charles-Gagnon:
Apologies for the large amount of nit comments - I mostly left them here to point out the different types of cleanup to be on the lookout for when you're working in this repo (most of the newer styles such as async/await weren't in use when this stuff was originally made). I'll tone it down in future PRs - and feel free to just mark those as follow up items if you want. 

There are a couple of actual issues/comments to look at too though.

cheenamalhotra:
> Apologies for the large amount of nit comments - I mostly left them here to point out the different types of cleanup to be on the lookout for when you're working in this repo (most of the newer styles such as async/await weren't in use when this stuff was originally made). I'll tone it down in future PRs - and feel free to just mark those as follow up items if you want.
> 
> There are a couple of actual issues/comments to look at too though.

I agree there's a lot of improvement scope in this repository, will address things that are in scope of the PR here.

src/models/connectionStore.ts
cheenamalhotra: I also have a limitation that 'loadAllConnections' is a non-async function.
Should I do `Promise.all(...).then(() => { });` in that case? Any other way of calling async function synchronously?

Charles-Gagnon: No, unfortunately there's no way to do that in TS (wish there was a .Wait() like C#...)

So you'd have to make all calling functions async to fully fix it.

src/controllers/vscodeWrapper.ts
```diff
@@ -295,6 +295,20 @@ export default class VscodeWrapper {
 		return vscode.window.showWarningMessage(Constants.extensionName + ': ' + msg);
 	}
 
+	/**
+	 * Formats and shows a vscode warning message with items
+	 */
+	public showWarningMessageAdvanced(msg: string, messageOptions: vscode.MessageOptions, items: any[]): Thenable<string> {
+		return vscode.window.showWarningMessage(Constants.extensionName + ': ' + msg, messageOptions, ...items);
```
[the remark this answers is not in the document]
cheenamalhotra: That's been the current design so I brought over but will address it.

src/models/connectionStore.ts
cheenamalhotra: I also noticed profilesConfiguration.map would proceed in parallel and causes inconsistent behavior.. I did make the method async and also used for loop to update 1 profile at a time.

src/controllers/mainController.ts
```diff
@@ -230,7 +238,9 @@ export default class MainController implements vscode.Disposable {
 		// Init connection manager and connection MRU
 		this._connectionMgr = new ConnectionManager(this._context, this._statusview, this._prompter);
 
-		this.showReleaseNotesPrompt();
+		// Shows notifications on new extension installtion
+		// This call is intentionally not awaited o avoid blocking extension activation
```
Charles-Gagnon, line 242:
```suggestion
		// This call is intentionally not awaited to avoid blocking extension activation
```

src/controllers/vscodeWrapper.ts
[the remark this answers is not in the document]
Charles-Gagnon: Out of curiosity - I saw you changed it but then reverted that back. What was breaking when you removed this?

src/models/connectionInfo.ts
```diff
@@ -50,17 +52,29 @@ export function fixupConnectionCredentials(connCreds: IConnectionInfo): IConnect
...
-		// Ensure minumum connection timeout if connecting to Azure SQL
+		// Ensure minumum connection timeout when connecting to Azure SQL
 		if (connCreds.connectTimeout < Constants.azureSqlDbConnectionTimeout) {
 			connCreds.connectTimeout = Constants.azureSqlDbConnectionTimeout;
 		}
 	}
 	return connCreds;
 }
 
+export function updateEncrypt(connection: IConnectionInfo): { connection: IConnectionInfo, status: boolean } {
```
Charles-Gagnon, line 66: Consider renaming status to `updatePerformed` to more clearly indicate what happened.

src/models/connectionStore.ts
```diff
@@ -202,38 +199,29 @@ export class ConnectionStore {
...
-				}).then(resolved => {
-					// Add necessary default properties before returning
-					// this is needed to support immediate connections
-					ConnInfo.fixupConnectionCredentials(profile);
-					resolve(profile);
-				}, err => {
-					reject(err);
-				});
-		});
+		await self._connectionConfig.addConnection(savedProfile)
```
Charles-Gagnon, line 216: nit: Another opportunity for clean (especially the catch which just catches and immediately rethrows...)

src/models/connectionStore.ts
```diff
@@ -383,10 +371,10 @@ export class ConnectionStore {
 	 * @param {Boolean} keepCredentialStore optional value to keep the credential store after a profile removal
 	 * @returns {Promise<boolean>} true if successful
 	 */
-	public removeProfile(profile: IConnectionProfile, keepCredentialStore: boolean = false): Promise<boolean> {
+	public async removeProfile(profile: IConnectionProfile, keepCredentialStore: boolean = false): Promise<boolean> {
 		const self = this;
-		return new Promise<boolean>((resolve, reject) => {
-			self._connectionConfig.removeConnection(profile).then(profileFound => {
+		return new Promise<boolean>(async (resolve, reject) => {
```
Charles-Gagnon, line 376: nit: This could be greatly cleaned up with async/await

src/models/connectionStore.ts
```diff
@@ -459,6 +446,14 @@ export class ConnectionStore {
 		// Connections defined in workspace scope are unioned with the Connections defined in user scope
 		let profilesInConfiguration = this._connectionConfig.getConnections(true);
 
+		// Update encrypt property value for each profile synchronously
```
Charles-Gagnon, line 449: Would be good to explain why doing synchronous updates is required

src/views/connectionUI.ts
```diff
@@ -85,32 +85,32 @@ export class ConnectionUI {
...
-					resolve(this.handleSelectedConnection(selection[0]));
-				} else {
-					resolve(undefined);
-				}
-			});
-			connectionProfileQuickPick.onDidHide(() => {
-				connectionProfileQuickPick.dispose();
-				resolve(undefined);
-			});
+		let connectionInfo: IConnectionInfo;
```
Charles-Gagnon, line 88:
```suggestion
		let connectionInfo: IConnectionInfo | undefined;
```

src/views/connectionUI.ts
```diff
@@ -85,32 +85,32 @@ export class ConnectionUI {
 	 * @returns The connectionInfo choosen or created from the user, or undefined if the user cancels the prompt.
 	 */
 	public async promptForConnection(ignoreFocusOut: boolean = false): Promise<IConnectionInfo | undefined> {
```
Charles-Gagnon, line 87: This method won't work correctly anymore - it always returns undefined. Situations like this you need to keep the Promise and resolve/reject because those are done in callbacks. Removing the Promise means it'll go set up the callbacks but then never actually wait for them to be called and instead return immediately.

src/views/connectionUI.ts
```diff
@@ -347,7 +347,7 @@ export class ConnectionUI {
 					connectFunc = this.fillOrPromptForMissingInfo(selection);
 				}
 
-				connectFunc.then((resolvedConnectionCreds) => {
+				await connectFunc.then((resolvedConnectionCreds) => {
```
Charles-Gagnon, line 350: FYI - the await here doesn't really do anything since this whole thing is dependent on the resolve being called (it doesn't really matter when the inner function itself completes). 

It'll still technically work correctly but what was there before was just fine.

src/views/connectionUI.ts
```diff
@@ -674,24 +674,19 @@ export class ConnectionUI {
 	}
 
 	// Prompts the user to pick a profile for removal, then removes from the global saved state
-	public removeProfile(): Promise<boolean> {
+	public async removeProfile(): Promise<boolean> {
 		let self = this;
```
Charles-Gagnon, line 678: While you're looking at this stuff - most of these self = this calls can be removed. That was a pattern required when using function variable callbacks but nowadays it's much simpler to just use lambda (arrow) functions which capture this without needing to do the self=this. So I like to remove this whenever I come across it in code I'm modifying.

Charles-Gagnon (changes requested):
ConnectionUI has an issue you should resolve, and there's some other stuff you should look at too.

src/controllers/vscodeWrapper.ts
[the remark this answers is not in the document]
cheenamalhotra: Because all our prompts include extension name, so I don't think we want to make it fall out of sync:

src/views/connectionUI.ts
cheenamalhotra: Keeping it the same is leading to test failures, as promise is not resolving with current implementation. Any thoughts?

Charles-Gagnon: Oof, yeah so the root cause of this is your change to make loadAllConnections async (because you persist the profiles when you update the connection encrypt field). 

promptForConnection calls getPickListItems which calls loadAllConnections. Previously this was all sync calls so it would complete all the calls in promptForConnection before returning the promise. But now with the async call it hits the await and then moves on to other tasks, which includes returning the created promise (which isn't done setting up all the calls yet such as the onDidChangeSelection callback). 

The test then moves on to firing the event, but because it hasn't hooked up the onDidChangeSelection callback yet that code is never hit so it doesn't resolve the promise like expected and thus the test times out. 

Here's the order of events:
1. Test calls promptForConnection
2. It creates the Promise which then in the constructor calls the callback
3. In the callback we call getPickListItems - because it's a promise it puts it in the task queue and moves on to other stuff
4. promptForConnection gets control again and returns the promise (which is still in the process of being set up)
5. The test fires the event
...some time later the call to getPickListItems completes and then it finally finishes setting up the promise (which includes hooking up the onDidChangeSelection callback)

With the way this code is currently written we pretty much can't have any awaitable code inside promptForConnection because of this.

My suggestion to fix this is don't check/update the encrypt value every single time we call loadAllConnections. This seems like it should just be a one-time thing anyways (done on extension load) so maybe you can make a separate call to just fix up all the encrypt values on extension load instead of doing it every single time that we call loadAllConnections?

src/views/connectionUI.ts
```diff
@@ -85,8 +85,10 @@ export class ConnectionUI {
 	 * @returns The connectionInfo choosen or created from the user, or undefined if the user cancels the prompt.
 	 */
 	public async promptForConnection(ignoreFocusOut: boolean = false): Promise<IConnectionInfo | undefined> {
-		return await new Promise<IConnectionInfo | undefined>((resolve, _) => {
-			let connectionProfileList = this._connectionStore.getPickListItems();
+		// Let this design use Promise and resolve/reject pattern instead of async/await
+		// because resolve/reject is done in in callback events.
+		return await new Promise<IConnectionInfo | undefined>(async (resolve, _) => {
```
Charles-Gagnon, line 90: You also need to remove the await here. This is going to force the function to wait until the promise resolves (which means the user has picked an item) which is breaking the test as well.

cheenamalhotra: That had been there before as well..

Charles-Gagnon: Oh yeah, don't know what I was thinking here. It would only be a problem if the test awaited the returned promise first. So yeah nevermind.

(although personally I don't like "return await" - it's just more verbose and unnecessarily unwrapping and re-wrapping the returned value. The only reason you'd want to do it like this is to keep the callstack exactly correct but that isn't something I've ever seen as being super necessary)

src/controllers/mainController.ts
```diff
@@ -230,7 +239,9 @@ export default class MainController implements vscode.Disposable {
 		// Init connection manager and connection MRU
 		this._connectionMgr = new ConnectionManager(this._context, this._statusview, this._prompter);
 
-		this.showReleaseNotesPrompt();
+		// Shows notifications on new extension installtion
```
Charles-Gagnon, line 242:
```suggestion
		// Shows first time notifications on extension installation or update
```

src/controllers/mainController.ts
```diff
@@ -262,6 +273,11 @@ export default class MainController implements vscode.Disposable {
 					conn.password = '';
 					profileChanged = true;
 				}
+				// Fixup 'Encrypt' property if needed
+				let result = ConnInfo.updateEncrypt(conn);
+				if (result.updateStatus) {
+					await this.connectionManager.connectionStore.saveProfile(result.connection as IConnectionProfile);
```
Charles-Gagnon, line 279: Do you need to call saveProfile here vs just setting profileChanged to true? Maybe even call saveProfile instead of setConfiguration directly since that seems like what we want to be doing anyways...

src/models/connectionStore.ts
```diff
@@ -137,29 +137,26 @@ export class ConnectionStore {
...
+		} else if (self.shouldLookupSavedPassword(<IConnectionProfile>credentialsItem.connectionCreds)) {
+			let credentialId = ConnectionStore.formatCredentialIdForCred(credentialsItem.connectionCreds, credentialsItem.quickPickItemType);
+			const savedCred = await self._credentialStore.readCredential(credentialId);
+			if (savedCred) {
+				credentialsItem.connectionCreds.password = savedCred.password;
+				return credentialsItem;
 			} else {
-				// Already have a password, no need to look up
-				resolve(credentialsItem);
+				Promise.reject(new Error('No saved password found'));
```
Charles-Gagnon, line 154: This isn't doing anything - Promise.reject just returns a promise that's rejected with the given value. But it's up to you to do something with that promise. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject

```suggestion
				throw new Error('No saved password found');
```

src/controllers/connectionManager.ts
```diff
@@ -462,6 +477,25 @@ export default class ConnectionManager {
...
+			{ modal: false },
+			[
+				LocalizedConstants.enableTrustServerCertificate,
+				LocalizedConstants.readMore,
+				LocalizedConstants.cancel
+			]);
+		if (selection === LocalizedConstants.enableTrustServerCertificate) {
+			profile.encrypt = EncryptOptions.Mandatory;
+			profile.trustServerCertificate = true;
+			await reconnectAction(profile);
```
Charles-Gagnon, line 492: Make sure you're being careful with these callbacks - if an error occurs in them (such as saveProfile throwing an error) then it'll silently fail. It'd be good to at the very least log the error in that case.

src/controllers/mainController.ts
cheenamalhotra: SaveProfile is needed here, just for the same reason as 'saveProfilePasswordIfNeeded' is called above.

src/controllers/connectionManager.ts
cheenamalhotra: saveProfile is done as part of 'reconnectAction' and the delegate takes care of throwing error.

src/views/connectionUI.ts
```diff
@@ -335,8 +337,8 @@ export class ConnectionUI {
 		});
 	}
 
-	private handleSelectedConnection(selection: IConnectionCredentialsQuickPickItem): Promise<IConnectionInfo> {
-		return new Promise<IConnectionInfo>((resolve, reject) => {
+	private async handleSelectedConnection(selection: IConnectionCredentialsQuickPickItem): Promise<IConnectionInfo> {
+		return new Promise<IConnectionInfo>(async (resolve, reject) => {
```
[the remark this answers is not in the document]
cheenamalhotra: I can remove the second async from lambda function.

```suggestion
	private async handleSelectedConnection(selection: IConnectionCredentialsQuickPickItem): Promise<IConnectionInfo> {
		return new Promise<IConnectionInfo>((resolve, reject) => {
```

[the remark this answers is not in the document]
Charles-Gagnon: Async is just syntactic sugar - all it does is allow you to use the await keyword and automatically wrap return values in promises. So if you don't use await then you can remove async since it doesn't actually do anything in that case. 

All async functions must return a promise, but not all functions that return a promise must be async.

src/controllers/mainController.ts
Charles-Gagnon: I was more asking why we have different ways of "saving" a profile in this function. It just makes it confusing to try and follow and can easily lead to errors down the line if someone doesn't realize how these two differ.

And I would more think that we should be calling saveProfile in all cases when a profile is changed. I'd have to spend more time following the code path to understand why we'd want to save the profile in one case but not another.

If you've done that yourself and it makes sense then that's fine (although i would love to see some comments added to help future people looking at this...).

src/controllers/connectionManager.ts
Charles-Gagnon: Yes, but nothing is handling the error. So if the callback/delegate throws then it will just silently fail.

src/views/connectionUI.ts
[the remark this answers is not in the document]
cheenamalhotra: It doesn't matter, since there's no awaited call here, will remove it. ☮️

src/controllers/connectionManager.ts
cheenamalhotra: What do you mean? Reconnect action is awaited.

Charles-Gagnon: But the calls to showInstructionTextAsWarning are not. 

Here's an example - you call showInstructionTextAsWarning here :

https://github.com/microsoft/vscode-mssql/pull/17484/files/050876a8d911f5456edd4cc4db39786a29bacc5d#diff-5eeace56def153787efa337fcd2f2e2771a202d038a1da07302cb41f6672a255R136

That call isn't awaited so it happily continues execution in that function. 

Meanwhile this function is running and eventually calls the reconnectAction. In this example let's say that the call to [saveProfile](https://github.com/microsoft/vscode-mssql/pull/17484/files/050876a8d911f5456edd4cc4db39786a29bacc5d#diff-5eeace56def153787efa337fcd2f2e2771a202d038a1da07302cb41f6672a255R141) throws an error. There's no catch in the delegate so it bubbles back up to this function here. This also doesn't have anything to catch it so it continues up the chain.

But because you didn't await (or have a .catch) on the call that error is never handled and lost forever.

cheenamalhotra: Yeah, I can't await either. I did test and noticed the error is thrown in console. I can add catch blocks to the calls, bt I don't think that's going to help either.

Charles-Gagnon: Typically in that you'd either retry or display an error to the user so they at least know something went wrong. Displaying an error is probably easiest - although keep in mind that for this stuff one of the things you're trying to do is reconnect which is much more likely to fail (due to any number of issues from network problems to server being shut off....) and so that might impact the usefulness of what you're trying to do here. 

If you did consider a retry you could show an error dialog with the error and have a button for "Retry" that will allow them to continue retrying it as many times they wanted in case it seems like a transient error.

cheenamalhotra: Bug to track this: https://github.com/microsoft/vscode-mssql/issues/17501

cheenamalhotra: I've fixed this behavior for this PR, so we have it working, and on failure to save profile, we would get:

image

I will fix the issue in main connection flow with the issue #17501

PR #17487, opened 2022-11-29
Don't package unnecessary files

Charles-Gagnon:
Noticed we were including a few test run artifacts in the final packaged VSIX - removing those to keep the size down

(technically we could probably remove some of the md files too like the changelog since that isn't used by VS Code. But those are small enough that it's fine for now and people might actually want to be able to see that stuff in the unpacked extension directory anyways)

After removal : 

image

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -1,19 +1,23 @@
 .github
 .vscode
 build
+coverage
+coverage-remapped
 lib
 localization
 out/test
+packages
 samples
 src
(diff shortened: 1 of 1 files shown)


cheenamalhotra (approved):

alanrenmsft:
what is the size difference?

Charles-Gagnon:
7,585 KB -> 7,143 KB

PR #17490, opened 2022-11-29
Updating qs

aasimkhan30:


Charles-Gagnon:
@aasimkhan30 - @cheenamalhotra is already doing this in https://github.com/microsoft/vscode-mssql/pull/17489

aasimkhan30:
@cheenamalhotra make sure you also update qs@~6.5.2. Thank you!

PR #17492, opened 2022-12-01
Adjusts pipeline to include extension code signing

lewis-sanchez:


build/build.yml
```diff
@@ -105,6 +105,45 @@ steps:
       BuildDropPath: '$(Build.ArtifactStagingDirectory)'
       PackageName: 'MSSQL for VS Code'
 
+  - task: UseDotNet@2
```
Charles-Gagnon, line 108: Do we actually need this? The agents come with stuff already installed so this might not be necessary

build/build.yml
```diff
@@ -105,6 +105,45 @@ steps:
...
+    inputs:
+      packageType: sdk
+      version: 5.0.x
+      installationPath: $(Agent.ToolsDirectory)/dotnet
+
+  - task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@1
+    inputs:
+      ConnectedServiceName: 'Code Signing'
+      FolderPath: '$(Build.SourcesDirectory)/.build'
+      Pattern: 'extensions/*.vsix,langpacks/*.vsix'
```
Charles-Gagnon, line 119: Wrong path here, you need to point it at the output path where the VSIX is generated for this repo

build/build.yml
```diff
@@ -105,6 +105,45 @@ steps:
...
+          {
+            "keyCode": "CP-233016",
+            "operationSetCode": "OpcVerify",
+            "parameters": [],
+            "toolName": "sign",
+            "toolVersion": "1.0"
+          }
+        ]
+      SessionTimeout: 120
+    displayName: 'Signing Extensions and Langpacks'
```
Charles-Gagnon, line 144:
```suggestion
    displayName: 'Signing VSIX'
```

build/build.yml
```diff
@@ -105,6 +105,45 @@ steps:
...
+    displayName: 'Install .NET Core sdk for signing'
+    inputs:
+      packageType: sdk
+      version: 5.0.x
+      installationPath: $(Agent.ToolsDirectory)/dotnet
+
+  - task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@1
+    inputs:
+      ConnectedServiceName: 'Code Signing'
+      FolderPath: '$(Build.SourcesDirectory)/.build'
```
Charles-Gagnon, line 118: There is no .build directory in this repo

Charles-Gagnon (changes requested):

Charles-Gagnon:
I would suggest you get an ad-hoc build with signing enabled done first to make sure things are done correctly

Charles-Gagnon (approved):
Make sure you've produced a test VSIX using the adhoc build (with signed = true) and verified that it is : 

1. Signed
2. Installable by VS Code without any issues

build/build.yml
```diff
@@ -51,6 +51,80 @@ steps:
       summaryFileLocation: $(System.DefaultWorkingDirectory)/coverage/cobertura-coverage.xml
       additionalCodeCoverageFiles: $(System.DefaultWorkingDirectory)/coverage-remapped/coverage.json
 
+  - task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@1
```
Charles-Gagnon, line 54: This shouldn't be done as a separate step - you can just add the signing operations to the one below.

(and all the entries in the file patterns list shouldn't be necessary. We're only signing the VSIX's here)

Charles-Gagnon: Also - did you verify that authenticode signing doesn't affect the ability to install and run the extensions from VS Code?

lewis-sanchez:
> Make sure you've produced a test VSIX using the adhoc build (with signed = true) and verified that it is :
> 
> 1. Signed
> 2. Installable by VS Code without any issues

I verified that the extension is signed and I was able to install the extension on VSCode without trouble to run queries against local DB's and Azure SQL DB's successfully.

PR #17512, opened 2023-01-09
Improve error handling and update profile when returned

cheenamalhotra:
Improves error handling experience to avoid double prompts and updates profile properly for re-attempted connection.
Fixes issue for SQL Database Projects where profile wasn't updated on return.

**Updated experience:**

VScode_NewExp

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -86,6 +86,7 @@ export default class ConnectionManager {
 		Map<IConnectionInfo, ServerInfo>;
 	private _uriToConnectionPromiseMap: Map<string, Deferred<boolean>>;
 	private _failedUriToFirewallIpMap: Map<string, string>;
+	private _failedUriToSSLMap: Map<string, string>;
 	private _accountService: AccountService;
 	private _firewallService: FirewallService;
 	public azureController: AzureController;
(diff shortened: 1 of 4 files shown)


kisantia:
I tried this out with the sql projects vscode extension, but `vscodeMssqlApi.promptForConnection(true)` is returning undefined before the "Enable Trust Certificate" prompt shows. Can vscode-mssql handle this before returning the result from `promptForConnection()`?
https://user-images.githubusercontent.com/31145923/211367764-c7043d52-ddb0-419a-974d-373a108f4f24.mp4

src/controllers/connectionManager.ts
```diff
@@ -496,6 +501,16 @@ export default class ConnectionManager {
 		}
 	}
 
+	public async handleSSLError(uri: string, profile: IConnectionProfile): Promise<IConnectionInfo> {
+		let updatedConn: IConnectionInfo | undefined;
+		this.showInstructionTextAsWarning(profile, async updatedConnection => {
```
Charles-Gagnon, line 506: You aren't waiting for the callback to actually finish so this isn't going to work as expected (it will immediately move on and return updatedConn which is currently undefined). You need to await the call to showInstructionTextAsWarning

```suggestion
		await this.showInstructionTextAsWarning(profile, async updatedConnection => {
```

src/controllers/connectionManager.ts
```diff
@@ -496,6 +501,16 @@ export default class ConnectionManager {
 		}
 	}
 
+	public async handleSSLError(uri: string, profile: IConnectionProfile): Promise<IConnectionInfo> {
```
Charles-Gagnon, line 504:
```suggestion
	public async handleSSLError(uri: string, profile: IConnectionProfile): Promise<IConnectionInfo | undefined> {
```

src/controllers/connectionManager.ts
cheenamalhotra: Makes sense, previously it couldn't be awaited so I didn't update that after moving.. thanks for catching!

kisantia (approved):
I verified that sql projects publish flow no longer exits when using a new connection with the changes in this branch. Thanks for fixing this Cheena!

PR #17543, opened 2023-02-14
Adding Group by schema to OE

aasimkhan30:
Menu item:
image
image

When feature enabled:
image
When feature disabled:
image

Setting will auto refresh:
GIF:
Code_zRwQATuUar

Light Mode:
image
Dark Mode:
image
HC:
image


Charles-Gagnon (commented):
This isn't very discoverable. Have you looking into adding inline action items and/or right click menu item support?

aasimkhan30:
@charles, added a menu item to connection OE tree

package.json
```diff
@@ -247,6 +247,18 @@
           "when": "view == queryHistory",
           "title": "%mssql.clearAllQueryHistory%",
           "group": "secondary"
+        },
+        {
+          "command": "mssql.objectExplorer.enableGroupBySchema",
+          "when": "view == objectExplorer && !config.mssql.objectExplorer.groupBySchema",
+          "title": "%mssql.objectExplorer.enableGroupBySchema%",
+          "group": "secondary"
```
Charles-Gagnon, line 255: For VS Code MSSQL seems like it'd be better to have it as a navigation item so it shows up on the title bar. That's a lot easier to see and work with - the secondary menu is still a bit too hidden IMO. 

@erinstellato-ms thoughts on this?

aasimkhan30: IMO I don't think people will toggle this setting a lot. Therefore, I am not sure about making it visible all the time.  They will either stick with schema tree or regular tree. We can enable this setting by default in the next version and let users disable it if they want to.

Charles-Gagnon: It's not like there isn't plenty of room there. And I just think that hiding it in the secondary menu means that most people aren't going to know it exists unless they're really diligent about reading the release notes. 

Just seems a shame to put all the work into this feature and then make it difficult for people to discover it. We can always hide the action later if people don't like it there.

alanrenmsft:
1. any reason we are changing the icon for schema? previously it looks like this 
image
2. I don't think we should be add the menu to the connection view as I mentioned earlier it is only applicable to MSSQL. Adding it to the connection context menu/inline action seems to be more appropriate, user might think it is a per connection setting, maybe we will see requests asking for per connection setting support.

erinstellato-ms:
@Charles-Gagnon We (@aasimkhan30 and I) talked about how to make it more discoverable.  I am planning to collect feedback from users once it is in the Insiders build to see what ideas/suggestions they have.

@alanrenmsft in terms of the icon, I honestly did like the one we had, it looked like a b-tree index to me.  The new icon is (to me) more intuitive because it has the database + schema.  Again, something I was planning to ask for user feedback on.  

Happy to have a quick chat tomorrow if ya'll want to discuss more!

dzsquared:
As far as the ordering of the schemas - the default `db_` schemas (everything before dbo) are the least likely item for users to engage with and probably should be artificially shuffled to the end.

Charles-Gagnon:
@alanrenmsft I'm suggesting the inline action button here because we only support MSSQL connections in the VS Code extension. I agree that finding some other way (such as context menu items) for ADS seems reasonable there.

package.json
```diff
@@ -1137,6 +1159,11 @@
           "type": "boolean",
           "description": "%mssql.ignorePlatformWarning%",
           "default": false
+        },
+        "mssql.objectExplorer.groupBySchema": {
+          "type": "boolean",
+          "description": "%mssql.objectExplorer.groupBySchema%",
+          "default": true
```
Charles-Gagnon, line 1166: Are we sure we want to default to true? That seems like it would just annoy people who don't care/don't want it grouped by schema. In general we should probably strive to not change existing defaults unless there's a really good reason to.

src/controllers/mainController.ts
```diff
@@ -1178,6 +1187,19 @@ export default class MainController implements vscode.Disposable {
 
 			await this.sanitizeConnectionProfiles();
 
+			if (e.affectsConfiguration(Constants.cmdObjectExplorerGroupBySchemaFlagName)) {
+				this._objectExplorerProvider.rootNodeConnections.forEach(async (conn) => {
+					if (this.connectionManager.isActiveConnection(conn)) {
+						const rootNodes = await this._objectExplorerProvider.getChildren();
+						rootNodes.forEach(async (rootNode: TreeNodeInfo) => {
+							if (rootNode.connectionInfo.accountId === conn.accountId) {
```
Charles-Gagnon, line 1195: What is this check for? Shouldn't we be able to just refresh the root nodes themselves regardless and have it pick up the new grouping that way?

src/controllers/mainController.ts
```diff
@@ -1178,6 +1187,19 @@ export default class MainController implements vscode.Disposable {
 
 			await this.sanitizeConnectionProfiles();
 
+			if (e.affectsConfiguration(Constants.cmdObjectExplorerGroupBySchemaFlagName)) {
+				this._objectExplorerProvider.rootNodeConnections.forEach(async (conn) => {
+					if (this.connectionManager.isActiveConnection(conn)) {
+						const rootNodes = await this._objectExplorerProvider.getChildren();
+						rootNodes.forEach(async (rootNode: TreeNodeInfo) => {
```
Charles-Gagnon, line 1194: Currently doing this means that if any errors occur while refreshing the nodes they'll be lost. I would suggest at least logging the errors if they happen (which means try/catch around the awaits)

src/controllers/mainController.ts
```diff
@@ -1178,6 +1187,19 @@ export default class MainController implements vscode.Disposable {
 
 			await this.sanitizeConnectionProfiles();
 
+			if (e.affectsConfiguration(Constants.cmdObjectExplorerGroupBySchemaFlagName)) {
+				this._objectExplorerProvider.rootNodeConnections.forEach(async (conn) => {
+					if (this.connectionManager.isActiveConnection(conn)) {
+						const rootNodes = await this._objectExplorerProvider.getChildren();
```
Charles-Gagnon, line 1193: You're already looping over rootNodeConnections - do you really have to call getChildren each time? It seems redundant to have to do that again for each node.

aasimkhan30:
> 1. any reason we are changing the icon for schema? previously it looks like this
>    image
> 2. I don't think we should be add the menu to the connection view as I mentioned earlier it is only applicable to MSSQL. Adding it to the connection context menu/inline action seems to be more appropriate, user might think it is a per connection setting, maybe we will see requests asking for per connection setting support.

@alanrenmsft , it is fine to add icon to connections view here because MSSQL extension only supports sql server in vscode.

package.json
alanrenmsft: agreed, default should be false.

src/controllers/mainController.ts
```diff
@@ -1178,6 +1187,17 @@ export default class MainController implements vscode.Disposable {
 
 			await this.sanitizeConnectionProfiles();
 
+			if (e.affectsConfiguration(Constants.cmdObjectExplorerGroupBySchemaFlagName)) {
+				(await this._objectExplorerProvider.getChildren()).forEach((n: TreeNodeInfo) => {
+					try {
+						this._objectExplorerProvider.refreshNode(n);
+					} catch (e) {
+						Utils.logDebug(e.toString());
```
alanrenmsft, line 1195: probably not needed since the information is the same as the error message.

src/controllers/mainController.ts
```diff
@@ -1178,6 +1187,17 @@ export default class MainController implements vscode.Disposable {
...
+						Utils.showErrorMsg(e.toString());
```
Charles-Gagnon, line 1196: Currently if something causes a couple of the refreshes to fail then this will get really spammy. Two options I would suggest you consider : 

1. Just logging to console
2. Only display a single message to the user if any of the node expansions failed - and optionally direct them to the console/output window logs for further information

aasimkhan30: Don't errors get automatically logged to console?

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -587,6 +587,9 @@
         <trans-unit id="columnWidthMustBePositiveError">
             <source xml:lang="en">Width cannot be 0 or negative</source>
         </trans-unit>
+        <trans-unit id="objectExplorerNodeRefreshError">
+            <source xml:lang="en">Error occured while refreshing nodes</source>
```
Charles-Gagnon, line 591: Error messages displayed to the user should be as helpful as possible - this is too generic to be much use. Something like `An error occurred refreshing the {NodeName} node. See the [MSSQL output](command:<command here>)` would be much more useful IMO. 

Although I suppose we generally already have logic to display an error in the tree view itself if node expansion fails - so it would probably be worth it to check if that happens when refreshing nodes as well (in which case displaying a separate message to the user is probably unnecessary)

localization/xliff/enu/localizedPackage.json.enu.xlf
```diff
@@ -392,6 +392,15 @@
         <trans-unit id="mssql.deviceCode.description">
             <source xml:lang="en">Allows users to sign in to input-constrained devices.</source>
         </trans-unit>
+        <trans-unit id="mssql.objectExplorer.groupBySchema">
+            <source xml:lang="en">When enabled, the database objects in Object explorer will be categorized by schema.</source>
```
kburtram, line 396: nit `Object Explorer`

kburtram (approved):
LGTM.

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -587,6 +587,12 @@
         <trans-unit id="columnWidthMustBePositiveError">
             <source xml:lang="en">Width cannot be 0 or negative</source>
         </trans-unit>
+        <trans-unit id="objectExplorerNodeRefreshError">
+            <source xml:lang="en">An error occurred refreshing nodes. See the MSSQL output</source>
```
Charles-Gagnon, line 591:
```suggestion
            <source xml:lang="en">An error occurred refreshing nodes. See the MSSQL output channel for more details.</source>
```

Charles-Gagnon (approved):
I still think you should have this be a primary action on the title bar, but you and the PMs can make that decision.

PR #17563, opened 2023-03-01
Updating vscode-mssql types with SqlProjects service types

Benjin:


typings/vscode-mssql.d.ts
```diff
@@ -433,6 +438,297 @@ declare module 'vscode-mssql' {
 		validateStreamingJob(packageFilePath: string, createStreamingJobTsql: string): Thenable<ValidateStreamingJobResult>;
 	}
 
+	/**
+	 * Interface for working with .sqlproj files
+	 */
+	export interface ISqlProjectsService {
```
Charles-Gagnon, line 444: I wonder if at some point it would be easier to split the sql proj typings off into their own typings file just to keep everything related in one place.

Charles-Gagnon (approved):
Bit more than just a typings update here 😅 

(you also have a few linting issues to fix it looks like....)

typings/vscode-mssql.d.ts
Benjin: Yeah... I start getting an itch to refactor a file around 1k lines, and we're certainly hitting that here.

kisantia (approved):

PR #17562, opened 2023-02-28
MSAL Authentication support + code cleanup

cheenamalhotra:
_Companion PR in STS: https://github.com/microsoft/sqltoolsservice/pull/1860_
This PR brings a lot of changes and bug fixes with existing ADAL design and introduces MSAL authentication support.

### New Settings
- Setting: **Mssql: Azure Authentication Library**: ADAL | MSAL
  - Allows users to toggle authentication library - requires reloading VS Code - requires reloading VS Code on modification.
again, without re-adding profiles.
- Setting: **Mssql: Enable Sql Authentication Provider**: true | false
  - Enables silent token acquisition from backend STS using MSAL.NET instead of MSAL.js
  - Only supported for MSAL Authentication mode, and is enabled by default - requires reloading VS Code on modification.

### New Command
- Command: **MS SQL: Add Azure Account**
  - Triggers adding Azure account flow without the need to make a connection.
  - This will be useful if connection cannot be made with currently added account, users can remove and add their account 

### Features
- Implements Auth Code Grant and Device Code auth modes with MSAL along-with persistent token caching.
- Implements HttpClient to enable proxy authentication support in the extension. This implementation is taken from msal.js code sample: https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-node-samples/custom-INetworkModule-and-network-tracing#implement-a-custom-inetworkmodule as previous implementation used axios that doesn't support proxy.

### Bug Fixes
- Fixes #17557 
- Fixes #17560 
- Fixes #17559 
  (with MSAL & SQL Authentication Provider enabled - default behavior)

### Others
- Code health / ES lint recommendations, etc.
- `http-proxy-agent` and `https-proxy-agent` npm packages caused conflicts with MSAL npm packages, and therefore needed to be updated. For more info: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4879#issuecomment-1444047007

### Dependencies
- **@microsoft/ads-adal-library** npm package will also require an update from lib\ads-adal-library to bring in fixes.

PR #17579, opened 2023-03-06
Resolve with higher version of 'lru-cache'

### LINKED ISSUE #17578: extension cannot be activated due to missing module
download the latest build: https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=192754&view=artifacts&pathAsName=false&type=publishedArtifacts

the extension won't launch:

image

cheenamalhotra:
Fixes #17578 

It's strange this issue doesn't occur in local development/testing.
With this change, I'm resolving higher version of 'lru-cache' that was added in PR #17562:

- @azure/msal-node@^1.15.0
  - jsonwebtoken@^9.0.0
    - semver@^7.3.8
      - lru-cache@^6.0.0

Tested by packaging and installing locally, issue seems to be resolved:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,9 @@
     "vscode-nls": "^2.0.2",
     "zone.js": "^0.6.26"
   },
+  "resolutions": {
+    "lru-cache": "^6.0.0"
+  },
   "capabilities": {
     "untrustedWorkspaces": {
       "supported": true
--- a/yarn.lock
+++ b/yarn.lock
@@ -3647,15 +3647,7 @@ loud-rejection@^1.0.0:
     currently-unhandled "^0.4.1"
     signal-exit "^3.0.0"
 
-lru-cache@^4.1.5:
-  version "4.1.5"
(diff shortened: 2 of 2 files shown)


alanrenmsft (approved):

alanrenmsft:
I'll test it a bit later this evening.

alanrenmsft:
tested on Windows. works fine.

Charles-Gagnon:
Seems like some issue with nested dependencies. There's multiple versions of lru-cache currently - one is from `gulp-json-editor` (4.1.5) and the other from `@azure/msal-node`. This works fine because locally you end up with nested dependencies - e.g. 

`C:\src\vscode-mssql\node_modules\jsonwebtoken\node_modules\lru-cache`

So the top level one is the 4.1.5 and the nested on is 6.0.0.

But for some reason it looks like VSCE isn't packaging up the nested 6.0.0 dependency. Maybe it just picks up the first one it sees and thus skips the nested one.

Anyways this fix should be fine as a quick fix since the 4.1.5 is just from the gulp-json-editor but given that we aren't even really using that (it was just for some gulp tasks to enable/disable code coverage - which isn't something we currently use) a cleaner fix would probably be just to remove that package completely.

alanrenmsft:
are you waiting for something? please also port the fix to the release branch.

PR #17581, opened 2023-03-07
Update STS to 4.5.1.1

kburtram:
Update STS in March release to match bump in ADS.


kburtram:
This is being merged as part of the "finish user management feature" changes & keeping vscode-mssql & ADS in sync.

PR #17611, opened 2023-03-11
Skip tenant question with sql auth provider enabled

### LINKED ISSUE #17609: User is prompted to select tenant when SQL Auth Provider is enabled
- MSSQL Extension Version: 1.18.0

According to https://github.com/microsoft/azuredatastudio/issues/22298#issuecomment-1464313921 this isn't supposed to prompt the user for the tenant, but I'm still seeing the prompt. 

It does seem to correctly ignore the selection though (I can connect to a server no matter the tenant I select) but the tenant ID is still saved in the settings. That's probably the behavior I was seeing that made me file https://github.com/microsoft/vscode-mssql/issues/17598 - but for that it feels like the "correct" thing here is to not prompt for the tenant ID at all (which means tenantId would be undefined in the settings)

cheenamalhotra:
When `mssql:enabledSqlAuthenticationProvider` is enabled, tenant question is not needed, as tenant is received from server during login as part of Authority URI. Fixes #17609

### DIFF
--- a/src/models/connectionProfile.ts
+++ b/src/models/connectionProfile.ts
@@ -109,6 +109,8 @@ export class ConnectionProfile extends ConnectionCredentials implements IConnect
 				message: LocalizedConstants.azureChooseTenant,
 				choices: tenantChoices,
 				default: defaultProfileValues ? defaultProfileValues.tenantId : undefined,
+				// Need not prompt for tenant question when 'Sql Authentication Provider' is enabled,
+				// since tenant information is received from Server with authority URI in the Login flow.
 				shouldPrompt: (answers) => profile.isAzureActiveDirectory() && tenantChoices.length > 1 && !getEnableSqlAuthenticationProviderConfig(),
 				onAnswered: (value: ITenant) => {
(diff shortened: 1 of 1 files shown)


src/models/connectionProfile.ts
```diff
@@ -107,7 +108,8 @@ export class ConnectionProfile extends ConnectionCredentials implements IConnect
 				name: LocalizedConstants.tenant,
 				message: LocalizedConstants.azureChooseTenant,
 				choices: tenantChoices,
-				shouldPrompt: (answers) => profile.isAzureActiveDirectory() && tenantChoices.length > 1,
+				default: defaultProfileValues ? defaultProfileValues.tenantId : undefined,
+				shouldPrompt: (answers) => profile.isAzureActiveDirectory() && tenantChoices.length > 1 && !getEnableSqlAuthenticationProviderConfig(),
```
Charles-Gagnon, line 112: Would suggest adding a comment here explaining why this isn't necessary when SQL auth provider is enabled

Charles-Gagnon (approved):

Issue #16967, opened 2021-03-09
Native VScode on M1 (Apple Silicon) breaks extension 


roelrens:
- MSSQL Extension Version:v1.10.1
- VSCode Version: 1.55.0-insider (Universal)
- OS Version: 11.2.3 (20D91)

Steps to Reproduce:

1. use the universal binary of VSCode
2. extension support broken.

davemcdermid:
Looks like this is can be fixed with a small change to platform mapping, I've created a PR: https://github.com/microsoft/vscode-mssql/pull/16969

leighghunt:
I've just run into this whilst setting up a new M1 Air - can help test if required

dzsquared:
There is now a preview release available here: https://github.com/microsoft/vscode-mssql/releases/tag/v1.10.2

Next stable release is planned for end of October

brunovieira97:
Any news on this?

leighghunt:
> Any news on this?

I'm not on the team or anything, but I'm successfully using v1.13.0, just downloaded from the marketplace if I recall correctly - and i can see the change that was made is in the v1.13.0 code branch - https://github.com/microsoft/vscode-mssql/blob/v1.13.0/src/models/platform.ts#L292

brunovieira97:
@leighghunt I see. Thanks for the heads up. I was checking up on this because it seems to be a dependency for Azure Data Studio to be properly ported to Apple Silicon. Glad to know this is taken care of.

nick-gibb:
+1, would be great if we could get this to run using ARM64.

Nova-Logic:
+1 waiting m1 version

Issue #1791, opened 2020-11-09
Failure starting client on Windows 10


fonnesbeck:
- MSSQL Extension Version: 1.9.0
- VSCode Version: 1.51.0
- OS Version: Windows 10 Pro

Steps to Reproduce:

1. Install SQL Server extension
2. Click on SQL Server extension in side bar

When the extension is run for the first time, it fails. The SQLToolsService output contains the following error:

```
[Error - 9:40:56 AM] Starting client failed
Error: spawn UNKNOWN
    at ChildProcess.spawn (internal/child_process.js:394:11)
    at Object.spawn (child_process.js:549:9)
    at c:\Users\fonne\.vscode\extensions\ms-mssql.mssql-1.9.0\node_modules\vscode-languageclient\lib\main.js:354:40
```

![image](https://user-images.githubusercontent.com/81476/98609856-e87e1700-22b3-11eb-9037-c72f5f30a1e9.png)

fonnesbeck:
Is there any reason 1.10.0 should not be available to me? I am at 1.9.0, but there is no update available, yet the README says 1.10.0 is the most recent version.

abist:
@fonnesbeck 1.10.0 will be releasing today or tomorrow.

fonnesbeck:
Thanks. Will this address the issue above?

abist:
@fonnesbeck You shouldn't be seeing this issue. Could you Toggle Developer Tools from the Help Menu button and click on Console and see if there are any errors there?

fonnesbeck:
I see the following, in addition to the error reported above:

```
Couldn't start client AzureResourceProvider
onDidChangeNotification @ breadcrumbsModel.ts:39

Couldn't start client SQLToolsService
onDidChangeNotification @ breadcrumbsModel.ts:39

notificationsAlerts.ts:40 mssql: Failed to load Sql Tools Service
onDidChangeNotification @ breadcrumbsModel.ts:39

Activating extension 'ms-mssql.mssql' failed: spawn UNKNOWN.
_logMessageInConsole @ abstractExtensionService.ts:663

rejected promise not handled within 1 second: Error: spawn UNKNOWN
t.log @ console.ts:140

```

DCAL12:
I'm having the same issue and same error output, also on version 1.9.0 with a fresh WIndows 10 install and VS Code version 1.51.1. When is version 1.10.0 expected to be available?

fonnesbeck:
[...]

PR #17613, opened 2023-03-13
use native arm64 STS bits

alanrenmsft:
This PR fixes: #16967
This PR fixes: #1791

win-arm64: 
image

osx-arm64:
image

also added platform specific builds for win-arm64 and osx-arm64
image

I'll work with Erica to update the release steps to make sure the new flavors are added to the release.


kburtram (approved):

alanrenmsft:
I just checked the release pipeline, since it doesn't have a fixed list of things to upload, we should be good there.
image

src/models/platform.ts
```diff
@@ -232,7 +236,9 @@ export class PlatformInformation {
 
 	private static getWindowsArchitecture(): Promise<string> {
 		return new Promise<string>((resolve, reject) => {
-			if (process.env.PROCESSOR_ARCHITECTURE === 'x86' && process.env.PROCESSOR_ARCHITEW6432 === undefined) {
+			if (process.env.PROCESSOR_ARCHITECTURE === 'ARM64') {
```
Charles-Gagnon, line 239: This doesn't look like it actuallly needs to return a promise

Charles-Gagnon (approved):

src/models/platform.ts
alanrenmsft: the method itself doesn't have to be, but the place it is called also needs to be modified, so I'd leave it as is.
image

PR #17625, opened 2023-03-15
MSAL cache encryption support

cheenamalhotra:
Introduces synchronized encryption algorithm for MSAL file system cache accessed from VSCode-MSSQL extension (MSAL.js) and STS (MSAL.NET) 

Companion PR: https://github.com/microsoft/sqltoolsservice/pull/1945

Also added prompt when credential could not be saved on Windows:

Text updated to:
_Keys for token cache could not be saved in credential store, this may cause Azure access token persistence issues and connection instabilities. Its likely that SqlTools has reached credential storage limit on Windows, please clear atleast 2 credentials created by Microsoft.SqlTools in Windows Credential Manager and reload._


kburtram (approved):

alanrenmsft (approved):

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -410,6 +410,9 @@
         <trans-unit id="msgRefreshTokenError">
             <source xml:lang="en">Error when refreshing token</source>
         </trans-unit>
+        <trans-unit id="msgAzureCredStoreSaveFailedError">
+            <source xml:lang="en">Keys for token cache could not be saved in credential store, this may cause Azure access token persistence issues and connection instabilities. Its likely that VS Code has reached credential storage limit on Windows, please clear atleast 2 credentials created by VS Code in Windows Credential Manager and reload.</source>
```
Charles-Gagnon, line 414:
```suggestion
            <source xml:lang="en">Keys for token cache could not be saved in credential store, this may cause Azure access token persistence issues and connection instabilities. It's likely that VS Code has reached credential storage limit on Windows, please clear at least 2 credentials created by VS Code in Windows Credential Manager and reload.</source>
```

Charles-Gagnon, line 414: Not sure we want to be suggesting to clear "credentials created by VS Code"...VS Code stores a lot of different things in cred man and clearing them without knowing what they are could break other functionality (for example long credentials are split among multiple entries, so deleting some of them could corrupt the entry)

If we're going to keep this suggestion we should at least try to make sure the user understands the repercussions of doing this - but I'm not sure how to do that in a succinct enough manner for this notification. What about writing something in the ADS/VS Code MSSQL Wiki that this can link to where we can more fully explain the situation?

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -410,6 +410,9 @@
         <trans-unit id="msgRefreshTokenError">
             <source xml:lang="en">Error when refreshing token</source>
         </trans-unit>
+        <trans-unit id="msgAzureCredStoreSaveFailedError">
+            <source xml:lang="en">Keys for token cache could not be saved in credential store, this may cause Azure access token persistence issues and connection instabilities. It's likely that SqlTools has reached credential storage limit on Windows, please clear at least 2 credentials created by Microsoft.SqlTools in Windows Credential Manager and reload.</source>
```
Charles-Gagnon, line 414: Little clearer, there isn't actually a "created by" field or anything which might confuse people.

```suggestion
            <source xml:lang="en">Keys for token cache could not be saved in credential store, this may cause Azure access token persistence issues and connection instabilities. It's likely that SqlTools has reached credential storage limit on Windows, please clear at least 2 credentials that start with "Microsoft.SqlTools|" in Windows Credential Manager and reload.</source>
```

Charles-Gagnon (approved):

PR #17630, opened 2023-03-17
Update metadataService.ts

sachamazouz:
Use destructuring for cleaner code: Instead of using metadataParams.ownerUri, we can use destructuring to extract ownerUri directly from metadataParams That's a cleaner way to do :smile:


kburtram (approved):
Looks like a non-functional change to syntax.  The alternate syntax appears a little cleaner in my opinion.

PR #17631, opened 2023-03-17
Allow using STS credential store on Linux for encryption keys

cheenamalhotra:
Noticed an issue on Linux where Encryption keys are not available in .NET backend when `enableSqlAuthenticationProvider` is enabled. Since we use VS Code secret storage for Linux, STS backend is not able to read these credentials.

Since we need these keys for MSAL cache, the only way it will work is if we store them via STS credential service.
Opening this PR for any feedback before I also open the same in ADS..


cheenamalhotra:
Closing this as discussed a few ideas offline

PR #17642, opened 2023-03-24
Show message indicating that ADAL is being deprecated

cssuh:
Similar to https://github.com/microsoft/azuredatastudio/pull/22432, 
This will display a message to the user upon selection of the ADAL Auth Library indicating that ADAL is being deprecated:

Screenshot 2023-03-23 at 5 26 54 PM

Also updated one place that still had EULA as the license when it should be MIT.


cheenamalhotra (approved):
Please add deprecationMessage here as well.

PR #17658, opened 2023-04-21
update sql projects/dacfx changes in vscode-mssql.d.ts

kisantia:
I did a diff of the vscode-mssql.d.ts files in ADS and vscode-mssql and found several differences. This brings over the sql projects/dacfx changes that were made in ADS, but weren't brought over to the vscode-mssql repo. After this is merged, I'll make a PR in ADS so that the two files will be the same.

### DIFF
--- a/src/services/dacFxService.ts
+++ b/src/services/dacFxService.ts
@@ -87,7 +87,7 @@ export class DacFxService implements mssql.IDacFxService {
 		upgradeExisting: boolean,
 		ownerUri: string,
 		taskExecutionMode: mssql.TaskExecutionMode,
-		sqlCommandVariableValues?: Record<string, string>,
+		sqlCommandVariableValues?: Map<string, string>,
 		deploymentOptions?: mssql.DeploymentOptions): Thenable<mssql.DacFxResult> {
 		const params: mssql.DeployParams = {
 			packageFilePath: packageFilePath,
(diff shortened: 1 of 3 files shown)


src/services/dacFxService.ts
```diff
@@ -145,4 +145,8 @@ export class DacFxService implements mssql.IDacFxService {
 		};
 		return this._client.sendRequest(dacFxContracts.ValidateStreamingJobRequest.type, params);
 	}
+
+	savePublishProfile(profilePath: string, databaseName: string, connectionString: string, sqlCommandVariableValues?: Map<string, string>, deploymentOptions?: mssql.DeploymentOptions): Thenable<mssql.ResultStatus> {
+		throw new Error('Method not implemented.');
```
kisantia, line 150: @SakshiS-harma I left this as not implemented since it will be part of adding publish profile support in vscode

SakshiS-harma: Yes, I will add it later with the other publish profile changes.

Charles-Gagnon (approved):

Charles-Gagnon:
If you want to remove that max line length check I'd be ok with that - it doesn't really matter IMO and is just annoying to deal with

https://github.com/Microsoft/vscode-mssql/blob/main/tslint.json#L29

typings/vscode-mssql.d.ts
```diff
@@ -1078,6 +1079,14 @@ declare module 'vscode-mssql' {
 		defaultDeploymentOptions: DeploymentOptions;
 	}
 
+	export interface SavePublishProfileParams {
+		profilePath: string;
+		databaseName: string;
+		connectionString: string;
+		sqlCommandVariableValues?: Record<string, string>;
```
Benjin, line 1086: @SakshiS-harma should this be a `Map` after the conversion PR from earlier this week?

Benjin (approved):

typings/vscode-mssql.d.ts
kisantia: I just updated it in the latest commit. Lmk if it shouldn't be swapped over to Map @SakshiS-harma

PR #17663, opened 2023-04-26
Remove LGTM badge

Charles-Gagnon:
We removed this a while ago and so the badge is not active - removing it to clean up our extension page in VS Code. 

image


alanrenmsft (approved):

ankit-gautam23 (commented):
LGTM

PR #17669, opened 2023-05-01
Remove version compatibility check

### LINKED ISSUE #17668: mssql: Client is not compatible with the service layer
I tried installing the latest vscode-mssql extension from the product build pipeline and saw this error when the extension is activated:

image

cheenamalhotra:
Fixes #17668


src/languageservice/serviceclient.ts
```diff
@@ -12,7 +12,6 @@ import {
 import * as path from 'path';
 import VscodeWrapper from '../controllers/vscodeWrapper';
 import * as Utils from '../models/utils';
-import { VersionRequest } from '../models/contracts';
```
Charles-Gagnon, line 15: Can you remove this type too?

kisantia (approved):

Charles-Gagnon (approved):

PR #17676, opened 2023-05-06
Validate server name before sending client request

### LINKED ISSUE #17674: Choosing connection profile for DB projects displays connection error
- MSSQL Extension Version: main

Steps to Reproduce:

1. Install latest SQL DB projects extension from the ADS repo (not released yet)
2. Go to the Database Projects view and select "Create Project from Database" in the more actions menu

This error notification shows up:

`mssql: Failed to connect: null`

In the console I see these messages:

```
console.ts:137 [Extension Host] rejected promise not handled within 1 second: ServerName cannot be null or empty
I @ console.ts:137
$logExtensionHostMessage @ mainThreadConsole.ts:39
N @ rpcProtocol.ts:455
M @ rpcProtocol.ts:440
H @ rpcProtocol.ts:370
G @ rpcProtocol.ts:296
(anonymous) @ rpcProtocol.ts:161
invoke @ event.ts:862
deliver @ event.ts:1075
fire @ event.ts:1031
fire @ ipc.net.ts:671
q.onmessage @ localProcessExtensionHost.ts:583
log.ts:441   ERR ServerName cannot be null or empty
```

cheenamalhotra:
Fixes #17674 

FixError

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -788,6 +788,9 @@ export default class ConnectionManager {
 			title: LocalizedConstants.connectProgressNoticationTitle,
 			cancellable: false
 		}, async (_progress, _token) => {
+			if (!connectionCreds.server) {
+				throw new Error(LocalizedConstants.serverNameMissing);
+			}
 			// Check if the azure account token is present before sending connect request (only with SQL Auth Provider is not enabled.)
 			if (connectionCreds.authenticationType === Constants.azureMfa) {
 				if (AzureController.isTokenInValid(connectionCreds.azureAccountToken, connectionCreds.expiresOn)) {
--- a/localization/xliff/enu/constants/localizedConstants.enu.xlf
+++ b/localization/xliff/enu/constants/localizedConstants.enu.xlf
@@ -41,6 +41,9 @@
         <trans-unit id="msgConnectionError2">
             <source xml:lang="en">Failed to connect: {0}</source>
         </trans-unit>
+        <trans-unit id="serverNameMissing">
+            <source xml:lang="en">Server name not received.</source>
+        </trans-unit>
         <trans-unit id="msgConnectionErrorPasswordExpired">
             <source xml:lang="en">Error {0}: {1} Please login as a different user and change the password using ALTER LOGIN.</source>
         </trans-unit>


src/controllers/connectionManager.ts
```diff
@@ -788,6 +788,9 @@ export default class ConnectionManager {
 			title: LocalizedConstants.connectProgressNoticationTitle,
 			cancellable: false
 		}, async (_progress, _token) => {
+			if (!connectionCreds.server) {
```
Charles-Gagnon, line 791: Why is connect even being called? The user is still choosing a connection at that point, so it shouldn't even be trying to connect to anything.

Benjin: Looking at the code (but not slapping a debugger on it), I'm guessing the callstack is:

```
DatabaseProject:
createProjectFromDatabaseQuickpick.ts
  createNewProjectFromDatabaseWithQuickpick() ->
  vscodeMssqlApi.promptForConnection() ->

Vscode-Mssql:
connectionUI.ts
  promptForConnection() ->
  handleSelectedConnection() ->
  createAndSaveProfile() ->
  validateAndSaveProfile()
```

Which would attempt to connect to connect to anything in the dropdown that gets highlighted, I believe.

Benjin (approved):

src/controllers/connectionManager.ts
cheenamalhotra: Not exactly, as first 'connect' is always called (this is where I'm validating server now) and 'err' is expected that triggers promptForConnection:

https://github.com/microsoft/azuredatastudio/blob/0d8e1882fa7f974b27bc4b68fe6203e4a627aae3/extensions/sql-database-projects/src/dialogs/createProjectFromDatabaseQuickpick.ts#L35C1-L46

```ts
// Get the list of databases now to validate that the connection is valid and re-prompt them if it isn't
try {
	connectionUri = await vscodeMssqlApi.connect(connectionProfile);
	dbs = (await vscodeMssqlApi.listDatabases(connectionUri))
		.filter(db => !constants.systemDbs.includes(db)); // Filter out system dbs
} catch (err) {
	// The mssql extension handles showing the error to the user. Prompt the user
	// for a new connection and then go and try getting the DBs again
	connectionProfile = await vscodeMssqlApi.promptForConnection(true);
	if (!connectionProfile) {
		// User cancelled
		return undefined;
	}

}
```

cheenamalhotra: Fixing in vscode-mssql is eitherway a good thing I guess, to ensure servername is not empty before reaching STS.

SakshiS-harma (approved):

Issue #1014, opened 2017-10-31
Nested SQL comments don't function as expected


weaslevdmd:
- MSSQL Extension Version: 1.2.0
- VSCode Version: 1.17.2
- OS Version: MacOS Sierra 10.12.6

Steps to Reproduce:

1. Create any SQL file with nested comments
/* -- Comment entire block
--/* -- Comment block 1
Code block 1
/* -- Comment block 1A
Code block 1A
--*/ -- End block 1A comment
--/* -- Comment block 1B
Code block 1B
--*/ -- End block 1B comment
--*/ -- End block 1 comment
--*/ -- End comment entire block

2. The above code should comment the entire comment block, however, in VSCode mssql only the code block 1A actually becomes commented. Looking at the same file in other IDEs (such as Visual Studio) shows the entire block commented out, but VSCode does not.

When attempting to run the code when some of the comment structure is incorrectly formatted will result in a parsing error (and failure to execute the SQL). Typically this error only occurs when the mismatched comments are all within a larger comment block that should not even be part of the query - but due to comment formatting issues does generate an error (see attached).

![screen shot 2017-10-31 at 09 13 54](https://user-images.githubusercontent.com/3435038/32235297-4e7416ae-be1c-11e7-9add-a66700b8ac0d.png)

NOTE: To generate the error, simply start putting random /* comment start characters inside other comment blocks.

codykonior:
This is still an issue in the current version and affects me also, because code flags with issues in VS Code that runs fine in SSMS.

ObiRonzo:
Been almost 2 years, still a problem, still not assigned

kburtram:
Could you please clarify the issue a bit?  When I put the example snippet into VS Code it appears to display the entire block as a comment (similar to SSMS).  The execution fails in both VS Code and SSMS since it appears that `/*` and `*/` are expected to be matched-up, at least against the server I tested with.

Is the issue that we show the error as a notification (which is a bug) instead of in the Messages pane?  Or are there cases where we don't execute queries in VS Code that work in SSMS?

![image](https://user-images.githubusercontent.com/599935/63617440-4da9f780-c59e-11e9-8f4e-f16302e28f05.png)

ObiRonzo:
Problem I encountered (and interpreted this to cover) is illustrated by these two illustrations of the same code, one in Visual Stuido Code and one in SQL Management Studio. Executing them results in same thing (nothing). It is just the color coding I am reporting.

/*
[...]

PR #17708, opened 2023-05-26
Fix nested comment colorization

Charles-Gagnon:
Fixes https://github.com/microsoft/vscode-mssql/issues/1014 and https://github.com/microsoft/azuredatastudio/issues/19205 (will close ADS one when I port it over there)

Fix here is to take what was suggested in https://github.com/microsoft/vscode-mssql/issues/1014#issuecomment-1172876040 -split out comment blocks into their own section and add themselves as possible children. Thank you @RedCmd for the fix suggestion!

I renamed the section to just comment-block, I don't see why the .c was added in the first place and it seems to work just fine without it. 

Before:
image

After:
image

### DIFF
--- a/syntaxes/SQL.plist
+++ b/syntaxes/SQL.plist
@@ -675,20 +675,32 @@
 					</array>
 				</dict>
 				<dict>
-					<key>begin</key>
-					<string>/\*</string>
-					<key>captures</key>
-					<dict>
-						<key>0</key>
-						<dict>
-							<key>name</key>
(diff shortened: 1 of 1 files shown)


Charles-Gagnon:
Here's the example query they mentioned in the original comment too as potentially being a problem if it was valid : 

image

But as you can see it's not valid since the block comment isn't correctly closed. So that seems to be behaving correctly to me.

kburtram (approved):

PR #17749, opened 2023-07-18
Remove all ADAL code

cssuh:
Removed all references to ADAL code, as MSAL will be the only option going forward.


lib/ads-adal-library/package.json
cheenamalhotra, line 1: Let's make sure we mark this package as deprecated in npm repo after merge (post July release).

cheenamalhotra (approved):

cheenamalhotra (commented):
You'll have to clean up all build scripts as well..

PR #17756, opened 2023-07-27
Update iKey

Charles-Gagnon:
This was never updated to the newer iKey, and the old one is no longer valid so no events were making it through.


Charles-Gagnon:
@aasimkhan30 Note that you'll need to do another release of the extension if you want to start getting telemetry now.

PR #17758, opened 2023-07-27
Add no-invalid-this linting

Charles-Gagnon:
Follow up to https://github.com/microsoft/vscode-mssql/pull/17757 - this will fail linting if an invalid call to this is made like was broken there. This will prevent such issues from happening again in the future.

@cssuh @cheenamalhotra I definitely recommend looking at tslint/eslint rules every time you come across a bug like this. There's a lot of really useful rules out there and while enabling every single one isn't super feasible, it's still good to at least check if one exists so we can verify other instances of the issue don't exist. (luckily in this case there weren't any such other issues outside of test code, although the fact that they weren't failing before might indicate some issues with what the tests are doing themselves, I didn't investigate that far since the tests are passing as is)

https://palantir.github.io/tslint/rules/no-invalid-this/

### DIFF
--- a/test/initialization.test.ts
+++ b/test/initialization.test.ts
@@ -19,7 +19,7 @@ function ensureExtensionIsActive(): Promise<any> {
 function waitForExtensionToBeActive(resolve): void {
 	if (typeof (vscode.extensions.getExtension('ms-mssql.mssql')) === 'undefined' ||
 		!vscode.extensions.getExtension('ms-mssql.mssql').isActive) {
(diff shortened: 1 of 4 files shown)


cheenamalhotra (approved):

Charles-Gagnon:
Welp, not sure what's going on but the tests are failing with my original changes in the pipeline but passing locally. Not really concerned about test code so just disabling the rule checking for them for now...it's more important we catch stuff going forward.

PR #17759, opened 2023-07-28
Enable more lint rules

cheenamalhotra:
Enables lint rules:
- function-constructor (replaces deprecated rule: no-function-constructor-with-string-args)
- file-name-casing
- no-misused-new
- static-this

I'll continue enabling more rules as needed, gradually..


Charles-Gagnon (approved):
🥳

PR #17787, opened 2023-09-08
Fix encryption mode in status bar

### LINKED ISSUE #17671: Status bar doesn't show correct 'Encryption' status for connection
The status bar shows active connection, but doesn't show correct value for encryption:

image

cheenamalhotra:
Fixes #17671 

Updated encrypt setting enum values and description to include 'Strict' encryption mode
image

Update status bar:

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -32,7 +32,7 @@
 			"outFiles": [
 				"${workspaceRoot}/out/src/**/*.js"
 			],
-			// "preLaunchTask": "build",
+			"preLaunchTask": "build",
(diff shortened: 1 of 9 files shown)


erinstellato-ms (approved):
Strict is also supported with Azure SQL DB, correct?  Do we need to call that out, or only SQL 2022 because the functionality is version-specific?

cheenamalhotra:
Yes, it supported, but we mention it to call out for on-premise servers since it's a new setting. As it becomes the new default for Azure, we can slowly fade the version info away.

PR #17793, opened 2023-09-11
Fix case comparisons for encryption options

### LINKED ISSUE #17792: Encryption mode case comparisons need fix
From the PR #17787, encryption mode case comparison needs adjustment to make it work properly:

https://github.com/microsoft/vscode-mssql/pull/17787/files#diff-3b4cca091c1d28b53446b9dc1efb5b12dda49604d103ea96f06b32bd0eef89e8R213

cheenamalhotra:
Fixes #17792


Charles-Gagnon (approved):
Some tests wouldn't be a bad idea (and something to keep in mind going forward to prevent issues like this in the first place)

kisantia (approved):

PR #17804, opened 2023-09-15
Support firewall rule name + fix firewall rule issues

### LINKED ISSUE #17607: Create Firewall Rule notification issues
A couple issues with the current "Create Firewall Rule" notification.

image

1. `Account signed in. Create new firewall rule?` doesn't really indicate what the problem is, and doesn't really seem like an accurate message to display to the user. Can we just add it to the error notification that also pops up to also remove the multiple notifications?
2. Because it's an info notification it disappears after ~15sec. I think this would be better to have either as an error as suggested above - or maybe even better as a follow-up quick input prompt (since the user was already there for profile creation anyways, and the create firewall rule uses those to get info from the user)

### LINKED ISSUE #17802: Azure Firewall rule creation doesn't work as expected
- MSSQL Extension Version: main
- VSCode Version: Latest
- OS Version: Windows 11

For below use cases, we have firewall creation failing:
- When creating rule for account with multiple tenants
- When expanding connection node that needs firewall rule, firewall rule creation flow doesn't trigger.

These issues have existed for a while now, so not recent regressions.
cc @cssuh

### LINKED ISSUE #17803: Support setting firewall rule name
When setting firewall rule for an Azure server, it is possible to set Firewall Rule name - must be supported as in ADS.

cheenamalhotra:
Fixes #17803 
Fixes #17802 
Fixes #17607 

Updated message:

image

### DIFF
--- a/localization/xliff/enu/constants/localizedConstants.enu.xlf
+++ b/localization/xliff/enu/constants/localizedConstants.enu.xlf
@@ -351,7 +351,7 @@
             <source xml:lang="en">Your client IP address does not have access to the server. Add an Azure account and create a new firewall rule to enable access.</source>
         </trans-unit>
         <trans-unit id="msgPromptRetryFirewallRuleSignedIn">
-            <source xml:lang="en">Your client IP Address '{0}' does not have access to the server '{1}' you're attempting to connect. Would you like to create new firewall rule?</source>
+            <source xml:lang="en">Your client IP Address '{0}' does not have access to the server '{1}' you're attempting to connect to. Would you like to create new firewall rule?</source>
         </trans-unit>
         <trans-unit id="msgPromptRetryFirewallRuleAdded">
             <source xml:lang="en">Firewall rule successfully added. Retry profile creation? </source>
--- a/localization/xliff/enu/constants/localizedConstants.enu.xlf
+++ b/localization/xliff/enu/constants/localizedConstants.enu.xlf
@@ -93,10 +93,13 @@
             <source xml:lang="en">Database name</source>
         </trans-unit>
         <trans-unit id="startIpAddressPrompt">
-            <source xml:lang="en">Start IP</source>
+            <source xml:lang="en">Start IP Address</source>
         </trans-unit>
         <trans-unit id="endIpAddressPrompt">
-            <source xml:lang="en">End IP</source>
+            <source xml:lang="en">End IP Address</source>
+        </trans-unit>
+        <trans-unit id="firewallRuleNamePrompt">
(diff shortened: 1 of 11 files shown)


localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -348,7 +351,7 @@
             <source xml:lang="en">Your client IP address does not have access to the server. Add an Azure account and create a new firewall rule to enable access.</source>
         </trans-unit>
         <trans-unit id="msgPromptRetryFirewallRuleSignedIn">
-            <source xml:lang="en">Account signed In. Create new firewall rule? </source>
+            <source xml:lang="en">Your client IP Address '{0}' does not have access to the server '{1}' you're attempting to connect. Would you like to create new firewall rule?</source>
```
Charles-Gagnon, line 354:
```suggestion
            <source xml:lang="en">Your client IP Address '{0}' does not have access to the server '{1}' you're attempting to connect to. Would you like to create new firewall rule?</source>
```

src/constants/constants.ts
```diff
@@ -114,6 +114,7 @@ export const azureAccountExtensionId = 'ms-vscode.azure-account';
 export const databaseString = 'Database';
 export const localizedTexts = 'localizedTexts';
 export const ipAddressRegex = /\b(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\b/;
+export const ruleNameRegex = /^[a-zA-Z0-9][a-zA-Z0-9_.-]{0,78}[a-zA-Z0-9_]?$/;
```
Charles-Gagnon, line 117: Add link to where you got this from if you have such a thing - or at least a brief description of what this should be doing is typically helpful for regexes. 

(tests wouldn't hurt either)

src/views/connectionUI.ts
```diff
@@ -597,31 +626,73 @@ export class ConnectionUI {
...
+				shouldPrompt: () => tenantChoices.length > 1,
+				onAnswered: (value: ITenant) => {
+					tenantIdAnswer = value.id;
+				}
 			}
 		];
 
 		// Prompt and return the value if the user confirmed
-		return this._prompter.prompt(questions).then((answers: { [questionId: string]: string }) => {
+		return this._prompter.prompt(questions).then(async (answers: { [questionId: string]: string }) => {
```
Charles-Gagnon, line 670: Converting to async/await pattern would be great

Charles-Gagnon (approved):

PR #17814, opened 2023-09-19
Added fix to path for localization (using last index replace)

smartguest:
Now replaces the last instances of the folder in the path (to prevent edge cases where user folders may share the same name/string.)

### DIFF
--- a/tasks/localizationtasks.js
+++ b/tasks/localizationtasks.js
@@ -208,7 +208,18 @@ function dictionaryMapping(file, packageAllKeys = undefined) {
 
 	// Make the new file create on root.
 	// the original directory path was 'vscode-mssql\localization\xliff\<lang>'
-	file.dirname = file.dirname.replace(language, '').replace('localization', '').replace('xliff', '');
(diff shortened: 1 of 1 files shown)


cssuh (approved):
vscode-mssql builds for me now, thanks Alex!

tasks/localizationtasks.js
```diff
@@ -208,7 +208,18 @@ function dictionaryMapping(file, packageAllKeys = undefined) {
 
 	// Make the new file create on root.
 	// the original directory path was 'vscode-mssql\localization\xliff\<lang>'
-	file.dirname = file.dirname.replace(language, '').replace('localization', '').replace('xliff', '');
+	file.dirname = replaceLastString(file.dirname, `\\${language}`);
```
Charles-Gagnon, line 211: This is not platform-aware, are you sure that the path separators will always be `\`?

smartguest: The separators are unnecessary as I found out

Charles-Gagnon (approved):

PR #17845, opened 2023-11-07
Update to Node 18 and VS Code 1.83

cheenamalhotra:
Upgrades VSCode version to 1.83.1, to bring Node version in sync with VS Code and ADS.

Currently building with Node 18 leads to below issue, which is now addressed in the PR as well:
https://github.com/ivogabe/gulp-typescript/issues/683

By specifying resolution as recommended:

```
"resolutions": {
    "gulp-typescript/source-map": "0.7.4"
  },
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "homepage": "https://github.com/Microsoft/vscode-mssql/blob/master/README.md",
   "engines": {
-    "vscode": "^1.78.0"
+    "vscode": "^1.83.1"
   },
   "categories": [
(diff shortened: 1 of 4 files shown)


Charles-Gagnon (approved):
Did you see actual issues with this or just general cleanup?

cheenamalhotra:
Main issue I liked above is what we see when we compile with Node 18. And since we're now using Node 18 with ADS, we have to keep on switching to Node 16 for working with this extension, which is why I thought of upgrading it.

PR #17852, opened 2023-12-12
add two keywors to SQL.plist of VsCode.

jingzhao65:
Added teo key words, connection, shortest_path, to SQL.plist


Charles-Gagnon (approved):
Do you know why the graph functions (MATCH, SHORTEST_PATH, etc.) are being marked as keywords? I would have expected them to be colored as functions instead (pink in SSMS)

SSMS:
image

I also noticed that other graph functions such as MATCH aren't colorized in ADS/VS Code MSSQL - if you're adding these ones it may be worth spending a few extra minutes to go through the rest of the graph functions and make sure the rest are being done correctly too.

ADS:
image

If you'd rather just stick with these two as they are for now then that's ok, just let me know.

Charles-Gagnon:
Tests are failing for an unrelated reason, merging in since the tests don't validate the grammar anyways.

PR #17857, opened 2024-01-04
Upgrade test-electron and node version

cheenamalhotra:


build/build.yml
```diff
@@ -8,9 +8,9 @@ pool:
 
 steps:
   - task: NodeTool@0
-    displayName: "Use Node 14.x"
+    displayName: "Use Node 16.x"
```
kisantia, line 11: the @types/node version in the package.json is 18.11.9. Should the pipeline be updated to node 18 instead of 16?

cheenamalhotra: Sure, let's test :)

kisantia (approved):

PR #17890, opened 2024-04-04
Update to sqltoolsservice 5.0.0.4 and remove specific Linux distros

maxvolk:
Currently, the most recent Ubuntu version is 16. This PR adds those LTS versions up to the most recently released one (22).

### DIFF
--- a/src/configurations/config.json
+++ b/src/configurations/config.json
@@ -16,6 +16,9 @@
       "RHEL_7": "rhel-x64-net7.0.tar.gz",
       "Ubuntu_14": "rhel-x64-net7.0.tar.gz",
       "Ubuntu_16": "rhel-x64-net7.0.tar.gz",
+      "Ubuntu_18": "rhel-x64-net7.0.tar.gz",
+      "Ubuntu_20": "rhel-x64-net7.0.tar.gz",
+      "Ubuntu_22": "rhel-x64-net7.0.tar.gz",
       "Linux_ARM64": "linux-arm64-net7.0.tar.gz"
     },
     "installDir": "../sqltoolsservice/{#version#}/{#platform#}",
--- a/src/configurations/config.json
+++ b/src/configurations/config.json
@@ -1,53 +1,21 @@
 {
   "service": {
     "downloadUrl": "https://github.com/Microsoft/sqltoolsservice/releases/download/{#version#}/microsoft.sqltools.servicelayer-{#fileName#}",
-    "version": "4.11.0.2",
+    "version": "5.0.0.4",
     "downloadFileNames": {
-      "Windows_86": "win-x86-net7.0.zip",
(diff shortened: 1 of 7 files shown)


maxvolk:

Charles-Gagnon:
@kisantia @kburtram FYI - what would probably be even better is to remove the specific distros and instead move to the generic "Linux" target that Alan added support for in the downloader a while back. That will give us forward compatibility too.

e.g. what we do in MSSQL extension : https://github.com/microsoft/azuredatastudio/blob/main/extensions/mssql/config.json

maxvolk:
@Charles-Gagnon I removed the specific distros. Also, I moved to the current release of sqltoolsservice.

src/configurations/config.json
```diff
@@ -1,50 +1,21 @@
...
+      "Linux_ARM64": "linux-arm64-net8.0.tar.gz"
     },
     "installDir": "../sqltoolsservice/{#version#}/{#platform#}",
     "executableFiles": [
       "MicrosoftSqlToolsServiceLayer.exe",
       "MicrosoftSqlToolsServiceLayer",
       "MicrosoftSqlToolsServiceLayer.dll"
     ]
-  },
-  "v1Service": {
```
Charles-Gagnon, line 28: This is still used for downlevel Mac versions so we should keep it as-is for now

https://github.com/Microsoft/vscode-mssql/blob/79da38b0e928ade250a846a204977519e4f120cf/src/languageservice/serviceclient.ts#L233

Charles-Gagnon:
Thanks for going through and updating all this @maxvolk! To confirm - is this a blocking issue for you on Ubuntu 22 right now?

maxvolk:
> Thanks for going through and updating all this @maxvolk! To confirm - is this a blocking issue for you on Ubuntu 22 right now?

This issue is not blocking me on Ubuntu right now, it stems more from curiosity.

src/configurations/config.json
maxvolk: What would be the rule or trade-off when one could remove such support for these old Mac OS versions? Or is in general the goal to offer maximum support to older versions?

Charles-Gagnon: We've never really set a goal or deprecation timeline, but that's definitely something we should work on! For now though continuing to support all previous versions is something we strive to adhere to.

Charles-Gagnon:
Gotcha, thanks! Since it's not blocking we might take a bit to get to this due to some other higher priority stuff, but this is definitely work we'll want to take so we'll get to it once we have some time! If this becomes a higher priority issue for any reason let us know and I can see what we can do.

Benjin:
I moved this linux distro consolidation work into #18179 so we don't lose it, but can also close this PR since it's been sitting around for a while

PR #17917, opened 2024-07-02
Sets up E2E tests with playwright and adds tests

lewis-sanchez:
This PR sets up E2E tests with playwright. The PR adds tests to verify that the extension appears in the activity bar, connects to a local SQL DB, and disconnects from that SQL DB.


test/e2e/connection.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+		let isConnectionButtonVisible = await addConnectionButton.isVisible();
+		if (!isConnectionButtonVisible) {
+			await vsCodePage.click('a[aria-label="SQL Server (Ctrl+Alt+D)"]');
+		}
+
+		await expect(addConnectionButton).toBeVisible({ timeout: 10000 });
+		await addConnectionButton.click();
+
+		// await vsCodePage.click('input[aria-label="input"]');
+		await vsCodePage.fill('input[aria-label="input"]', '(localdb)\\MSSqlLocalDb');
```
aasimkhan30, line 31: Ideally, we would like to take the connection details from command line and fill them up here instead of hard coding it in the test. I am guessing when we add this to pipelines, it will take the server url and password for the local docker db that we spin up for that run. It will be nice to make a class that does this so that we can reuse it across all tests.

test/e2e/connection.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+		const isDiconnectOptionVisible = await disconnectOption.isVisible()
+		if (isDiconnectOptionVisible) {
+			await disconnectOption.click();
+		}
+
+		await addedSqlConnection.click({ button: 'right' });
+		await expect(disconnectOption).toBeHidden({ timeout: 10000 });
+	});
+
+	test.afterAll(async () => {
```
aasimkhan30, line 57: We should also take screenshot on errors.
Something like this should suffice.
https://github.com/microsoft/playwright/issues/14854#issuecomment-1155667859

test/e2e/utils/utils.ts
```diff
@@ -0,0 +1,35 @@
+/*---------------------------------------------------------------------------------------------
```
aasimkhan30, line 1: Maybe name this file as launchVscodeWithMsqqlExt.ts? as it is already in utils folder.

aasimkhan30:
@lewis-sanchez please don't forget to add any instructions in the wiki page for running smoke tests.

lewis-sanchez:
Need to address merge conflicts

test/e2e/.env.example
```diff
@@ -0,0 +1,22 @@
...
+# The name of the database server to connect to.
+# This value is the name of the server that will be used by the tests.
+SERVER_NAME=(localdb)\MSSqlLocalDb
+
+# The name of the database to connect to.
+# This value is optional and can be left empty.
+DATABASE_NAME=
+
+# The authentication type to use to connect to the server.
+# Options are: "SQL Login", "Integrated", "Microsoft Entra Id - Universal w/ MFA Support"
```
aasimkhan30, line 17: Can you please add password variable too?

aasimkhan30 (approved):

kburtram (approved):

PR #17918, opened 2024-07-03
Create devcontainer.json

Daverosales93:


kburtram:
Would it make more sense to add `devcontainer.json` to .gitignore so that everyone can define their own devcontainer configuration as desired?

kburtram:
I think that the user's devcontainer settings seems like something that would be configured on a per-user basis, as far as I understand.  So probably we should add this file to .gitignore to avoid it showing up in the changed file list if someone modifies it for their own environment.  I'll close this out on that basis, please let me know if I missed something with this.

PR #17919, opened 2024-07-04
Integrating React App into extension src files and switching from 'Vite+Rollup' to Esbuild

aasimkhan30:
Esbuild is a Go-based JavaScript bundler that offers significant performance improvements over other bundlers. It is even faster than tsc, which only compiles TypeScript without bundling. You can learn more about Esbuild and its speed advantages [here](https://esbuild.github.io/faq/#why-is-esbuild-fast). 
**TLDR:** its speed comes from being a native, compiled Go application, as opposed to other bundlers that are written in JavaScript and run in a JavaScript virtual machine. It also skips type-checking which is being run in parallel for now.

I have also integrated the React files into the extension source, so it now shares the same `package.json` with the extension source code and resides in `src\reactviews` folder.

The React files have their own TypeScript configuration file, `tsconfig.react.json`, and their own build step `ext:compile-reactviews` in the Gulp file . This step uses esbuild to bundle the webviews into .js and .css files.

I've integrated esbuild into our extension src as well, and it's been performing well without major issues so far. Before fully replacing our current extension watch, which uses tsc, I'll conduct more testing. Currently, esbuild is set up as an unused task `ext:bundle-src` in the gulp file. All functionalities, including localization and debugging, are working properly, but additional testing is still needed.

These improvements should improve overall dev productivity as the time to compile when watch  has reduced by almost 95%

I got the initial idea to use Esbuild when I was reading this guide from vscode team for bundling extensions: [Link](https://code.visualstudio.com/api/working-with-extensions/bundling-extension#using-esbuild). Which means there is official guidance for the using esbuild in vscode extension.

#### For React files:
- Using esbuild: 587ms
image
- Using Vite with Rollup.js: 8+ seconds
image

#### For the extension (currently disabled):
- Using esbuild: 650ms
image
- Existing tsc task: 6.5 seconds
image

As you can see the improvements are good enough to eventually switch every compilation step to esbuild.


kburtram (approved):

kburtram:
Thanks for making these updates!

PR #17925, opened 2024-07-10
Fix changing timestamp issue when switching query result tabs

### LINKED ISSUE #17916: Timestamp changes when switching tabs
Switch to other tabs and then return to the current tab, the timestamp will change.

lewis-sanchez:
This PR fixes #17916

The PR stops the timestamp and total elapsed time measurements from changing when the query results tab is changed to another.

### DIFF
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -307,7 +307,6 @@ export default class QueryRunner {
 			const batchSet = this.batchSets[batchId];
 			this.eventEmitter.emit('batchStart', batchSet);
 			let executionTime = <number>(Utils.parseTimeString(batchSet.executionElapsed) || 0);
-			this._totalElapsedMilliseconds += executionTime;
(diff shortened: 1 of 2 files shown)


cssuh (approved):

src/controllers/queryRunner.ts
```diff
@@ -307,7 +307,6 @@ export default class QueryRunner {
 			const batchSet = this.batchSets[batchId];
 			this.eventEmitter.emit('batchStart', batchSet);
 			let executionTime = <number>(Utils.parseTimeString(batchSet.executionElapsed) || 0);
-			this._totalElapsedMilliseconds += executionTime;
```
lewis-sanchez, line 310: When this line is present, execution time continues to change on tab switch. Here's a gif of this happening:

Total Execution Time - Change



When this line is not present, execution time doesn't change as seen in this other gif:

Total Execution Time - No change

PR #17931, opened 2024-07-17
Fix e2e tests for Linux

lewis-sanchez:
This PR adjusts E2E tests for them to run successfully on Linux.

### DIFF
--- a/test/e2e/connection.spec.ts
+++ b/test/e2e/connection.spec.ts
@@ -30,10 +30,7 @@ test.describe('MSSQL Extension - Database Connection', async () => {
 		await addDatabaseConnection(vsCodePage, serverName, databaseName, authType, userName, password, savePassword, profileName);
 
 		await openNewQueryEditor(vsCodePage, profileName, password);
-		await new Promise(r => setTimeout(r, 2 * 1000));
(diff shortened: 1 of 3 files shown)


test/e2e/connection.spec.ts
```diff
@@ -29,26 +29,14 @@ test.describe('MSSQL Extension - Database Connection', async () => {
...
-		await disconnectOption.click();
-		const isDiconnectOptionVisible = await disconnectOption.isVisible()
-		if (isDiconnectOptionVisible) {
-			await disconnectOption.click();
-		}
-
-		await addedSqlConnection.click({ button: 'right' });
-		await expect(disconnectOption).toBeHidden({ timeout: 10000 });
+		await openNewQueryEditor(vsCodePage, profileName, password);
+		await new Promise(r => setTimeout(r, 2 * 1000));
```
Charles-Gagnon, line 33: Hardcoded timeouts like this aren't great since they tend to be kind of flakey - could you not wait for a specific element to be visible/hidden with a reasonable timeout instead? (e.g. what you did for the disconnected status below)

aasimkhan30 (approved):

PR #17943, opened 2024-07-24
Start gulp watch before launching the extension for debugging

lewis-sanchez:
This PR starts gulp watch before the MSSQL extension is launched. This allows refreshes to load the latest changes without needing to do a complete build like it does currently.

watch_and_launch_extension

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -434,7 +434,7 @@ gulp.task('build', gulp.series('clean', 'ext:build', 'ext:install-service'));
 
 gulp.task('quick-build', function(done) {
 	if (!require('fs').existsSync('out')) {
-		return run('gulp build')();
+		run('gulp build')();
 	}
 
 	done();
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -432,6 +432,14 @@ gulp.task('clean', function (done) {
 
 gulp.task('build', gulp.series('clean', 'ext:build', 'ext:install-service'));
 
+gulp.task('quick-build', function(done) {
```
aasimkhan30, line 435: How do we know if out contains the latest code in the extension? It can also be stale code out, right? Why don't we just run watch?

lewis-sanchez: I tried using watch, but watch never completes since it's on going and that prevents the pre-launch task from "completing".

aasimkhan30 (approved):

aasimkhan30:
@lewis-sanchez quick question. What happens if we stop debugging and relaunch the extension. Will it create duplicate watch tasks in background?

lewis-sanchez:
> @lewis-sanchez quick question. What happens if we stop debugging and relaunch the extension. Will it create duplicate watch tasks in background?

It just relaunches the extension without a need to start a new watch task. Check out the gif:
watch_launch_and_relaunch_extension

aasimkhan30 (approved):

PR #17950, opened 2024-07-29
moving some deps to devdeps to reduce package size (#17949)

kburtram:


kburtram:
Merging to queue rc build.

PR #17954, opened 2024-07-30
Bumping MSAL and Node to current/active versions

Benjin:


kburtram (approved):

kburtram:
will this require everyone to update to Node.JS 22x on their dev environment?

Benjin:
@kburtram 
> will this require everyone to update to Node.JS 22x on their dev environment?

Nope, you can continue using Node 18 if you want, but it's probably a good idea to switch your dev environment to 20 or 22.  I bumped this because MSAL 1.x doesn't support higher than Node 18, which is in maintenance while 20 and 22 are recommended.

PR #17972, opened 2024-08-07
Update SQL.plist to add `RETURNING` keyword

andrewroazen:
`RETURNING` keyword added for `RETURNING columnname INTO :varname` clauses. This has already been tested by directly modifying `extensions/sql/sql.tmLanguage.json`.


kburtram (commented):
If I understand correctly, this is PL/SQL syntax, but not valid T-SQL.  Should this be added here or to the Oracle extension?

kburtram:
Closing since this appears to be PL/SQL and it isn't clear why this would be included here.

PR #17976, opened 2024-08-13
Adding pr comment bot that alerts if the size has increased.

aasimkhan30:
Comments on the PR about size changes in bundle and vsix files and fails if the size has increased by more than 5% or the vsix file is above 25MB.

This should help us detect any size changes in the PR and help us decide if those changes are necessary during the PR review process.

### DIFF
--- a/.github/workflows/file-size-check.yml
+++ b/.github/workflows/file-size-check.yml
@@ -9,7 +9,7 @@ on:
       - main
 
 permissions:
-  issues: write
+  content: write
 
 jobs:
   size-check:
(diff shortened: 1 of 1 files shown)


.github/workflows/file-size-check.yml
```diff
@@ -0,0 +1,133 @@
...
+            echo "Main branch VSIX size: $main_size bytes"
+            echo "PR branch VSIX size: $pr_size bytes"
+            echo "Size difference: $size_diff bytes"
+            echo "Percentage change: $percentage_change%"
+
+            echo "main_vsix_size=$main_size" >> $GITHUB_ENV
+            echo "pr_vsix_size=$pr_size" >> $GITHUB_ENV
+            echo "vsix_percentage_change=$percentage_change" >> $GITHUB_ENV
+
+      -name: Calculate webview bundle sizes
```
Benjin, line 76: We chatted about this offline, but to capture it here: let's add some documentation with the minutia of the webview bundling and how that impacts how webview module dependencies should be consumed.

Benjin (approved):

PR #17982, opened 2024-08-16
Updating readme to add instructions about disabling telemetry.

aasimkhan30:
Had an old branch that adds readme instructions for disabling telemetry.


kburtram (approved):
LGTM, thanks for writing that up!

PR #17983, opened 2024-08-19
Execution Plan View

laurennat:
View Execution Plans on files with .sqlplan attachment:

Entry point: 
Screenshot 2024-08-19 163945

Button on top right corner of .sqlplan files

executionPlanView


package.json
```diff
@@ -490,6 +505,15 @@
           "light": "media/changeConnection.svg"
         }
       },
+      {
+        "command": "mssql.executionPlan",
+        "title": "View Execution Plan",
```
kburtram, line 510: maybe "Open Execution Plan File" is more precise description of the command behavior.

package.json
```diff
@@ -258,6 +268,11 @@
           "command": "mssql.changeDatabase",
           "when": "editorLangId == sql",
           "group": "navigation@4"
+        },
+        {
+          "command": "mssql.executionPlan",
```
kburtram, line 273: maybe "mssql.openExecutionPlanFile" will help disambiguate this from future potential execution plan related commands (such as "mssql.toggleActualExecutionPlan").

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,168 @@
...
+import { WebviewRoute } from "../sharedInterfaces/webviewRoutes";
+import * as LocalizedConstants from "../constants/localizedConstants";
+
+export class ExecutionPlanWebViewController extends ReactWebViewPanelController<
+  ep.ExecutionPlanWebViewState,
+  ep.ExecutionPlanReducers
+> {
+  constructor(
+    context: vscode.ExtensionContext,
+    private _executionPlanService: ep.ExecutionPlanService,
```
kburtram, line 18: inconsistent use of "_" for private variables.  I won't call this out every time in this PR, but please either use the prefix or not consistently for all private variables (at least within a single file)>

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,168 @@
...
+    private executionPlanContents: string,
+    // @ts-ignore
+    private fileName: string
```
kburtram, line 21: how is this going to work for interactive query plans returned as XML with the query results, when there isn't a file?  Is there going to be a different controller for that?  If this controller is only for SQLPLAN files than perhaps the name of the class should reflect that.

src/controllers/mainController.ts
```diff
@@ -911,6 +917,27 @@ export default class MainController implements vscode.Disposable {
 		}
 	}
 
+	public async onExecutionPlan() {
+		const planContents = vscode.window.activeTextEditor.document.getText();
+		// remove file extension from name
+		let docName = vscode.window.activeTextEditor.document.fileName;
+		console.log("File name: ", docName);
```
kburtram, line 924: should we use the Logger class for debug logging, assuming we want to keep this in the code?

src/controllers/mainController.ts
```diff
@@ -911,6 +917,27 @@ export default class MainController implements vscode.Disposable {
 		}
 	}
 
+	public async onExecutionPlan() {
```
kburtram, line 920: `onOpenExecutionPlanFile`?

src/controllers/mainController.ts
```diff
@@ -911,6 +917,27 @@ export default class MainController implements vscode.Disposable {
...
+		const planContents = vscode.window.activeTextEditor.document.getText();
+		// remove file extension from name
+		let docName = vscode.window.activeTextEditor.document.fileName;
+		console.log("File name: ", docName);
+		let lastSlash = docName.lastIndexOf('\\');
```
kburtram, line 925: maybe something like `path.sep` would simplify this to not need the fallback?

src/controllers/mainController.ts
```diff
@@ -911,6 +917,27 @@ export default class MainController implements vscode.Disposable {
...
+		if (lastSlash === -1) {
+			lastSlash = docName.lastIndexOf('/');
+		}
+		docName = docName.substring(lastSlash + 1, docName.lastIndexOf('.'));
```
kburtram, line 929: Do we need to worry about files that don't have an extension?  perhaps forced into SQLPLAN mode using the language selection command.

.vscode/launch.json
```diff
@@ -32,7 +33,6 @@
 			"outFiles": [
 				"${workspaceRoot}/out/src/**/*.js",
 			],
-			"preLaunchTask": "watch",
```
aasimkhan30, line 35: Can you please undo this change?

.vscode/settings.json
```diff
@@ -28,5 +28,5 @@
         "typescriptreact",
         "typescript"
     ],
-    "svg.preview.background": "dark-transparent",
```
aasimkhan30, line 31: This change too.

src/reactviews/pages/ExecutionPlan/iconMenu.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+const useStyles = makeStyles({
+	iconStack: {
+		right: 0,
+		display: "flex",
+		flexDirection: "column",
+		alignItems: "center",
+		width: "25px",
+		opacity: 1,
+		zIndex: "9999",
```
kburtram, line 20: how did we arrive at this value for zIndex?  i.e., was it recommend by vscode extensibility documentation as a way to always be on top of all their UI or something?

src/controllers/executionPlanWebviewController.ts
kburtram: further in the code it looks like this is the documentTitle, so that makes sense.  Though cleaning up class and parameter names would help with code readability.

kburtram (changes requested):
About a quarter of the way through.  I'll send more comments later today.

src/sharedInterfaces/webviewRoutes.ts
```diff
@@ -6,4 +6,5 @@
 export enum WebviewRoute {
 	tableDesigner = 'tableDesigner',
 	connectionDialog = 'connectionDialog',
+	executionPlan = 'executionPlan'
```
kburtram, line 9: maybe 'executePlanDocument' is better.  I'm worried that "executionPlan" is too generic and will be confused with the execution plan graph that is embedded in the query editor.  Please throughout the code try to make it clear which part of the code is generic to all the execution plan code paths, and the parts specific to SQLPLAN documents.

kburtram:
could you please add some unit test coverage for this PR?

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,168 @@
...
+      this.state.totalCost += payload.totalCost;
+
+      return {
+        ...state,
+        totalCost: this.state.totalCost
+      };
+    });
+  }
+
+  private async getExecutionPlan() {
```
kburtram, line 136: typically "get" methods return a value, maybe createExecutionPlanGraphs or something would better describe what's happening in the method.

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
```diff
@@ -0,0 +1,203 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { useContext, useEffect, useState } from 'react';
+import { ExecutionPlanContext } from "./executionPlanStateProvider";
+import * as utils from './queryPlanSetup';
+import * as azdataGraph from 'azdataGraph/dist/build';
```
kburtram, line 9: do we need to reference `dist/build` to import this control?  Can't we do something like `import * as azdataGraphModule from 'azdataGraph';` like we do in ADS?

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
```diff
@@ -0,0 +1,203 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { useContext, useEffect, useState } from 'react';
+import { ExecutionPlanContext } from "./executionPlanStateProvider";
+import * as utils from './queryPlanSetup';
+import * as azdataGraph from 'azdataGraph/dist/build';
+import 'azdataGraph/src/css/common.css';
```
kburtram, line 10: this seems odd pattern to me.  Shouldn't the control import its own CSS files.

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
```diff
@@ -0,0 +1,203 @@
...
+		// @ts-ignore
+		window['mxResourceExtension'] = '.txt';
+		// @ts-ignore
+		window['mxLoadStylesheets'] = false;
+		// @ts-ignore
+		window['mxBasePath'] = './src/reactviews/pages/ExecutionPlan/mxgraph';
+
+		const mxClient = azdataGraph.default();
+		// console.log(mxClient);
+		console.log("Execution Plan State: ", executionPlanState);
```
kburtram, line 97: should we use Logger for debug output?

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
```diff
@@ -0,0 +1,203 @@
...
+		if (executionPlanView) {
+			executionPlanView.setZoomLevel(zoomNumber);
+			setExecutionPlanView(executionPlanView);
+			setZoomNumber(executionPlanView.getZoomLevel());
+		}
+		setCustomZoomClicked(false);
+	};
+
+	const getQueryCostPercentage = () => {
+		const percentage = (cost / executionPlanState!.totalCost!) * 100;
```
kburtram, line 157: Is this every a divide by zero situation?

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
```diff
@@ -0,0 +1,203 @@
...
+			</div>
+			<IconStack executionPlanView={executionPlanView} setExecutionPlanView={setExecutionPlanView} setZoomNumber={setZoomNumber} setCustomZoomClicked={setCustomZoomClicked} setFindNodeClicked={setFindNodeClicked} setHighlightOpsClicked={setHighlightOpsClicked} query={query}/>
+		</div>
+	);
+};
+
+/**
+ * Format a string. Behaves like C#'s string.Format() function.
+ */
+function formatString(str: string, ...args: any[]): string {
```
kburtram, line 190: should we put this in `utils.ts` to it can be used elsewhere as needed?

src/reactviews/pages/ExecutionPlan/findNodes.tsx
```diff
@@ -0,0 +1,128 @@
...
+import * as ep from './executionPlanInterfaces';
+
+const useStyles = makeStyles({
+	inputContainer: {
+		position: "absolute",
+		top: 0,
+		right: "35px",
+		padding: "10px",
+		border: "1px solid #ccc",
+		zIndex: "100",
```
kburtram, line 21: why 100 here?

src/reactviews/pages/ExecutionPlan/executionPlanView.ts
```diff
@@ -0,0 +1,421 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as ep from './executionPlanInterfaces';
+
+export class ExecutionPlanView {
+	private _diagram: any;
+	// private _diagramModel: ep.AzDataGraphCell;
```
kburtram, line 10: dead code?

src/reactviews/pages/ExecutionPlan/executionPlanView.ts
```diff
@@ -0,0 +1,421 @@
...
+				label: ''
+			};
+		});
+	}
+
+	private createGraphElementId(): string {
+		return `element-${this.getNonce()}`;
+	}
+
+	private getNonce(): string {
```
kburtram, line 173: seems like this should be shared utility function to be done consistently across views?

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,168 @@
...
+        graphFileContent: this.executionPlanContents,
+        graphFileType: ".sqlplan",
+      };
+      this.state.executionPlan =
+        await this._executionPlanService.getExecutionPlan(planFile);
+      this.state.executionPlanGraphs = this.state.executionPlan.graphs;
+    }
+  }
+
+  private async fileExists(
```
kburtram, line 148: maybe this should be in `utils.ts` to be reused.  There appears to be a similar method there already, but this one seems to use vscode URI instead of paths, so maybe we need both?

kburtram (changes requested):
Looks good overall, but still working through the code.  Please take a look through comments and I'll check back through code in next iterations.

src/reactviews/pages/ExecutionPlan/iconMenu.tsx
laurennat: I changed it to a more reasonable value.

src/reactviews/pages/ExecutionPlan/executionPlanView.ts
laurennat: yes, removed!

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
laurennat: Not really. it's only zero if there are no graphs returned by the api call, but one of the prereq conditions of the executionPlanGraph component even showing up is that there is a graph for it to show up.

src/reactviews/pages/ExecutionPlan/findNodes.tsx
laurennat: fixed to a more reasonable value

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,168 @@
...
+
+export class ExecutionPlanWebViewController extends ReactWebViewPanelController<
+  ep.ExecutionPlanWebViewState,
+  ep.ExecutionPlanReducers
+> {
+  constructor(
+    context: vscode.ExtensionContext,
+    private _executionPlanService: ep.ExecutionPlanService,
+    private executionPlanContents: string,
+    // @ts-ignore
```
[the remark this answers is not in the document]
laurennat: for some reason it doesn't recognize that filename is getting used in the super call. I added a comment about it

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,168 @@
...
+      await this.getExecutionPlan();
+
+      return {
+        ...state,
+        executionPlan: this.state.executionPlan,
+        executionPlanGraphs: this.state.executionPlanGraphs
+      };
+    });
+    this.registerReducer("saveExecutionPlan", async (state, payload) => {
+      const homeDir = require("os").homedir(); // Get the user's home directory
```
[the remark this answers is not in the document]
laurennat: nope, i changed it to the regular way!

src/controllers/mainController.ts
laurennat: for now the entry point only shows up on files with .sqlplan as an extension. I can definitely add in support for other files later.

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
laurennat: The import statement and require() didn't work in this context. So this was the next best thing we came up with

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
laurennat: same reason as the above comment, this was the only way to load in stylings.

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
laurennat: Basically, when we tried to import it normally, typescript would not be able to find a type definition for it. When we added one in, the package would not load correctly, and was not usable. the require didn't work for similar reasons. You can probably ask @aasimkhan30  for more context

package.json
```diff
@@ -258,6 +268,11 @@
...
+          "when": "editorLangId == sqlplan && config.mssql.enableExperimentalFeatures",
+          "group": "navigation@1"
```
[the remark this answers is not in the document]
laurennat: I reorganized them so they're together

package.json
laurennat: changed it

package.json
laurennat: changed it

src/controllers/mainController.ts
laurennat: changed!

src/controllers/executionPlanWebviewController.ts
laurennat: changed it!

src/controllers/mainController.ts
laurennat: I'm just going to remove them. they were more just there for while i was developing. Unless for some reason I should log them?

src/sharedInterfaces/webviewRoutes.ts
laurennat: changed it

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
laurennat: removed

src/controllers/executionPlanWebviewController.ts
laurennat: i fixed this for both of the files it was inconsistent in: executionPlanWebviewController and executionPlanView

src/reactviews/pages/ExecutionPlan/executionPlanView.ts
laurennat: I moved it to src\reactviews\common\utils.ts

src/controllers/executionPlanWebviewController.ts
laurennat: moved this to src\reactviews\common\utils.ts

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
laurennat: I moved it to src\reactviews\common\utils.ts

aasimkhan30:
@kburtram , I noticed the size of the js file increase by a lot due to embedding all the execution plan images in the js file. I have a new PR that addresses this issue. Instead of embedding all images in js files, now we only add links to them. @laurennat please go ahead with your PR and I will merge mine afterwards. 
https://github.com/microsoft/vscode-mssql/pull/17986

kburtram:
@laurennat is it possible to have the "File->Open File" command load the .sqlplan file into the visual document by default, perhaps using a custom editor (i.e. https://code.visualstudio.com/api/extension-guides/custom-editors).  It seems unusual to have a specific command for opening a particular file type to me.

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
kburtram: Do you know why the import syntax I showed working ADS?

kburtram:
For the document title, I would recommend we just show the file name, such as "plan.sqlplan" in the below screenshot (without the text 'Execution Plan'.  Basically, the document is just the standard editor for the file type in this scenario.

image

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
kburtram: hmm, is this something that should be fixed in the control?  It's fine to leave for initial check-in but would be good to fix at some point.

src/controllers/mainController.ts
kburtram: If I paste showplan XML into an untitled text document and then change the language mode to SQL Execution Plan, when I show the control the title just says "Execution Plan".  Is that the intended behavior?

image

kburtram:
@laurennat the '+' and '-' appear to be inverted compared to ADS.

VSCode
image

ADS
image

kburtram:
The toolbar icons seem a little bigger here than in ADS.  Could you please try to match the styles in ADS as closely as possible since that UX has already gone through a design review?

image

kburtram:
It can be done as a follow-up PR if needed, but let's open a task to fix the styling on the "Find Node" popup to more closely match ADS.  For example, the font and width of the dropdown control doesn't appear as good.

ADS
image

VSCode
image

aasimkhan30:
> @laurennat the '+' and '-' appear to be inverted compared to ADS.
> 
> VSCode image
> 
> ADS image

@laurennat can you update to the latest version from https://github.com/microsoft/azdataGraph `68` for the fix.

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
aasimkhan30: @kburtram , that's because we add specify that alias in the loader. 
https://github.com/microsoft/azuredatastudio/blob/db96392893698710b4a4280a938022b2b4807894/src/vs/code/browser/workbench/workbench.html#L119

We can probably add something similar in the esbuild options but I agree that the right fix will be in the control itself.

aasimkhan30:
@laurennat , can you make sure that the toolbar is focusable? Currently I cannot press tab to enter the toolbar from the graph. Moreover, can you please make sure whenever we open a find, search, zoom widget, we focus in the first field for the widget.

typings/azdataGraph.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module 'azdataGraph/dist/build' {
```
aasimkhan30, line 1: @kburtram , I will see if we can move some typing to the control. That should fix some of the import problems.

yarn.lock
```diff
@@ -2293,6 +2293,10 @@ axios@^1.6.0:
     form-data "^4.0.0"
     proxy-from-env "^1.1.0"
 
+"azdataGraph@github:Microsoft/azdataGraph#0.0.66":
```
aasimkhan30, line 2296: @laurennat, can you please upgrade to version 68

src/utils/utils.ts
```diff
@@ -14,4 +15,14 @@ export async function exists(path: string): Promise<boolean> {
 	}
 }
 
+export async function fileExists(uri: vscode.Uri, filename: string): Promise<boolean> {
```
aasimkhan30, line 18: Maybe we can use https://code.visualstudio.com/updates/v1_37#_vscodeworkspacefs `vscode-fs`

This works in ssh and wsl scenarios as well.

src/reactviews/pages/ExecutionPlan/iconMenu.tsx
```diff
@@ -0,0 +1,171 @@
...
+		}
+		else {
+			setFindNodeClicked(false);
+			setHighlightOpsClicked(true);
+			setCustomZoomClicked(false);
+		}
+	}
+
+	return (
+		<div id="iconStack" className={classes.iconStack} style={{outline: `2px solid ${utils.seperator(executionPlanState!.theme!)}`, background:`${utils.background(executionPlanState!.theme!)}`}}>
```
aasimkhan30, line 132: @laurennat, Can we switch to toolbar from fluent ui? https://react.fluentui.dev/?path=/docs/components-toolbar--docs

This toolbar is not accessible in its current form. 

@kburtram, while it doesn't look like ads toolbar but at least it will be accessible and responsive.

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,157 @@
...
+    context: vscode.ExtensionContext,
+    private executionPlanService: ep.ExecutionPlanService,
+    private executionPlanContents: string,
+    // needs ts-ignore because linter doesn't recognize that fileName is being used in the call to super
+    // @ts-ignore
+    private xmlPlanFileName: string
+  ) {
+    super(
+      context,
+      `${xmlPlanFileName} ${LocalizedConstants.executionPlan}`,  // Sets the webview title
```
aasimkhan30, line 28: Can't you just force this as string instead of using ts-ignore?

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,157 @@
...
+        ),
+      }
+    );
+    this.state.isLoading = true;
+    this.initialize();
+  }
+
+  private async initialize() {
+    this.state.sqlPlanContent = this.executionPlanContents;
+	  this.state.theme = vscode.window.activeColorTheme.kind === vscode.ColorThemeKind.Dark ? "dark" : "light";
```
aasimkhan30, line 51: @laurennat , why is this needed? I thought this is already done by the base class. If there is a bug, please fix it in the based class so that all web views get fixed.

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,157 @@
...
+
+      return {
+        ...state,
+        executionPlan: this.state.executionPlan,
+        executionPlanGraphs: this.state.executionPlanGraphs
+      };
+    });
+    this.registerReducer("saveExecutionPlan", async (state, payload) => {
+      const homeDir = homedir();
+      const documentsFolder = vscode.Uri.file(`${homeDir}/Documents`);
```
aasimkhan30, line 72: What about other OSes? Will this work for Mac and linux too? Also, can you please use join path instead. This might give some problems across different oses.

`vscode.Uri.joinPath `

kburtram:
I'm seeing the query plan document get stuck loading with a notification about open/closing file.  I don't see an exception in the log.  Also, I don't think we need the notification when loading the file, IMO. 

image

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,157 @@
...
+        content: payload.sqlPlanContent,
+        language: 'xml'
+      });
+
+      await vscode.window.showTextDocument(planXmlDoc);
+
+      return state;
+    });
+    this.registerReducer("showQuery", async (state, payload) => {
+      const sqlDoc = await vscode.workspace.openTextDocument({
```
aasimkhan30, line 119: Can you use the existing service to open sql documents?

For example:
https://github.com/microsoft/vscode-mssql/blob/4c4f13274f0881a613af0c822fd0b0d49fcc2b2d/src/tableDesigner/tableDesignerWebViewController.ts#L215

src/controllers/reactWebviewController.ts
```diff
@@ -5,6 +5,7 @@
 
 import * as vscode from 'vscode';
 import { WebviewRoute } from '../sharedInterfaces/webviewRoutes';
+import { getNonce } from '../reactviews/common/utils';
```
aasimkhan30, line 8: Can we keep react views folder for only things used within the react app? I will try to add a linter rule to prevent such import.

src/reactviews/pages/ExecutionPlan/highlightExpensiveOperations.tsx
```diff
@@ -0,0 +1,176 @@
...
+			</Combobox>
+			<Button onClick={handleHighlightExpensiveOperation} icon={<Checkmark20Regular />} />
+			<Button icon={<Dismiss20Regular />} onClick={handleHighlightClose}/>
+		</div>
+
+	);
+};
+
+
+function getExpensiveOperationDelegate(selectedExpensiveOperationType: ep.ExpensiveMetricType): (cell: ep.AzDataGraphCell) => number | undefined {
```
aasimkhan30, line 98: I am guessing you picked these functions from ads. Maybe we can move them to the azdatagraph to prevent duplication in the next iteration.

src/reactviews/pages/ExecutionPlan/highlightExpensiveOperations.tsx
```diff
@@ -0,0 +1,176 @@
...
+		gap: "2px",
+		opacity: 1
+	},
+	dropdown: {
+		maxHeight: "200px"
+	}
+})
+
+interface HighlightExpensiveOperationsProps {
+	executionPlanView: any;
```
aasimkhan30, line 34: Can we use proper types instead of any?

aasimkhan30: You should be able to import and use vscodeMssql types here.

src/reactviews/pages/ExecutionPlan/findNodes.tsx
```diff
@@ -0,0 +1,128 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { useContext, useState } from 'react';
```
aasimkhan30, line 6: @laurennat , have you tried using something like popover from fluent control?
https://react.fluentui.dev/?path=/docs/components-popover--docs

src/reactviews/pages/ExecutionPlan/highlightExpensiveOperations.tsx
laurennat: added proper typings!

laurennat:
> The toolbar icons seem a little bigger here than in ADS. Could you please try to match the styles in ADS as closely as possible since that UX has already gone through a design review?
> 
> image

Fixed this to be closer in size

kburtram:
Another issue to look at is the sizing of the plan control when the sqlplan file contains multiple queries.  The document should look similar to in ADS (shown below) where the graphs take a proportional amount of space based on the size of the plan tree.  In VS Code the first graph consumes the whole document space regardless of plan contents, and second plan is below but not initially visible without scrolling.

ADS
image

VS Code
image

src/controllers/executionPlanWebviewController.ts
laurennat: i don't think it was a problem of it being defined or not, it just straight didn't register that it was being used

src/controllers/executionPlanWebviewController.ts
laurennat: the base class does have the theme, but it doesn't specify whether the theme is light or dark, which I need to know to load in the icons

src/controllers/executionPlanWebviewController.ts
```diff
@@ -0,0 +1,157 @@
...
+        executionPlan: this.state.executionPlan,
+        executionPlanGraphs: this.state.executionPlanGraphs
+      };
+    });
+    this.registerReducer("saveExecutionPlan", async (state, payload) => {
+      const homeDir = homedir();
+      const documentsFolder = vscode.Uri.file(`${homeDir}/Documents`);
+
+      let filename: vscode.Uri;
+      let counter = 1;
```
kburtram, line 75: nit: I wonder if the counter should start at 2, so the second SQLPLAN file is "plan2.sql"?

src/controllers/mainController.ts
```diff
@@ -1411,4 +1418,38 @@ export default class MainController implements vscode.Disposable {
 	public onClearAzureTokenCache(): void {
 		this.connectionManager.onClearTokenCache();
 	}
+
+	private ExecutionPlanCustomEditorProvider = class implements vscode.CustomTextEditorProvider {
```
kburtram, line 1422: it may make sense to organize this differently, i.e., in it's own file in a ExecutionPlan directory or something like that.  But since it's a short code block and there's only one custom editor, this could be cleaned up later.

src/controllers/mainController.ts
```diff
@@ -1411,4 +1418,38 @@ export default class MainController implements vscode.Disposable {
...
+			service: ExecutionPlanService,
+        ) {
+			this.context = context;
+			this.service = service;
+        }
+
+        public async resolveCustomTextEditor(
+            document: vscode.TextDocument
+        ): Promise<void> {
+			await vscode.commands.executeCommand('workbench.action.closeActiveEditor');
```
kburtram, line 1436: what is this about?  Do we need to manually close the XML file or something?

src/controllers/reactWebviewController.ts
kburtram: maybe more common utility functions can go into 'src/utils/utils.ts'?

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
kburtram: ok, makes sense.  Maybe this is something we can clean-up later in the azgraph module.

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
kburtram: It looks like the total cost can be 0 with a query such as "SELECT 1", which causes the percent to display as NaN.

image

src/controllers/executionPlanWebviewController.ts
kburtram: this is same behavior as ADS and no user hasn't opened a suggestion to change yet.  So I'll resolve this comment.

src/controllers/mainController.ts
laurennat: by default, vscode first opens the document in it's text editor, and then calls the custom editor to handle it. when we make a new react webview, it opens in another tab, so I'm closing the old tab and then opening the react view. I'm trying to see if there's a way to open the content in the existing tab

kburtram: I'm a little worried about race conditions with this snippet since we aren't explicitly expressing which editor to close.  For example, if the active editor changes between the time the temp doc is opened and this call is run.  Ideally, we could use the initial document directly if possible, or if not at least use an API that allows us to provide the URI of the document to close.

laurennat:
> I'm seeing the query plan document get stuck loading with a notification about open/closing file. I don't see an exception in the log. Also, I don't think we need the notification when loading the file, IMO.
> 
> image

this is fixed

src/controllers/mainController.ts
laurennat: addressed this by moving the closing command to within the onDidOpenTextDocument function so it's guaranteed to close the correct file

yarn.lock
aasimkhan30: Are you not updating this to 68 @laurennat

typings/azdataGraph.d.ts
kburtram: thanks!

src/reactviews/pages/ExecutionPlan/iconMenu.tsx
kburtram: @aasimkhan30 this is a great idea, though let's do it in a subsequent PR please.

src/controllers/mainController.ts
```diff
@@ -1411,4 +1422,41 @@ export default class MainController implements vscode.Disposable {
...
+        ): Promise<void> {
+			await this.onOpenExecutionPlanFile(document);
+        }
+
+		public async onOpenExecutionPlanFile(document: vscode.TextDocument) {
+			const planContents = document.getText();
+			let docName = document.fileName;
+			docName = docName.substring(docName.lastIndexOf(path.sep) + 1);
+
+			const executionPlanPanel = new ExecutionPlanWebViewController(
```
kburtram, line 1451: nit: why is a controller object named `executionPlanPanel`, shouldn't it be something like `executionPlanController`?

aasimkhan30:
@laurennat please push yarn.lock in to fix the build issue

src/utils/utils.ts
```diff
@@ -14,4 +15,26 @@ export async function exists(path: string): Promise<boolean> {
...
+    const path = vscode.Uri.joinPath(uri, filename);
```
kburtram, line 19: nit: it seems odd to have two methods in this file named `exists` and `fileExists` doing basically the same thing but with different arguments.  Maybe they should be overloads using the different `fs` implementation based on the parameters provided?

kburtram (approved):
LGTM all-up.  Please either resolve all pending issues prior to merging or copy the feedback to an issue for follow-up.

src/reactviews/pages/ExecutionPlan/iconMenu.tsx
laurennat: Issue here: https://github.com/microsoft/vscode-mssql/issues/17996

croblesm:
It looks very good 👍

Lasso66:
> Another issue to look at is the sizing of the plan control when the sqlplan file contains multiple queries.  The document should look similar to in ADS (shown below) where the graphs take a proportional amount of space based on the size of the plan tree.  In VS Code the first graph consumes the whole document space regardless of plan contents, and second plan is below but not initially visible without scrolling.
> 
> ADS
> image
> 
> VS Code
> image
>

PR #17990, opened 2024-08-23
Adding advanced connection settings pane to connection dialog

Benjin:
* Refactoring connection dialog form slightly
* Generating form items from STS call (paired with some fixups in https://github.com/microsoft/sqltoolsservice/pull/2386)
* Adding help text to each item, where available

https://github.com/user-attachments/assets/20821533-668d-4d1f-8567-f01699c345fb


package.json
```diff
@@ -149,6 +149,7 @@
     "@azure/msal-common": "^14.14.0",
     "@azure/msal-node": "^2.12.0",
     "@microsoft/ads-extension-telemetry": "^3.0.2",
+    "@types/azdata": "^1.46.6",
```
Benjin, line 152: remove this

src/connectionconfig/connectionDialogWebViewController.ts
```diff
@@ -314,38 +293,129 @@ export class ConnectionDialogWebViewController extends ReactWebViewPanelControll
...
+		const allConnectionFormComponents = new Map<string, {option: ConnectionOption, component: FormComponent}>();
+
+		for (const option of connectionOptions) {
+			allConnectionFormComponents.set(option.name, {option, component: this.convertToFormComponent(option)});
+		}
+
+		await this.completeFormComponents(allConnectionFormComponents);
+
+		// organize the main components and advanced components
+		const mainComponents: FormComponent[] = [];
```
Benjin, line 382: add comment saying that advanced components are grouped, while main components aren't.

src/connectionconfig/connectionDialogWebViewController.ts
```diff
@@ -314,38 +293,129 @@ export class ConnectionDialogWebViewController extends ReactWebViewPanelControll
...
+				};
+			}
+			return {
+				isValid: true,
+				validationMessage: ''
+			};
+		};
+	}
+
+	private _mainOptionNames = new Set<string>([
```
Benjin, line 350: consider changing this to an array to ensure consistent ordering

src/models/contracts/connection.ts
```diff
@@ -258,3 +259,15 @@ export namespace EncryptionKeysChangedNotification {
 export namespace ClearPooledConnectionsRequest {
 	export const type = new RequestType<object, void, void, void>('connection/clearpooledconnections');
 }
+
+//#region Connection capabilities
+
+export namespace GetCapabilitiesRequest {
```
Benjin, line 265: Add docstring

src/reactviews/pages/ConnectionDialog/connectionFormPage.tsx
```diff
@@ -0,0 +1,144 @@
...
+                    width: "200px",
+                    alignSelf: "center",
+                }}
+            >
+                Advanced
+            </Button>
+            <Button
+                appearance="primary"
+                disabled={
+                    connectionDialogContext.state.connectionStatus ===
```
Benjin, line 121: note to self: increase the linewidth for prettier to avoid unnecessary wrapping

src/reactviews/pages/ConnectionDialog/connectionPageContainer.tsx
```diff
@@ -0,0 +1,79 @@
...
+		case FormTabType.ConnectionString:
+			return <ConnectionStringPage />;
+	}
+}
+
+export const ConnectionInfoFormContainer = () => {
+	const connectionDialogContext = useContext(ConnectionDialogContext);
+	const classes = useStyles();
+
+	// TODO: if you want to generate components from a form spec, that's fine, but they shouldn't be part of the state.
```
Benjin, line 57: remove comment; I concede :)

src/sharedInterfaces/connectionDialog.ts
```diff
@@ -122,11 +129,11 @@ export interface FormComponentOptions {
 /**
  * Interface for a form event
  */
-export interface FormEvent {
+export interface FormEvent<T> {
```
Benjin, line 132: Pull `FormContextProps<T>`, `FormEvent<T>`, and baseclass for `ConnectionDialogReducers` to Form-dedicated file

kburtram:
@Benjin in terms of UX, I was expecting the advanced options to be an additional tab, like Connection String.  From the video the UX looks good to me (I'll sync the branch and test it out).  Though I'm interested if this "dialog on top of dialog within a document" pattern is what we want to follow across the product.  No concerns with this for the initial check-in, but we should do a UX review prior to public preview.

src/connectionconfig/connectionDialogWebViewController.ts
```diff
@@ -184,96 +180,75 @@ export class ConnectionDialogWebViewController extends ReactWebViewPanelControll
...
-				propertyName: 'password',
-				label: 'Password',
-				required: false,
-				type: FormComponentType.Password,
-			},
-			{
+					}),
+				};
+			default:
+				console.log(`Unhandled connection option type: ${connOption.valueType}`);
```
kburtram, line 233: should we be using `Logger` for error messages?  I see inconsistent usage of Logger and accessing console directly through the codebase, but that doesn't seem desirable unless I'm missing something.

src/connectionconfig/connectionDialogWebViewController.ts
```diff
@@ -314,38 +293,129 @@ export class ConnectionDialogWebViewController extends ReactWebViewPanelControll
...
-				type: FormComponentType.Checkbox,
-			},
+			}
+		});
+
+		components.set('profileName', {
+			option: undefined,
+			component: {
+			propertyName: 'profileName',
+			label: 'Profile Name',
```
kburtram, line 303: should these labels be localized?

src/connectionconfig/connectionDialogWebViewController.ts
Benjin: Never mind:
1) JS Maps are iterated in insertion order
2) insertion order (and therefore ordering) is controlled by STS.

src/connectionconfig/connectionDialogWebViewController.ts
```diff
@@ -184,96 +180,75 @@ export class ConnectionDialogWebViewController extends ReactWebViewPanelControll
...
-			},
-			{
-				// Hidden if connection string is set or if the authentication type is not SQL Login
-				propertyName: 'user',
-				label: 'User Name',
-				type: FormComponentType.Input,
-				required: true,
-				validate: (value: string) => {
-					if (this.state.connectionProfile.authenticationType === AuthenticationType.SqlLogin && !value) {
+	private convertToFormComponent(connOption: ConnectionOption): FormComponent {
```
kburtram, line 183: I wonder if it would be useful to have a more generic "convert metadata to UI components" API similar to this but without the connection details.  Though then you'd still need a "convert connection metadata to generic metadata" function, which would look similar to this.  Maybe if we need to do a bunch of "key/value" type dialogs in the future we can look into that.

src/connectionconfig/connectionDialogWebViewController.ts
```diff
@@ -314,38 +293,129 @@ export class ConnectionDialogWebViewController extends ReactWebViewPanelControll
...
+			component: {
+			propertyName: 'profileName',
+			label: 'Profile Name',
+			required: false,
+			type: FormComponentType.Input,
+			}
+		});
+
+		// add missing validation functions for generated components
+		components.get('server')!.component.validate = (value: string) => {
```
kburtram, line 310: it seems like this validation logic could be wired up more automatically using the "required" metadata property?  Since there's only a few properties I'm not sure it's worth refactoring though.

Benjin:
@kburtram Agreed that we should have Janki look at this before we go to public preview.  I put it as a drawer in the same tab (for now) for two reasons:

1. the two existing tabs - "Parameters" and "Connection String" - operate independently instead of converting between connection string form and per-property form of the same connection details.  I think the tabs should have consistent behavior (all tabs reflect properties/connection string of the same connection, or each tab operates independently), but don't have a preference as to which.  Something to chat with Janki about.

2. The ADS connection dialog throws a pane over the main options, so this matches that.  Just inspiration, though; not that we should be just copying ADS UX at every turn.

src/reactviews/common/forms/formUtils.tsx
```diff
@@ -0,0 +1,228 @@
...
+import { ConnectionDialogContextProps, FormComponent, FormComponentType, IConnectionDialogProfile } from "../../../sharedInterfaces/connectionDialog";
+import { ConnectionDialogContext } from "../../pages/ConnectionDialog/connectionDialogStateProvider";
+
+export const FormInput = ({ value, target, type }: { value: string, target: keyof IConnectionDialogProfile, type: 'input' | 'password' | 'textarea' }) => {
+	const connectionDialogContext = useContext(ConnectionDialogContext);
+	const [inputVal, setValueVal] = useState(value);
+	const [showPassword, setShowPassword] = useState(false);
+
+	useEffect(() => {
+		setValueVal(value);
```
kburtram, line 19: nit: not the most intuitive method name/call.  i.e. this is passing the value that will be the val passed to setValue which will update some value.  I guess since this is generic method maybe that is as precise as we can be here?

croblesm:
Hi @Benjin,

This looks promising, however I have some feedback:

* Change the `Parameter` vs `Connection string` option. Use a radio button for this option, and make `Parameter` the default and first option. 
* Change the `Advanced` and `Connect` button positioning, I would suggest putting them side by side.
* Reorder the Advanced options, show the most common (General) options first
* `Test connection` button is missing, but I'm not sure if you have that functionality ready in your backend
* `Cancel` button is missing

This is a concept design from Janki:
image

kburtram (commented):
I made it through about half the code.  Will take a look at the rest later today & will build/test the branch.  Please work with @aasimkhan30 to get a sign-off since he is most familiar with some of the code you're updating.

kburtram:
@Benjin makes sense on the design rationale.  I think this UI looks good to me; it just wasn't what I was expecting.  I'm interested in using these first UIs as an opportunity to establish UX patterns to copy in other features and this seems like a good 'test case.'  Though I'd definitely put this feedback in the "open a follow-up task to do in a separate PR" so that we can get this core functionality into next release.

src/connectionconfig/connectionDialogWebViewController.ts
Benjin: I think that's what Aasim was going for with the `FormComponent` and the helper functions this PR moves over to `formUtils.tsx`, though I need do finish up some of the refactoring ideas I was playing around with earlier this week in a different branch to fully divorce formUtils.ts from connection dialog components.

Maybe we're already doing what you're thinking about?
The idea is that `FormComponent` is a generic interface that can be used to generate a state-bound, validateable form composed of actual React components.

1. hardcode or generate an array of `FormComponent`s (happening here)
2. convert from `FormComponent[]` to renderable React components (happening in formUtils.tsx)

...in retrospect, maybe `FormComponent` is misnamed, because it's not a React _component_, but a spec for one.

src/connectionconfig/connectionDialogWebViewController.ts
Benjin: I definitely something worth looking into as the generic form structure gets refined.  It might need more prodding to correctly handle fields that are "conditionally required", like username, which is required, but only if auth is set to SqlLogin.  Lemme see what I can do.

src/connectionconfig/connectionDialogWebViewController.ts
Benjin: I've localized the ones in the controller for now, but Aasim's working on a larger localization effort (in `aasim/feat/newloc`) that'll enable localization in the webviews as well, as that isn't possible with the current system.

localization/xliff/enu/constants/localizedConstants.enu.xlf
```diff
@@ -677,6 +677,57 @@
         <trans-unit id="queryFailed">
             <source xml:lang="en">Query failed</source>
         </trans-unit>
+        <trans-unit id="connectionDialog">
+            <source xml:lang="en">ConnectionDialog</source>
```
aasimkhan30, line 681: Maybe Connection Dialog?

aasimkhan30 (approved):

aasimkhan30:
Having a drawer over the recent connections feels a bit awkward since the recent connections already resemble a drawer. Initially, I considered using toggle buttons for the parameters and connection string, along with tabs for the basic and advanced fields.


Example of toggle button 
image


Unfortunately, fluent doesn't support this ui component

Benjin:
@aasimkhan30 yup, the drawer

> Having a drawer over the recent connections feels a bit awkward since the recent connections already resemble a drawer. Initially, I considered using toggle buttons for the parameters and connection string, along with tabs for the basic and advanced fields.  Unfortunately, fluent doesn't support this ui component

@aasimkhan30 and I chatted offline, and I'll raise his ideas + concerns with Janki when we meet on Monday to iron out the differences between the mockups and what VSCode webviews can do.

src/connectionconfig/connectionDialogWebViewController.ts
kburtram: yeah, I think the FormComponent is doing the metadata-driven UI from the extension to the webview. But is still pretty UI-focused metadata (e.g. tooltip, label, etc.).  Where this function is converting from connection metadata to UI metadata.  I was wonder if we want a higher-level, more common metadata; though I suspect the connection metadata is really a special-case since I can't think of another similar example off-hand.

kburtram (approved):

PR #17991, opened 2024-08-24
Adding missing Always Encrypted setting

### LINKED ISSUE #17891: Column Encryption Setting=enabled
- MSSQL Extension Version:v1.22.1
- VSCode Version:1.88.0
- OS Version: Windows 10

There is no way to set up a connection string with Column Encryption Setting=enabled.

Benjin:
Fixes #17891

### DIFF
--- a/src/models/connectionCredentials.ts
+++ b/src/models/connectionCredentials.ts
@@ -74,6 +74,7 @@ export class ConnectionCredentials implements IConnectionInfo {
 		details.options['trustServerCertificate'] = credentials.trustServerCertificate;
 		details.options['hostNameInCertificate'] = credentials.hostNameInCertificate;
 		details.options['persistSecurityInfo'] = credentials.persistSecurityInfo;
+		details.options['columnEncryptionSetting'] = credentials.columnEncryptionSetting;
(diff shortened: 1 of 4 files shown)


src/models/connectionCredentials.ts
```diff
@@ -28,6 +28,7 @@ export class ConnectionCredentials implements IConnectionInfo {
 	public trustServerCertificate: boolean | undefined;
 	public hostNameInCertificate: string | undefined;
 	public persistSecurityInfo: boolean | undefined;
+	public columnEncryptionSetting: string | undefined;
```
kburtram, line 31: any need to restrict the possible values with a string literal type?

kburtram (approved):

lewis-sanchez (approved):

src/models/connectionCredentials.ts
Benjin: Given that other properties with a limited value-set (e.g. applicationIntent, authenticationType) are _not_, it seems like that isn't the assumption here.  Maybe something to look into changing, but might be more than just this PR.

Benjin:
Note: also appears to be missing properties for `attestationProtocol` and `enclaveAttestationUrl`

Issue #1753, opened 2020-09-30
SQL Messages need to persist and be accurate


cdchild:
- MSSQL Extension Version: 1.9.0
- VSCode Version: 1.49.2
- OS Version: Windows 10

Steps to Reproduce:

1. Open two SQL Scripts
2. Execute both SQL Scripts and note: the Messages section with the Total Row Count, any other message output, & Total Execution Time in the results window after each one is executed.
3. Switch between the results tabs of the two scripts.
Notice that the messages disappear (the Total Rows, statistics, Print output, etc.) disappears and the Total Execution Time increases by the original amount of time each time tabs are switched.

cdchild:
![image](https://user-images.githubusercontent.com/8571711/94712179-57a43b00-0306-11eb-8f90-39e4b75ba8dc.png)
First Results

![image](https://user-images.githubusercontent.com/8571711/94712249-73a7dc80-0306-11eb-91bc-19845ff26634.png)
Looks good so far

![image](https://user-images.githubusercontent.com/8571711/94712358-a05bf400-0306-11eb-8838-3d58b84f839e.png)
Original execution time was 00:00:00.021, now it's 00:00:00.042.
Row affected disappeared.

![image](https://user-images.githubusercontent.com/8571711/94712497-cb464800-0306-11eb-8cce-3973967c6a4e.png)
Original execution time was 00:00:00.018, now it's 00:00:00.036...
Row affected is no longer there either.

![image](https://user-images.githubusercontent.com/8571711/94713792-80c5cb00-0308-11eb-96a8-bdd34e4c59f9.png)
It'll keep going the more times you switch.

For simplicity in reporting/recreating I've omitted examples where the number of rows affected is helpful.  However, it is helpful especially when multi-tasking, waiting for longer/larger queries, etc.  We'll need to be able to switch between multiple results without worrying about the stats changing.  I have another request coming on statistics...

cdchild:
This also affects other statistics like the output from SET STATISTICS IO (https://docs.microsoft.com/en-us/sql/t-sql/statements/set-statistics-io-transact-sql).
Now you see it:
![image](https://user-images.githubusercontent.com/8571711/94715055-32193080-030a-11eb-9156-7b4a1227dc47.png)

Now you don't (after switching between tabs):
![image](https://user-images.githubusercontent.com/8571711/94715300-83292480-030a-11eb-872f-6a877d071b21.png)

RandomGHUser:
[...]

PR #17999, opened 2024-08-27
Retain query messages when switching tabs

kburtram:
Fixes https://github.com/microsoft/vscode-mssql/issues/1753


Charles-Gagnon:
@aasimkhan30 Is the -58% size difference for the webview bundle size comparison expected? It doesn't seem like the changes here should result in a change that massive. (even the -3% VSIX size difference seems like a lot for a couple line change)

cssuh (approved):

PR #18000, opened 2024-08-27
Adding filtering dialog to OE

aasimkhan30:
Node that has filtering options:
image

Filtering dialog: 
image

Once a filter is applied (name contains 'aasim):
1. Label is changed to add (filtered) suffix
2. Icon is changed to 'filled filter' indicating the node contains filter.
3. Remove filter icon appears that lets users clear filter without opening the dialog.
image


src/models/contracts/objectExplorer/nodeInfo.ts
```diff
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ObjectMetadata } from 'vscode-mssql';
+import * as vscodeMssql from 'vscode-mssql';
```
kburtram, line 6: nit: `vscodeMssql` isn't the easiest on the eyes when reading code. 😄 Maybe just `mssql` would "look" better.

src/objectExplorer/objectExplorerFilter.ts
```diff
@@ -0,0 +1,87 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { ReactWebViewPanelController } from "../controllers/reactWebviewController";
+import * as vscodeMssql from 'vscode-mssql';
+import * as vscode from 'vscode';
+import { WebviewRoute } from "../sharedInterfaces/webviewRoutes";
+import { TreeNodeInfo } from "./treeNodeInfo";
+import { ObjectExplorerFilterState, ObjectExplorerReducers } from "../sharedInterfaces/objectExplorerFilter";
+
+export class ObjectExplorerFilterReactWebviewController extends ReactWebViewPanelController<ObjectExplorerFilterState, ObjectExplorerReducers> {
```
kburtram, line 13: I wonder if we really care that this is a "ReactWebview" controller?  Isn't the fact that it uses react kinda an implementation details, or is the controller so coupled with React that it wouldn't make sense to use the framework without it?  Basically, should we ditch React in all the identifiers and just call it the `WebViewContoller`?

src/objectExplorer/objectExplorerFilter.ts
```diff
@@ -0,0 +1,87 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ReactWebViewPanelController } from "../controllers/reactWebviewController";
```
kburtram, line 6: We're inconsistently casing `WebView` and `Webview` in various identifier and filenames.  Would be great to be consistent there.  Looks like VS Code doesn't capitalize the `v` so I'd recommend going that direction.

src/objectExplorer/objectExplorerFilter.ts
```diff
@@ -0,0 +1,87 @@
...
+	private _onCancel: vscode.EventEmitter<void> = new vscode.EventEmitter<void>();
+	public readonly onCancel: vscode.Event<void> = this._onCancel.event;
+
+	constructor(
+		context: vscode.ExtensionContext,
+		data?: ObjectExplorerFilterState
+	) {
+		super(
+			context,
+			'Object Explorer Filter',
```
kburtram, line 26: does this require localization?

kburtram (approved):
Looks good to me.  We should get some UX feedback here for follow-up.  I'd like to see if we can rethink the dialogs a bit to make better use of the document space available, since we have so much more space available then with ADS dialogs, all the blank space doesn't look that great, IMO.

src/objectExplorer/objectExplorerFilter.ts
```diff
@@ -0,0 +1,87 @@
...
+			this._onCancel.fire();
+			this.panel.dispose();
+			return state;
+		});
+	}
+
+	public loadData(data: ObjectExplorerFilterState): void {
+		this.state = data;
+	}
+
```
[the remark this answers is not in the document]
Benjin: On that note, can you shrink the tabsize to 4 for the new files?  This is a little skewy to read.

croblesm:
Hi @aasimkhan30,

I'm not quite sure about the purpose of some of the filters, for example:
- owner
- durability type
- is memory optimized

Do we have telemetry to help us understand how popular these options are in ADS? TBH, these filters are probably more familiar to a data professional than to a developer.

cc @jdesai-24

Thanks!

src/objectExplorer/objectExplorerFilter.ts
aasimkhan30: There was already a webview controller. I added react prefix to this in order to separate from the old code. Once we get rid of those I can rename this class

src/objectExplorer/objectExplorerFilter.ts
```diff
@@ -0,0 +1,87 @@
...
+			'Object Explorer Filter',
+			WebviewRoute.objectExplorerFilter,
+			data ?? {
+				filterProperties: [],
+				existingFilters: [],
+				nodePath: ''
+			},
+			vscode.ViewColumn.Beside,
+			{
+				dark: vscode.Uri.joinPath(context.extensionUri, 'media', 'filter_inverse.svg'),
```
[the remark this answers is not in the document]
aasimkhan30: We cannot use @fluentui/react-icons in the extension code since it is react only library.

PR #18005, opened 2024-08-27
Fixing execution plan loading

aasimkhan30:
Switching setTimeout loops to message passing to eliminate any race conditions that might be happening. 

Also, added better loading states with support for error state when loading an invalid plan.
Error Page:
image

### DIFF
--- a/src/controllers/executionPlanWebviewController.ts
+++ b/src/controllers/executionPlanWebviewController.ts
@@ -12,7 +12,7 @@ import { homedir } from "os";
 import { exists } from "../utils/utils";
 import UntitledSqlDocumentService from '../controllers/untitledSqlDocumentService';
 import * as path from 'path';
-import { ApiStatus } from "../sharedInterfaces/connectionDialog";
+import { ApiStatus } from "../sharedInterfaces/webview";
 
 export class ExecutionPlanWebViewController extends ReactWebViewPanelController<
   ep.ExecutionPlanWebViewState,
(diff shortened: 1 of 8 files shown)


src/controllers/executionPlanWebviewController.ts
```diff
@@ -154,4 +165,8 @@ export class ExecutionPlanWebViewController extends ReactWebViewPanelController<
     }
     return sum;
   }
+
+  private updateState() {
```
laurennat, line 169: how does this change the state?

src/controllers/executionPlanWebviewController.ts
```diff
@@ -12,6 +12,7 @@ import { homedir } from "os";
 import { exists } from "../utils/utils";
 import UntitledSqlDocumentService from '../controllers/untitledSqlDocumentService';
 import * as path from 'path';
+import { ApiStatus } from "../sharedInterfaces/connectionDialog";
```
kburtram, line 15: should execution plan take a dependency on connection dialog?  i.e. should we move ApiStatus to a common shared location?

kburtram (approved):
LGTM, we can move ApiStatus separately if that's easier.

src/controllers/executionPlanWebviewController.ts
aasimkhan30: Will move to a new file since it is a common interface

laurennat (approved):

Issue #17901, opened 2024-05-10
mssql is not working 


Himanshukumar2812:
Type: <b>Bug</b>

not getting the options to connect the database, not even show the icon on the side bar for the connection please check the issue 

Extension version: 1.22.1
VS Code version: Code 1.89.1 (dc96b837cf6bb4af9cd736aa3af08cf8279f7685, 2024-05-07T05:13:33.891Z)
OS version: Windows_NT x64 10.0.19045
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz (8 x 1805)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|undefined|
|Memory (System)|15.39GB (6.50GB free)|
|Process Argv|--crash-reporter-id 975af454-5bea-47c2-8695-fee123d578da|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vspor879:30202332
vspor708:30202333
vspor363:30204092
tftest:31042121
vscoreces:30445986
vscod805cf:30301675
binariesv615:30325510
vsaa593cf:30376535
py29gd2263:31024239
c4g48928:30535728
azure-dev_surveyone:30548225
2i9eh265:30646982
962ge761:30959799
pythongtdpath:30769146
welcomedialogc:30910334
pythonidxpt:30866567
pythonnoceb:30805159
asynctok:30898717
pythontestfixt:30902429
pythonregdiag2:30936856
pyreplss1:30897532
pythonmypyd1:30879173
pythoncet0:30885854
h48ei257:31000450
pythontbext0:30879054
accentitlementsc:30995553
dsvsc016:30899300
dsvsc017:30899301
dsvsc018:30899302
cppperfnew:31000557
dsvsc020:30976470
pythonait:31006305
jchc7451:30973076
[...]

PR #18016, opened 2024-08-29
Remove accounts in account store with no keys

cssuh:
Adds a null check on the account.key field, so we can filter out any stale/broken account objects in extensionContext.globalState.  

Previously, attempting to remove an account with no key resulted in this error: `An error occurred while removing Microsoft Entra account: Cannot read properties of undefined (reading 'id')`

fixes: #18019

### DIFF
--- a/src/azure/accountStore.ts
+++ b/src/azure/accountStore.ts
@@ -41,7 +41,13 @@ export class AccountStore {
 			this._logger.error('Azure Account key not received for removal request.');
 		}
 		let configValues = this.getAccounts();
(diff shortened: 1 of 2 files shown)


Benjin (approved):

Benjin:
Is there a bug/issue to attach to this?

cssuh:
> Is there a bug/issue to attach to this?

https://github.com/microsoft/vscode-mssql/issues/17901 this addresses a bug that is in the discussion of this issue, although I'm not sure it addresses the base issue itself so I didn't link it.  I will open a new issue to attach to this PR.

PR #18017, opened 2024-08-29
Adding code splitting to webviews to reduce individual webview sizes.

aasimkhan30:
Code splitting analyzes bundles with multiple entry points and divides code into chunks such that 
a) a given piece of code is only ever in one chunk and 
b) each entry point doesn't download code that it will never use.
You can read more about it in the official architecture: https://github.com/evanw/esbuild/blob/main/docs/architecture.md#code-splitting

This PR separates the webviews into multiple entry points, allowing esbuild to optimally create chunks based on the two conditions mentioned above.

This approach should enhance the overall performance of smaller webviews, such as the connection dialog (which primarily uses Fluent controls) as it prevents VS Code from loading unnecessary libraries for other webviews.

Before this change, the entire webview was bundled into a single JavaScript file containing all the libraries. This meant that even for something simple, like the connection dialog, all libraries were loaded. For instance, we unnecessarily loaded libraries like azdataGraph and the Monaco editor for the connection dialog, which didn't require them at all.

Here is the structure of the bundle file before running this optimization:
image

With code splitting enabled, we now have separate chunks—a main entry chunk for each feature and additional chunks for shared code
image

For example, as shown in the image, the azdataGraph library is only included in the executionPlan.js chunk, while the Monaco editor is only included in the tableDesigner.js chunk. This approach ensures that each feature loads only the necessary libraries, reducing the overall load and improving performance.

The libraries that are shared across all three files (connection dialog, table designer, and execution plan), such as Fluent and React, are now placed in a separate chunk. This ensures that common dependencies are loaded only once, reducing redundancy and improving load times for the webviews.

Load times for connection dialog before optimization:
image
Load times for connection dialog after code splitting:
image

I don't believe this will have a significant impact on subsequent loads after VS Code caches the files. However, it does make a big difference during the initial few loads.

Will leave comments in the PR explaining some of my changes better.


.github/workflows/file-size-check.yml
```diff
@@ -77,24 +77,6 @@ jobs:
           echo "webview_bundle_percentage_change=$percentage_change" >> $GITHUB_ENV
           webview_text_color=$(if [ $percentage_change -gt 0 ]; then echo "red"; else echo "green"; fi)
           echo "webview_text_color=$webview_text_color" >> $GITHUB_ENV
-          main_js_file=$(find ./main/out/src/reactviews/assets -name "*.js")
```
aasimkhan30, line 80: Now file check just returns the file of the entire bundle folder.

gulpfile.js
```diff
@@ -220,7 +220,9 @@ async function generateReactWebviewsBundle() {
 		 * for each entry point, to be used by the webview's HTML content.
 		 */
 		entryPoints: {
-			mssqlwebview: 'src/reactviews/index.tsx'
+			'connectionDialog': 'src/reactviews/pages/ConnectionDialog/index.tsx',
```
aasimkhan30, line 223: Adding entry points for each features. New features will have to have their own index.tsx file and we will have to add them here to let esbuild optimally create chunks.

gulpfile.js
```diff
@@ -240,25 +242,26 @@ async function generateReactWebviewsBundle() {
 			typecheckPlugin()
 		],
 		sourcemap: prod ? false : 'inline',
-		metafile: !prod,
+		metafile: true,
```
aasimkhan30, line 245: I am always creating a metafile now. This file won't be included in the vsix file as it present in .vscodeignore.

caohai (approved):

gulpfile.js
```diff
@@ -240,25 +242,26 @@ async function generateReactWebviewsBundle() {
...
 		minify: prod,
 		minifyWhitespace: prod,
 		minifyIdentifiers: prod,
+		format: 'esm',
```
aasimkhan30, line 249: The default format for esbuild is 'cjs'. However, because esbuild doesn't support code splitting with 'cjs', I had to switch to 'esm'.

gulpfile.js
```diff
@@ -240,25 +242,26 @@ async function generateReactWebviewsBundle() {
...
+		splitting: true,
```
aasimkhan30, line 250: Flag that enabled code splitting.

src/controllers/reactWebviewBaseController.ts
```diff
@@ -60,33 +58,32 @@ export abstract class ReactWebviewBaseController<State, Reducers> implements vsc
...
 				</head>
 				<body>
-				  <div id="root"></div>
-				  <script>
-				 	const assetPathVscodeUri = "${this.resourceUrl(['']).toString()}";
-				  </script>
-				  <script nonce="${nonce}" src="${scriptUri}"></script>
+					<link rel="stylesheet" href="${this._sourceFile}.css">
+					<div id="root"></div>
+				  	<script type="module" nonce="${nonce}" src="${this._sourceFile}.js"></script>
```
aasimkhan30, line 84: Since we are using 'esm' format, we have to let chrome know that the script is of type module. Without this, chrome won't support esm syntax in the script like 'import'

src/reactviews/pages/ConnectionDialog/connectionHeader.tsx
```diff
@@ -27,7 +26,7 @@ export const ConnectionHeader = () => {
 					padding: '10px',
 				}
 			}
-				src={connectionDialogContext?.theme === webLightTheme ? loadImage(sqlServerImage) : loadImage(sqlServerImageDark)} alt='SQL Server' height={60} width={60} />
+				src={connectionDialogContext?.theme === webLightTheme ? sqlServerImage : sqlServerImageDark} alt='SQL Server' height={60} width={60} />
```
Benjin, line 29: this alt-text probably should be localized

aasimkhan30: Well fix this in my webview localization PR

PR #18018, opened 2024-08-29
Refactoring form to decouple from underlying form target type

Benjin:
With this change, the form system is now general-purpose and can be reused in different contexts.


src/connectionconfig/connectionDialogWebViewController.ts
```diff
@@ -154,19 +155,19 @@ export class ConnectionDialogWebViewController extends ReactWebViewPanelControll
 		}
 	}
 
-	private getActiveFormComponents(): FormComponent[] {
+	private getActiveFormComponents(): FormItemSpec<IConnectionDialogProfile>[] {
```
Benjin, line 158: I renamed `FormComponent` to `FormItemSpec` because I thought that there might be confusion about whether this was a React component.

aasimkhan30 (approved):
Thanks for making this change.

lewis-sanchez (approved):

PR #18023, opened 2024-09-04
Switching to new vscode localization api and library. 

aasimkhan30:
# Localization

The extension has 2 main ways to localize the strings in the UI.

## 1. Static contributions in `package.nls.json`

The extension uses the `package.nls.json` file to localize the strings the contributions in the `package.json` file. The `package.nls.json` file is a simple JSON file that maps the keys to localized strings.

Example of `package.nls.json`

```json
{
  "mssql.key1": "Localized string 1",
  "mssql.key2": "Localized string 2"
}
```

Using in `package.json` to contribute to vscode extensibility points. For example: commands

```json
{
  "commands": [
    {
      "command": "mssql.testCommand",
      "title": "%mssql.key1%",
      "icon": "$(debug-start)"
    }
  ]
}
```

### Extension Code (excluding React webviews):

If you need to localize strings in the extension code, you can use the directly use the l10n api provided by vscode.

```ts
import { l10n } from 'vscode';

const test = l10n.t('Test loc');
```

After adding a new string to localize, you need to run `yarn localization` to update the xlf and l10n files.

### React webviews:

Since webviews do not have access to the vscode API, we need to use `@vscode/l10n` library to localize strings in the React components.

```ts
import * as l10n from '@vscode/l10n';

const test = l10n.t(`Test loc`);
```

After adding a new string to localize, you need to run `yarn localization` to update the xlf and l10n files.

### Step 1: Extracting Localization Strings
1. Run the command `yarn localization`.
2. This triggers a Gulp task that scans all the source files for both the extension and webview.
3. It extracts all `l10n.t` function calls and compiles them into a `bundle.l10n.json` file located in the `localization\l10n` folder.
4. The task then reads this `bundle.l10n.json` file, along with the `package.nls.json` file, and updates the `enu.xliff` file found in the `localization\xliff` folder.

### Step 2: Translating the XLIFF File
- The `enu.xliff` file is sent to translators for translation.
- After translation, the translated XLIFF files are placed back in the `localization\xliff` folder.

### Step 3: Generating Localization Files
1. As a part of build process we run the gulp task `ext:generate-runtime-localization-files`.
2. This task reads the translated XLIFF files and generates localized JSON files:
   - `bundle.l10n.{locale}.json` files are saved in the `localization\l10n` folder.
   - `package.nls.{locale}.json` files are saved in the root folder.

### Step 4: Using the Localization Files
- The extension uses these generated JSON files to localize the strings in the user interface (UI). The root of the l10n is provided in the `package.json` file with `l10n` key. For now it is set to `localization/l10n`.

## To test localization.

1. Run `yarn gulp ext:generate-pseudo-loc` to generate pseudo localization files.
2. This generates `bundle.l10n.qps-ploc.json` and `package.nls.qps-ploc.json` files in the `localization\l10n` and root folders respectively.
3. Install the `Pseudo Language Language Pack` extension in VS Code.
4. Set the display language to `Pseudo Language`.
5. Reload vscode and test the extension to see the pseudo localized strings in the UI.

Webview example: 
image


localization/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Step 4: Using the Localization Files
+- The extension uses these generated JSON files to localize the strings in the user interface (UI). The root of the l10n is provided in the `package.json` file with `l10n` key. For now it is set to `localization/l10n`.
+
+## To test localization.
+
+1. Run `yarn gulp ext:generate-pseudo-loc` to generate pseudo localization files.
+2. This generates `bundle.l10n.qps-ploc.json` and `package.nls.qps-ploc.json` files in the `localization\l10n` and root folders respectively.
+3. Install the `Pseudo Language Language Pack` extension in VS Code.
+4. Set the display language to `Pseudo Language`.
```
Benjin, line 86: Can you add that this is via the `Configure Display Language` command palette command?

localization/README.md
```diff
@@ -0,0 +1,87 @@
+# Localization
```
Benjin, line 1: Thanks for writing this up!

src/controllers/connectionManager.ts
```diff
@@ -1086,52 +1083,57 @@ export default class ConnectionManager {
...
+					this.vscodeWrapper.logToOutputChannel(
+						LocalizedConstants.msgRefreshTokenSuccess(
+							profile.connectionId,
+							uri,
+							JSON.stringify(this.getConnectionInfo(uri))
+						)
+					);
 					return;
 				} catch {
 					this.vscodeWrapper.showInformationMessage(Utils.formatString(LocalizedConstants.msgRefreshTokenError));
```
Benjin, line 1125: Is there a reason Utils.formatString is still needed here?

Benjin: Given that we have string interpolation for non-localized strings and function-based string construction for localized strings, think we could just get rid of `Utils.formatString` entirely?  Feels like a bad-practice function to keep around to me since it doesn't enforce or inform the number or order of string parameters.

src/reactviews/common/vscodeWebViewProvider.tsx
```diff
@@ -35,6 +36,10 @@ interface VscodeWebviewContext<State, Reducers> {
 	 * Routes for the webview.
 	 */
 	route: WebviewRoute;
+	/**
+	 * Localization status.
```
Benjin, line 40: What's the purpose of this localization effect?  Can you elaborate in this docstring?

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
```diff
@@ -154,11 +153,17 @@ export const ExecutionPlanGraph: React.FC<ExecutionPlanGraphProps> = ({
   		return percentage.toFixed(2) + '%';
 	};
 
+	const queryCostHeader = l10n.t({
```
Benjin, line 156: Can you add an example for this sort of function-based localization to the loc README file, too?

localization/README.md
```diff
@@ -0,0 +1,87 @@
...
+    {
+      "command": "mssql.testCommand",
+      "title": "%mssql.key1%",
+      "icon": "$(debug-start)"
+    }
+  ]
+}
+```
+
+## 2. Dynamic contributions in `*.ts` and `*.tsx` files
```
Benjin, line 32: There's also the mostly-dynamic method of defining it in the locConstants file, right?  Can you add a blub here mentioning that file as well, and providing a recommendation of when to use one over the other?  (shared strings vs. one-time-use strings, converted "legacy" localized strings vs. new localized strings)

src/reactviews/pages/TableDesigner/designerChangesPreviewButton.tsx
```diff
@@ -147,7 +147,7 @@ export const DesignerChangesPreviewButton = () => {
         <DialogSurface>
             <DialogBody>
                 <DialogTitle>
-                    Preview Database Updates
+                    {'Preview Database Updates'}
```
Benjin, line 150: Why this change?  Should this be `l10n.t('Preview Database Updates')`?

tasks/localizationtasks.js
```diff
@@ -24,228 +26,142 @@ const iso639_3_to_2 = {
...
+		if (content) {
+			fileContents.push(content);
 		}
-
-		// Return json with {id:{target,source}} format
-		dict[id] = { 'source': source, 'target': target };
 	}
-
-	return dict;
+	const result = await vscodel10n.getL10nJson(fileContents.map(f => {
```
Benjin, line 56: Do you know if this will pick up both of these?

```
l10n.t('hello')
```

and

```
const { t } = l10n;
...
t('hello');
```

tasks/localizationtasks.js
```diff
@@ -24,228 +26,142 @@ const iso639_3_to_2 = {
 	trk: 'tr'
 };
 
+const supportedLocLangs = [
```
Benjin, line 29: Can you add a comment with a link to the list of valid loc strings?

Benjin (approved):

tasks/localizationtasks.js
aasimkhan30: I just got it from the existing loc folders in the repo. I don't know if we have some list anywhere.

src/reactviews/pages/ExecutionPlan/executionPlanGraph.tsx
aasimkhan30: I think it will be best if I link to the official documentation for the API. Since I don't want to keep updating the readme.

src/controllers/connectionManager.ts
aasimkhan30: I don't know why this was added in the first place as there is nothing to format for the string.

tasks/localizationtasks.js
aasimkhan30: added this note in the readme.

localization/README.md
aasimkhan30: It doesn't matter where we define the strings now as long as the strings are in src folder in 'ts' or 'tsx' files. I am just using locConstants as a way to reduce the changes for the PR.

aasimkhan30:
Spoke to loc team. The changes look good to check in. Will wait for a couple of weeks to see if the translations are done as expected.

Issue #18029, opened 2024-09-06
Connection dialog has unexpected behavior when profileName is an empty string or undefined


Benjin:
If the profile name is set to `''`, then the connection dialog doesn't generate a friendly display name.

If the profile name is undefined, then the connection dialog unexpectedly sets the profileName property for the user as a way to have something to display.

PR #18030, opened 2024-09-06
Fixing some connection string bugs

### LINKED ISSUE #18028: Connection dialog crashes when a connection is a connectionString and doesn't use SQL Auth
Repro:
1. Create a connection with a connectionString, e.g. `Data Source=MyServer;Integrated Security=true;TrustServerCertificate=true`
2. Open the connection dialog

Expected:
Dialog opens successfully

Reality:
Dialog crashes because all connectionString connections are expected to have a password saved in the connection store

Benjin:
Fixes #18028 and #18029


src/connectionconfig/connectionDialogWebViewController.ts
```diff
@@ -106,28 +106,41 @@ export class ConnectionDialogWebViewController extends ReactWebViewPanelControll
...
-					connection.password = password.substring(passwordStart, passwordEnd);
+			// If the connection is a connection string connection with SQL Auth:
+			//   * the full connection string is stored as the "password" in the credential store
+			//   * we need to extract the password from the connection string
+			// If the connection is a connection string connection with a different auth type, then there's nothing in the credential store.
+
+			const connectionString = await this._mainController.connectionManager.connectionStore.lookupPassword(connection, isConnectionStringConnection);
+
+			if (connectionString) {
+				const passwordIndex = connectionString.indexOf('Password=') === -1 ? connectionString.indexOf('password=') : connectionString.indexOf('Password=');
```
lewis-sanchez, line 124: This could be simplified with `const passwordIndex = connectionString.toLowerCase().indexOf('password=');`

The search will only be performed once with this and will match `Password=` regardless of case.

aasimkhan30 (approved):

lewis-sanchez (approved):

PR #18033, opened 2024-09-10
Lewissanchez/feat/add language service switch

### LINKED ISSUE #17298: [Suggestion] Include an option to set MSSQL to none by default
Currently, there are options to disable error checking and disable intellisense. Is it possible to have an option to set the MSSQL intellisense option located at the status bar to none by default?

lewis-sanchez:
This PR fixes #17298

This PR adds a new setting to the MSSQL extension that allows users to have the language mode set to none by default
image

New query editors that are opened when this setting is enabled, will have "None" as the chosen language
image

Previously saved query files that are opened when this setting is enabled, will also have "None" as the chosen language
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1197,6 +1197,12 @@
           "description": "%mssql.intelliSense.lowerCaseSuggestions%",
           "scope": "window"
         },
+        "mssql.intelliSense.setQueryEditorLanguageToNone": {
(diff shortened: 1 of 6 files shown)


lewis-sanchez:
Need to fix PR issues with whitespace
image

Issue #801, opened 2017-03-31
Extension overrides Postgres syntax


BenWildeman:
- MSSQL Extension Version: 0.3.0
- VSCode Version: 1.10.2
- OS Version: Windows 10

Steps to Reproduce:

1. Install MSSQL
2. Write PG SQL within a .sql file
3. Cry because of all the "errors"

When having to write SQL for both MSSQL and Postgres for my job, it's rather annoying that this extension has no idea about other SQL syntax that's very much valid

kevcunnane:
@NitroGhost thanks for raising this. We're somewhat limited by how VSCode supports language extensions - for a given file type, 1 editor is chosen. What might work is if you modified the PostGre extension to have additional, more specific file type (e.g. `.pg.sql`) and that way it would get used instead of our extension when opening that file type. 

Regarding support for PostGre - this isn't in the plans for right now. With community involvement we could certainly look at not listing errors on syntax / allowing a setting to disable our parsing on files. Again there are some limitations since VSCode accepts a syntax file for a lot of its processing, but it's certainly interesting to think about.

kburtram:
@NitroGhost another possible workaround for this issue is to add "mssql.intelliSense.enableErrorChecking": false option to your settings file.  This will turn off error highlighting for SQL files, but you'll still be able to use other language service and query execution features.  

Longer term we can look at possibly detecting Postgres or MySQL syntax and disabling our error checker for those SQL files.

aiingstan:
@kburtram how's the progress now? I'm including some MySQL .sql scripts in my project, jump into this annoying issue. How about adding a configuration item to exclude specified file extension, or read the very first line of the .sql file to determine syntax? Like shebang `#!` in Linux world.

jcush:
@kburtram @kevcunnane 
Is there any hope of this arriving soon? I have to work with both T-SQL and MySQL daily and after 2-3 weeks had over 2k "problems" in VSCode

I can confirm that Karl's suggestion of setting `mssql.intelliSense.enableErrorChecking` to `false` in the settings file does suppress all errors, but I do mean _all_ errors (T-SQL included). It's not ideal, but if you know how to wrap queries in transactions then the only negative effect of this solution is inconvenience

Zireael:
vscode-mssql fails to recognise Postgres syntax and incorrectly reports errors. It's very annoying and for this reason I've had to disable the extension.
Please expand to recognise Postgres syntax, or find a way to stop reporting it as errors.

[...]

PR #18034, opened 2024-09-10
Adds lang service switch

### LINKED ISSUE #17298: [Suggestion] Include an option to set MSSQL to none by default

lewis-sanchez:
This PR fixes #17298

This PR adds a new setting to the MSSQL extension that allows users to have the language mode set to none by default
image

New query editors that are opened when this setting is enabled, will have "None" as the chosen language
image

Previously saved query files that are opened when this setting is enabled, will also have "None" as the chosen language
image

To enable the MSSQL language mode for an editor, first click on "None" in the status bar, and then choose MSSQL as the language
image

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -882,11 +882,11 @@ export default class ConnectionManager {
 					// Notify the language service that the editor with the specified URI doesn't have a language flavor to avoid error squiggles
 					if (flavor === Constants.noneProviderName) {
 						SqlToolsServerClient.instance.sendNotification(LanguageServiceContracts.LanguageFlavorChangedNotification.type,
-							<LanguageServiceContracts.DidChangeLanguageFlavorParams>{
+							{
 								uri: fileUri,
 								language: 'sql',
 								flavor: Constants.noneProviderName
-							});
+							} as LanguageServiceContracts.DidChangeLanguageFlavorParams);
 					}
 				}
 				this.vscodeWrapper.logToOutputChannel(
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -23,7 +23,7 @@ import { IAccount } from '../models/contracts/azure';
 import * as ConnectionContracts from '../models/contracts/connection';
 import { ClearPooledConnectionsRequest, ConnectionSummary } from '../models/contracts/connection';
 import * as LanguageServiceContracts from '../models/contracts/languageService';
-import { EncryptOptions, IConnectionProfile } from '../models/interfaces';
+import { EncryptOptions, IConnectionProfile, IntelliSenseConfig } from '../models/interfaces';
 import { PlatformInformation, Runtime } from '../models/platform';
 import * as Utils from '../models/utils';
 import { IPrompter, IQuestion, QuestionTypes } from '../prompts/question';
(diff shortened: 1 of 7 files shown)


src/controllers/mainController.ts
```diff
@@ -1288,8 +1288,21 @@ export default class MainController implements vscode.Disposable {
...
+			const intelliSenseConfig = configuration.get<IntelliSenseConfig>(Constants.configIntelliSense);
+			const flavor = intelliSenseConfig.defaultQueryEditorLanguageToNone ? Constants.noneProviderName : Constants.mssqlProviderName;
 			// set encoding to false
-			this._statusview.languageFlavorChanged(doc.uri.toString(true), Constants.mssqlProviderName);
+			this._statusview.languageFlavorChanged(doc.uri.toString(true), flavor);
+
+			// Notify the language service that the editor with the specified URI doesn't have a language flavor to avoid error squiggles.
+			if (flavor === Constants.noneProviderName) {
+				SqlToolsServerClient.instance.sendNotification(LanguageFlavorChangedNotification.type,
+					<DidChangeLanguageFlavorParams>{
```
Benjin, line 1300: nit: we should try to use `{...} as DidChangeLanguageFlavorParams` instead of `<DidChangeLanguageFlavorParams>{...}` to make potential future intermingling with jsx/tsx easier.

src/controllers/connectionManager.ts
```diff
@@ -872,9 +872,22 @@ export default class ConnectionManager {
...
+					const intelliSenseConfig = configuration.get<IntelliSenseConfig>(Constants.configIntelliSense);
+					const flavor = intelliSenseConfig.defaultQueryEditorLanguageToNone ? Constants.noneProviderName : Constants.mssqlProviderName;
+					this.statusView.languageFlavorChanged(fileUri, flavor);
 					this.statusView.connecting(fileUri, connectionCreds);
-					this.statusView.languageFlavorChanged(fileUri, Constants.mssqlProviderName);
+					this.statusView.languageFlavorChanged(fileUri, flavor);
+
+					// Notify the language service that the editor with the specified URI doesn't have a language flavor to avoid error squiggles
+					if (flavor === Constants.noneProviderName) {
+						SqlToolsServerClient.instance.sendNotification(LanguageServiceContracts.LanguageFlavorChangedNotification.type,
```
Benjin, line 884: Does the language service already support this, or does there need to be work over here, too?

Benjin (approved):

src/controllers/connectionManager.ts
lewis-sanchez: The language service already supports it..

aasimkhan30:
Maybe we need some PM feedback on this.

Currently this option sets intellisense language to 'none' for all sql editors.

However, there are 2 conditions which should always have 'mssql' intellisense enabled.
1. if we open a query editor from a OE context menu, it should launch as MSSQL editor always. 
image
2. if user specifically opens an MSSQL editor from command line, it should still have the language set to MSSQL.
image

Maybe this option should only apply to sql files opened from disk as those files can belong to any database provider (PGSQL, MYSQL....)

lewis-sanchez:
@aasimkhan30, here's another GitHub issue that is related to the concern you're pointing out. (https://github.com/microsoft/vscode-mssql/issues/801).

This PR was intended to add a switch to optionally set the language mode to "None" as the default, instead of MSSQL when opening editors. The new switch that is added by this PR is initially turned off, meaning that MSSQL will be the default language mode when editors are opened.

kburtram:
> @aasimkhan30, here's another GitHub issue that is related to the concern you're pointing out. (#801).
> 
> This PR was intended to add a switch to optionally set the language mode to "None" as the default, instead of MSSQL when opening editors. The new switch that is added by this PR is initially turned off, meaning that MSSQL will be the default language mode when editors are opened.

How does #801 address the point Aasim mentions regarding opening editors with an MSSQL connection?  I think in those scenarios the language mode should probably be MSSQL regardless of this setting.  I think this setting is more relevant for non-connected SQL files.

kburtram (changes requested):
I think we need to do a design review on Intellisense settings and how to handle connections for new SQL documents.  And how best to handle non-MSSQL compatible SQL files.  This seems like part of the solution but maybe should be a little different than currently implemented.

Benjin:
Closing PR as we revisit the design

PR #18037, opened 2024-09-11
Code Styling fixes.

aasimkhan30:
I have added the prettier eslint rules which makes the code a lot more readable.


kburtram (approved):
LGTM, we can discuss how best to merge with least impact to feature work branches.

Benjin:
Let's get a list of paths that feature groups are working in, and add those to the exclude list.  We can have the respective devs add their path back for eslint once they're at a good point.

PR #18039, opened 2024-09-12
Refactoring in prep for azure browse and fixing up the UI to better match the mockups

Benjin:
image


aasimkhan30 (approved):
@Benjin, do you know why some labels are colored differently? Like 'save password'

aasimkhan30:
@Benjin probable because it's not an info label

src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -1,655 +1,660 @@
...
-				await this.panel.dispose();
-			} catch (error) {
-				this.state.connectionStatus = ApiStatus.Error;
-				return state;
-			}
-			return state;
-		});
-	}
-}
+/*---------------------------------------------------------------------------------------------
```
Benjin, line 1: not sure what's going on with the diff for this file...  The highlight is that I reorganized how form components are stored in the state (one record for everything), and how the different modes/views reference the specific form components (by propertyName instead of duplicating the entire component.

The idea is that this makes the connection dialog more composable without keeping state copied around.

aasimkhan30 (approved):

PR #18040, opened 2024-09-12
Localizing webview display strings

aasimkhan30:
Some pseudo loc screenshots:
Execution Plans:
image
Connection Dialog: (The non-localized strings are from STS)
image

Table Designer: (The non-localized strings are from STS which doesn't support pseudo-loc)
image


src/reactviews/pages/ConnectionDialog/connectButton.tsx
```diff
@@ -31,7 +34,7 @@ export const ConnectButton = ({style}: {style?: CSSProperties}) => {
 			}
 			iconPosition="after"
 			icon={ connectionDialogContext.state.connectionStatus === ApiStatus.Loading ? <Spinner size='tiny' /> : undefined}>
-				Connect
+				{CONNECT}
```
Benjin, line 37: is there a reason to pull this out into its own const rather than just having `{l10n.t("Connect")`?

Benjin (approved):

PR #18047, opened 2024-09-17
Open Protocol Handlers for Opening Connection Dialog

### LINKED ISSUE #17985: Support same app protocol handler connection options as ADS
We should support the same app protocol handler connection options as ADS.  This is to better support an "Open in VS Code" launch experience from web applications.

lewis-sanchez:
This PR closes #17985

This PR adds a new URI handler, so that user's can open the VS Code extension using URI's like the following:

`vscode://ms-mssql.mssql/connect?server=(localdb)\MssqlLocalDb&database=msdb&user=lewis&authenticationType=SqlLogin`

URI's like the one above can be entered into a browser. Once the URI is submitted, VS Code will prompt the user if they would like the URI: 
image

Clicking on "Open" in the prompt will open the connection dialog for users to make additional adjustments to connect to their target connection.
image

### Handling URIs with MSSQL Extension Not Installed or Disabled

When the MSSQL Extension is not installed or disabled, VS Code will do the following:
- Installs the extension, if not installed
image

- Enables the extension, if disabled
image


src/mssqlProtocolHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+	private connect(uri: vscode.Uri): IConnectionInfo | undefined {
+		return this.readProfileFromArgs(uri.query);
+	}
+
+	private readProfileFromArgs(query: string): IConnectionInfo | undefined {
+		if (!query) {
+			return undefined;
+		}
+
+		const args = new URLSearchParams(query);
```
aasimkhan30, line 46: Are we planning to support connection strings?

src/mssqlProtocolHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+		/*
+			Authentication Type:
+			1. Take --authenticationType, if not
+			2. Take --integrated, if not
+			3. take --aad, if not
+			4. If user exists, and user has @, then it's azureMFA
+			5. If user exists but doesn't have @, then its SqlLogin
+			6. If user doesn't exist, then integrated
+		*/
+		const authenticationType =
```
aasimkhan30, line 59: While this is a good start, maybe we should think about supporting all options we have in parameters so user can also provide advanced options

src/mssqlProtocolHandler.ts
lewis-sanchez: This is the first iteration since Benjin asked me to keep things small to avoid a high number of merge conflicts since he's still actively working on connections.

I'd say that we would want to support connection strings.

src/controllers/mainController.ts
```diff
@@ -223,6 +224,18 @@ export default class MainController implements vscode.Disposable {
 			const providerInstance = new this.ExecutionPlanCustomEditorProvider(this._context, this.executionPlanService, this._untitledSqlDocumentService);
 			vscode.window.registerCustomEditorProvider('mssql.executionPlanView', providerInstance);
 
+			if (this.isExperimentalEnabled) {
```
aasimkhan30, line 227: NIT: Can you put the if condition inside the URI handler and always register this handler?

src/mssqlProtocolHandler.ts
```diff
@@ -0,0 +1,182 @@
...
+		const multiSubnetFailover = isNaN(multiSubnetFailoverValue) ? undefined : multiSubnetFailoverValue === 1;
+
+		const multipleActiveResultSetsValue = parseInt(args.get('multipleActiveResultSets'));
+		const multipleActiveResultSets = isNaN(multipleActiveResultSetsValue) ? undefined : multipleActiveResultSetsValue === 1;
+
+		const packetSizeValue = parseInt(args.get('packetSize'));
+		const packetSize = isNaN(packetSizeValue) ? undefined : packetSizeValue;
+
+		const typeSystemVersion = args.get('typeSystemVersion') ?? undefined;
+		const connectionString = args.get('connectionString') ?? undefined;
```
aasimkhan30, line 140: Maybe add a validation that checks if connection string is provided and other parameters should be empty?

src/mssqlProtocolHandler.ts
```diff
@@ -0,0 +1,182 @@
...
+		/*
+			Authentication Type:
+			1. Take --authenticationType, if not
+			2. Take --integrated, if not
+			3. take --aad, if not
+			4. If user exists, and user has @, then it's azureMFA
+			5. If user exists but doesn't have @, then its SqlLogin
+			6. If user doesn't exist, then integrated
+		*/
+		const authenticationType =
```
aasimkhan30, line 66: @Benjin, isn't there a way to get all connection properties from STS? Should we use that API here? That might make future options automatic.

src/mssqlProtocolHandler.ts
```diff
@@ -0,0 +1,182 @@
...
+		const database = args.get('database') ?? '';
+		const user = args.get('user') ?? '';
+		const email = args.get('email') ?? '';
+		const accountId = args.get('accountId') ?? '';
+		const tenantId = args.get('tenantId') ?? '';
+
+		const portValue = parseInt(args.get('port'));
+		const port = isNaN(portValue) ? 0 : portValue;
+
+		/*
```
aasimkhan30, line 57: Maybe we could consider adding a dedicated section for URIs in the README. While this might be too big of a change for this PR, we could start by adding JSDoc comments to the URI handler and then auto-generating the relevant section in the README based on the code. This way, we wouldn't have to manually update the README with future changes to the URI handler.

That said, we might need to design the handler in a way that facilitates this process and ensures all supported URI patterns and their behaviors are clearly documented in the code for the generator to pick up.

aasimkhan30:
@lewis-sanchez , looks good as the first PR. Maybe maintain a json object or list for all property names for easy iteration over args and creation of connection profile object. 

you can create a list of 
```ts
             const props: {
                name: keyof IConnectionProfile;
                type: "string" | "boolean" | "number";
            }[] = [
                {
                    name: "server",
                    type: "string",
                },
            ];

            let connectionProfile = {
            };
            for (const prop of props) {
                switch (prop.type) {
                    case "string":
                        connectionProfile[prop.name] = "";
                        break;
                    case "boolean":
                        connectionProfile[prop.name] = false;
                        break;
                    case "number":
                        connectionProfile[prop.name] = 0;
                        break;
                    default:
                        break
            }
```

aasimkhan30 (approved):
I have added an alternate code suggestion. However, if possible try to use the api for sts to get all property names. That might make the code simpler.

src/mssqlProtocolHandler.ts
Benjin: Yeah, we should swap to that and remove the guesswork here.  @lewis-sanchez 

Here's a place we're using that STS call: https://github.com/microsoft/vscode-mssql/blob/main/src/connectionconfig/connectionDialogWebviewController.ts#L529

src/mssqlProtocolHandler.ts
```diff
@@ -0,0 +1,267 @@
...
+     * @returns The connection information object or undefined if the query is empty.
+     */
+    private readProfileFromArgs(query: string): IConnectionInfo | undefined {
+        if (!query) {
+            return undefined;
+        }
+
+        const args = new URLSearchParams(query);
+
+        const connectionString = args.get("connectionString") ?? undefined;
```
Benjin, line 77: Is there a reason we can't/shouldn't just map directly to the properties in the IConnectionInfo interface? That way, we have one centralized spot for validating which properties are allowed with which others instead of duplicating logic that reasons about, say, authentication type.

I think the right move here is to just loop over `args.keys()` and if present, set the associated properties in `IConnectionInfo`.

----

If we can't do that for whatever reason and need to grab properties, let's discuss.  We should at least try to employ the connection capabilities STS API: https://github.com/microsoft/vscode-mssql/blob/main/src/connectionconfig/connectionDialogWebviewController.ts#L529

src/controllers/mainController.ts
```diff
@@ -336,6 +337,24 @@ export default class MainController implements vscode.Disposable {
                 providerInstance,
             );
 
+            const self = this;
+            const uriHandler: vscode.UriHandler = {
+                handleUri(uri: vscode.Uri): vscode.ProviderResult<void> {
+                    if (self.isExperimentalEnabled) {
```
Benjin, line 343: Why put the `isExperimentalEnabled` check inside the handler, instead of outside like some of the other registrations?  I think inside makes sense if we have two different implementations instead of swapping between no implementation and something.

src/mssqlProtocolHandler.ts
```diff
@@ -0,0 +1,267 @@
...
+
+    /**
+     * Handles the given URI and returns connection information if applicable. Examples of URIs handled:
+     * - vscode://ms-mssql.mssql/connect?server=myServer&database=dbName&user=sa&authenticationType=SqlLogin
+     * - vscode://ms-mssql.mssql/connect?connectionString=Server=myServerAddress;Database=myDataBase;User Id=myUsername;Password=myPassword;
+     *
+     * @param uri - The URI to handle.
+     * @returns The connection information or undefined if not applicable.
+     */
+    public handleUri(uri: vscode.Uri): IConnectionInfo | undefined {
```
Benjin, line 29: Can you add a test to validate that this checks and extracts the correct IConnectionInfo?

src/mssqlProtocolHandler.ts
Benjin: If we keep this logic here in any capacity, we definitely need to add tests.  But at first glance, I think we should simplify this away.

PR #18065, opened 2024-09-19
Fixing pre-commit hooks and running yarn localization in precommit

aasimkhan30:
The `prepare` command is a reserved `npm` command that runs automatically after `yarn install`. I'm configuring Husky within the prepare script so that whenever someone runs yarn install, Husky will set up the pre-commit hook. This pre-commit hook runs `yarn lint` and `yarn localization` before allowing code to be committed. While this might slow down the commit process slightly, it ensures that code checks are performed upfront, saving time compared to waiting for checks to pass after submitting a pull request.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,2 @@
 yarn lint
+yarn localization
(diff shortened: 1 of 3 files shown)


caohai (approved):
This is great!

PR #18075, opened 2024-09-25
Fixing table designer bugs.

aasimkhan30:
I am bring a bunch of styling and functional fixes to table designer UI

## Fixing issue deep linking from issues view
Added issues deep linking from the issues list. Previously it wasn't working for more complicated issues but now all issues are properly deep linked.

## Result Pane Script As Create: 
Added option to copy script and added ability to maximize the panel size. Resizing works exactly as vscode terminal resize logic.

Before: 
image
After:
image

## Result Pane Issues: 

Switched from tabular form to a cleaner list based approach. Clicking to list takes you the problematic element. Panel resize toggle is also visible here.

Before:
image
After:
image

## Added ability to edit table 
Changed the table name to editable option at the top of table designer

Before:
image
After:
image

## Moved `About table` tab to last and renamed it to `General`

Before:
image
After:
image

## Made tables more compact and added color scheming to make them look cleaner

Before:
image
image

## Drag and reorder rows have a better visual indicator (blue line to show where it will dropped)
https://github.com/user-attachments/assets/ab969ba0-1f16-44a4-bb44-d5089e9f4b60

## Better error indication for error on table
Inputs
image
Dropdowns
image

## Cleaner Properties view
Before:
image
After:
image

## Cleanup Publish Screen
Added confirmation button to before update database becomes available and also changed the styling for markdown report.
Before:
image
After:
image


src/reactviews/pages/TableDesigner/designerPropertiesPane.tsx
```diff
@@ -92,108 +114,176 @@ export const DesignerPropertiesPane = () => {
...
+                                                    (i) =>
+                                                        (group === "General" &&
+                                                            !i.group) ||
+                                                        group === i.group,
+                                                )
+                                                .filter(
+                                                    (item) =>
+                                                        item.showInPropertiesView !==
+                                                        false,
+                                                )
```
Benjin, line 187: When the code is indented this far, it gets annoying to read.  Some ideas to clean it up:

* Pull something (Accordion or AccordionPanel, maybe) out into a helper function/component.
* increase the prettier pagewidth
* pull the `.map()` contents out into a helper function

src/reactviews/pages/TableDesigner/designerResultPane.tsx
```diff
@@ -119,67 +111,116 @@ export const DesignerResultPane = () => {
...
-        [
-            useTableColumnSizing_unstable({
-                columnSizingOptions: columnSizingOption,
-            }),
-        ],
-    );
-    const rows = getRows();
+        let elementToFocus: HTMLElement | undefined = undefined;
+        switch (issuePath.length) {
+            case 1:
```
Benjin, line 155: Lots of magic constants here... What do these numbers represent?

At a minimum, they need comments, but making them named constants would be even better.

src/reactviews/pages/TableDesigner/designerTable.tsx
```diff
@@ -379,97 +343,170 @@ export const DesignerTable = ({
...
-                                    {column.renderHeaderCell()}
-                                </fluentui.TableHeaderCell>
+                                    {columnsDef.map((column, columnIndex) => {
+                                        const borderColor =
+                                            state?.provider.getErrorMessage([
+                                                ...componentPath,
+                                                row.rowId,
+                                                column.columnId,
+                                            ])
+                                                ? "1px solid var(--vscode-errorForeground)"
```
Benjin, line 478: Is this where the indication for an error on the table is?  A few design thoughts on that front:

* To make it look a little nicer, maybe it can have a slight border radius.
* Since the error border takes up 1px of space, maybe there should be a 1px invisible border if there's no error so that everything stays aligned, or the error border should be an inner border, or the non-error cells get an extra pixel of margin
* or maybe better yet: since the designerXYZ components are fields, set those fields into error states and let the fluent library handle rendering the error border natively for each control?

aasimkhan30: image
image

Updated styling

kburtram (approved):

PR #18083, opened 2024-09-25
Fixing bug where launching connection dialog from button can result in unpopulated webview

### LINKED ISSUE #18074: Connection dialog doesn't open when using 'Add connection' button
- MSSQL Extension Version:
- VSCode Version:
- OS Version:

Steps to Reproduce:

1. Open sql server OE 

3. Click on `Add connection` button at the top of the bar.
image

3. The connection dialog opens with error and none of the fields are visible.
image

This seems to happen on a newly installed extension with no existing connections

Benjin:
Fixes #18074 

When opening the connection dialog from a VS Code view button (like the "+"), the command args are populated with info about the source button.  After #18047, the args were assumed to be a valid connection info object, but needed to have a typeguard.


localization/l10n/bundle.l10n.json
```diff
@@ -99,9 +99,6 @@
   "Move Up": "Move Up",
   "Move Down": "Move Down",
   "Delete": "Delete",
-  "Error loading designer": "Error loading designer",
```
Benjin, line 102: Automatically got removed because they're duplicates (see lines 162, 164, and 168)

aasimkhan30 (approved):

PR #18086, opened 2024-09-26
Eslint improvements.

aasimkhan30:
1. Running eslint and localization in parallel during pre-commit hook for better perf.
2. Upgrading eslint to latest
3. Adding caching to eslint to make subsequent runs faster.  This also involved removing lint gulp task and using cli to get caching option.
4. Adding eslint cache file to ignore.
5. Re-enable prettier as error. It was set to warning leading to some files not being formatted correctly.
6. Reformatting those files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "build": "gulp build",
     "compile": "gulp ext:compile",
     "watch": "gulp watch",
-    "lint": "gulp lint",
+    "lint": "eslint --quiet --cache",
     "localization": "gulp ext:extract-localization-strings",
     "smoketest": "gulp ext:smoke",
     "test": "gulp test:cover --log",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -53,12 +53,13 @@
     "build": "gulp build",
     "compile": "gulp ext:compile",
     "watch": "gulp watch",
-    "lint": "gulp lint",
+    "lint": "eslint --quiet --cache",
```
Benjin, line 56: does `--fix` work with this, too?

src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -26,7 +26,10 @@ import {
 import { ConnectionOption } from "azdata";
 import { Logger } from "../models/logger";
 import VscodeWrapper from "../controllers/vscodeWrapper";
-import { ConnectionDialog as Loc, refreshTokenLabel } from "../constants/locConstants";
+import {
```
Benjin, line 29: Are there any code changes in these files, or is it just formatting?

Benjin (approved):

Benjin (commented):
ADO PR build is failing due to looking for the lint task; make sure to update that, too.

package.json
aasimkhan30: Yeah, fix should work as before.

src/connectionconfig/connectionDialogWebviewController.ts
aasimkhan30: Just prettier formatting.

PR #18096, opened 2024-09-27
Linting only staged files in precommit

aasimkhan30:
Since linting was slow and took a while to complete, we are now linting only the staged files in the precommit-hook. This is the same approach we followed in ADS. However, instead of writing a custom script to do that, I’m using [lint-staged](https://github.com/lint-staged/lint-staged?tab=readme-ov-file#configuration).

Here is the custom script we had in ADS (forked from vscode): https://github.com/microsoft/azuredatastudio/blob/main/build/hygiene.js


caohai (approved):
This makes a lot of sense and we have the Github CI to catch repo wide linting issues (if any).

PR #18101, opened 2024-10-01
Improving Azure server loading experience and performance

Benjin:
* adds loading spinners for subscription and server calls
* switches server call to load from all subscriptions in parallelish (not awaiting until all promises have started)
* adds shortcut to immediately load servers for a subscription if the user selects that subscription before it's been populated in its turn

Note: there's a known bug where the user is prevented from typing in the "server" combobox; will fix in a later iteration/PR: #18102

### DIFF
--- a/src/reactviews/pages/ConnectionDialog/connectionFormPage.tsx
+++ b/src/reactviews/pages/ConnectionDialog/connectionFormPage.tsx
@@ -12,7 +12,6 @@ import { IConnectionDialogProfile } from "../../../sharedInterfaces/connectionDi
 import { ConnectButton } from "./components/connectButton.component";
 import { locConstants } from "../../common/locConstants";
 import { AdvancedOptionsDrawer } from "./components/advancedOptionsDrawer.component";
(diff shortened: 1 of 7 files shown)


src/reactviews/pages/ConnectionDialog/connectionFormPage.tsx
```diff
@@ -63,9 +62,6 @@ export const ConnectionFormPage = () => {
                     {locConstants.connectionDialog.advancedSettings}
                 </Button>
                 <div className={formStyles.formNavTrayRight}>
-                    <TestConnectionButton
```
aasimkhan30, line 66: Is there a reason why we are removing this?

aasimkhan30 (approved):

src/reactviews/pages/ConnectionDialog/connectionFormPage.tsx
Benjin: It's not wired up to anything yet, so I'm removing it now until I have a chance to actually implement it.

PR #18124, opened 2024-10-01
Added Query Plan to results pane

laurennat:
queryPlanInResultsPane


src/queryResult/queryResultWebViewController.ts
```diff
@@ -60,17 +72,105 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+                while (await exists(`plan${counter}.sqlplan`, folder)) {
+                    counter += 1;
+                }
+                filename = vscode.Uri.joinPath(
+                    folder,
+                    `plan${counter}.sqlplan`,
+                );
+            } else {
+                filename = vscode.Uri.joinPath(folder, "plan.sqlplan");
+            }
```
Benjin, line 106: Can you add some comments here explaining what the logic is intending to do?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -60,17 +72,105 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+                sqlPlanContent: this._executionPlanContents,
+                executionPlan: this.state.executionPlan,
+                executionPlanGraphs: this.state.executionPlanGraphs,
+            };
+        });
+        this.registerReducer("saveExecutionPlan", async (state, payload) => {
+            let folder = vscode.Uri.file(homedir());
+            if (await exists("Documents", folder)) {
+                folder = vscode.Uri.file(path.join(folder.path, "Documents"));
+            }
```
Benjin, line 92: .sqlplan files aren't documents, so let's not put them there.  I think homedir is fine as a default location.

src/queryResult/queryResultWebViewController.ts
```diff
@@ -60,17 +72,105 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+                await vscode.workspace.fs.writeFile(
+                    saveUri,
+                    Buffer.from(payload.sqlPlanContent),
+                );
+            }
+
+            return state;
+        });
+        this.registerReducer("showPlanXml", async (state, payload) => {
+            const planXmlDoc = await vscode.workspace.openTextDocument({
```
Benjin, line 127: Is the xml that we get back from Tools Service already formatted?  It would be nice to have it formatted in a readable way when we display or save it.

src/queryResult/queryResultWebViewController.ts
```diff
@@ -60,17 +72,105 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+            const saveUri = await vscode.window.showSaveDialog({
+                defaultUri: filename,
+                filters: {
+                    "SQL Plan Files": ["sqlplan"],
+                },
+            });
+
+            if (saveUri) {
+                // Write the content to the new file
+                await vscode.workspace.fs.writeFile(
```
Benjin, line 118: since nothing is going to use the result from this call, we don't need to block and it can be `void vscode.workspace.fs.writeFile()` instead of `async`

src/queryResult/queryResultWebViewController.ts
```diff
@@ -60,17 +72,105 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+
+            return state;
+        });
+        this.registerReducer("showPlanXml", async (state, payload) => {
+            const planXmlDoc = await vscode.workspace.openTextDocument({
+                content: payload.sqlPlanContent,
+                language: "xml",
+            });
+
+            await vscode.window.showTextDocument(planXmlDoc);
```
Benjin, line 132: can this be `void` instead of `async`?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -60,17 +72,105 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+
+            return state;
+        });
+        this.registerReducer("showQuery", async (state, payload) => {
+            await this.untitledSqlDocumentService.newQuery(payload.query);
```
Benjin, line 137: can this be `void` instead of `async`?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -60,17 +72,105 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+
+            return state;
+        });
+        this.registerReducer("updateTotalCost", async (state, payload) => {
+            this.state.totalCost += payload.totalCost;
```
Benjin, line 142: The arg is called `totalCost`, which makes it sound like this should be an assignment instead of an addition.  Should this be `this.state.totalCost = payload.totalCost` or `this.state.totalCost += addedCost`?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -60,17 +72,105 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+    public addQueryResultState(uri: string, isExecutionPlan?: boolean): void {
+        let currentState = {
             value: "",
             messages: [],
             tabStates: {
                 resultPaneTab: qr.QueryResultPaneTabs.Messages,
             },
             uri: uri,
-        });
+            isExecutionPlan: isExecutionPlan ?? false,
```
Benjin, line 159: instead, you can set a default value for `isExecutionPlan` in the param

src/queryResult/queryResultWebViewController.ts
```diff
@@ -87,4 +187,47 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+                this.state.executionPlanGraphs =
+                    this.state.executionPlan.graphs;
+                this.state.loadState = ApiStatus.Loaded;
+                this.state.totalCost = this.calculateTotalCost();
+            } catch (e) {
+                this.state.loadState = ApiStatus.Error;
+                this.state.errorMessage = e.toString();
+            }
+        }
+        this.updateState();
```
Benjin, line 219: Does state still need to be updated if `this.state.executionPlan === undefined`?  Should `updateState()` be inside the if block?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -87,4 +187,47 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+    }
+
+    private calculateTotalCost(): number {
+        let sum = 0;
+        for (const graph of this.state.executionPlanGraphs!) {
```
Benjin, line 224: Doing `executionPlanGraphs!` without confirming that it's defined makes me nervous.  What about having an `if (undefined) return 0;` before the summation?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -87,4 +187,47 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+            sum += graph.root.cost + graph.root.subTreeCost;
```
Benjin, line 225: what's the difference between cost and subTreeCost?  I don't know this bit of code.

src/reactviews/pages/ExecutionPlan/executionPlanPage.tsx
```diff
@@ -33,9 +33,16 @@ const useStyles = makeStyles({
     },
 });
 
-export const ExecutionPlanPage = () => {
+interface ExecutionPlanPageProps {
+    context?: any;
```
Benjin, line 37: is there a better type that can use here?

src/reactviews/pages/ExecutionPlan/highlightExpensiveOperations.tsx
```diff
@@ -40,13 +40,19 @@ interface HighlightExpensiveOperationsProps {
     executionPlanView: any;
     setExecutionPlanView: any;
     setHighlightOpsClicked: any;
+    context?: any;
```
Benjin, line 43: Can this be something better than `any`? (same for the others)

src/reactviews/pages/ExecutionPlan/iconMenu.tsx
```diff
@@ -46,6 +46,7 @@ interface IconStackProps {
     setHighlightOpsClicked: any;
     setPropertiesClicked: any;
     query: any;
+    context?: any;
```
Benjin, line 49: same question about `any` here (and the rest of these props)

src/reactviews/pages/ExecutionPlan/properties.tsx
```diff
@@ -109,14 +109,16 @@ const useStyles = makeStyles({
 interface PropertiesPaneProps {
     executionPlanView: any;
     setPropertiesClicked: any;
+    context?: any;
```
Benjin, line 112: more specific type here?

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -200,6 +213,14 @@ export const QueryResultPane = () => {
                     >
                         {MESSAGES}
                     </Tab>
+                    {metadata.resultSetSummary && metadata.isExecutionPlan && (
+                        <Tab
+                            value={qr.QueryResultPaneTabs.ExecutionPlan}
+                            key={qr.QueryResultPaneTabs.ExecutionPlan}
+                        >
+                            {"Query Plan"}
```
Benjin, line 221: localize

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -25,10 +25,12 @@ import { useContext, useEffect, useRef, useState } from "react";
 import { OpenFilled } from "@fluentui/react-icons";
 import { QueryResultContext } from "./queryResultStateProvider";
 import * as qr from "../../../sharedInterfaces/queryResult";
+import * as ep from "../ExecutionPlan/executionPlanInterfaces";
```
Benjin, line 28: import these with more useful names please

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -173,6 +176,16 @@ export const QueryResultPane = () => {
 
     const gridRef = useRef<SlickGridHandle>(null);
 
+    const getExecutionPlanGraphs = async (contents: string) => {
+        if (!state?.state?.executionPlanGraphs!.length) {
```
Benjin, line 180: can you rename the context var to actually be `context`?  `state.state` is a bit misleading to read

Benjin: Also, if `context` is undefined, this entire component can just return immediately instead of getting down to here.

Benjin: Also also, is executionPlanGraphs ever undefined?  if so, that should be checked (and returned?) if it's undefined.  If not, maybe that property shouldn't be nullable.

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -256,6 +276,11 @@ export const QueryResultPane = () => {
                                             var columnLength =
                                                 metadata?.resultSetSummary
                                                     ?.columnInfo?.length;
+                                            if (metadata?.isExecutionPlan) {
+                                                getExecutionPlanGraphs(
+                                                    r.rows[0][0].displayValue,
```
Benjin, line 281: can you add comments explaining what the zeros achieve?  something like `// first column's name`

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -355,6 +380,16 @@ export const QueryResultPane = () => {
                         </Table>
                     </div>
                 )}
+                {metadata.tabStates!.resultPaneTab ===
```
Benjin, line 383: why a null assertion here?

src/reactviews/pages/QueryResult/queryResultStateProvider.tsx
```diff
@@ -41,7 +42,45 @@ const QueryResultStateProvider: React.FC<QueryResultContextProps> = ({
...
+                        });
+
+                        if (webViewState?.state.executionPlan) {
+                            return Promise.reject(
+                                new Error("Execution plan is undefined"),
+                            );
+                        }
+
+                        return Promise.resolve(
+                            webViewState!.state!.executionPlan!,
```
Benjin, line 59: So many assertions... These are generally code smells for me - if you have to assert because the type checker didn't catch something, it's _usually_ a good indication that the types aren't quite set up right, or that you should add explicit checks for null/undefined  to help it along.  Not always, but usually.

src/sharedInterfaces/queryResult.ts
```diff
@@ -3,6 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { ApiStatus } from "../sharedInterfaces/webview";
+import * as ep from "../reactviews/pages/ExecutionPlan/executionPlanInterfaces";
```
Benjin, line 7: import name

src/sharedInterfaces/queryResult.ts
```diff
@@ -11,11 +14,27 @@ export enum QueryResultLoadState {
...
     setResultTab: (tabId: QueryResultPaneTabs) => void;
+    /**
+     * Gets the execution plan graph from the provider for a given plan file
+     * @param planFile file that contains the execution plan
+     */
+    getExecutionPlan(
+        planFile: ep.ExecutionPlanGraphInfo,
+    ): Thenable<ep.GetExecutionPlanResult>;
+
+    saveExecutionPlan(sqlPlanContent: string): void;
```
Benjin, line 25: add docstrings to the rest

src/models/sqlOutputContentProvider.ts
```diff
@@ -239,7 +250,10 @@ export class SqlOutputContentProvider {
                 await this.createWebviewController(uri, title, queryRunner);
             }
         } else {
-            this._queryResultWebviewController.addQueryResultState(uri);
+            this._queryResultWebviewController.addQueryResultState(
+                uri,
+                executionPlanOptions?.includeEstimatedExecutionPlanXml ?? false,
```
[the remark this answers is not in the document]
laurennat: I made an item for this here:  https://msdata.visualstudio.com/SQLToolsAndLibraries/_workitems/edit/3428410

src/sharedInterfaces/queryResult.ts
```diff
@@ -28,12 +47,35 @@ export interface QueryResultWebviewState {
     value?: string;
     messages: IMessage[];
     tabStates?: QueryResultTabStates;
+    isExecutionPlan?: boolean;
+    sqlPlanContent?: string;
+    executionPlan?: ep.GetExecutionPlanResult;
+    executionPlanGraphs?: ep.ExecutionPlanGraph[];
+    theme?: string;
```
caohai, line 54: iirc the webview state already has theme field

src/sharedInterfaces/queryResult.ts
```diff
@@ -28,12 +47,35 @@ export interface QueryResultWebviewState {
     value?: string;
     messages: IMessage[];
     tabStates?: QueryResultTabStates;
+    isExecutionPlan?: boolean;
```
caohai, line 50: Can we encapsulate execution plan related fields into one field like `executionPlanStates: ExecutionPlanStates` and put these fields under `ExecutionPlanStates`?

caohai:
I noticed that the column name under `Results` tab for query plan has a TODO tag and is not localized, I've included that fix in my PR here https://github.com/microsoft/vscode-mssql/pull/18131/commits/fa292acb4c09cc9273b26dc93bfbe4c8174552ea

src/sharedInterfaces/queryResult.ts
laurennat: this is a theme that specifically says whether it's dark or light, no just the hex code

src/queryResult/queryResultWebViewController.ts
laurennat: made an issue for this: https://github.com/microsoft/vscode-mssql/issues/18132

src/sharedInterfaces/queryResult.ts
caohai: https://github.com/microsoft/vscode-mssql/blob/main/src/reactviews/pages/QueryResult/queryResultUtils.ts
Here's how we use `context.theme` to provide different icons for dark and light themes.

src/reactviews/pages/ExecutionPlan/highlightExpensiveOperations.tsx
laurennat: I updated the context and executionPlanView type. the others are different variations of React.Dispatch<React.SetStateAction<type>, so I wasn't sure whether to update those

src/queryResult/queryResultWebViewController.ts
laurennat: both of these are se tby the call to sqltoolsservice- the first is the root node's cost, and the second is the calculated cost of the entire root node's subtree of nodes

src/reactviews/pages/QueryResult/queryResultPane.tsx
laurennat: I was mostly just following what they did for the other tabs, but if i'm understanding correctly, it's for the initial load in when the state might not be defined yet.

caohai:
Can you add/update the screenshot/gif with a multi-batch query like this?
```sql
select * from sys.objects
go
select * from sys.objects
go
```

src/reactviews/pages/ExecutionPlan/executionPlanPage.tsx
```diff
@@ -33,10 +34,18 @@ const useStyles = makeStyles({
...
+    context?: QueryResultState;
+}
+
+export const ExecutionPlanPage: React.FC<ExecutionPlanPageProps> = ({
+    context,
+}) => {
     const classes = useStyles();
-    const provider = useContext(ExecutionPlanContext);
-    const loadState = provider?.state?.loadState ?? ApiStatus.Loading;
+    console.log(context);
```
caohai, line 45: Do we want to keep this log for prod or is it just for debugging?

src/sharedInterfaces/queryResult.ts
```diff
@@ -11,11 +18,43 @@ export enum QueryResultLoadState {
 
 export interface QueryResultReactProvider {
     setResultTab: (tabId: QueryResultPaneTabs) => void;
+    /**
```
aasimkhan30, line 21: Maybe extract it out to a parent interface shared between the 2 webviews.
(execution plan and query result)

src/sharedInterfaces/queryResult.ts
```diff
@@ -27,12 +66,37 @@ export interface QueryResultWebviewState {
     resultSetSummaries: { [key: number]: ResultSetSummary };
     messages: IMessage[];
     tabStates?: QueryResultTabStates;
+    isExecutionPlan?: boolean;
+    executionPlanState: {
```
aasimkhan30, line 70: Why not make an interface out of it and share it with the regular plan webview. Let's try avoid duplicate types since they tend to get out of sync with each other.

src/sharedInterfaces/queryResult.ts
```diff
@@ -27,12 +66,37 @@ export interface QueryResultWebviewState {
...
+        sqlPlanContent?: string;
+        executionPlan?: GetExecutionPlanResult;
+        executionPlanGraphs?: ExecutionPlanGraph[];
+        theme?: string;
```
aasimkhan30, line 74: Theming variable is provided by vscodeWebviewProvider. Please try to use that instead. Things like theming, telemetry and localization should be kept centralized as any bug fixes made to them will be applied everywhere.

src/sharedInterfaces/queryResult.ts
```diff
@@ -27,12 +66,37 @@ export interface QueryResultWebviewState {
...
+        loadState?: ApiStatus;
+        errorMessage?: string;
+    };
 }
 
 export interface QueryResultReducers {
     setResultTab: {
         tabId: QueryResultPaneTabs;
     };
+    getExecutionPlan: {
```
aasimkhan30, line 85: Please centralize the reducers as well

src/reactviews/pages/QueryResult/queryResultStateProvider.tsx
```diff
@@ -41,7 +45,50 @@ const QueryResultStateProvider: React.FC<QueryResultContextProps> = ({
                             tabId: tabId,
                         });
                     },
+                    getExecutionPlan: function (
```
aasimkhan30, line 48: If these are copied from the execution plan provider, it will be good to have one implementation of it.

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -85,6 +87,7 @@ const RESULTS = l10n.t("Results");
 const MESSAGES = l10n.t("Messages");
 const TIMESTAMP = l10n.t("Timestamp");
 const MESSAGE = l10n.t("Message");
+const QUERY_PLAN = l10n.t("Query Plan");
```
aasimkhan30, line 90: Maybe use the locFile in `reactviews/common/locConstants.ts`

aasimkhan30: @caohai and @laurennat, FYI, this approach won’t work because the webview initially loads with default strings. It then requests the extension to retrieve the current language file and updates the strings accordingly. Since these are constant variables, they cannot be updated dynamically after initialization.

src/controllers/executionPlanWebviewController.ts
```diff
@@ -30,16 +29,18 @@ export class ExecutionPlanWebviewController extends ReactWebviewPanelController<
...
-                    vscode.ColorThemeKind.Dark
-                        ? "dark"
-                        : "light",
-                loadState: ApiStatus.Loading,
-                executionPlan: undefined,
-                executionPlanGraphs: [],
-                totalCost: 0,
+                executionPlanState: {
+                    sqlPlanContent: executionPlanContents,
+                    theme:
```
aasimkhan30, line 34: Please use the theme variable in vscodeWebviewContextProvider. This implementation already doesn't support high contrast theme.

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -372,6 +404,17 @@ export const QueryResultPane = () => {
...
                 )}
+                {metadata.tabStates!.resultPaneTab ===
+                    qr.QueryResultPaneTabs.ExecutionPlan &&
+                    Object.keys(metadata.resultSetSummaries).length > 0 && (
+                        <div
+                            id={"executionPlanResultsTab"}
+                            className={classes.queryResultContainer}
+                            style={{ height: "100%", minHeight: "300px" }}
+                        >
+                            <ExecutionPlanPage context={state} />
```
aasimkhan30, line 415: I don't think passing context in this way is the best design approach. It looks like you've special-cased the context variable in all the ExecutionPlan children components:
```
var context = context ?? useContext(ExecutionPlanProvider)
```
The purpose of using context is to avoid prop drilling (i.e., passing props down through multiple child components). A more effective way to handle this would be:
```
<ExecutionPlanContextProvider
   props={queryResultState.executionPlan}
>
   <ExecutionPlanPage />
</ExecutionPlanContextProvider>
```
By properly initializing all the props in the provider, you wouldn't need to modify the existing ExecutionPlan components.

src/reactviews/pages/QueryResult/queryResultPane.tsx
caohai: Thanks @aasimkhan30 ! I created a PR fixing this https://github.com/microsoft/vscode-mssql/pull/18136

src/controllers/executionPlanWebviewController.ts
```diff
@@ -104,7 +109,7 @@ export class ExecutionPlanWebviewController extends ReactWebviewPanelController<
 
             if (saveUri) {
                 // Write the content to the new file
-                await vscode.workspace.fs.writeFile(
+                vscode.workspace.fs.writeFile(
```
Benjin, line 112: if `writeFile` is async and we don't want to await it, then we should explicitly use the `void` keyword.

src/controllers/executionPlanWebviewController.ts
```diff
@@ -118,49 +123,58 @@ export class ExecutionPlanWebviewController extends ReactWebviewPanelController<
                 language: "xml",
             });
 
-            await vscode.window.showTextDocument(planXmlDoc);
+            vscode.window.showTextDocument(planXmlDoc);
```
Benjin, line 126: void keyword

src/controllers/executionPlanWebviewController.ts
```diff
@@ -118,49 +123,58 @@ export class ExecutionPlanWebviewController extends ReactWebviewPanelController<
...
 
             return state;
         });
         this.registerReducer("showQuery", async (state, payload) => {
-            await this.untitledSqlDocumentService.newQuery(payload.query);
+            this.untitledSqlDocumentService.newQuery(payload.query);
```
Benjin, line 131: void keyword

caohai:
@laurennat FYI you'd need to merge main to resolve the `pr-checks` failure.

src/reactviews/pages/ExecutionPlan/properties.tsx
```diff
@@ -117,7 +118,7 @@ export const PropertiesPane: React.FC<PropertiesPaneProps> = ({
 }) => {
     const classes = useStyles();
     const state = useContext(ExecutionPlanContext);
-    const executionPlanState = state?.state;
+    const theme = state!.theme;
```
aasimkhan30, line 121: NIT: fix type

src/reactviews/pages/ExecutionPlan/queryPlanSetup.ts
```diff
@@ -395,162 +395,179 @@ export function getCollapseExpandPaths() {
     };
 }
 
-export const save = (theme: string) => {
+// this function determines whether to use dark or light icons,
```
aasimkhan30, line 398: What's this? 

Why don't you do something like this? 

https://github.com/microsoft/vscode-mssql/blob/516b87fa66016d4ec92f3e2325dcc378fadb03fc/src/reactviews/pages/TableDesigner/designerResultPane.tsx#L307

You can get the theme from here
https://github.com/microsoft/vscode-mssql/blob/516b87fa66016d4ec92f3e2325dcc378fadb03fc/src/reactviews/pages/TableDesigner/designerResultPane.tsx#L307

You can set this variable like this.
https://github.com/microsoft/vscode-mssql/blob/516b87fa66016d4ec92f3e2325dcc378fadb03fc/src/reactviews/pages/TableDesigner/tableDesignerStateProvider.tsx#L185-L186

src/reactviews/pages/QueryResult/queryResultStateProvider.tsx
laurennat: made an issue to track: https://github.com/microsoft/vscode-mssql/issues/18186

aasimkhan30 (approved):

src/controllers/executionPlanWebviewController.ts
```diff
@@ -59,28 +53,30 @@ export class ExecutionPlanWebviewController extends ReactWebviewPanelController<
...
         this.registerReducer("saveExecutionPlan", async (state, payload) => {
             let folder = vscode.Uri.file(homedir());
-            if (await exists("Documents", folder)) {
-                folder = vscode.Uri.file(path.join(folder.path, "Documents"));
-            }
-
             let filename: vscode.Uri;
+
+            // make the default filename of the plan to be saved-
+            // start ad plan.sqlplan, then plan1.sqlplan, ...
```
Benjin, line 79: typo: "start ad" -> "start at"

src/controllers/executionPlanWebviewController.ts
```diff
@@ -118,49 +114,54 @@ export class ExecutionPlanWebviewController extends ReactWebviewPanelController<
...
-                this.state.loadState = ApiStatus.Loaded;
-                this.state.totalCost = this.calculateTotalCost();
-            } catch (e) {
-                this.state.loadState = ApiStatus.Error;
-                this.state.errorMessage = e.toString();
-            }
+    private async createExecutionPlanGraphs(content: string) {
+        const planFile: ep.ExecutionPlanGraphInfo = {
+            graphFileContent: content,
+            graphFileType: ".sqlplan",
```
Benjin, line 142: Can you move the file extension into a constants file and swap all the literal strings?

src/controllers/mainController.ts
```diff
@@ -1525,11 +1548,13 @@ export default class MainController implements vscode.Disposable {
...
+
             await self._outputContentProvider.runQuery(
                 self._statusview,
                 uri,
                 querySelection,
                 title,
+                self._executionPlanOptions,
             );
         } catch (err) {
             console.warn(`Unexpected error running query : ${err}`);
```
Benjin, line 1560: I know it's not your change, but can you change `${err}` to `${getErrorMessage(err)}`?

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -176,6 +176,7 @@ export class QueryHistoryProvider implements vscode.TreeDataProvider<any> {
                 uri,
                 undefined,
                 title,
+                {},
```
Benjin, line 179: Can you add a comment saying which parameter this is meant to be, since it isn't otherwise apparent?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -78,17 +89,104 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+                payload.xmlPlan,
+            ];
+            return state;
+        });
+        this.registerReducer("saveExecutionPlan", async (state, payload) => {
+            let folder = vscode.Uri.file(homedir());
+            let filename: vscode.Uri;
+
+            // make the default filename of the plan to be saved-
+            // start ad plan.sqlplan, then plan1.sqlplan, ...
```
Benjin, line 111: same typo ad -> at

src/queryResult/queryResultWebViewController.ts
```diff
@@ -78,17 +89,104 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+            let counter = 1;
```
Benjin, line 112: Is this the same logic as in executionPlanWebviewController?  Can you pull it into a shared helper/utils method?  Something like

`getUniquePath(fileBaseName: string, extension: string, path: Uri): Uri`

src/queryResult/queryResultWebViewController.ts
```diff
@@ -78,17 +89,104 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+                    totalCost: this.state.executionPlanState.totalCost,
+                },
+            };
+        });
     }
 
-    public addQueryResultState(uri: string): void {
-        this._queryResultStateMap.set(uri, {
+    public addQueryResultState(uri: string, isExecutionPlan?: boolean): void {
+        let currentState = {
```
Benjin, line 172: const

src/queryResult/queryResultWebViewController.ts
```diff
@@ -105,4 +203,60 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+        service: UntitledSqlDocumentService,
+    ): void {
+        this.untitledSqlDocumentService = service;
+    }
+
+    private async createExecutionPlanGraphs(
+        state: qr.QueryResultWebviewState,
+        xmlPlans: string[],
+    ) {
+        let newState = {
```
Benjin, line 221: const?

src/reactviews/pages/ExecutionPlan/executionPlanPage.tsx
```diff
@@ -33,10 +34,17 @@ const useStyles = makeStyles({
...
+    context?: QueryResultState;
+}
+
+export const ExecutionPlanPage: React.FC<ExecutionPlanPageProps> = ({
+    context,
+}) => {
     const classes = useStyles();
-    const provider = useContext(ExecutionPlanContext);
-    const loadState = provider?.state?.loadState ?? ApiStatus.Loading;
+    const provider = context ? context : useContext(ExecutionPlanContext);
```
Benjin, line 45: I thought the point of `useContext` was to remove the need to pass the context in as a property.  @aasimkhan30?

No need to change for this PR since this isn't just a nit, but something we should circle back to.

Benjin, line 45: after this, you can probably just do

```
if (provider === undefined)
return undefined;
```

and avoid many of the null propagation `?`s below

src/reactviews/common/utils.ts
```diff
@@ -20,3 +26,14 @@ export function formatString(str: string, ...args: any[]): string {
     }
     return result;
 }
+
+export const getVscodeTheme = (theme: Theme) => {
```
Benjin, line 30: should this be `getVscodeThemeType()` instead?

Benjin, line 30: For functions that aren't just local functions, it's more readable to use the normal `export function getVscodeThemeType() { }` syntax over `export const getVsCodeThemeType = ...`

Benjin, line 30: add a docstring to this function

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -269,6 +278,32 @@ export const QueryResultPane = () => {
...
+            }
+        };
+
+        if (
+            metadata &&
+            metadata.resultSetSummaries &&
+            metadata.executionPlanState.xmlPlans &&
+            Object.keys(metadata.resultSetSummaries).length > 0 &&
+            Object.keys(metadata.resultSetSummaries).length ===
+                metadata.executionPlanState.xmlPlans.length
```
Benjin, line 301: Can you add some comments to this long boolean statements to say what each is meaning to check since the prop drilling doesn't match up (and therefore isn't immediately intuitive)?

src/reactviews/pages/QueryResult/queryResultPane.tsx
Benjin: The issue where state may not be defined yet _should_ be fixed now with [this](https://github.com/microsoft/vscode-mssql/pull/18147/files#diff-c03515c44e865bf8d0d3ed19221a87854595b4505ce58559e3e5f0d9f13d8f59) change, but no need to go through and remove all the checks in this PR.

Unresolving this comment for vis, but you can just mark it as resolved again.

src/queryResult/queryResultWebViewController.ts
Benjin: Tracking in #18186

Benjin (approved):

PR #18126, opened 2024-10-02
Fixing vsix not running

aasimkhan30:
Apparently lint-staged is causing some issues in the created vsix file. I am not able to find the root cause but removing it makes the vsix file work.

https://msdata.visualstudio.com/SQLToolsAndLibraries/_build/results?buildId=143623244&view=results


gulpfile.js
```diff
@@ -70,6 +70,34 @@ const cssLoaderPlugin = {
 	},
 };
 
+gulp.task('ext:lint-staged', async (done) => {
+	const fix = (argv.fix === undefined) ? false : true;
+	const {stdout, stderr} = await exec("git diff --cached --name-only");
+	const stagedFiles = stdout.split('\n').filter(file => file.endsWith('.ts') || file.endsWith('.tsx'));
```
Benjin, line 76: .css as well?

gulpfile.js
```diff
@@ -70,6 +70,34 @@ const cssLoaderPlugin = {
 	},
 };
 
+gulp.task('ext:lint-staged', async (done) => {
```
Benjin, line 73: Odd that lint-staged affected this... and yet packaging the vsix locally was fine.

Can you add a comment on this gulp task explaining why this ended up having to be done manually rather than using the lint-staged package for posterity?

gulpfile.js
aasimkhan30: eslint doesn't work on css. I will have to get prettier running here somehow.

caohai (approved):

kburtram (approved):

PR #18133, opened 2024-10-03
Fixing vsix and updating vsce

aasimkhan30:
1. Introduced `signal-exit` as  `lockfile` depends on the package and wasn't working
1. `tar` was upgraded as I was getting `Class extends value #<Object> is not a constructor or null\n\tat Object.<anonymous>` from the previous version.
1. `yallist` was required by the latest tar but was not found in node_modules. 

I don't exactly know why these indirect dependencies are not being found in the node_modules folder. 

Also updated vsce to latest version. The package is moved from `vsce` to `@vscode\vsce`

### DIFF
--- a/.github/workflows/daily-build-and-test.yml
+++ b/.github/workflows/daily-build-and-test.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: '18'
+          node-version: '22'
 
       - name: Set up xvfb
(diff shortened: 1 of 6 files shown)


caohai:
We are still using the `lint-staged` pkg, right? Does the tar issue come from us using a different version in our code already?

Benjin (approved):

aasimkhan30:
@caohai that's correct. The original issues of vsix not working was due to tar package not working.

Issue #18054, opened 2024-09-19
Running a failed query doesn't focus on messages tab


cssuh:
- MSSQL Extension Version: main
- VSCode Version: 1.93.1
- OS Version: Sonoma 14.6.1

Steps to Reproduce:

1. Open a query editor a run a query that fails (due to syntax error or some other reason).
2. The (react) query results pane will not open the messages pane automatically to display the error message.

PR #18143, opened 2024-10-07
Fix query result tab behavior without results

caohai:
This PR fixes https://github.com/microsoft/vscode-mssql/issues/18054
We should switch to the result tab only if results are available.

### DIFF
--- a/src/models/sqlOutputContentProvider.ts
+++ b/src/models/sqlOutputContentProvider.ts
@@ -482,9 +482,17 @@ export class SqlOutputContentProvider {
                                 uri,
                             );
                         vscode.commands.executeCommand("queryResult.focus");
+                        const tabState =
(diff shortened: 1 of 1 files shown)


src/models/sqlOutputContentProvider.ts
```diff
@@ -482,9 +482,17 @@ export class SqlOutputContentProvider {
                                 uri,
                             );
                         vscode.commands.executeCommand("queryResult.focus");
+                        const tabState =
```
aasimkhan30, line 485: NIT: Maybe disable/hide the results tab when there are no results.

caohai: It's already hidden when no result is available, I just missed correctly setting the tab state here.
image

PR #18147, opened 2024-10-08
Add prompt for enabling Trust Server Certificate

### LINKED ISSUE #18110: Potential Improvement: Suggest enabling "Trust Server Certificate" when adding a new non-prod connection
Repro: connect to a sql-sever docker container without checking "Trust Server Certificate"
image

In ADS, we have this:
image

Benjin:
Fixes #18110 

Only this specific connection error results in a dialog because it has actionable buttons; all others still display via the banner

Wording taken directly from ADS

image


src/reactviews/common/vscodeWebviewProvider.tsx
```diff
@@ -158,7 +162,10 @@ export function VscodeWebviewProvider<State, Reducers>({
                 }}
                 theme={theme}
             >
-                {children}
+                {
+                    // don't render webview unless necessary dependencies are initialized
+                    isInitialized() && children
```
Benjin, line 167: I made this change after chatting with @aasimkhan30.  It results in all webviews having defined state (with the default value set by each controller) instead of having to sprinkle undefined checks in each component that uses context.

src/reactviews/pages/ConnectionDialog/components/trustServerCertificateDialog.component.tsx
```diff
@@ -0,0 +1,81 @@
...
+                        <Link href={connectionCertValidationReadMoreUrl}>
+                            {locConstants.connectionDialog.readMore}
+                        </Link>
+                    </DialogContent>
+                    <DialogActions>
+                        <Button
+                            appearance="primary"
+                            onClick={() => {
+                                context.cancelTrustServerCertDialog();
+                                context.formAction({
```
Benjin, line 55: this should also check if it's in connectionString input mode, and add/set the property there

aasimkhan30 (approved):

PR #18169, opened 2024-10-09
Add ability to filter subscriptions for browsing Azure in connection dialog

### LINKED ISSUE #18107: Connection Dialog missing ability to adjust subscription filters in-box
Currently requires Azure Resources extension or manual editing of preferences JSON

image

or

```json"
azureResourceGroups.selectedSubscriptions": [
    "<tenant ID>/<subscription ID>"
],
```

Benjin:
Fixes #18107

The filter uses the `azureResourceGroups.selectedSubscriptions` config key (hardcoded in the VS Code Azure APIs), and takes the form `<tenantId>/<subscriptionId>`.  Therefore, if a user has filter their visible/available subscriptions, they'll get a consistent experience across all the Azure-related extensions.

image


src/reactviews/pages/ConnectionDialog/components/connectButton.component.tsx
```diff
@@ -29,7 +29,6 @@ export const ConnectButton = ({
                 connectionDialogContext.state.connectionStatus ===
                 ApiStatus.Loading
             }
-            shape="square"
```
Benjin, line 32: The vscode aesthetic is shifting to match that of the Fluent default controls with their gently-rounded corners, so I've swapped the buttons to match.

localization/i18n/chs/constants/localizedConstants.i18n.json
aasimkhan30, line 1: Why are i18n files added?

aasimkhan30 (approved):

localization/i18n/chs/constants/localizedConstants.i18n.json
Benjin: They showed up in my added files, so my default was to include them.  Should I remove them?  If they shouldn't be checked in, then I'll update `.gitignore` to exclude them.

PR #18173, opened 2024-10-09
Adding error handling to publish errors in table designer.

### LINKED ISSUE #18166: Table Designer: Error not reported during publishing, causing infinite loading "Publishing Changes"
Steps to repro:
1. Connect to a db and start table designer to add a new table
2. Shutdown the db before clicking "Update Database"
3. Click "Update Database"

No error is reported, and the spinner loads infinitely.
image
I was able to close the publishing dialog by pressing Esc multiple times, not sure if this intended design as well.

We can also see the same behavior for any script error during publishing (I was able to repro the same issue with memory-optimized table).

ADS reports back the error.
image

aasimkhan30:
image

This fixes #18166

### DIFF
--- a/src/reactviews/common/locConstants.ts
+++ b/src/reactviews/common/locConstants.ts
@@ -133,6 +133,8 @@ export class LocConstants {
                     args: [objectType],
                     comment: ["{0} is the object type"],
                 }),
+            retryUpdatingDatabase: l10n.t("Retry updating database"),
+            backToPreview: l10n.t("Back to preview report"),
         };
     }
 
--- a/src/reactviews/common/locConstants.ts
+++ b/src/reactviews/common/locConstants.ts
@@ -134,6 +134,7 @@ export class LocConstants {
                     comment: ["{0} is the object type"],
                 }),
             backToPreview: l10n.t("Back to preview"),
+            copy: l10n.t("Copy"),
         };
     }
 
--- a/src/reactviews/common/locConstants.ts
+++ b/src/reactviews/common/locConstants.ts
@@ -133,8 +133,7 @@ export class LocConstants {
                     args: [objectType],
                     comment: ["{0} is the object type"],
                 }),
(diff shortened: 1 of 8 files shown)


caohai:
Can we also add something like a `Copy Details` button that allow user to copy the error message?

cssuh:
Ignore the tag from my PR, it was an accident

src/reactviews/common/locConstants.ts
```diff
@@ -132,6 +133,8 @@ export class LocConstants {
                     args: [objectType],
                     comment: ["{0} is the object type"],
                 }),
+            retryUpdatingDatabase: l10n.t("Retry updating database"),
```
Benjin, line 136: The text for these buttons are a little long.  Maybe something simpler like "Retry" and "Back to preview".

@jdesai-24

Benjin:
That `( ! )` icon is huge.  What about displaying the error with a <MessageBar> like this dialog does?  #18147

src/reactviews/pages/TableDesigner/designerChangesPreviewButton.tsx
```diff
@@ -130,13 +146,48 @@ export const DesignerChangesPreviewButton = () => {
                 </>
             );
         }
+        if (metadata?.apiState?.publishState === LoadState.Error) {
```
Benjin, line 149: these LoadState cases feel like they're screaming for refactoring the dialog content out into helper functions/components.

src/reactviews/common/locConstants.ts
jdesai-24: @Benjin, I agree, we can add "Retry" and "Back to preview". Do we need a close button, I think a cross icon would suffice. 
For the error message, can we break down the message into different points which will be easier to read?

caohai (approved):

lewis-sanchez (approved):

PR #18177, opened 2024-10-10
Fixing a bunch of Table designer bugs. 

### LINKED ISSUE #18159: Table Designer initial values
When the table designer opens, automatically place the cursor in the ```Table Name``` textbox. This will prompt the user to immediately enter a table name, making it clear that naming the table is the first step. Instead of using a default value for the table name (e.g., "NewTable"), leave the ```Table Name``` textbox empty to encourage users to provide a meaningful name from the start.

For the **first column** that appears in the designer, update the default name from **"column_1"** to **"id."** Ensure that the Primary Key option is selected by default for this **"id"** column. Add the Is Identity option immediately after Primary Key, and leave it enabled by default with Identity ```Seed = 1``` and ```Identity Increment = 1```.

Move the ```Schema``` dropdown list directly after the Table Name field. Table Name and Schema should be the first two fields the user completes before moving on to the column management section.

Tagging @jdesai-24 for visibility

### LINKED ISSUE #18160: Move Identity property to column management
Move the ```Is Identity``` option right after Primary Key, in the main panel. Leave the identity enabled by default for the initial column (```Id```). Use Identity ```Seed = 1``` and ```Identity Increment = 1``` by default.

Tagging @jdesai-24 for visibility

### LINKED ISSUE #18161: Column properties panel updates
Start by collapsing all the sections by default. Only leave the ```General``` section expanded, as it is more likely that developers would like to modify values just by using that section. Also consider updating the text for the ```Advanced Type``` in the General section to ```Data Type```. 

The ```Precision``` and ```Scale``` options in the General section are limited only to Numeric data types. It is very confusing to show these options even if my data type is `int` or `varchar`. Only show these options when a relevant data type (such as decimal or numeric) is chosen.

Also question, why ```Default Constraint Name``` is greyed out? If this option were a modifiable action from the user, I would rather group it with Primary Key and create a new section for clarity (Collapsed by default).

### LINKED ISSUE #18163: Repurpose the General tab in Table designer
Rename the ```General``` tab to ```Advanced options``` in the Table designer main panel. Use this tab for settings that are not immediately needed when creating the table, such as ```Description```, ```Type```, ```Memory Optimized``` and ```System Versioning``` (removed **enabled** from the label).

aasimkhan30:
Fixes #18163 
1. Renamed `General` tab to `Advanced Options` in the Table designer main panel. 
image

2. Moved `table name` and `schema` out of the tab to the top of the designer 
image

3. Renamed `System versioning Enabled` to `System Versioning`.
image

Fixes #18160 
1. Moved Is Identity to columns table
2. Leaving identity enabled for first column and also making it primary key
3. Naming first column 'Id
image

Fixes #18161 
1. Leaving all panels collapsed except 'General'
image
2. Chaning Advanced Type prop to Data Type
image
3. Precision and Scale are only visible for applicable data types now
int
image
numberic/decimal
image
4. Hiding not 'Grayed out' props
image

Fixes #18159 partially
Things fixed: 
1. Focusing on table name property first.
2. First column name is 'id'. Primary key and identity is enabled by default.
3. Schema is below table name field.
image

Things not fixed:
Leaving default name Empty as it is not so simple to do from the backend (Will require DacFx update).

Companion PR in STS: https://github.com/microsoft/sqltoolsservice/pull/2404

### DIFF
--- a/src/reactviews/pages/TableDesigner/designerPropertiesPane.tsx
+++ b/src/reactviews/pages/TableDesigner/designerPropertiesPane.tsx
@@ -260,7 +260,7 @@ export const DesignerPropertiesPane = () => {
                 />
             </div>
             <div className={classes.stack}>
-                <Accordion multiple collapsible defaultOpenItems={groups}>
+                <Accordion multiple collapsible defaultOpenItems={[groups[0]]}>
                     {data &&
                         groups?.map((group) => {
                             const groupItems = parentTableProperties
--- a/src/reactviews/pages/TableDesigner/designerPropertiesPane.tsx
+++ b/src/reactviews/pages/TableDesigner/designerPropertiesPane.tsx
@@ -21,6 +21,7 @@ import { DesignerDropdown } from "./designerDropdown";
 import { DesignerTable } from "./designerTable";
 import {
     CheckBoxProperties,
+    DesignerDataPropertyInfo,
     DesignerTableProperties,
     DropDownProperties,
     InputBoxProperties,
@@ -113,7 +114,10 @@ export const DesignerPropertiesPane = () => {
         );
     }
 
-    const renderAccordionItem = (group: string | undefined) => {
+    const renderAccordionItem = (
+        group: string | undefined,
+        groupItem: DesignerDataPropertyInfo[],
+    ) => {
         if (!group) {
             return undefined;
         }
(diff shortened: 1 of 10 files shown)


src/reactviews/pages/TableDesigner/designerPropertiesPane.tsx
```diff
@@ -122,93 +126,81 @@ export const DesignerPropertiesPane = () => {
...
-                                                key={`${group}-${item.propertyName}`}
-                                            />
-                                        );
-                                }
-                            })}
+                        {groupItem.map((item) => {
+                            if (!data) {
+                                return undefined;
+                            }
+                            const modelValue = data![item.propertyName];
```
Benjin, line 133: is this `!` still necessary after the check on L130?

Benjin (approved):

lewis-sanchez (approved):

PR #18181, opened 2024-10-11
Cleans up open protocol handler and adds unit tests

lewis-sanchez:
This PR cleans up the open protocol handler logic and adds unit tests.

The cleaned up logic continues to handle MSSQL protocol URIs as before. A connection string like:

`vscode://ms-mssql.mssql/connect?server=(localdb)\MssqlLocalDb&database=msdb&user=lewis&authenticationType=SqlLogin&profileName=profile1`

The URi can be entered into a browser. Once submitted, VS Code will prompt the user if they would like to open the URI:
image

Clicking on "Open" in the prompt will open the connection dialog for users to make any additional adjustments to connect to their target connection.
image

### DIFF
--- a/test/unit/mssqlProtocolHandler.test.ts
+++ b/test/unit/mssqlProtocolHandler.test.ts
@@ -186,4 +186,45 @@ suite("MssqlProtocolHandler Tests", () => {
         assert.equal(connInfo.connectTimeout, 15);
         assert.isTrue(connInfo.trustServerCertificate);
     });
+
+    test("handleUri - with connect command and query with invalid bool value for trust server cert - trust server cert is false and parses valid params", async () => {
+        const connInfo = await mssqlProtocolHandler.handleUri(
+            Uri.parse(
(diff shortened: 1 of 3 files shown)


test/unit/mssqlProtocolHandler.test.ts
```diff
@@ -0,0 +1,189 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as TypeMoq from "typemoq";
+import { assert } from "chai";
+import { MssqlProtocolHandler } from "../../src/mssqlProtocolHandler";
+import SqlToolsServiceClient from "../../src/languageservice/serviceclient";
+import { Uri } from "vscode";
+
+suite("MssqlProtocolHandler Tests", () => {
```
aasimkhan30, line 12: Thanks for adding tests. 

Maybe add tests that test invalid values for some attributes. That might increase the code coverage for the protocolhandler file.

-Invalid boolean value
-Invalid numberic value
-Invalid parameter name

aasimkhan30 (approved):
Looks good.

PR #18192, opened 2024-10-12
Support query result auto-switch based on editor context

caohai:
This PR adds automatic result switch based on the current active editor.
If the current active editor has an associated result, the query result panel automatically switches to that result, otherwise a placeholder message is displayed.
Users can hide the panel by clicking on the "Click here to hide this panel" link.
Oct-11-2024 16-59-53


src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -284,6 +282,8 @@ export const QueryResultPane = () => {
             metadata &&
             // makes sure result sets are defined
             metadata.resultSetSummaries &&
+            // makes sure this is an execution plan
+            metadata.isExecutionPlan &&
```
caohai, line 286: @laurennat lmk if this looks correct

aasimkhan30:
Looks good to me. Since we are adding adding hide panel view, should we also add a show query panel view in the editor?

aasimkhan30 (approved):

localization/l10n/bundle.l10n.json
```diff
@@ -256,6 +256,8 @@
   "Save as CSV": "Save as CSV",
   "Save as Excel": "Save as Excel",
   "Save as JSON": "Save as JSON",
+  "No result found for the current active editor, please run a query or switch to another editor.": "No result found for the current active editor, please run a query or switch to another editor.",
```
Benjin, line 259: How about:

> No result found for the active editor; please run a query or switch to another editor.

removes the redundancy between "current" and "active" and switches the comma to a semicolon since they're different clauses.

localization/l10n/bundle.l10n.json
```diff
@@ -256,6 +256,8 @@
...
+  "Click here to hide this panel": "Click here to hide this panel",
```
Benjin, line 260: I think just "Hide this panel" is good.  It should be obvious that the user can click there because it's a link.

caohai:
> Looks good to me. Since we are adding adding hide panel view, should we also add a show query panel view in the editor?

@aasimkhan30 If the user runs another query / switch to a tab with query result, the panel will show up automatically. Otherwise they can always manually toggle panel visibility by running command `workbench.action.togglePanel`
image

Are you talking about having an extra button somewhere to allow user to explicitly show the panel?

src/controllers/reactWebviewBaseController.ts
```diff
@@ -239,6 +239,17 @@ export abstract class ReactWebviewBaseController<State, Reducers>
                 return undefined;
             }
         };
+
+        this._webviewRequestHandlers["executeCommand"] = async (message) => {
+            if (!message?.command) {
+                console.log("No command provided to execute");
```
Benjin, line 245: if there's no command, should this `return` instead of attempting to execute a commandless command?

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -300,7 +300,22 @@ export const QueryResultPane = () => {
         }
     });
 
-    return (
+    return !metadata || !hasResultsOrMessages(metadata) ? (
+        <div>
+            <div>{locConstants.queryResult.noResultMessage}</div>
```
Benjin, line 305: Based on the screenshot, this pane need some margins so that the text has some breathing room.

Benjin: @jdesai-24 for some design feedback here.

Benjin (approved):

localization/l10n/bundle.l10n.json
caohai: updated

localization/l10n/bundle.l10n.json
caohai: updated

src/reactviews/pages/QueryResult/queryResultPane.tsx
caohai: Yeah the styling of this placeholder message definitely needs some improvement, I'll create an issue tracking this.

cssuh (commented):
I tried out the branch, and if I switch back and forth between two query editors rapidly, occasionally it flashes the "No result found" view before loading the correct results.  Any idea why this might be happening?

cssuh (approved):

src/controllers/reactWebviewBaseController.ts
caohai: forgot about this when I changed from `throw` to `console.log` :), updated

src/reactviews/pages/QueryResult/queryResultPane.tsx
caohai: https://github.com/microsoft/vscode-mssql/issues/18193

caohai:
> I tried out the branch, and if I switch back and forth between two query editors rapidly, occasionally it flashes the "No result found" view before loading the correct results. Any idea why this might be happening?

will look into it next week

aasimkhan30:
@caohai , I was thinking about an icon in the toolbar that lets you do that.

caohai:
> @caohai , I was thinking about an icon in the toolbar that lets you toggle visibility of the query pane.

makes sense, I think this can be useful when a user has the result panel hidden then switches to a tab with result, they can toggle the panel by themselves which is less aggressive. I'll look into this option and send a separate PR.

PR #18222, opened 2024-10-15
Improving first-run prompt flow

### LINKED ISSUE #18210: Too many notifications on first launch after update
Ideally we should only have one notification, but we have two.  I would suggest removing release notes prompt, but since the notes will have details on the new features, we probably do want that prompt.  Maybe the "Learn More" link can substitute for what we want to accomplish with the Release Notes prompt.  

Also, for wording the prompt, I like the more casual tone, but I might rephrase from "GUI-features" to like "The MSSQL for VS Code extension is introducing new modern data development features!  Would you like to enable them?"  I guess the "GUI" part seems more incidental to me.

Image

Benjin:
Fixes #18210

vscode now reloads the extensions immediately after the users accepts the prompt to enable the new UX

Users should now see a maximum of one prompt at launch, with the prompt to enable new features taking the place of the typical release notes prompt if users have neither enabled or suppressed it.  Users will be prompted to enable the new UX each launch until they pick one, but the release notes prompt will (still) only be shown once per extension update.

Once the new UX is GAed, the new prompt will be removed and the behavior will be reverted.

### DIFF
--- a/src/constants/locConstants.ts
+++ b/src/constants/locConstants.ts
@@ -643,10 +643,15 @@ export let executionPlan = l10n.t("Execution Plan");
 export let scriptCopiedToClipboard = l10n.t("Script copied to clipboard");
 export let copied = l10n.t("Copied");
 
-export let enableRichExperiencesPrompt = l10n.t(
-    "The MSSQL extension has some new GUI-based experiences! Would you like to enable them?",
-);
-export let enableRichExperiences = l10n.t("Enable Rich Experiences");
+export function enableRichExperiencesPrompt(learnMoreUrl: string) {
+    return l10n.t({
+        message:
+            "The MSSQL for VS Code extension is introducing new modern data development features! Would you like to enable them? [Learn more]({0})",
+        args: [learnMoreUrl],
+        comment: ["{0} is a url to learn more about the new features"],
+    });
+}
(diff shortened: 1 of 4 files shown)


aasimkhan30 (approved):

src/constants/locConstants.ts
```diff
@@ -643,10 +643,15 @@ export let executionPlan = l10n.t("Execution Plan");
 export let scriptCopiedToClipboard = l10n.t("Script copied to clipboard");
 export let copied = l10n.t("Copied");
 
-export let enableRichExperiencesPrompt = l10n.t(
-    "The MSSQL extension has some new GUI-based experiences! Would you like to enable them?",
-);
-export let enableRichExperiences = l10n.t("Enable Rich Experiences");
+export function enableRichExperiencesPrompt(learnMoreUrl: string) {
```
aasimkhan30, line 646: NIT: Maybe rename the variable to `enableModernFeatures` to match the prompt.

Benjin: I like that better.  "rich" always felt a bit buzwordy to me

Benjin: I think I'm going to not change it because it's not user-facing (outside of manually editing the config json), and is kind of a sprawling (albeit simple find-replace) change.

Issue #18212, opened 2024-10-15
Exception executing query with new result grid


kburtram:
Steps to Reproduce:

1. Run a couple queries and interact with messages pane and switch around in the UI, like hiding the panel/showing it again.  I'll try to narrow down the steps exactly
3. Run another query

Expected: All the queries work

Actual: I hit the below error.

```
[ExtensionListenerError] Extension 'ms-mssql.mssql' FAILED to handle event: TypeError: Cannot read properties of undefined (reading 'webview'): Error: [ExtensionListenerError] Extension 'ms-mssql.mssql' FAILED to handle event: TypeError: Cannot read properties of undefined (reading 'webview')
    at gh.value (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:156:35184)
    at T.B (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:29:732)
    at T.C (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:29:802)
    at T.fire (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:29:1019)
    at gh.value (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:155:61240)
    at T.B (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:29:732)
    at T.fire (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:29:950)
    at nw.acceptDocumentsAndEditorsDelta (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:121:11570)
    at nw.$acceptDocumentsAndEditorsDelta (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:121:10040)
    at Zb.S (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:31:113896)
    at Zb.Q (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:31:113676)
    at Zb.M (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:31:112765)
    at Zb.L (file:///c:/vscode/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:31:111870)
[...]

PR #18227, opened 2024-10-16
Port: bump STS to bring in a few fixes #18224

caohai:
Changes: https://github.com/microsoft/sqltoolsservice/compare/5.0.20241010.3...5.0.20241015.3
image

PR to main: https://github.com/microsoft/vscode-mssql/pull/18224
This PR fixes: #18212


Benjin:
We're reverting MDS to 5.1.6 because 5.2.0 introduced a bug during query execution (see #18212 for info).  Rather than rolling forward to the hotfixed 5.2.2, we decided to return to the known stable build for now.

Benjin (approved):

Issue #18233, opened 2024-10-16
Add a button to toggle query result visibility


caohai:
https://github.com/microsoft/vscode-mssql/pull/18192#issuecomment-2408284005

Issue #18213, opened 2024-10-15
Multiple result sets in a batch don't work correctly


kburtram:
Steps to Reproduce:

1. Run multiple queries in a batch, such as:

```SQL
 SELECT 1

 SELECT 2

 SELECT 3
```

Expected: Three resultsets with 1, 2, 3

Actual: Only the final result set.  If we add GO between the statements everything works as expected.

![Image](https://github.com/user-attachments/assets/6f2d6a56-1502-47bb-abd1-a05dd7fdc219)

kburtram:
@caohai this one looks pretty impactful too.  Could you please take a look when you have a chance?

caohai:
@kburtram Sure this is a bug I missed, will send out a fix this afternoon

Issue #18140, opened 2024-10-07
Query Result grid header styling


caohai:
Currently the header row has the same styling as the data rows.

ADS:
<img width="1051" alt="image" src="https://github.com/user-attachments/assets/fe4c7359-f979-487f-82f7-d321ed393650">

caohai:
1. header background color
2. font and font size (ration between text size and cell size)
3. scroll bar

Issue #18059, opened 2024-09-19
Remove lines between messages in Messages tab of Query Result


caohai:
and general styling/font

PR #18234, opened 2024-10-16
Fix a few Query Result bugs:

caohai:
This PR fixes https://github.com/microsoft/vscode-mssql/issues/18213
Screenshot 2024-10-15 at 22 16 22

This PR fixes: https://github.com/microsoft/vscode-mssql/issues/18233 by adding a button next to the Run Query button to reveal the query result panel, this button only shows up when the current language is sql and the `queryResult' View is not visible.

Oct-15-2024 22-31-28

This PR fixes https://github.com/microsoft/vscode-mssql/issues/18140
Before:
Screenshot 2024-10-15 at 22 35 23
Now:
Screenshot 2024-10-15 at 22 19 36

This PR fixes https://github.com/microsoft/vscode-mssql/issues/18193, still not the best-looking message but better the initial version
Before:
image
Now:
Screenshot 2024-10-15 at 22 20 46

This PR fixes https://github.com/microsoft/vscode-mssql/issues/18059
Before:
image

Now:
Screenshot 2024-10-15 at 22 25 36


src/queryResult/queryResultWebViewController.ts
```diff
@@ -259,6 +258,19 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
         return res;
     }
 
+    public addResultSetSummary(
+        uri: string,
+        resultSetSummary: qr.ResultSetSummary,
+    ) {
+        let state = this.getQueryResultState(uri);
```
Benjin, line 265: does this need an early return in case there's no query result state for that uri?

package.json
```diff
@@ -309,6 +309,11 @@
           "when": "editorLangId == sql && resourcePath in mssql.runningQueries",
           "group": "navigation@2"
         },
+        {
+          "command": "mssql.revealQueryResultPanel",
+          "when": "editorLangId == sql && view.queryResult.visible == false",
```
Benjin: does this `when` clause also need to check `config.mssql.enableRichExperiences`?

src/queryResult/queryResultWebViewController.ts
caohai: getQueryResultState will throw if no state is found for the uri

package.json
Benjin: hah, you fixed it already

caohai: I've actually updated this clause, you might need to refresh the page.

cssuh (approved):

Benjin (approved):

Issue #18254, opened 2024-10-16
After viewing a query plan, the XML in the results pane is flickering


Benjin:
Repro:
1. Run a query (e.g. `select * from sys.objects`)
2. Click the "view estimated query plan" button in the top-right corner of the editor window

_"Query Plan" tab in results pane is automatically selected_

3. Select "Results" tab in results pane

Expected:
Results pane shows the XML with a link that opens the plan in its own tab

Reality: 
* link on XML is not able to be clicked
* query results webview eventually freezes - switching to another pane then back to Query Result shows a blank panel, and vscode must be reloaded
https://github.com/user-attachments/assets/35c74e30-ca05-445c-8d7e-8677b5946f7b

PR #18258, opened 2024-10-17
Fix query plan flickering bug

laurennat:
Fixed the flickering bug by preventing grid refreshes if the query is an execution plan

issue: https://github.com/microsoft/vscode-mssql/issues/18254

eventually, I'll move the state change responsible for the query plan rendering to the extension side rather than the client side: https://github.com/microsoft/vscode-mssql/issues/18257

### DIFF
--- a/src/reactviews/pages/QueryResult/queryResultPane.tsx
+++ b/src/reactviews/pages/QueryResult/queryResultPane.tsx
@@ -285,7 +285,7 @@ export const QueryResultPane = () => {
 
     const renderGridPanel = () => {
         const grids = [];
-        gridRefs.current.forEach((r) => r?.refreshGrid());
+        // gridRefs.current.forEach((r) => r?.refreshGrid());
         let totalResultCount = 0;
         Object.values(metadata?.resultSetSummaries ?? []).forEach((v) => {
             totalResultCount += Object.keys(v).length;
--- a/src/reactviews/pages/QueryResult/queryResultPane.tsx
+++ b/src/reactviews/pages/QueryResult/queryResultPane.tsx
@@ -285,7 +285,9 @@ export const QueryResultPane = () => {
 
     const renderGridPanel = () => {
         const grids = [];
-        // gridRefs.current.forEach((r) => r?.refreshGrid());
+        if (!metadata?.isExecutionPlan) {
+            gridRefs.current.forEach((r) => r?.refreshGrid());
+        }
         let totalResultCount = 0;
         Object.values(metadata?.resultSetSummaries ?? []).forEach((v) => {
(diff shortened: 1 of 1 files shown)


caohai:
Can you add a gif/screenshot to the PR description?

aasimkhan30:
@laurennat , since this is a PR to the main branch do not use `fixes #issue` as it will close the issue. Just mention the issue like `Issue: #issue`. We still need the issue to be opened until this gets ported to the release.

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -285,7 +285,9 @@ export const QueryResultPane = () => {
 
     const renderGridPanel = () => {
         const grids = [];
-        gridRefs.current.forEach((r) => r?.refreshGrid());
+        if (!metadata?.isExecutionPlan) {
+            gridRefs.current.forEach((r) => r?.refreshGrid());
```
aasimkhan30, line 289: Also the correct fix will be to prevent refreshing the grid all the times. IIUC, this just means that the grid is not refreshed repeatedly when the query plan is present. This won't fix the issue when query plan isn't present, right? 

@caohai , what do you think?

Benjin: Any idea why the query plan being refreshed constantly resulted in flickering, but a regular result set didn't flicker?  (not so say it didn't also have the issue you're investigating, but I'm curious why it manifested differently)

Benjin:
Don't forget to address the second part of bug #18254:
> Results pane shows the XML **with a link that opens the plan in its own tab**

When I check out this branch, Looks like the link still isn't able to be opened.

src/reactviews/pages/QueryResult/queryResultPane.tsx
caohai: IIRC the repeated refreshing issue only happens when the query plan is present.

Benjin (changes requested):
Don't forget to address the second part of bug #18254:
> Results pane shows the XML **with a link that opens the plan in its own tab**

When I check out this branch, Looks like the link still isn't able to be opened.

laurennat:
the link stuff was always meant to be in a later PR, but i can make an issue for it @Benjin

src/reactviews/pages/QueryResult/queryResultPane.tsx
laurennat: execution plans only load after reading the resulting xml showplan of the query. therefore, it updates the state once the results are loaded, which causes a rendering loop if the grid gets refreshed

Benjin (approved):

PR #18264, opened 2024-10-17
Fix event subscription in webview

caohai:
Currently every time a page is re-rendered, the new page would subscribe to the events, but we are not cleaning up the stale subscribers.
This PR removes these stale subscribers by adding a `callerId` field and thus each caller can only have one subscription for one method.


Benjin (approved):
LGTM, but @aasimkhan30 is the one that really matters here

PR #18269, opened 2024-10-17
Fixing nps telemetry

aasimkhan30:
Adding ids to the questions to make them easier to query


Benjin (approved):
Thanks!

src/reactviews/pages/UserSurvey/userSurveyPage.tsx
```diff
@@ -108,29 +108,23 @@ export const UserSurveyPage = () => {
                             <NSATQuestion
                                 key={index}
                                 question={question}
-                                onChange={(d) =>
-                                    onAnswerChange(question.label, d)
-                                }
+                                onChange={(d) => onAnswerChange(question.id, d)}
```
Benjin, line 111: Looks like the NSAT and NPS values are numbers packed into strings, then everything goes into additonalProperties.  Those should be split out into `additionalMeasures` so that we can compute over the responses.  Easiest way would be to parse the known numeric questions into numbers at emission time.

But if you wanted to go even further and fix this properly, then ideally <NPSQuestion> and <NSATQuestion> would have numeric callbacks and values instead of strings, user answers becomes `Record<string, string | number>` and then the actual telemetry emission call divides the entries in the record between props and measures depending on their type.

Benjin (changes requested):

Benjin (approved):

PR #18271, opened 2024-10-17
Add min size & fix resize behavior

### LINKED ISSUE #18248: Result sets have no minimum vertical size when displayed in query results pane
Repro: run a query with 10+ result sets, e.g.

```
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
select * from sys.objects
```

Expected:
query pane is still usable.  each result set has a minimum height (that's at least the size of the sidebar buttons), and a scrollbar appears to scroll to see the overflow

Reality:
help.
Image

cssuh:
Adds a min size to query results grids & fixes resize behavior

Fixes #18248

minSize

### DIFF
--- a/src/reactviews/pages/QueryResult/queryResultPane.tsx
+++ b/src/reactviews/pages/QueryResult/queryResultPane.tsx
@@ -226,7 +226,6 @@ export const QueryResultPane = () => {
         batchId: number,
         resultId: number,
         gridCount: number,
-        totalResultCount: number,
     ) => {
         const divId = `grid-parent-${batchId}-${resultId}`;
         return (
@@ -325,7 +324,7 @@ export const QueryResultPane = () => {
         ) {
             var batch = metadata?.resultSetSummaries[i];
             for (let j = 0; j < Object.keys(batch ?? []).length; j++) {
-                grids.push(renderGrid(i, j, count, totalResultCount));
+                grids.push(renderGrid(i, j, count));
                 count++;
             }
         }
--- a/src/reactviews/pages/QueryResult/queryResultPane.tsx
+++ b/src/reactviews/pages/QueryResult/queryResultPane.tsx
@@ -100,6 +100,15 @@ const useStyles = makeStyles({
     },
 });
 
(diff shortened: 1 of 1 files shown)


src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -100,6 +100,15 @@ const useStyles = makeStyles({
     },
 });
 
+const MIN_GRID_HEIGHT = 273; // Minimum height for a grid
```
caohai, line 103: Is this a random pick or did we use the same number in ADS?

cssuh: Using the same number as ADS

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -201,16 +230,7 @@ export const QueryResultPane = () => {
...
         const divId = `grid-parent-${batchId}-${resultId}`;
         return (
-            <div
-                id={divId}
-                className={classes.queryResultContainer}
-                style={{
-                    height:
-                        totalResultCount === 1
-                            ? "100%"
-                            : (100 / totalResultCount).toString() + "%",
```
caohai, line 211: You can remove totalResultCount parameter since it's no longer used.
image

caohai:
If I run a query with 2 result sets then run a single result set query in the same editor, the single result set doesn't take the entire space 
image

cssuh:
> If I run a query with 2 result sets then run a single result set query in the same editor, the single result set doesn't take the entire space image

fixed in latest update

caohai (approved):

Issue #18240, opened 2024-10-16
Search Azure subscriptions not working


amthomas46:
Please add "Area" label and repro steps before submitting.

I logged in with Azure (was prompted by the extension to log in) but typing doesnt narrow the subscriptions

![Image](https://github.com/user-attachments/assets/2d99be48-953a-4a11-aa4c-98ce6237d56c)

amthomas46:
i used the other filtering option but nothing shows up for RGs
![Image](https://github.com/user-attachments/assets/e1666f15-ab35-4570-967d-a14d8624c969)

Benjin:
After some investigation, I found that I was getting an error message saying that the resource filter argument I was passing for server-side processing was invalid... for some subscriptions, but not all (or even most).  Curiously, it was consistently the same subscriptions that would return the error, and I didn't notice any pattern to it.  My mind went to permissions, but the query for _all_ resources under that sub is consistently working fine for all subs so that doesn't make sense.

PR #18274, opened 2024-10-18
Changes to client-side azure resource filtering

Benjin:
Addresses #18240

I found that I was getting an error message saying that the resource filter argument I was passing for server-side processing was invalid... for some subscriptions, but not all (or even most). Curiously, it was consistently the same subscriptions that would return the error, and I didn't notice any pattern to it. My mind went to permissions, but the query for all resources under that sub is consistently working fine for all subs so that doesn't make sense.

In any case, I switched to doing the filtering client-side in order to solve the issue.

### DIFF
--- a/src/connectionconfig/azureHelper.ts
+++ b/src/connectionconfig/azureHelper.ts
@@ -5,9 +5,19 @@
 
 import * as vscode from "vscode";
 import { l10n } from "vscode";
-import { ConnectionDialogWebviewState } from "../sharedInterfaces/connectionDialog";
-import { getErrorMessage } from "../utils/utils";
-import { VSCodeAzureSubscriptionProvider } from "@microsoft/vscode-azext-azureauth";
+import {
+    AzureSqlServerInfo,
+    ConnectionDialogWebviewState,
+} from "../sharedInterfaces/connectionDialog";
+import { getErrorMessage, listAllIterator } from "../utils/utils";
+import {
+    AzureSubscription,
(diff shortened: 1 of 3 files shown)


aasimkhan30 (approved):

src/connectionconfig/azureHelper.ts
```diff
@@ -97,3 +107,76 @@ export async function getQuickPickItems(
...
+    sub: AzureSubscription,
+): Promise<AzureSqlServerInfo[]> {
+    const result: AzureSqlServerInfo[] = [];
+    const client = new ResourceManagementClient(
+        sub.credential,
+        sub.subscriptionId,
+    );
+
+    // for some subscriptions, supplying a `resourceType eq 'Microsoft.Sql/servers/databases'` filter to list() causes an error:
+    // > invalid filter in query string 'resourceType eq "Microsoft.Sql/servers/databases'"
```
caohai, line 125: Is it possible to fall back to client-side filtering only if Azure API side filtering is failing with this error>

Benjin: Yeah, I thought about doing that, but in my light testing on a subscription with 460+ total resources where only 75 were databases, the performance difference is negligible.  There seems to be a high fixed cost of querying the Azure management plane, and the number of resources returned doesn't appear to impact it as much, so filtering doesn't have a big benefit.

I have a telemetry point for how long it takes to load the Azure info that we can look at to see if it's worth doing perf testing to optimize further here.

PR #18281, opened 2024-10-18
Improving the azure browse filter dropdown logic

Benjin:
Refining and consolidating the logic to have more intuitive default logic.  The overall logic is now:
* `server` always selects the first one in the list to make sure it's always populated, unless the user has explicitly cleared it
* all others default to no filter
* dropdowns are automatically scoped by any upstream filters that are set (subscriptions/resource groups/locations/servers)
* if a value is already selected and the upstream filters change, the selection stays if it's still in-filter or is cleared if not

Also adding several telemetry more points around Azure load times and errors

Note: because contents of subscriptions are loaded asynchronously, different servers may end up being selected each time depending on which subscriptions finish first.  I thought it would be more confusing to swap the selection around to (e.g.) the first alphabetically as new entries loaded, so /shrug.  The other dropdowns don't have that issue because they aren't set to always make a selection when available.

### DIFF
--- a/src/connectionconfig/connectionDialogWebviewController.ts
+++ b/src/connectionconfig/connectionDialogWebviewController.ts
@@ -1211,6 +1211,7 @@ export class ConnectionDialogWebviewController extends ReactWebviewPanelControll
                 );
             } else {
                 state.loadingAzureServersStatus = ApiStatus.Loading;
+                state.azureServers = [];
                 this.updateState();
 
                 const startTime = Date.now();
--- a/src/connectionconfig/connectionDialogWebviewController.ts
+++ b/src/connectionconfig/connectionDialogWebviewController.ts
@@ -1138,6 +1138,8 @@ export class ConnectionDialogWebviewController extends ReactWebviewPanelControll
                         string[] | undefined
                     >(azureSubscriptionFilterConfigKey) !== undefined;
 
+            const startTime = Date.now();
+
             this._azureSubscriptions = new Map(
                 (await auth.getSubscriptions(shouldUseFilter)).map((s) => [
                     s.subscriptionId,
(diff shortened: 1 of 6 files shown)


Benjin:
I am _not_ looking to get this in for the October release, hence draft mode

src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -1138,14 +1138,16 @@ export class ConnectionDialogWebviewController extends ReactWebviewPanelControll
                         string[] | undefined
                     >(azureSubscriptionFilterConfigKey) !== undefined;
 
+            const startTime = Date.now();
```
aasimkhan30, line 1141: Maybe we should create a helper function to track time. It would return another function that, when called, automatically measures the elapsed time and sends the action event with the mentioned properties.

aasimkhan30 (approved):

kburtram (approved):
Looks like a solid improvement to me!

src/reactviews/common/comboboxHelper.ts
```diff
@@ -0,0 +1,57 @@
...
+    /** callback to set the displayed value (not guaranteed to be valid if the user has manually typed something) */
+    setValue: (v: string) => void,
+    /** list of valid options */
+    optionList: string[],
+    /** behavior for choosing the default selected value */
+    defaultSelectionMode: DefaultSelectionMode = DefaultSelectionMode.AlwaysSelectNone,
+) {
+    // if there is no current selection or if the current selection is no longer in the list of options (due to filter changes),
+    // then select the only option if there is only one option,
+    // or either the first option or none if there are multiple options, depending on how shouldSelectIfAny is set
```
Benjin, line 30: update comment text

src/connectionconfig/connectionDialogWebviewController.ts
Benjin: I like this thought.  I'll make an issue to add some util stuff like that.

Issue #18289, opened 2024-10-21
Sizing of results grid is incorrect


kburtram:
Steps to Reproduce:

1. Run a query

Expected: Scollbars only show if needed, there are only two scrollbars when needed for multiple resultsets, the toolbar buttons are always visible.

Actual: Double scrollbars, and toolbar buttons are obsured.

![Image](https://github.com/user-attachments/assets/df7c9909-e610-4dd5-b5cb-80227760f8fe)

kburtram:
I merged the fix to release/1.25, which fixes the main issue here but there are a couple details we still need to address for Nov.

1. The scrollbars are different widths on different platforms, causing extra padding on mac, which we should tighten-up.
2. The boundary condition of when to adjust for the double-scrollbar, i.e., `gridCount * MIN_GRID_HEIGHT >= availableHeight`, is off by some pixels, causing the toolbar to be obscured within a narrow range while resizing.

cssuh:
Thanks @kburtram, I'll take it from here!

dombarnes:
Not sure how related, but the ability to double click a column resize bar and it stretch to the maximum width needed seems to be gone. Since you can only stretch it to the width of the available width in the query results window, it makes resizing later columns very difficult.

cssuh:
Opened https://github.com/microsoft/vscode-mssql/issues/18383 and https://github.com/microsoft/vscode-mssql/issues/18382 for tracking remaining items

PR #18290, opened 2024-10-21
Reverting "Add min size & fix resize behavior (#18271) (#18272)" and "Port Fix a few Query Result bugs: (#18234) #18238"

Benjin:
Reverting #18272 (a port of #18271) and #18238 (a port of #18234) due to #18289


aasimkhan30 (approved):

Benjin:
Reverting just this commit wasn't enough to resolve the issue; checking a few more...

Benjin:
I'd like to chat more about the bugs that the were fixed in #18234 before merging this PR into `release/1.25`.

kburtram (changes requested):
It seems like we're reverting more than we'd like to here.  Let's investigate and see if there's a more targeted fix for this issue.

PR #18296, opened 2024-10-22
Adding some webview panel and user survey tests. 

aasimkhan30:
Will add more user survey tests in a separate PR after this https://github.com/microsoft/vscode-mssql/pull/18295 PR gets merged in.

### DIFF
--- a/src/controllers/reactWebviewBaseController.ts
+++ b/src/controllers/reactWebviewBaseController.ts
@@ -105,18 +105,17 @@ export abstract class ReactWebviewBaseController<State, Reducers>
 
     protected _getHtmlTemplate() {
         const nonce = getNonce();
-        const baseUrl =
-            this._getWebview()
-                .asWebviewUri(
-                    vscode.Uri.joinPath(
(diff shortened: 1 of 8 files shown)


Benjin (approved):

src/controllers/reactWebviewBaseController.ts
```diff
@@ -105,18 +105,17 @@ export abstract class ReactWebviewBaseController<State, Reducers>
...
-                        this._context.extensionUri,
-                        "out",
-                        "src",
-                        "reactviews",
-                        "assets",
-                    ),
-                )
-                .toString() + "/";
+
+        const baseUrl = this._getWebview().asWebviewUri(
```
caohai, line 109: Are we using this `baseUrl` somewhere else? Just curious if there's something special about this change.

aasimkhan30: The base URL sets the root of the webview. Any resource referred by the webview will be relative to this base url. We use it to load images within the webviews.

PR #18303, opened 2024-10-22
Added Query Plan Telemetry

laurennat:
Added  a telemetry point for opening query plans

### DIFF
--- a/src/controllers/executionPlanWebviewController.ts
+++ b/src/controllers/executionPlanWebviewController.ts
@@ -12,6 +12,11 @@ import UntitledSqlDocumentService from "../controllers/untitledSqlDocumentServic
 import { exists } from "../utils/utils";
 import { homedir } from "os";
 import { sqlPlanFileExtension } from "../constants/constants";
+import {
+    TelemetryActions,
+    TelemetryViews,
+} from "../sharedInterfaces/telemetry";
+import { sendActionEvent } from "../telemetry/telemetry";
 
 export class ExecutionPlanWebviewController extends ReactWebviewPanelController<
     ep.ExecutionPlanWebviewState,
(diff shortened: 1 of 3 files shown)


src/queryResult/queryResultWebViewController.ts
```diff
@@ -327,6 +331,10 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
             } catch (e) {
                 newState.loadState = ApiStatus.Error;
                 newState.errorMessage = e.toString();
+                sendActionEvent(
```
Benjin, line 334: `sendErrorEvent` for errors; also add metrics

src/queryResult/queryResultWebViewController.ts
```diff
@@ -305,6 +305,10 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
         state: qr.QueryResultWebviewState,
         xmlPlans: string[],
     ) {
+        sendActionEvent(
```
Benjin, line 308: `sendActionEvent`) (and `sendErrorEvent()`) both have optional props that can be set: `additionalMeasurements` (most numeric metrics) and `additionalProperties` (everything else).

Capturing something like the number of plans as a metric here would be good so we can see if, say, people are frequently viewing plans for 25 queries at once, or if we always get errors when it hits some threshold.

src/controllers/executionPlanWebviewController.ts
```diff
@@ -141,6 +146,10 @@ export class ExecutionPlanWebviewController extends ReactWebviewPanelController<
     }
 
     private async createExecutionPlanGraphs(content: string) {
+        sendActionEvent(
+            TelemetryViews.QueryPlan,
+            TelemetryActions.OpenQueryPlan,
+        );
```
Benjin, line 152: sending an event at the beginning can be useful, but we generally care more about the end state: did it succeed or fail?  if it failed, what error did we get?  how many execution plans were in the XML?  how long did generating the plan from the XML take?

Benjin (changes requested):

laurennat:
Closing this PR; I'm adding the telemetry changes with my other pr that addresses query plan issues

Issue #18211, opened 2024-10-15
Cannot see the selected rows clearly in new result grid


kburtram:
Steps to Reproduce:

1. Run a query with new results grid

Expected: Selected regions are clear

Actual:  They are not.  Can we use the previous styling for this?

![Image](https://github.com/user-attachments/assets/4a5c5dc0-9c17-4caf-9dab-db25dd3c1496)

kburtram:
@caohai @cssuh do we know if a fix is coming for this?  It seems like "baseline functionality" we need here, right?

caohai:
@kburtram This should be fixed in this PR, the gif in the PR shows the region selection styling.
https://github.com/microsoft/vscode-mssql/pull/18206

kburtram:
thanks @caohai!  I figured a fix was coming but wanted to open "just in case".  I couldn't find an existing issue.

cssuh:
> [@kburtram](https://github.com/kburtram) This should be fixed in this PR, the gif in the PR shows the region selection styling. [#18206](https://github.com/microsoft/vscode-mssql/pull/18206)

That PR linked doesn't change much around region selection styling, but I can address it in a future PR.

caohai:
@cssuh This issue is fixed, right

Issue #18094, opened 2024-09-27
Support Selection based "Save as..."


cssuh:
Support selection based "Save as..." if the user has selected a subset of the grid instead of saving the whole data set displayed on the grid.

PR #18305, opened 2024-10-22
Adding selection based save as & improve cell selector visibility

cssuh:
- Adds selection based save as
- Increases the size of the cell selector dashed line
- Fixes focused active selection color

Cell selector dashed line:
Screenshot 2024-10-22 at 6 02 06 PM

Focused active selection:
Screenshot 2024-10-23 at 11 26 43 AM

Non-focused active selection:
Screenshot 2024-10-23 at 11 26 58 AM

Fixes https://github.com/microsoft/vscode-mssql/issues/18211
Fixes https://github.com/microsoft/vscode-mssql/issues/18094


src/reactviews/pages/QueryResult/table/plugins/cellSelectionModel.plugin.ts
```diff
@@ -34,17 +40,28 @@ export class CellSelectionModel<T extends Slick.SlickData>
...
+    ) {
+        this.webViewState = webViewState;
         this.options = mixin(this.options, defaults, false);
         if (this.options.cellRangeSelector) {
             this.selector = this.options.cellRangeSelector;
         } else {
             // this is added by the node requires above
             this.selector = new CellRangeSelector({
-                selectionCss: { border: "2px dashed grey" },
+                selectionCss: { border: "4px dashed grey" },
```
Benjin, line 64: I think the issue with the visibility is less the border thickness and more the color.  What about using `theme.colorBrandBackgroundHover` here?

src/reactviews/pages/QueryResult/table/plugins/cellSelectionModel.plugin.ts
```diff
@@ -132,6 +149,14 @@ export class CellSelectionModel<T extends Slick.SlickData>
 
         this.ranges = this.removeInvalidRanges(ranges);
         this.onSelectedRangesChanged.notify(this.ranges);
+        this.webViewState.state.selection = JSON.parse(
```
kburtram, line 152: do you think this could introduce perf delays with large resultsets?  Could you please test with very large resultsets (like cross join "sys.objects" a few times of something, and then select all)?  If it's too slow, maybe there's a way to avoid the deep copy here?

kburtram (approved):

caohai:
Can you update the `Save As` telemetry event to include the origin of the operation (from command bar or context menu)? You can ping me if we need to log the telemetry from the webview side.

src/reactviews/pages/QueryResult/table/plugins/cellSelectionModel.plugin.ts
caohai: If we need this selection info to be included in the state, we might just include it into the `QueryResultWebviewState` definition (or did I miss this?).
And ADS keeps selection for each grid, do we want this or just keep selection for the grid in focus?
image

cssuh: I did add it to the QueryResultWebviewState definition.  I think we should keep a selection for each grid, since we'll eventually want to be able to copy results from multiple grids at once.

src/reactviews/pages/QueryResult/table/plugins/cellSelectionModel.plugin.ts
cssuh: ok, will figure out how to get access to the theming here

Benjin: the webview state should have a `theme` property will all the values

cssuh: Fixed color & updated screenshot

Benjin:
Can you swap the screenshot for a video and show us all the states?  I think the "selected" state may also need an update.

Benjin (approved):

Issue #18246, opened 2024-10-16
Run a query with a loop is painfully slow to execute


kburtram:
Steps to Reproduce:

1. Run a query with a loop, like below.

```SQL
-- Query that prints tons of messages
DECLARE @j INT = 0;
WHILE @j < 10000 -- Adjust this number to control how many messages are printed
BEGIN
    PRINT 'Message ' + CAST(@j AS NVARCHAR(10)) + ': ' + CAST(NEWID() AS NVARCHAR(36));
    SET @j = @j + 1;
END
```

Expected: The query runs and renders quickly.

Actual: It does not.  Compare the execution of this snippet between vscode & SSMS.  It takes a like 100x as long in my env.

Benjin:
Pri 1 for investigation; if complex, defer to Nov/Dec

caohai:
I did some quick investigation yesterday and I believe the main issue comes from the way we handle message event, each call to the event handler triggers a call to the webview, which can get slow with a large number of msg event. Rendering all the messages once takes a few seconds, which is not super quick but is not the bottleneck for now.

I'll move this issue to Nov and continue the investigation/fix.

Issue #18276, opened 2024-10-18
Wrong grid height when switching between panel tabs


caohai:
Steps to repro:
1. Run a query with 2 results:
```
select * from sys.objects
go
select * from sys.objects
```
2. switch to `Terminal` then back to `Query Results` tab
![Image](https://github.com/user-attachments/assets/831bcf16-ad00-48de-84e1-6d27e3eb971a)

This issue goes away when you:
1. Resize
2. Re-run the query
3. Switch to a different editor tab then switch back
4. Switch to `Messages` tab then back to `Results` tab.

PR #18314, opened 2024-10-24
Fix Message Pane Performance

caohai:
This PR fixes https://github.com/microsoft/vscode-mssql/issues/18246

This PR does the following two things to fix the message pane performance issue:
1. Adds virtualization to the message pane by using `react-data-grid-react-window`
    > A variant of the Fluent UI [DataGrid](https://react.fluentui.dev/?path=/docs/components-datagrid--default) that is virtualized using [react-window](https://react-window.vercel.app/#/examples/list/fixed-size).

    Now we only render the rows that can be seen (plus a few more). This improves the performance of the initial rendering as well as rendering during text editor switching, Results/Messages tab switching, panel tab switching.
    Screenshot 2024-10-24 at 10 01 00
2. Set a fixed interval for setting state in `message` event handler. Ideally, we should separate messages out of the state and batch incremental updates to the webview but it would still require some sort of cooldown for that communication. And this simple change seems to be good enough. Now it takes <1s for the 10K messages query in https://github.com/microsoft/vscode-mssql/issues/18246 to finish in UI. 

This PR also updates the message row styling to match ADS styling.
Now:
image

Before:
image


src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -309,29 +261,114 @@ export const QueryResultPane = () => {
...
             for (let j = 0; j < Object.keys(batch ?? []).length; j++) {
-                grids.push(renderGrid(i, j, count));
-                count++;
+                grids.push(renderGrid(i, j));
             }
         }
         return grids;
     };
 
+    // ========================= MESSAGE GRID =========================
```
aasimkhan30, line 278: Can you just move this into a separate component (in a follow up PR)? Both the message and result grid? It is not very effecient to have big monolith components. Re-rendering becomes more expensive

aasimkhan30 (approved):
Looks good to me.

src/reactviews/pages/QueryResult/queryResultPane.tsx
caohai: Yes this is what I'm planning to do, I don't want to do it in this PR so just re-ordered things and added the section comments.

cssuh:
Something seems off with the boundaries for each message, are they too big? - if I run `select * from sys.objects` and then hover over various messages, the hover box for each individual message seems to extend beyond what I would expect.  It can make it difficult to select individual messages:
Screenshot 2024-10-24 125059
I think this is the box for the message saying `133 rows affected` - is there a reason it has a min height of 44 px?

cssuh:
Also, the resize performance seems worse on the actual results grid - it seems noticeably slower than before.

cssuh:
One last thing I noticed is that the result grid size calculation seems a bit off - the last row of data gets cut off

caohai:
> Something seems off with the boundaries for each message, are they too big? - if I run `select * from sys.objects` and then hover over various messages, the hover box for each individual message seems to extend beyond what I would expect. It can make it difficult to select individual messages: Screenshot 2024-10-24 125059 I think this is the box for the message saying `133 rows affected` - is there a reason it has a min height of 44 px?

Taking a look.



> One last thing I noticed is that the result grid size calculation seems a bit off - the last row of data gets cut off
That's weird, I didn't change the calculation logic, just used the old numbers to set grid props, will take a look

caohai:
@cssuh I've reverted all the result grid related changes, I'm seeing https://github.com/microsoft/vscode-mssql/issues/18276 with any vertical resizing.
Screenshot 2024-10-24 at 14 22 12

src/models/sqlOutputContentProvider.ts
```diff
@@ -474,14 +477,18 @@ export class SqlOutputContentProvider {
...
-                    this._queryResultWebviewController.getQueryResultState(
-                        uri,
-                    ).tabStates.resultPaneTab = QueryResultPaneTabs.Messages;
-                    this._queryResultWebviewController.state =
-                        this._queryResultWebviewController.getQueryResultState(
-                            uri,
-                        );
-                    vscode.commands.executeCommand("queryResult.focus");
+
+                    // Set state for messages at fixed intervals to avoid spamming the webview
```
Benjin, line 481: @aasimkhan30 think this could be solved at a more generic level in the webviews by adding a debounce in the right spot?

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -193,40 +183,9 @@ export const QueryResultPane = () => {
...
-        },
-        [
-            useTableColumnSizing_unstable({
-                columnSizingOptions: columnSizingOption,
-            }),
-        ],
-    );
-    const rows = getRows();
 
+    // ========================= RESULT GRID =========================
```
Benjin, line 187: FYI: you can do this if you want vscode to recognize this as a region for code collapsing

```
//#region Result Grid
stuff here
//#endregion
```

Benjin (approved):

caohai:
Waiting on fix for #18276 as this change make the bug happen more frequently.

src/reactviews/pages/QueryResult/queryResultPane.tsx
caohai: updated

caohai:
> Something seems off with the boundaries for each message, are they too big? - if I run `select * from sys.objects` and then hover over various messages, the hover box for each individual message seems to extend beyond what I would expect. It can make it difficult to select individual messages: Screenshot 2024-10-24 125059 I think this is the box for the message saying `133 rows affected` - is there a reason it has a min height of 44 px?
Fixed in the latest iteration, column width is off, will fix in a different PR.
image

Issue #18186, opened 2024-10-11
Centralize execution plan reducers


laurennat:
If these are copied from the execution plan provider, it will be good to have one implementation of it.

_Originally posted by @aasimkhan30 in https://github.com/microsoft/vscode-mssql/pull/18124#discussion_r1788147671_

Issue #18132, opened 2024-10-02
Make query plan xml formatted


laurennat:
Is the xml that we get back from Tools Service already formatted?  It would be nice to have it formatted in a readable way when we display or save it.

_Originally posted by @Benjin in https://github.com/microsoft/vscode-mssql/pull/18124#discussion_r1783674009_

PR #18323, opened 2024-10-28
Fixes Query Plan Issues and adds telemetry to query plan

laurennat:
Fixes: https://github.com/microsoft/vscode-mssql/issues/18257, https://github.com/microsoft/vscode-mssql/issues/18257, https://github.com/microsoft/vscode-mssql/issues/18186, https://github.com/microsoft/vscode-mssql/issues/18132

A lot of the line changes detected are just code removed from the executionPlan and queryResult controllers and centralized in executionPlanUtils

New popover styling: 
Screenshot 2024-10-28 110849
Screenshot 2024-10-28 110920


src/controllers/sharedExecutionPlanUtils.ts
```diff
@@ -0,0 +1,184 @@
...
+    // start at plan.sqlplan, then plan1.sqlplan, ...
+    let counter = 1;
+    if (await exists(`plan.sqlplan`, folder)) {
+        while (await exists(`plan${counter}.sqlplan`, folder)) {
+            counter += 1;
+        }
+        filename = vscode.Uri.joinPath(folder, `plan${counter}.sqlplan`);
+    } else {
+        filename = vscode.Uri.joinPath(folder, "plan.sqlplan");
+    }
```
Benjin, line 41: Can you pull this into a helper function in `utils.ts`?  I think it'll be useful elsewhere:

`function getUniqueFilePath(folder, baseName, extension) { ... }`

src/controllers/sharedExecutionPlanUtils.ts
```diff
@@ -0,0 +1,184 @@
+/*---------------------------------------------------------------------------------------------
```
Benjin, line 1: Love this refactor!

src/controllers/sharedExecutionPlanUtils.ts
```diff
@@ -0,0 +1,184 @@
...
+        filename = vscode.Uri.joinPath(folder, `plan${counter}.sqlplan`);
+    } else {
+        filename = vscode.Uri.joinPath(folder, "plan.sqlplan");
+    }
+
+    // Show a save dialog to the user
+    const saveUri = await vscode.window.showSaveDialog({
+        defaultUri: filename,
+        filters: {
+            "SQL Plan Files": ["sqlplan"],
```
Benjin, line 47: "SQL Plan Files" probably needs to be localized

src/controllers/sharedExecutionPlanUtils.ts
```diff
@@ -0,0 +1,184 @@
...
+            newState.loadState = ApiStatus.Loaded;
+
+            sendActionEvent(
+                TelemetryViews.QueryPlan,
+                TelemetryActions.OpenQueryPlan,
+                {},
+                {
+                    numberOfPlans:
+                        state.executionPlanState.executionPlanGraphs.length,
+                    LoadTimeInMs: performance.now() - startTime,
```
Benjin, line 128: keep consistent casing (lowercase "load")

src/controllers/sharedExecutionPlanUtils.ts
```diff
@@ -0,0 +1,184 @@
...
+                },
+            );
+        } catch (e) {
+            newState.loadState = ApiStatus.Error;
+            newState.errorMessage = e.toString();
```
Benjin, line 133: use `getErrorMessage(err)` since anything can be thrown as an error, and this function will handle more cases cleanly.

src/controllers/sharedExecutionPlanUtils.ts
```diff
@@ -0,0 +1,184 @@
...
+            sendErrorEvent(
+                TelemetryViews.QueryPlan,
+                TelemetryActions.OpenQueryPlan,
+                e,
+                true, // includeErrorMessage
```
Benjin, line 138: Can you confirm that this error message won't have any PII in it?  Otherwise, you may need to try to extract something useful out of it for a property and set this to `false`.

src/controllers/sharedExecutionPlanUtils.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    let sum = 0;
+    for (const graph of state.executionPlanState.executionPlanGraphs) {
+        sum += graph.root.cost + graph.root.subTreeCost;
+    }
+    return sum;
+}
+
+export function formatXml(xmlContents: string): string {
+    let formattedXml = "";
```
Benjin, line 164: Did you get this logic from somewhere?  If there's a package library that does XML formatting, that might be a less-risky option than custom code.

Perhaps it'd be a good idea to wrap it all in a try-catch where catch just returns the original XML just in case.

src/queryResult/queryResultWebViewController.ts
```diff
@@ -80,13 +84,22 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
 
     private registerRpcHandlers() {
         this.registerRequestHandler("getRows", async (message) => {
-            return await this._sqlOutputContentProvider.rowRequestHandler(
+            let result = await this._sqlOutputContentProvider.rowRequestHandler(
```
Benjin, line 87: `let` if it's not getting changed later, otherwise `const`

src/queryResult/queryResultWebViewController.ts
```diff
@@ -80,13 +84,22 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
                 message.batchId,
                 message.resultId,
                 message.rowStart,
                 message.numberOfRows,
             );
+            let currentState = this.getQueryResultState(message.uri);
+            if (currentState.isExecutionPlan) {
+                currentState.executionPlanState.xmlPlans =
+                    currentState.executionPlanState.xmlPlans.concat(
+                        result.rows[0][0].displayValue,
```
Benjin, line 98: add a comment: what should we expect in the first cell?  First column's name? Plan name?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -164,10 +177,23 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
-            state.executionPlanState.loadState = ApiStatus.Loaded;
-            state.tabStates.resultPaneTab =
-                qr.QueryResultPaneTabs.ExecutionPlan;
+            const currentResultState = this.getQueryResultState(payload.uri);
+            if (
+                currentResultState.executionPlanState.xmlPlans.length &&
+                currentResultState.executionPlanState.xmlPlans.length ===
+                    Object.keys(currentResultState.resultSetSummaries).length &&
+                currentResultState.executionPlanState.executionPlanGraphs
+                    .length === 0
```
Benjin, line 186: add a comment: what's the overall goal of this compound boolean?

src/controllers/sharedExecutionPlanUtils.ts
laurennat: i didn't get the code from somewhere, but i'll wrap it in a try catch and investigate packages to format it

src/controllers/sharedExecutionPlanUtils.ts
laurennat: me too

src/constants/locConstants.ts
```diff
@@ -640,6 +640,8 @@ export let queryFailed = l10n.t("Query failed");
 export let parameters = l10n.t("Parameters");
 export let loading = l10n.t("Loading");
 export let executionPlan = l10n.t("Execution Plan");
+export let executionPlanFileFilter = l10n.t("SQL Plan Files");
+export let executionPlanFileBaseName = l10n.t("plan");
```
aasimkhan30, line 644: should this be localized?

src/sharedInterfaces/telemetry.ts
```diff
@@ -17,6 +17,7 @@ export enum TelemetryViews {
     UserSurvey = "UserSurvey",
     General = "General",
     ConnectionDialog = "ConnectionDialog",
+    QueryPlan = "QueryPlan",
```
aasimkhan30, line 20: Query plan or execution plan? I see that the code uses a lot of interfaces contains 'ExecutionPlan'. Let's try to keep the feature name consistent. Helps in searching and refactoring stuff.

src/sharedInterfaces/queryResult.ts
```diff
@@ -55,9 +55,7 @@ export interface QueryResultWebviewState extends ExecutionPlanWebviewState {
     tabStates?: QueryResultTabStates;
     isExecutionPlan?: boolean;
     selection?: ISlickRange[];
-    executionPlanState: ExecutionPlanState & {
-        xmlPlans?: string[];
-    };
+    executionPlanState: ExecutionPlanState;
```
aasimkhan30, line 58: Probably a good followup task will be to add js doc comments to all the interfaces and properties.

src/reactviews/pages/ExecutionPlan/executionPlanPage.tsx
```diff
@@ -35,9 +35,22 @@ const useStyles = makeStyles({
...
+            state &&
+            state.provider &&
+            executionPlanState &&
+            // checks if execution plans have already been gotten
+            executionPlanState.executionPlanGraphs &&
+            !executionPlanState.executionPlanGraphs.length
+        ) {
+            state.provider.getExecutionPlan(executionPlanState.xmlPlans!);
+        }
+    }, [state, executionPlanState]);
```
aasimkhan30, line 52: Do we really need state here? Can you make this more targeted?

src/controllers/sharedExecutionPlanUtils.ts
aasimkhan30: Is there a need to log these error? What will we do once we get these errors. It is rare that execution plan parser has errors in it, most likely it will be malformed plan file.

src/constants/locConstants.ts
laurennat: do we usually localize file name related things?

src/sharedInterfaces/queryResult.ts
laurennat: i added some for now on the less self-explanatory ones

src/constants/locConstants.ts
Benjin: I think there isn't a need to localize the basename or file extension, but we do want to localize "SQL Plan Files" since that's meant to be a human-readable string to describe the extension.

src/controllers/sharedExecutionPlanUtils.ts
aasimkhan30: @Benjin , @laurennat

Benjin: I think we should attempt to detect if it's a malformed XML error and _not_ call `sendErrorEvent()` in that case.  Seems silly to have an opportunity to detect unknown errors here and not emit an error event if they aren't user errors.

src/constants/constants.ts
```diff
@@ -220,7 +219,7 @@ export const windowsResourceClientPath = "SqlToolsResourceProviderService.exe";
 export const unixResourceClientPath = "SqlToolsResourceProviderService";
 export const microsoftPrivacyStatementUrl =
     "https://www.microsoft.com/en-us/privacy/privacystatement";
-export const sqlPlanFileExtension = ".sqlplan";
+export const sqlPlanLanguageId = "sqlplan";
```
Benjin, line 222: out of curiosity, why "language ID" instead of the old "file extension"?  Is this being used for more than just the file-picker?

laurennat: language id is being used for the custom editor registration in the package.json, and I figured we don't need two constants of almost the exact same thing- i can just add the dot in manually when needed

src/controllers/sharedExecutionPlanUtils.ts
laurennat: i was looking at this more and I think the only errors actually thrown after the refactor are the malformed plans. I'll just remove the send error event

aasimkhan30 (approved):

PR #18328, opened 2024-10-29
Adding start and end activity function

aasimkhan30:
Fixes: #18322


src/controllers/reactWebviewBaseController.ts
```diff
@@ -47,36 +52,68 @@ export abstract class ReactWebviewBaseController<State, Reducers>
...
-                this.postMessage({ type: "response", id: message.id, result });
-                const endTime = Date.now();
-                sendActionEvent(
-                    TelemetryViews.WebviewController,
-                    TelemetryActions.WebviewRequest,
-                    {
+                try {
+                    const result = await handler(message.params);
+                    this.postMessage({
+                        type: "response",
```
lewis-sanchez, line 73: Are there other types in addition to `"response"`?

src/sharedInterfaces/telemetry.ts
```diff
@@ -56,3 +56,30 @@ export enum TelemetryActions {
...
+export type FinishActivityFailed = (
+    error?: Error,
+    includeErrorMessage?: boolean,
+    errorCode?: string,
+    errorType?: string,
+    additionalProperties?: Record<string, string>,
+    additionalMeasurements?: Record<string, number>,
+) => void;
+
+export type FinishActivityObject = {
```
Benjin, line 81: `ActivityObject` might be better so that we can add more to it than just ending the activity.

src/telemetry/telemetry.ts
```diff
@@ -110,3 +115,70 @@ export function sendErrorEvent(
...
+    }
+
+    sendActionEvent(telemetryView, telemetryAction, additionalProps, {
+        ...additionalMeasurements,
+        startTime: Math.round(startTime),
+    });
+
+    return {
+        correlationId: correlationId,
+        end: (
```
Benjin, line 138: Can you swap this (and `endFailed()`) to a deconstruction so it's cleaner to call without setting everything?

Benjin, line 138: IMO inline functions that span more than a couple lines are tough to read. Can you make this a local function and then reference that here?

src/sharedInterfaces/telemetry.ts
```diff
@@ -56,3 +56,30 @@ export enum TelemetryActions {
...
+    end: FinishActivity;
```
Benjin, line 82: I think `timeElapsed` would be useful here, too.  Sometimes you want to emit midpoint events in addition to just the starts and ends.

src/controllers/reactWebviewBaseController.ts
aasimkhan30: For now, there isn't any other type. I have been meaning to add stronger types but haven't done that yet.

Benjin (approved):

lewis-sanchez (approved):

Issue #18245, opened 2024-10-16
Message pane should have Select All for copying


kburtram:
Steps to Reproduce:

1. Run a query that generates a lot of message, for example below is what I was testing with.

```SQL
-- Query that prints tons of messages
DECLARE @j INT = 0;
WHILE @j < 10000 -- Adjust this number to control how many messages are printed
BEGIN
    PRINT 'Message ' + CAST(@j AS NVARCHAR(10)) + ': ' + CAST(NEWID() AS NVARCHAR(36));
    SET @j = @j + 1;
END
```

Expected: It's easy to select all the messages and copy/paste them.

Actual: No "Select All" or similar command to get everything selected for copy and UI isn't idea for doing this.  And also, if nothing is selected and you click on a "message row, and click Copy, nothing goes to clipboard.  I would expect the clicked-on message entry to be copied, or if nothing will be copied that the menu item would be disabled.  FWIW, ADS uses "Copy All", which is also totally fine for me.

PR #18332, opened 2024-10-29
Support copying all messages in the Message pane

caohai:
This PR fixes https://github.com/microsoft/vscode-mssql/issues/18245
This PR adds the `Copy All` context menu to the messages pane. This context menu uses the VSCode `webview/context` contrib point so it's included in the native context menu.
https://code.visualstudio.com/api/extension-guides/webview#context-menus
I've also tested the 10k message copy and the perf is good.
Oct-29-2024 13-36-49

### DIFF
--- a/package.json
+++ b/package.json
@@ -562,6 +562,12 @@
           "command": "mssql.editTable",
           "when": "view == objectExplorer && viewItem =~ /\\btype=(Table)\\b/ && config.mssql.enableRichExperiences"
         }
+      ],
+      "webview/context": [
+        {
+          "command": "mssql.copyAll",
+          "when": "webviewId == 'queryResult' && webviewSection == 'queryResultMessagesPane'"
+        }
       ]
     },
     "commands": [
@@ -581,6 +587,11 @@
         "title": "%mssql.cancelQuery%",
         "category": "MS SQL",
         "icon": "$(debug-stop)"
+      },
+      {
+        "command": "mssql.copyAll",
+        "title": "%mssql.copyAll%",
+        "category": "MS SQL"
       },
       {
         "command": "mssql.revealQueryResultPanel",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -562,6 +562,12 @@
           "command": "mssql.editTable",
           "when": "view == objectExplorer && viewItem =~ /\\btype=(Table)\\b/ && config.mssql.enableRichExperiences"
         }
+      ],
+      "webview/context": [
+        {
+          "command": "mssql.copyAll",
+          "when": "webviewId == 'queryResult' && webviewSection == 'queryResultMessagesPane'"
```
caohai, line 569: Not including the rich feature flag here as the `queryResult` view inherently requires the flag to be enabled.

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -438,7 +438,13 @@ export const QueryResultPane = () => {
                     renderGridPanel()}
                 {metadata.tabStates!.resultPaneTab ===
                     qr.QueryResultPaneTabs.Messages && (
-                    <div className={classes.messagesContainer}>
+                    <div
+                        className={classes.messagesContainer}
+                        data-vscode-context={JSON.stringify({
```
aasimkhan30, line 443: Can you make a helper function for this?

aasimkhan30 (approved):

lewis-sanchez (approved):

Benjin (approved):

src/reactviews/pages/QueryResult/queryResultPane.tsx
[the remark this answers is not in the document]
caohai: This is the context that will be used by VSCode to provide native context menus.
https://code.visualstudio.com/api/extension-guides/webview#context-menus:~:text=HTML%20using%20the-,data%2Dvscode%2Dcontext,-data%20attribute%20(or
> The data-vscode-context value is a JSON object that specifies the contexts to set when the user right-clicks on the element.

caohai: I couldn't find a good way to make a helper function for this, it's just stringifying an object. We can look into this if we have more usage of this attribute.

PR #18342, opened 2024-10-31
Fix tab switch resize bug

### LINKED ISSUE #18276: Wrong grid height when switching between panel tabs

cssuh:
Fixes #18276


cssuh:
after further testing, this didn't fix the bug completely.

PR #18345, opened 2024-11-01
Added enable/disable actual plan for query plans

laurennat:
actualPlans


package.json
```diff
@@ -338,6 +338,16 @@
           "command": "mssql.showExecutionPlanInResults",
           "when": "editorLangId == sql && config.mssql.enableRichExperiences",
           "group": "navigation@4"
+        },
+        {
+          "command": "mssql.enableActualPlan",
+          "when": "editorLangId == sql && config.mssql.enableRichExperiences && !(resource in mssql.executionPlan.urisWithActualPlanEnabled)",
```
kburtram, line 344: this condition seems a little specific, but I guess we need it "resource in mssql.executionPlan.urisWithActualPlanEnabled"?

src/controllers/mainController.ts
```diff
@@ -1632,6 +1644,30 @@ export default class MainController implements vscode.Disposable {
...
 
+    public onToggleActualPlan(
+        isEnable: boolean,
+        callbackThis?: MainController,
+    ): void {
+        // the 'this' context is lost in retry callback, so capture it here
+        const self: MainController = callbackThis ? callbackThis : this;
+        const uri = self._vscodeWrapper.activeTextEditorUri;
+
+        if (isEnable && !this._actualPlanStatuses.includes(uri)) {
```
kburtram, line 1655: so we only need self for the "vscodeWrapper" but the rest of the step is fine from whatever "this"?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -93,7 +93,14 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
                     message.numberOfRows,
                 );
             let currentState = this.getQueryResultState(message.uri);
-            if (currentState.isExecutionPlan) {
+            if (
+                currentState.isExecutionPlan &&
```
kburtram, line 97: is this like our "prettier" wanting the code like this?  I don't know, it feels like LISP or something, ok, whatever. 😄

src/queryResult/queryResultWebViewController.ts
```diff
@@ -309,4 +325,22 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+
+    public getNumExecutionPlanResultSets(
+        resultSetSummaries: qr.QueryResultWebviewState["resultSetSummaries"],
+        actualPlanEnabled: boolean,
+    ): number {
+        const summariesLength = Object.keys(resultSetSummaries).length;
+        if (!actualPlanEnabled) {
+            return summariesLength;
+        }
+        // if actual plan is enabled, then
```
kburtram, line 337: Is this statement always true, like even for queries that have no results, like PRINTs or INSERTs, etc?  I don't know, if just seems like we may be "taking a leap of faith" here.

src/queryResult/queryResultWebViewController.ts
```diff
@@ -309,4 +325,22 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
         const messageText = messages.join("\n");
         await this._vscodeWrapper.clipboardWriteText(messageText);
     }
+
+    public getNumExecutionPlanResultSets(
```
kburtram, line 329: Do you think we should add some unit tests for this new functionality?

src/controllers/mainController.ts
```diff
@@ -86,6 +86,7 @@ export default class MainController implements vscode.Disposable {
         includeEstimatedExecutionPlanXml: false,
         includeActualExecutionPlanXml: false,
     };
+    private _actualPlanStatuses: string[] = [];
```
kburtram, line 89: Consider adding comments around _actualPlanStatuses usage in different functions to clarify the state updates, especially in cleanup and toggle functions.

kburtram (approved):

src/queryResult/queryResultWebViewController.ts
laurennat: Not LISP! but yeah the default formatter did that

package.json
laurennat: it's to make sure we know which queries have the plan enabled  if there are multiple open

src/queryResult/queryResultWebViewController.ts
laurennat: yes! that'll be my next pr

package.json
Benjin: if only json supported comments 🙃

src/controllers/mainController.ts
Benjin, line 89: Do you think it would make more sense to track these in queryResultsController, and call into that from here?

src/controllers/mainController.ts
```diff
@@ -1951,6 +1987,16 @@ export default class MainController implements vscode.Disposable {
             this._outputContentProvider.onDidCloseTextDocument(doc);
         }
 
+        if (this._actualPlanStatuses.includes(closedDocumentUri)) {
+            const index = this._actualPlanStatuses.indexOf(closedDocumentUri);
+            this._actualPlanStatuses.splice(index, 1); // Remove the URI
```
Benjin, line 1992: The JS community tends to favor `list.filter(x => x != closedDocumentUri)` since it's more readable

src/queryResult/queryResultWebViewController.ts
Benjin: Agreed.  Seems better to actually inspect each summary.  What does ADS do to differentiate between types?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -93,7 +93,14 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
             let currentState = this.getQueryResultState(message.uri);
-            if (currentState.isExecutionPlan) {
+            if (
+                currentState.isExecutionPlan &&
+                // check if the current result set is the result set that contains the xml plan
+                currentState.resultSetSummaries[message.batchId][
+                    message.resultId
+                ].columnInfo[0].columnName ===
+                    "Microsoft SQL Server 2005 XML Showplan"
+            ) {
```
Benjin, line 103: If Karl's comment below ("is this always true?") means checking each summary, this may be a good thing to extract out to a utility function

Benjin:
In playing around with your branch, I found a bug that results in the plan tab being completely stuck.  Probably a good one to fix before merging in:

Repro:
1. open new query, and write something e.g. `select * from sys.objects`
2. Click "Enable actual plan"
3. Click "Estimated Plan"
4. Change results pane tab to "Query Plan"

Expected:
estimated query plan is displayed

Reality:
* stuck forever on "Loading execution plan"
* Console repeated displaying "No query result state found for uri undefined"
image

src/queryResult/queryResultWebViewController.ts
laurennat: I changed it to manually check

laurennat:
> In playing around with your branch, I found a bug that results in the plan tab being completely stuck. Probably a good one to fix before merging in:
> 
> Repro:
> 
> 1. open new query, and write something e.g. `select * from sys.objects`
> 2. Click "Enable actual plan"
> 3. Click "Estimated Plan"
> 4. Change results pane tab to "Query Plan"
> 
> Expected: estimated query plan is displayed
> 
> Reality:
> 
> * stuck forever on "Loading execution plan"
> * Console repeated displaying "No query result state found for uri undefined"
>   image

This was fixed by having the manual check :)

src/queryResult/queryResultWebViewController.ts
laurennat: i'm not sure if it makes sense to me to extract the new implementation of it into a utility function. in my head, this was already a utility function. Let me know though if you feel that I should still make a separate utility function with the new implementation!

src/queryResult/queryResultWebViewController.ts
Benjin: Yeahh... I want to bump the target printwidth for prettier so that it's a little less aggressive about wrapping lines.

src/queryResult/queryResultWebViewController.ts
```diff
@@ -309,4 +330,31 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+        if (!actualPlanEnabled) {
+            return summariesLength;
+        }
+        // count the amount of xml showplans in the result summaries
+        let total = 0;
+        Object.values(resultSetSummaries).forEach((batch) => {
+            Object.values(batch).forEach((result) => {
+                // Check if any column in columnInfo has the specific column name
+                if (
+                    result.columnInfo.some(
```
Benjin, line 348: Why is this checking all columns to see if they have the showplan string, but on line 103, it's only checking the first column?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -309,4 +330,31 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+                        (column) =>
+                            column.columnName ===
+                            "Microsoft SQL Server 2005 XML Showplan",
```
Benjin, line 351: since this string is very specific and is used twice, let's extract it to a shared const.

src/queryResult/queryResultWebViewController.ts
```diff
@@ -93,7 +94,14 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
                     message.numberOfRows,
                 );
             let currentState = this.getQueryResultState(message.uri);
-            if (currentState.isExecutionPlan) {
+            if (
```
Benjin, line 97: Small readability thing, but if the bulk of the logic is inside the condition, then it's nice to invert it and have it bail out early.  That way, all the long longer logic isn't indented as deeply.

instead of
```
if (something) {
  doStuff()
  doStuff()
  doStuff()
  doStuff()
  doStuff()
  doStuff()
  doStuff()
}
return state;
```

it's easier to read if it's:
```
if (!something) {
return state; // short null/bail case is first
}

doStuff()
doStuff()
doStuff()
doStuff()
doStuff()
doStuff()
doStuff()
return state;
```

PR #18346, opened 2024-11-02
Fix tab switch behavior

### LINKED ISSUE #18276: Wrong grid height when switching between panel tabs

cssuh:
Fixes #18276

TODO:

- [x] Cleanup PR
- [x] Fix multiple resize height
- [x] Fix sizing issue with last row

### DIFF
--- a/src/reactviews/pages/QueryResult/queryResultPane.tsx
+++ b/src/reactviews/pages/QueryResult/queryResultPane.tsx
@@ -194,8 +194,11 @@ export const QueryResultPane = () => {
 
         observer.observe(gridParent);
 
-        return () => observer.disconnect();
-    }, [metadata?.resultSetSummaries]);
+        return () => {
+            console.log("disconnect");
+            observer.disconnect();
+        };
+    }, [metadata?.resultSetSummaries, gridParentRef.current]);
     const [columns] =
         useState<TableColumnDefinition<qr.IMessage>[]>(columnsDef);
     const items = metadata?.messages ?? [];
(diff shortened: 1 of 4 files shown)


caohai:
Is there anything special about the height calculation on Windows?

cssuh:
> Is there anything special about the height calculation on Windows?

Updated, it isn't a Windows-related thing

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -144,55 +145,85 @@ export const QueryResultPane = () => {
             return; // Exit if there are no grids to render
         }
 
-        const gridParent = gridParentRef.current;
-        if (!gridParent) {
+        const resultPaneParent = resultPaneParentRef.current;
+        if (!resultPaneParent) {
+            console.log("resultPaneParent null");
```
Benjin, line 150: `console.log()` often means that we should think more about what we want to log and how:
* should it be removed because it was just for development debugging?
* should it instead be `console.warn()`, if we want users to be able to find it if they need to file a bug?
* is this an error that we should be emitting telemetry for?

Benjin: Same for other instances of `console.log`

cssuh: I removed these log statements, they were just for debugging purposes

src/reactviews/pages/QueryResult/resultGrid.tsx
```diff
@@ -81,7 +82,18 @@ const ResultGrid = forwardRef<ResultGridHandle, ResultGridProps>(
             }
             setRefreshKey((prev) => prev + 1);
         };
+        // const gridParent = props.gridParentRef?.current;
```
caohai, line 85: do we still need this?

caohai (approved):

src/reactviews/pages/QueryResult/resultGrid.tsx
cssuh: no, I'll remove it

kburtram (approved):

PR #18347, opened 2024-11-02
Fixing Theming for webviews. 

aasimkhan30:
Connection Dialog in different themes.
image


src/reactviews/common/utils.ts
```diff
@@ -29,21 +25,21 @@ export function formatString(str: string, ...args: any[]): string {
 
 /**
  * Gets the theme type based on the theme passed in
- * @param theme the theme of the react webview
+ * @param themeKind the theme of the react webview
```
Benjin, line 28: Can you update the comments on this (and related) utility methods that state what they're mapping from/to?  e.g. "maps from VS Code theme type to Fluent theme type", with a link to the docs page that says what the valid fluent types are.

src/reactviews/common/utils.ts
```diff
@@ -29,21 +25,21 @@ export function formatString(str: string, ...args: any[]): string {
...
-        case webDarkTheme:
+export function getVscodeThemeType(themeKind: ColorThemeKind): string {
+    switch (themeKind) {
+        case ColorThemeKind.Dark:
             return "vs-dark";
-        case teamsHighContrastTheme:
+        case ColorThemeKind.HighContrast:
             return "hc-black";
         default:
             return "light";
```
Benjin, line 37: Do we care about the high-contrast light theme in VS Code?  Or maybe Fluent doesn't have an appropriate mapping for it?

Benjin (approved):

src/reactviews/common/utils.ts
aasimkhan30: This doesn't do any mapping now. It is just exposing the vscode theme kind.

src/reactviews/common/utils.ts
aasimkhan30: I am using this in the table designer monaco editor which doesn't implement the hc-light yet. That's why I am just handling hc-light as light.

kburtram (approved):
looks great!

Issue #18312, opened 2024-10-23
Telemetry for enabling experiences is sending localized repsonses


Benjin:
Should be unlocalized responses

PR #18353, opened 2024-11-05
Unlocalizes "enable modern UX?" prompt response in telemetry

Benjin:
Fixes https://github.com/microsoft/vscode-mssql/issues/18312


aasimkhan30 (approved):
Thanks for fixing

PR #18354, opened 2024-11-05
Add back line to refresh grid

### LINKED ISSUE #18276: Wrong grid height when switching between panel tabs

cssuh:
We need to refresh each grid when we render the grid panel

Fixes #18276 along with this PR: #18346

### DIFF
--- a/src/reactviews/pages/QueryResult/queryResultPane.tsx
+++ b/src/reactviews/pages/QueryResult/queryResultPane.tsx
@@ -349,9 +349,7 @@ export const QueryResultPane = () => {
         // of the query. therefore, it updates the state once the results
         // are loaded, which causes a rendering loop if the grid
         // gets refreshed
-        if (!metadata?.isExecutionPlan) {
-            gridRefs.current.forEach((r) => r?.refreshGrid());
-        }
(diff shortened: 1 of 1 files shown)


src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -345,6 +345,13 @@ export const QueryResultPane = () => {
 
     const renderGridPanel = () => {
         const grids = [];
+        // execution plans only load after reading the resulting xml showplan
+        // of the query. therefore, it updates the state once the results
+        // are loaded, which causes a rendering loop if the grid
+        // gets refreshed
+        if (!metadata?.isExecutionPlan) {
```
caohai, line 352: @laurennat Do we still need this if check here?

laurennat: nope we don't! take out the check and it should be fine.

laurennat (approved):

caohai (approved):

PR #18361, opened 2024-11-06
Adding more theme mappings to fix some issues

aasimkhan30:
This fixes the error message box styling in connection Dialog
Before:
image

Now:
image
image
image

Also fixes, how disabled elements are highlighted in table designer.
Allow null in the first row is disabled and primary key and is identity in the second row is enabled.
Before:
image
Now:
image


aasimkhan30:
Will add more comments on the other variables in a follow up PR.

Benjin (approved):

PR #18367, opened 2024-11-07
Setting the min height for a connection error message

### LINKED ISSUE #18280: Handle long error messages correctly
- MSSQL Extension Version: 1.25

Steps to Reproduce:

1. Connect to a server that requires a firewall rule update

There are two issues:

1. The error message is too big for the container
2. The error message should be focused on the screen when it appears (I didn't see it until I scrolled up in the connection dialog)

 
Image

Benjin:
Fixes #18280

This was rendering fine on Windows, but not macOS before.

Before:
image

After:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "@azure/core-paging": "^1.6.2",
     "@eslint/compat": "^1.1.0",
     "@eslint/js": "^9.5.0",
-    "@fluentui/react-components": "^9.54.13",
+    "@fluentui/react-components": "^9.55.1",
     "@fluentui/react-list-preview": "^0.3.6",
     "@fluentui-contrib/react-data-grid-react-window": "^1.2.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -81,7 +81,7 @@
     "@azure/core-paging": "^1.6.2",
     "@eslint/compat": "^1.1.0",
     "@eslint/js": "^9.5.0",
-    "@fluentui/react-components": "^9.54.13",
+    "@fluentui/react-components": "^9.55.1",
```
aasimkhan30, line 84: Just curious, does the update has anything to do with the fix?

aasimkhan30 (approved):

package.json
Benjin: Nope.  I updated the package in case that was the explanation for the difference between windows and mac.  It wasn't, but I felt like it worth keeping.

PR #18370, opened 2024-11-08
Add parameter to not include headers

### LINKED ISSUE #18318: Header always included when copy from result pane
- MSSQL Extension Version: 1.25.0
- VSCode Version: 1.94.2
- OS Version: Win11

Steps to Reproduce:

1. run sql that returns results
2. in result pane, select one or more cells
3. right click on selected cells, click "Copy" (not "Copy with Headers")
4.  when you paste the results, the header is included

https://github.com/user-attachments/assets/289ed14e-6c16-4fef-a050-50b89f47aec0

cssuh:
Before:

Copy was picking up the user setting: Image

After:

Since we have both copy & copy with headers options in the context menu, pass includeHeaders = false in order to copy without headers

Fixes #18318

### DIFF
--- a/src/queryResult/queryResultWebViewController.ts
+++ b/src/queryResult/queryResultWebViewController.ts
@@ -149,6 +149,7 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
                 message.batchId,
                 message.resultId,
                 message.selection,
+                false,
             );
         });
         this.registerRequestHandler("copyWithHeaders", async (message) => {


Benjin:
I wonder why this config option existed if we had context menu items for both "Copy" and "Copy with Headers"...  Perhaps we can remove it after the old UX is fully replaced/removed.

Benjin (approved):

cssuh:
> I wonder why this config option existed if we had context menu items for both "Copy" and "Copy with Headers"... Perhaps we can remove it after the old UX is fully replaced/removed.

Yeah I agree, we should remove the setting once the angular implementation is gone

Issue #18250, opened 2024-10-16
Result grid client-side sort popup has a Cancel option


kburtram:
Steps to Reproduce:

1. Run a query and sort one of the columns client-side

Expected: Sort Ascending/Sort Descending and maybe like "Reset"

Actual: Cancel item, that doesn't appear to do anything

![Image](https://github.com/user-attachments/assets/c9baf5c6-1c01-43b0-9c38-22e06706932e)

Benjin:
Does cancel the context menu or cancel the sort?  unclear

Issue #18063, opened 2024-09-19
Support header filtering in result grid for Query Result


caohai:

cssuh:
Fixed in https://github.com/microsoft/vscode-mssql/pull/18373

PR #18373, opened 2024-11-11
Add header filtering to query result grid

cssuh:
Addresses: https://github.com/microsoft/vscode-mssql/issues/18063
Addresses: https://github.com/microsoft/vscode-mssql/issues/18250

Adds filtering to the header menu

filterFeatures

TODO:

- [x] Add context switching (currently we lose applied filters if we switch tabs)


caohai:
Can you update the PR title to be more specific? Probably something like `Add header filtering to query result grid`

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -238,6 +376,86 @@ export class HeaderFilter<T extends Slick.SlickData> {
         this.setFocusToColumn(columnDef);
     }
 
+    private async createFilterList(): Promise<void> {
```
aasimkhan30, line 379: Does this list need to be virtualized?

src/reactviews/common/locConstants.ts
```diff
@@ -316,6 +316,12 @@ export class LocConstants {
             copy: l10n.t("Copy"),
             copyWithHeaders: l10n.t("Copy with Headers"),
             copyHeaders: l10n.t("Copy Headers"),
+            null: l10n.t("NULL"),
+            blankString: l10n.t("Blanks"),
+            apply: l10n.t("Apply"),
+            clear: l10n.t("Clear"),
+            search: l10n.t("Search..."),
+            selected: l10n.t("0 Selected"),
```
Benjin, line 324: When is this shown?  My gut says that this either ought to be `"{num} selected"` or `"None selected"` (word rather than number for the special "zero-case" feels nicer here to me?)

Benjin, line 324: If this is only shown when none are selected, then the var name should be `noneSelected` or `zeroSelected` instead of `selected`

Benjin:
Can you add a video to the PR description so we can see it in action?

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -93,14 +103,22 @@ export class HeaderFilter<T extends Slick.SlickData> {
...
             .addClass("slick-header-menubutton")
             .data("column", column);
+        if (
+            this.theme === ColorThemeKind.Dark ||
+            this.theme === ColorThemeKind.HighContrast
+        ) {
+            $el.addClass("vs-dark");
+        } else {
+            $el.addClass("vs-light");
+        }
```
Benjin, line 113: Do you know how `"vs-light"` differs from just `"light"`?  Could this just use `utils.resolveVscodeThemeType(themeKind)` instead?

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
caohai: @cssuh Can you test filtering with something like 10k different rows?

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
aasimkhan30: Quick question, why are we adding classes to the component?

Is using vscode css variables not enough in this case? We should generally just stick to them for all of our coloring and styling needs.

Currently, this doesn't support vs Hight Contrast light theme.

cssuh: This is because we in table.css we show either a light or dark version of the filter icon and this is controlled by these classes. I add the class like this now - any issues with this approach?
Screenshot 2024-11-19 at 8 35 17 AM

src/reactviews/common/locConstants.ts
cssuh: ok, I'm removing this part for now because it's not used at the moment.

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -183,25 +216,81 @@ export class HeaderFilter<T extends Slick.SlickData> {
...
         this.activePopup = $popup;
         const checkboxContainer = $popup.find("#checkbox-list");
-        if (this.listData) {
-            this.listData.forEach((option) => {
-                const checkbox = jQuery(
-                    `<label><input type="checkbox" value="${option.value}"> ${option.displayText}</label>`,
+        this._list = new VirtualizedList(
+            checkboxContainer.get(0),
+            this._listData,
+            (itemContainer, item) => {
```
Benjin, line 222: Can you pop this long function and the next one out into named local functions so that this constructor invocation is easier to read?

src/reactviews/common/virtualizedList.ts
```diff
@@ -0,0 +1,188 @@
...
+                this.updateFocusedItemIndex(i);
+                this._itemSelected(itemDiv, item);
+                e.preventDefault();
+                e.stopPropagation();
+            });
+
+            this._eventManager.addEventListener(itemDiv, "keydown", (e) => {
+                const event = e as KeyboardEvent;
+                let handled = false;
+                if (event.key === "ArrowDown") {
```
Benjin, line 154: extract all these keycode constants to a constants file, e.g. `Keys.ArrowDown` or `Keys.Space`

Benjin: Also, looks like the copyKeybind plugin uses the deprecated `.keyCode === 67` for "c"; can you update that?

src/reactviews/pages/QueryResult/resultGrid.tsx
```diff
@@ -98,6 +100,11 @@ const ResultGrid = forwardRef<ResultGridHandle, ResultGridProps>(
             table?.layout(dimension);
         };
         useEffect(() => {
+            const filter = async () => {
```
Benjin, line 103: nit: I think using the `function` keyword is nicer to read:

```
async function filter() {
  ...
}
```

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -86,8 +113,9 @@ export class HeaderFilter<T extends Slick.SlickData> {
             args.node.classList.add("slick-header-sortable");
             append(args.node, $("span.slick-sort-indicator"));
         }
-
+        let theme: string = resolveVscodeThemeType(this.theme);
```
Benjin, line 116: const?

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -97,20 +125,27 @@ export class HeaderFilter<T extends Slick.SlickData> {
...
-        $el.on("click", async (e: JQuery.ClickEvent) => {
-            e.stopPropagation();
-            e.preventDefault();
-            this.showFilter($el[0]);
-        });
+        const elDivElement = $el.get(0);
+        if (elDivElement) {
+            this._eventManager.addEventListener(
+                elDivElement,
+                "click",
```
Benjin, line 132: extract to const (and elsewhere)

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -194,18 +287,172 @@ export class HeaderFilter<T extends Slick.SlickData> {
...
                 closePopup($popup);
                 this.activePopup = null;
+                this.grid.setSortColumn(this.columnDef.id!, false);
+                this.columnDef.sorted = SortProperties.DESC;
             },
         );
 
+        jQuery(document).on("click", "#apply", async () => {
+            // try to append columnDef.id to apply button ID to differentiate buttons
+            //TODO: only apply click to correct header filter
```
Benjin, line 297: is this TODO still active?

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -194,18 +287,172 @@ export class HeaderFilter<T extends Slick.SlickData> {
...
+        const dataView = this.grid.getData() as IDisposableDataProvider<T>;
+        if (instanceOfIDisposableDataProvider(dataView)) {
+            await dataView.filter(this.grid.getColumns());
+            this.grid.invalidateAllRows();
+            this.grid.updateRowCount();
+            this.grid.render();
+        }
+        this.onFilterApplied.notify({ grid: this.grid, column: columnDef });
+        this.setFocusToColumn(columnDef);
+        //TODO: add search term
```
Benjin, line 434: still active?

src/reactviews/pages/QueryResult/table/table.ts
```diff
@@ -179,9 +190,25 @@ export class Table<T extends Slick.SlickData> implements IThemable {
             console.log("oncolumnresize"),
         );
         this.style(styles);
+        console.log(`table.ts: ${this.queryResultState.state.filterState}`);
```
Benjin, line 193: This should either be removed or swapped to `console.debug`

src/reactviews/common/virtualizedList.ts
cssuh: > Also, looks like the copyKeybind plugin uses the deprecated `.keyCode === 67` for "c"; can you update that?

Sure, I'll fix this one in a separate PR

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
cssuh: @Benjin I'm not quite sure what you mean for this one, could you clarify?

src/queryResult/utils.ts
```diff
@@ -217,6 +218,26 @@ export function registerCommonRequestHandlers(
             return state;
         },
     );
+    webviewController.registerReducer(
+        "setFilterState",
+        async (state, payload) => {
+            if (!state.filterState[payload.filterState.columnDef]) {
```
caohai, line 224: what does this if check cover? Wouldn't the part below this (Line232 - 237) already work if `state.filterState` doesn't have `columnDef` property?

caohai:
LGTM overall.
Is the screenshot still up to date? I'd suggest adding a gif/video that demos all the filtering features added in this PR.

src/reactviews/pages/QueryResult/resultGrid.tsx
```diff
@@ -60,6 +61,7 @@ export interface ResultGridProps {
         QueryResultReducers
     >;
     gridParentRef?: React.RefObject<HTMLDivElement>;
+    state: QueryResultState;
```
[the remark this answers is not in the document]
caohai: theme and state set can already be done via `webViewState` (a.k.a `state.state`) iirc, doesn't really matter so you can decide.

caohai (approved):

PR #18384, opened 2024-11-13
Remove min grid height for 1 grid

cssuh:
Removes the min grid height for scenarios with just 1 grid, this is consistent with the ADS behavior.
Before:
minSizeBefore

After:
minSize


caohai:
Can you add a screenshot/gif for the difference?

Benjin (approved):

Issue #18376, opened 2024-11-11
Actual Plan is broken for some queries


caohai:
`select 1`

PR #18411, opened 2024-11-14
Fixed actual plan not working for some queries

laurennat:
Fixes https://github.com/microsoft/vscode-mssql/issues/18376
actualPlanQueryBugFix

### DIFF
--- a/src/queryResult/queryResultWebViewController.ts
+++ b/src/queryResult/queryResultWebViewController.ts
@@ -216,7 +216,7 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
                     loadState: ApiStatus.Loading,
                     executionPlanGraphs: [],
                     totalCost: 0,
-                    xmlPlans: [],
+                    xmlPlans: {},
                 },
             }),
         };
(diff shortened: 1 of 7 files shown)


src/queryResult/queryResultWebViewController.ts
```diff
@@ -336,4 +336,8 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
         });
         return total;
     }
+
+    public recordLength(record: any): number {
```
Benjin, line 340: Move this to `utils`

src/queryResult/utils.ts
```diff
@@ -86,16 +86,36 @@ export function registerCommonRequestHandlers(
...
+                    1 &&
+            message.resultId ===
+                webviewViewController.recordLength(
+                    currentState.resultSetSummaries[message.batchId],
+                ) -
+                    1 &&
+            (!currentState.executionPlanState?.xmlPlans ||
+                !webviewViewController.recordLength(
+                    currentState.executionPlanState.xmlPlans,
+                ))
```
Benjin, line 115: Thanks for adding the comment block at the top!  For cascading boolean conditions like this one, I think it's _also_ nice to have comments on each "clause" in the conditional.

Might be worth checking to see if adding more parens tells Prettier to format it more nicely (it's weird that the `1` is on a different line from the measure it's being subtracted from)

laurennat: prettier removes "unnecessary" parens, so that's a no go

laurennat: for what it's worth, I do think moving the recordLength function to utils helps

Benjin (approved):

PR #18419, opened 2024-11-15
Change context menu button text to "Close"

cssuh:
Addresses https://github.com/microsoft/vscode-mssql/issues/18250
Before:
Screenshot 2024-11-15 at 11 35 02 AM

After:
Screenshot 2024-11-15 at 11 32 05 AM

### DIFF
--- a/src/reactviews/common/locConstants.ts
+++ b/src/reactviews/common/locConstants.ts
@@ -309,6 +309,7 @@ export class LocConstants {
             copy: l10n.t("Copy"),
             copyWithHeaders: l10n.t("Copy with Headers"),
             copyHeaders: l10n.t("Copy Headers"),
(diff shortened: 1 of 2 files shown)


aasimkhan30:
LGTM, why can't we just close it when user clicks the sort ascending or sort descending button?

cssuh:
> LGTM, why can't we just close it when user clicks the sort ascending or sort descending button?

We do that too, I can just get rid of this option entirely if it doesn't make sense

Benjin (approved):

Benjin:
There's still a bug where a user could end up opening multiple context menus, so until we fix that, I think we should keep the "close" option.  But I agree that clicking one of the sort options should also close it.

PR #18429, opened 2024-11-16
Change results pop-out prompt so that it doesn't block the results actually opening in the tab 

### LINKED ISSUE #18428: Prompt for setting default query result pane location shouldn't delay the pane from opening
When the user is prompted for a default behavior, the query results being opened in the new tab are delayed/blocked until the user has selected an option.  Since the user has already indicated that they want to pop-out the current result set, there's no reason to delay opening them.  The query results should pop up immediately, and the prompt should be shown asynchronously.

Benjin:
Fixes #18428

### DIFF
--- a/src/queryResult/queryResultWebViewController.ts
+++ b/src/queryResult/queryResultWebViewController.ts
@@ -116,6 +116,8 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
 
     private registerRpcHandlers() {
         this.registerRequestHandler("openInNewTab", async (message) => {
+            void this.createPanelController(message.uri);
+
             if (this.shouldShowDefaultQueryResultToDocumentPrompt) {
                 const response =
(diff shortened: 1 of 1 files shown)


caohai (approved):

src/queryResult/queryResultWebViewController.ts
```diff
@@ -116,6 +116,8 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
 
     private registerRpcHandlers() {
         this.registerRequestHandler("openInNewTab", async (message) => {
+            void this.createPanelController(message.uri);
```
caohai, line 119: should be still await herre?

Benjin: nope; intentionally not awaiting it so that it the prompt doesn't wait for the tab to fully open

PR #18446, opened 2024-11-21
Centralizing some common loc strings

Benjin:


aasimkhan30:
Since the xlf files don’t have duplicates for the same string, I’m not sure it would add much benefit. It could also cause issues down the line if someone changes a shared string without realizing it’s used in multiple places.

aasimkhan30 (approved):

Benjin:
Hmm, good point about the xlf files not duplicating strings.  I was mostly thinking this might still help us to be more consistent across the board with minor things like casing, grammar, or wording of promtps which in turn gives an air of quality, so I'll keep this, but I'll add a comment to the Common areas with a note about the impact of changing those strings.

PR #18459, opened 2024-11-27
Fixing typings for azdataGraph lib

aasimkhan30:


typings/azdataGraph.d.ts
```diff
@@ -1,6 +0,0 @@
-declare module 'azdataGraph/dist/build' {
```
Benjin, line 1: Double-checking: this was intentionally removed, and doesn't need to be replaced?

Issue #18003, opened 2024-08-27
Unexpected "Save Untitled File" prompt when saving XML as SQLPLAN file


kburtram:
Steps to Reproduce:

1. Run a query that generates query plan XML, for example the following snippet.

```SQL
SET SHOWPLAN_XML ON
GO
SELECT TOP 100 *
FROM sys.all_objects o
CROSS JOIN sys.all_views
ORDER BY o.object_id
```
2. Click the showplan XML link in the result grid to open a new document with the raw execution plan XML
3. Save the untitled XML to a .sqlplan file

Expected: The SQLPLAN custom editor is displayed, try this scenario in ADS for expected behavior.

Actual: SQLPLAN custom editor tries to show, but there are two SQLPLAN documents open (though they don't load correctly, likely due to [different issue](https://github.com/microsoft/vscode-mssql/issues/18002)).  Then an unexpected prompt is displayed to save "<ShowPlanXML" document, which no longer appears to exist as raw XML has already been saved to .sqlplan file.  This prompt appears related to the call to close the active document in the custom editor implementation.

The following screenshot shows the UX matching the above description.

![image](https://github.com/user-attachments/assets/84fa510d-e564-47aa-b6c7-8c97cc4eee12)

PR #18480, opened 2024-12-06
Fixes some query plan issues

### LINKED ISSUE #18349: Query plan properties title and table header should be sticky
Steps:
1. Open a query plan properties windows. 
2. Scroll down.

Current behavior:

1. The header and toolbar gets hidden when the properties are scrolled.
Start position:

Image

End position:

Image

Expected behavior:

1. Only table content should scroll. Rest all things (Title, toolbar and table header) should be sticky.

laurennat:
Fixes #18349, #18003

Misc query plan bug fixes

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -2062,12 +2062,6 @@ export default class MainController implements vscode.Disposable {
             );
         }
 
-        if (doc && doc.languageId === Constants.sqlPlanLanguageId) {
-            vscode.commands.executeCommand(
-                "workbench.action.closeActiveEditor",
-            );
-        }
-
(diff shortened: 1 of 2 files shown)


Benjin:
Looks like this fixes two of the three things for https://github.com/microsoft/vscode-mssql/issues/18349 (title and toolbar), but the table heading (`Name | Value`) still scroll out of view.  How easy is to for those to be stickied, too?

Benjin:
Looks like the default file format in the "save as" window is `.xml` instead of `.sqlplan`, which should be fixed.

lewis-sanchez (approved):

Benjin (approved):

Issue #18452, opened 2024-11-25
Message issues


dotvihar:
# Technical info

## MSSQL Extension Version:
> Name: SQL Server (mssql)
> Id: ms-mssql.mssql
> Version: 1.26.0

## VSCode Version:
> Version: 1.95.3 (user setup)
> Commit: f1a4fb101478ce6ec82fe9627c43efbf9e98c813
> Date: 2024-11-13T14:50:04.152Z
> Electron: 32.2.1
> ElectronBuildId: 10427718
> Chromium: 128.0.6613.186
> Node.js: 20.18.0
> V8: 12.8.374.38-electron.0
> OS: Windows_NT x64 10.0.22631

## OS Version:
> Edition: Windows 11 Enterprise
> Version: 23H2
> OS build: 22631.4460
> Experience: Windows Feature Experience Pack 1000.22700.1047.

---
# Message BUGs
## 1. Do not receive any messages during executions
I have a query that runs for about an hour. There are print statements in the code to help me track its execution, but I do not receive any messages during the execution.
```sql
-- code example
print @msg_body + ' finished. (' + convert(varchar, @current_dttm) + ')'
print 'Time elapsed: '
print '_ millisecond: ' + convert(varchar, (datediff(ms, @start_dttm, @current_dttm)))  + ', '
print '_ second: '      + convert(varchar, (datediff(s, @start_dttm, @current_dttm)))   + ', '
print '_ minute: '      + convert(varchar, (datediff(mi, @start_dttm, @current_dttm)));
```

## 2. The lines overlap in Messages pane
![Image](https://github.com/user-attachments/assets/cf57a8ec-d885-489c-b8e4-ade095f6e93c)

caohai:
Thanks @dotvihar for creating the issue!
We are aware of these two bugs (no msg && rendering issue) and we are actively working on a fix. 
For the no msg bug, you can track the progress of the fix at https://github.com/microsoft/vscode-mssql/issues/18447

sistefi:
## Hola el mismo problema, sin mensajes.
### No se da en todos los casos pero es muy largo el proseso de tener que verificar en la base de datos, para ver si todo quedo bien.

> Me acomode muy bien con la APP Grecias.

Issue #18447, opened 2024-11-21
[Rich Experience] Message pane empty after running command


EdCallahan:
- MSSQL Extension Version: 1.26.0
- VSCode Version: 1.95.3
- OS Version: Win11

Steps to Reproduce:

1.  Close any open Query Results (Preview) panes
2.  Run a command such as USE MASTER or an ALTER SCHEMA. Message pane is not populated with messages
3.  Run a SELECT. Now Message pane contains messages
4.  At this point, USE MASTER or ALTER SCHEMA, etc, generate messages on the Message pane

Thanks for your work on this project, VSCode is a nice place to so SQL work.

https://github.com/user-attachments/assets/6ae73ef5-36c3-409a-b439-6c97365d58a8

Fred-Wu:
I was about to report the same.  

If `enableNewQueryResultFeature` is set as true,  it shows nothing even when a query is being executed. 
It should behaviour like the old one, telling you

![Image](https://github.com/user-attachments/assets/fadf872d-6e4a-43ce-986d-4203c016cf47)


Also for an empty table gets queried and written to a temporary table, the message pane also shows nothing, which could not tell what's wrong with the table in the first place.

caohai:
@EdCallahan @Fred-Wu Thanks for reporting the bug. I can confirm the repro. I'll take a look at the issue.

Fred-Wu:
> [@EdCallahan](https://github.com/EdCallahan) [@Fred-Wu](https://github.com/Fred-Wu) Thanks for reporting the bug. I can confirm the repro. I'll take a look at the issue.
> 
> > for an empty table gets queried and written to a temporary table
> 
> [@Fred-Wu](https://github.com/Fred-Wu) Are you talking about something like this with source table being an empty table?
> 
> ```
> SELECT *  
> INTO #TempDestinationTable
> FROM Source_Table
> WHERE ...
> ```

Yes.  In the old interface, the message would show how many rows affected. But nothing appears in the new interface.

Issue #18405, opened 2024-11-14
Query execution message text isn't positioned properly


kburtram:
1. Run a query and look at the message

Expected: Text more in the left of the visible pane.

Actual: Text is way too far to the right.

![Image](https://github.com/user-attachments/assets/5a9f4446-0130-4300-9ef1-6836760cea2c)

For reference here's what it looks like in ADS:

![Image](https://github.com/user-attachments/assets/486e13bd-bdd1-401b-82f9-9e9af1e58fc0)

laurennat:
also, if there are multiple lines in the message, it ends up not spacing properly

![Image](https://github.com/user-attachments/assets/fdce0f0d-a1c4-4922-a78a-2f9a9d9e4a00)

caohai:
> also, if there are multiple lines in the message, it ends up not spacing properly
> 
> ![Image](https://github.com/user-attachments/assets/fdce0f0d-a1c4-4922-a78a-2f9a9d9e4a00)

@laurennat Can you share the repro query for this?

laurennat:
select 2
for json auto

croblesm:
@caohai 
Same issue. This is the query I used:

```sql
INSERT INTO authors (id, first_name, middle_name, last_name) VALUES
(6, 'Frank', 'Patrick', 'Herbert'), 
(7, 'Orson', 'Scott', 'Card'), 
(8, 'Suzanne', NULL, 'Collins'),
(9, 'Andy', NULL, 'Weir'), 
(10, 'Yuval', 'Noah', 'Harari');
```

Result:
![Image](https://github.com/user-attachments/assets/efa60b34-f62b-430a-8c66-6e5331d18ee8)

PR #18483, opened 2024-12-06
Fix a few message pane issues

caohai:
This PR fixes https://github.com/microsoft/vscode-mssql/issues/18405
Before:
image
After:
image

This PR fixes https://github.com/microsoft/vscode-mssql/issues/18447
The issue was caused by `messageGridHeight` not getting set when there's no result set.
Before:

https://github.com/user-attachments/assets/ef77d6dd-cdbb-4f45-9958-6c8e86c7f583

After:

https://github.com/user-attachments/assets/0c3b9513-60da-45cc-83e1-09abf4dc5576

This PR fixes https://github.com/microsoft/vscode-mssql/issues/18452
There are two factors causing this issue: space wrapping and multiline message. The ideal solution would be dynamically setting the height of the message row based on the content to allow free resizing and it's non-trivial to implement and the datagrid with virtualization support in FluentUI doesn't support this, one workaround would be computing the height at run time which can be expensive. See discussion here: https://github.com/bvaughn/react-window/issues/6
For now I'm doing these two things to fix the issue which should be good enough for now:
1: disable space wrapping
2: split a multiline message into multiple messages
Created an issue here to track the ideal solution: https://github.com/microsoft/vscode-mssql/issues/18484
Before:
image
image

After:
image
image

This PR also fixes the missing line text in Copy All message button:
image

### DIFF
--- a/src/reactviews/pages/QueryResult/queryResultUtils.ts
+++ b/src/reactviews/pages/QueryResult/queryResultUtils.ts
@@ -5,6 +5,7 @@
 
 import { ColorThemeKind } from "../../common/vscodeWebviewProvider";
 import { QueryResultWebviewState } from "../../../sharedInterfaces/queryResult";
+import * as qr from "../../../sharedInterfaces/queryResult";
 
 export const saveAsCsvIcon = (theme: ColorThemeKind) => {
     return theme === ColorThemeKind.Light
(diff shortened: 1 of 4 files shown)


src/reactviews/pages/QueryResult/queryResultUtils.ts
```diff
@@ -30,3 +31,15 @@ export function hasResultsOrMessages(state: QueryResultWebviewState): boolean {
         state.messages.length > 0
     );
 }
+
+export const splitMessages = (messages: qr.IMessage[]): qr.IMessage[] => {
```
cssuh, line 35: Could you add a comment explaining this function? I'm not super familiar with the .split function syntax with /r & /n

cssuh (approved):

src/reactviews/pages/QueryResult/queryResultUtils.ts
caohai: added

PR #18500, opened 2024-12-11
Apply mssql font settings to result grid

### LINKED ISSUE #18351: New Query Results pane does not follow existing font size and font family settings
> the new pane does not seem to respect the following settings:
> mssql.resultsFontFamily
> mssql.resultsFontSize
> 
> Thank you for pointing me in the right direction. 

 _Originally posted by @gwduvalljr-AT in [#1553](https://github.com/microsoft/vscode-mssql/issues/1553#issuecomment-2453482551)_

cssuh:
Fixes the query result grid to use the `mssql.resultsFontFamily` and `mssql.resultsFontSize` settings.
Example with font size: 20 & mac default font family
Screenshot 2024-12-11 at 3 09 08 PM

Fixes #18351

### DIFF
--- a/src/queryResult/queryResultWebViewController.ts
+++ b/src/queryResult/queryResultWebViewController.ts
@@ -55,6 +55,7 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
             },
             executionPlanState: {},
             filterState: {},
+            fontSettings: {},
         });
 
         void this.initialize();
@@ -74,6 +75,7 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
                         isExecutionPlan: false,
                         executionPlanState: {},
                         filterState: {},
+                        fontSettings: {},
                     };
                 }
             });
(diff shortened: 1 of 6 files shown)


Benjin:
Can you add a screenshot showing the change if these are set to a different font and larger size?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -170,6 +170,18 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
         this.registerRequestHandler("getWebviewLocation", async () => {
             return qr.QueryResultWebviewLocation.Panel;
         });
+        this.registerRequestHandler("getFontFamily", async () => {
+            console.log("getFontFamily");
+            return this._vscodeWrapper
+                .getConfiguration(Constants.extensionName)
+                .get(Constants.extConfigResultKeys[3]);
```
Benjin, line 177: indexing into an array for a specific value feels like a recipe for bugs when values change.  Can you update the `extConfigResultKeys` array to instead be a string-type enum?

```
export const extConfigResultKeys = [
    shortcuts = "shortcuts",
    ...
];
```

src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -110,15 +108,39 @@ function getAvailableHeight(
 }
 
 export const QueryResultPane = () => {
-    const classes = useStyles();
     const state = useContext(QueryResultContext);
+    const [fontFamily, setFontFamily] = useState("");
+    const [fontSize, setFontSize] = useState("");
+
+    useEffect(() => {
+        const getFontFamily = async () => {
```
Benjin, line 116: what does this look like if these settings aren't set?  they should default to `var(--vscode-editor-font-size)` and `var(--vscode-editor-font-family)`

Benjin: Looks like the default behavior is the UI font, not the editor font.

Without these settings, it should look like this:
image

src/queryResult/queryResultWebviewPanelController.ts
```diff
@@ -74,6 +74,18 @@ export class QueryResultWebviewPanelController extends ReactWebviewPanelControll
...
             return qr.QueryResultWebviewLocation.Document;
         });
+        this.registerRequestHandler("getFontFamily", async () => {
+            console.log("getFontFamily");
+            return this._vscodeWrapper
+                .getConfiguration(Constants.extensionName)
+                .get(Constants.extConfigResultKeys[3]);
+        });
+        this.registerRequestHandler("getFontSize", async () => {
+            console.log("getFontFamily");
```
kburtram, line 84: looks like we're logging the wrong function name?  Maybe we should use the logger class rather than console calls?

cssuh: Yeah, this logging was just for my debugging, I'll be removing it before I merge the PR and I will add more detailed logs using the logger class in the future

cssuh:
Chatted with @Benjin  and we are expanding the scope of this fix to also adjust row / column sizes dynamically based on the font size that is input, so I will be working on adding that in a larger PR that will be for the January release.

Issue #18489, opened 2024-12-10
Running a query steals focus from active tab


link-transport-mickoz:
- MSSQL Extension Version: 1.26.0
- VSCode Version: 1.95.3
- OS Version: Windows 11

Hi,

Running a query steals focus from active tab. I found [this issue](https://github.com/microsoft/vscode-mssql/issues/1540) from 2019 that fixed this problem, but it seems it came back.

Steps to Reproduce:

1. Run a query
2. Tab or query pane has focus instead of editor tab with the query.

Expected behaviour:

After running a query, tab with SQL query should be in focus.

cssuh:
Hi @link-transport-mickoz, do you have this feature (enable rich experiences) enabled or disabled when you see this happen?
![Image](https://github.com/user-attachments/assets/a64ff274-aa3b-41ea-aaff-256f6a716324)

link-transport-mickoz:
I have it enabled.

I disabled it and focus stays with the editor.

cssuh:
@link-transport-mickoz thanks for the quick response! We will work on getting this fixed.

caohai:
This issue has been fixed by #18503 and #18505 and it will go out with next release.

PR #18503, opened 2024-12-12
Preserve focus during query execution

caohai:
This PR fixes https://github.com/microsoft/vscode-mssql/issues/18489
This PR fixes the focus changes during query execution with the new query result feature.
With this change, the focus is preserved during:
1. Query Execution with "Run Query" button for both the bottom panel and the new tab.
2. Clicking "Hide this panel link" when there's no result in the bottom result area.
3. Clicking "Reveal Query Result" button
4. Clicking "Open in New Tab" button


caohai:
taking a look at the test failure

Benjin (approved):

PR #18533, opened 2024-12-20
DRAFT PR - Edit Data

lewis-sanchez:


package.json
```diff
@@ -484,6 +484,11 @@
           "when": "view == objectExplorer && viewItem =~ /\\btype=(AggregateFunction|PartitionFunction|ScalarValuedFunction|StoredProcedure|TableValuedFunction|View)\\b/",
           "group": "MS_SQL@4"
         },
+        {
+          "command": "mssql.editData",
+          "when": "view == objectExplorer && viewItem =~  /\\btype=(Table)\\b/",
```
Benjin, line 489: if you gate the feature behind the "enable preview features" flag as well, then it'll be easier to check in changes as they're ready.  One thing we want to avoid is to wait until the feature is almost done, then have a huge PR because that's more difficult to effectively review.

src/editData/editDataWebViewController.ts
```diff
@@ -0,0 +1,259 @@
...
+        context: vscode.ExtensionContext,
+        private node: TreeNodeInfo,
+        private readonly connectionManager: ConnectionManager,
+        private readonly editDataService: EditDataService,
+        data?: EditDataWebViewState,
+    ) {
+        super(
+            context,
+            "editData",
+            data ?? {
```
Benjin, line 36: This is the initial state for `EditDataWebViewState`, right?  Probably good to have this situated with the definition for EditDataWebViewState.

PR #18549, opened 2025-01-07
Add Auto Column Size

### LINKED ISSUE #18089: Support auto column width in query result grid

### LINKED ISSUE #18534: Double clicking result pane width button doesn't expand result
- MSSQL Extension Version: MSSQL Extension Version: 1.27.0
- VSCode Version: VSCode Version: 1.95.1
- OS Version: OS Version: Windows 11 Enterprise 10.0.22631 Build 22631

Steps to Reproduce:

1. When running a query and viewing the results, I used to be able to double-click the border of the column that I wanted to expand and it would open to fit the length of the text, but now it doesn't appear to be working.

Also, is there any option to just have this expand the width automatically instead of having to do it manually every time? It gets tiresome when doing data analysis to constantly have to expand the width to see my results.

cssuh:
Gives the proper space for column widths.  Also enables double clicking on a column border to snap to the correct size.

In the first screenshot you can see that the column headers have an ellipsis but in the second screenshot, they are sized correctly.

Before:
Screenshot 2025-01-09 at 6 14 07 PM

After:
Screenshot 2025-01-09 at 6 13 43 PM

Note: Instead of adding an additional max column size config option, I hardcoded it at 400px to minimize the number of settings options we have for this extension.

Double-click feature:
autoColumnSize

Note: this feature will expand the column to fit the header or any visible row data (whichever is larger), but won't shrink a cell that's already been manually expanded past the size of the header or visible row data.

Fixes #18089
Fixes #18534


localization/xliff/vscode-mssql.xlf
```diff
@@ -1539,6 +1539,9 @@
     <trans-unit id="mssql.query.executionTimeout">
       <source xml:lang="en">An execution time-out of 0 indicates an unlimited wait (no time-out)</source>
     </trans-unit>
+    <trans-unit id="mssql.resultsGrid.autoSizeColumns">
+      <source xml:lang="en">Auto size the columns width on initial results. Could have performance problems with a large number of columns or large cells</source>
```
Benjin, line 1543: In your testing, did you find thresholds where this starts seeing performance issues?  We should know more about when we start hitting issues with this enabled.

Benjin:
Can you add a video of the double-click to expand a column to the PR description?

Benjin:
Last week, we'd talked about it remembering when a user adjusts a column width for that result set if they rerun the query; is that part of this PR?

src/reactviews/pages/QueryResult/table/utils.ts
```diff
@@ -75,3 +75,18 @@ export function tryCombineSelections(selections: ISlickRange[]): ISlickRange[] {
     }
     return [unifiedSelection];
 }
+
+export function deepClone<T>(obj: T): T {
```
Benjin, line 79: Looks like this is [from VSCode](https://github.com/microsoft/vscode/blob/5bd3d12fb18047ae33ac4b171dee3cd044b6f3d4/src/vs/base/common/objects.ts#L8)?  Can you add a comment with a link to the source in case we need to update it.

Benjin, line 79: This seems like it could be useful in other react views; let's put it in `reactviews/common/utils.ts`

src/reactviews/pages/QueryResult/table/table.ts
```diff
@@ -164,6 +166,13 @@ export class Table<T extends Slick.SlickData> implements IThemable {
             new CopyKeybind(this.uri, this.resultSetSummary, this.webViewState),
         );
 
+        this.registerPlugin(
+            new AutoColumnSize({
+                maxWidth: MAX_COLUMN_WIDTH_PX,
```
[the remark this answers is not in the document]
cssuh: No, I thought we discussed that we wouldn't include a user option for this because we don't want to add too many config options.  Should I change this?

package.nls.json
```diff
@@ -100,6 +100,7 @@
 "mssql.format.placeCommasBeforeNextStatement":"should commas be placed at the beginning of each statement in a list e.g. ', mycolumn2' instead of at the end e.g. 'mycolumn1,'",
 "mssql.format.placeSelectStatementReferencesOnNewLine":"Should references to objects in a select statements be split into separate lines? E.g. for 'SELECT C1, C2 FROM T1' both C1 and C2 will be on separate lines",
 "mssql.applyLocalization":"[Optional] Configuration options for localizing into VSCode's configured locale (must restart VSCode for settings to take effect)",
+"mssql.resultsGrid.autoSizeColumns": "Auto size the columns width on initial results. Could have performance problems with a large number of columns or large cells",
```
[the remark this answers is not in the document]
cssuh: I went with "Automatically adjust the column widths based on the visible rows in the result set"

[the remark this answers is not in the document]
cssuh: I'll do some perf testing for this to find out more

cssuh:
> Last week, we'd talked about it remembering when a user adjusts a column width for that result set if they rerun the query. Is that part of this PR? If so, can you add a short video of how that works? (btw, separate video per topic/feature is best)

This part slipped my mind, I'll add it shortly.

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+
+    /**
+     * For each column, find the max width of the texts in the first MAX_ROW_TO_SCAN rows.
+     * @param columnDefs Column definitions of all columns that need to be resized
+     * @param colIndices Column indices of all columns that need to be resized
+     * @returns An array of the max widths of each column
+     */
+    private getMaxColumnTextWidths(
```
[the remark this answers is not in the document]
cssuh: Yes I tested this and it seems fine

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,376 @@
...
+    private getMaxColumnTextWidth(
+        columnDef: Slick.Column<T>,
+        colIndex: number,
+    ): number {
+        let texts: Array<string> = [];
+        let rowEl = this.createRow();
+        let data = this._grid.getData() as Slick.DataProvider<T>;
+        let viewPort = this._grid.getViewport();
+        let start = Math.max(0, viewPort.top);
+        let end = Math.min(data.getLength(), viewPort.bottom);
```
kburtram, line 237: are we going through all the data in the result set, or the visible viewPort? I'm not sure what viewPort.bottom is.  But we probably shouldn't go through the whole result set, maybe 100 or 1000 rows is totally fine to get a general shape of the data.

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,376 @@
...
+        texts: string[],
+        columnDef: Slick.Column<T>,
+        colIndex: number,
+        data: Slick.DataProvider<T>,
+        rowEl: JQuery,
+    ): JQuery | HTMLElement | string {
+        let max = 0,
+            maxTemplate: JQuery | HTMLElement | string | undefined;
+        let formatFun = columnDef.formatter;
+        texts.forEach((text, index) => {
```
kburtram, line 300: this looks expensive at scale (e.g. lots of rows)

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
cssuh: It just goes through the visible viewPort

Benjin: Recording here to make sure it doesn't get lost: the suggestion from sync yesterday was to calculate the width using the first 50-100 rows of data, rather than just the viewport since the viewport is possibly only a small slice.

Benjin:
@cssuh 

> > Last week, we'd talked about it remembering when a user adjusts a column width for that result set if they rerun the query. Is that part of this PR? If so, can you add a short video of how that works? (btw, separate video per topic/feature is best)
> 
> This part slipped my mind, I'll add it shortly.

Also okay if that comes in a quick follow-up PR.

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
cssuh: @kburtram Did some perf testing on this, and since it just goes by the visible viewport, it only did the first 10 rows of each column for me (30 rows if I expand the grid to my full screen).  With 1024 columns (which seems like the upper recommended limit for # of columns), it took a couple seconds extra compared to if I had the setting off.

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
cssuh: @kburtram @Benjin increased it to 50 rows (don't want to make the calculation too expensive).

cssuh:
> @cssuh
> 
> > > Last week, we'd talked about it remembering when a user adjusts a column width for that result set if they rerun the query. Is that part of this PR? If so, can you add a short video of how that works? (btw, separate video per topic/feature is best)
> > 
> > 
> > This part slipped my mind, I'll add it shortly.
> 
> Also okay if that comes in a quick follow-up PR.

I think I'll add this in a follow-up PR, and I realized this is not for if they re-run the query (since that would generate a new query result set and result set grid), but for if they leave the result pane and then come back (like switching to messages tab and then returning)

src/reactviews/pages/QueryResult/table/table.ts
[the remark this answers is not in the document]
Benjin: No, you're right about that.  What I meant was that if I have a column with lots of text such that the content is wider than `MAX_COLUMN_WIDTH_PX`, autosize will limit itself to `MAX_COLUMN_WIDTH_PX`, but I - as the user - will still be able to drag the column to be even wider to see more of my data, right?

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
Benjin: Can you add a telemetry point with timing and number of columns here so that we can detect any egregious times to autosize?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -131,6 +118,16 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
                         this._queryResultStateMap.set(uri, state);
                     }
                 }
+                if (
+                    e.affectsConfiguration("mssql.resultsGrid.autoSizeColumns")
+                ) {
+                    for (const [uri, state] of this._queryResultStateMap) {
+                        state.autoSizeColumns = this._vscodeWrapper
```
Benjin, line 125: `getAutoSizeColumnsConfig()`

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,383 @@
...
+        let data = this._grid.getData() as Slick.DataProvider<T>;
+        let dataLength = data.getLength();
+        let viewPort = this._grid.getViewport();
+        let start = Math.max(0, viewPort.top);
+        let end = Math.min(dataLength, viewPort.bottom);
+        if (end < 50 && dataLength >= 50) {
+            end = 50;
+        } else if (end < 50 && dataLength < 50) {
+            end = dataLength;
+        }
```
Benjin, line 199: `50` should be pulled out into a constant

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,383 @@
...
+    private getMaxColumnTextWidths(
+        columnDefs: Slick.Column<T>[],
+        colIndices: number[],
+    ): number[] {
+        let data = this._grid.getData() as Slick.DataProvider<T>;
+        let dataLength = data.getLength();
+        let viewPort = this._grid.getViewport();
+        let start = Math.max(0, viewPort.top);
+        let end = Math.min(dataLength, viewPort.bottom);
+        if (end < 50 && dataLength >= 50) {
```
Benjin, line 195: add a comment explaining what's going on here, something like
`// limit column width calculation to 50 rows`

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
Benjin, line 199: I think this can be simplified to

```
if (end < 50) {
    end = Math.min(50, dataLength);
}
```

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,383 @@
...
+import { mixin } from "../objects";
+
+export interface IAutoColumnSizeOptions extends Slick.PluginOptions {
+    maxWidth?: number;
+    autoSizeOnRender?: boolean;
+    extraColumnHeaderWidth?: number;
+}
+
+const defaultOptions: IAutoColumnSizeOptions = {
+    maxWidth: 212,
```
Benjin, line 18: what's this `maxWidth`?  Should it be 400px?

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,383 @@
...
+            return;
+        }
+
+        e.preventDefault();
+        e.stopPropagation();
+
+        this.reSizeColumn(headerEl, columnDef);
+    }
+
+    private reSizeColumn(headerEl: JQuery, columnDef: Slick.Column<T>) {
```
Benjin, line 155: nit: "resize" instead of "reSize"

src/reactviews/pages/QueryResult/table/table.ts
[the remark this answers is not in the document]
cssuh: oh I see, yes that's correct - this max width is only how wide the auto-sizing will go, but users can drag the columns to be however wide they want.

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
cssuh: That's a good idea, I'll add that

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,416 @@
...
+    private getMaxColumnTextWidths(
+        columnDefs: Slick.Column<T>[],
+        colIndices: number[],
+    ): number[] {
+        let data = this._grid.getData() as Slick.DataProvider<T>;
+        let dataLength = data.getLength();
+        let viewPort = this._grid.getViewport();
+        let start = Math.max(0, viewPort.top);
+        let end = Math.min(dataLength, viewPort.bottom);
+        // limit column width calculation to 50 rows
```
Benjin, line 218: update comment to swap 50 for the new const

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,416 @@
...
+                max = length;
+                maxTemplate = template || text;
+            }
+        });
+        let endTime = Date.now();
+        let timeElapsed = endTime - startTime;
+        let telemetryEvent: WebviewTelemetryActionEvent = {
+            telemetryView: TelemetryViews.QueryResult,
+            telemetryAction: TelemetryActions.AutoColumnSize,
+            additionalMeasurements: {
```
Benjin, line 359: also include number of columns and number of rows scanned

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,423 @@
...
+        });
+        let column = allColumns[colIndex];
+
+        let autoSizeWidth =
+            Math.max(
+                headerWidth,
+                this.getMaxColumnTextWidth(columnDef, colIndex),
+            ) + 1;
+
+        // Only resize if the current width is smaller than the new width.
```
Benjin, line 195: In testing/playing around, this felt a bit funky (felt like a bug until I remembered that I'd seen a comment about it).  I know ADS does it this way, but Excel will resize both directions.

Benjin:
The autosizing appears to work pretty well.  No regressions compared to having the setting disabled.

I found a few other things that I'll record here as notes (not for this PR, I'll make issues):
* some console.log statements need cleaning up, or at least improvement to be more useful for debugging
* running `select * from sys.objects a cross join sys.objects b` (17k rows) performs okay, but running that twice in the same batch seems to scale poorly (worse than 2x)

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
cssuh: Do you think I should change this to resize in both directions then?

Benjin: Yeah, let's allow it to resize in both directions.

Benjin (approved):

lewis-sanchez (approved):

src/reactviews/pages/QueryResult/table/plugins/autoColumnSize.plugin.ts
```diff
@@ -0,0 +1,419 @@
...
+                maxTemplate = template || text;
+            }
+        });
+        let endTime = Date.now();
+        let timeElapsed = endTime - startTime;
+        let telemetryEvent: WebviewTelemetryActionEvent = {
+            telemetryView: TelemetryViews.QueryResult,
+            telemetryAction: TelemetryActions.AutoColumnSize,
+            additionalMeasurements: {
+                timeElapsed: timeElapsed,
```
Charles-Gagnon, line 361: @cssuh I'd strongly suggest adding units to this value - e.g. `timeElapsedMs`

cssuh: Will do

PR #18550, opened 2025-01-09
Option to automatically disable language service on detected Non-TSQL files

laurennat:
PR for giving users an option to automatically disable the language service when the extension detects non T-Sql syntax/ keywords.

The bulk of github's detected line changes come from the added non t-sql keyword list.

tooManySyntaxErrosCase
languageServiceCase1


croblesm:
Hi @laurennat just a quick comment about the prompt. Could you please update the text to say "non-SQL Server" instead of "MSSQL". Another option is to say "non-TSQL" as T-SQL is SQL Server SQL language name.

src/controllers/mainController.ts
```diff
@@ -2101,6 +2113,92 @@ export default class MainController implements vscode.Disposable {
         this._lastSavedUri = savedDocumentUri;
     }
 
+    /**
+     * Called by VS Code when a text document is changed.
+     * @param change The change event that was detected
+     */
+    public async onDidChangeTextDocument(
```
aasimkhan30, line 2120: Please add some unit tests for this code in a follow up PR.

src/languageservice/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    return keywords;
+}
+
+export enum LanguageServiceOptions {
+    SwitchDisabled = 0,
+    SwitchEnabled = 1,
+    SwitchUnset = 2,
+}
+
+export function hasNonTSqlKeywords(
```
aasimkhan30, line 24: Good candidate for some unit tests.

src/controllers/mainController.ts
```diff
@@ -2101,6 +2113,92 @@ export default class MainController implements vscode.Disposable {
...
+
+        const sources: vscode.Diagnostic[] = docErrors.filter(
+            (error) => error.source !== undefined,
+        );
+
+        if (
+            // This condition is true if there are any errors in the document
+            // from another sql language service
+            sources.length ||
+            docErrors.length >= 50 ||
```
aasimkhan30, line 2147: Do we want to make this number configurable?

src/languageservice/nonTSqlKeywords.ts
```diff
@@ -0,0 +1,694 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * This list was built by finding keywords in the MySql documentation,
+ * and  PostGreSql documentation that are not present in the TSQL documentation
+ * of keywords.
+ */
+export const keywords: Set<string> = new Set([
```
aasimkhan30, line 11: What if these keywords appear in object names? It's possible for keywords like 'cost' to be used as object names. With the current logic, this might trigger the prompt. I'm not sure what the best solution would be, but it's something to consider.

src/languageservice/nonTSqlKeywords.ts
```diff
@@ -0,0 +1,694 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * This list was built by finding keywords in the MySql documentation,
+ * and  PostGreSql documentation that are not present in the TSQL documentation
```
aasimkhan30, line 8: If possible, please add links to the docs.

src/constants/locConstants.ts
```diff
@@ -561,6 +561,9 @@ export let flavorDescriptionMssql = l10n.t(
 export let flavorDescriptionNone = l10n.t(
     "Disable intellisense and syntax error checking on current document",
 );
+export let autoDisableNonTSqlLanguageServicePrompt = l10n.t(
+    "Possible non-SQL Server SQL file detected. Automatically disable Intellisense for all non-SQL Server SQL files?",
```
Benjin, line 565: @croblesm, can you help refine this string to make sure the branding references are correct? Also, I feel like saying "SQL" so many times makes it a bit confusing.

src/languageservice/nonTSqlKeywords.ts
Benjin: Maybe we should trim some likely candidates for object names from this list?

src/languageservice/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    text: string,
+    nonTSqlKeywords: Set<string>,
+): boolean {
```
Benjin, line 27: For some reason, I envisioned this happening inside the Parser itself or within STS, rather than at the VS Code layer.  The Parser code is already parsing the entire document to identifying comments/literals/object names, and the STS code (see `HandleSyntaxParseRequest`) is enumerating errors.

I'm also concerned about the performance of using a slew of regexes on what could be a large script file.  Have you tried this with a long script (e.g. a deployment script)?

src/constants/locConstants.ts
Benjin: Looks like you already did in another comment; thanks!

Issue #18512, opened 2024-12-13
Attempting to highlight text in column option textbox ends up dragging the entire row for reordering


Benjin:
Repro:
1. open table designer for new table
2. Rename default "Id" column to something longer, e.g. "ThisIsAnIdentificationNumber"
3. Attempt to highlight just a portion of the column name, e.g. "Identification" with your mouse by clicking and dragging

Expected:
* Text is able to be highlighted just like a normal textbox

Reality:
* Cursor is placed, but dragging the mouse results in dragging the entire row for a reorder operation

https://github.com/user-attachments/assets/cd95db74-04ef-4f7d-9983-a2ed545ca784

PR #18551, opened 2025-01-14
Fixing text highlighting in table designer.

aasimkhan30:
Fixes: #18512 

Previously the drag handlers were placed on the entire row which was messing up text highlighting. This PR moves the drag handlers to the drag button in the `DesignerTable` component from the entire row.

https://github.com/user-attachments/assets/d9d30708-1bbe-4e09-8e0d-c614af9f47b8


Benjin:
Can you add before/after videos?

Benjin (approved):

Issue #18511, opened 2024-12-13
Schema list in Table Designer should be ordered logically


Benjin:
At the very least, alphabetically (by lowercased version, IMO, so that `TestSchema` and `testSchema2` are next to each other.

I think we could go a step further and put all the user-created schemas first, and put the system schemas at the end.  Maybe order by schema's `principal_id` if we have that property?

![Image](https://github.com/user-attachments/assets/0d3fe01f-8e0c-4bb8-80ec-e75ffe4f4350)

PR #18552, opened 2025-01-14
Logically ordering schemas in table designer.

aasimkhan30:
Fixes: #18511 
image


src/reactviews/pages/TableDesigner/designerMainPane.tsx
```diff
@@ -145,6 +145,41 @@ export const DesignerMainPane = () => {
         }
     }
 
+    function getSortedSchemaValues() {
```
Benjin, line 148: Can this be done in the controller so that it doesn't need to sort the schemas on every render?

Benjin (approved):

PR #18557, opened 2025-01-15
Automatically disable language service when non T-SQL sql language is detected

laurennat:
Allow users to disable language service if another non Tsql language flavor is detected.

STS PR: https://github.com/microsoft/sqltoolsservice/pull/2438

tooManySyntaxErrosCase
In this case, the prompt is triggered by having too many errors in the script, which would be typical of a longer, non TSql script.

languageServiceCase1
This script triggers the prompt by using keywords that aren't TSQL keywords


localization/l10n/bundle.l10n.json
```diff
@@ -665,6 +665,7 @@
   "Choose SQL Language": "Choose SQL Language",
   "Use T-SQL intellisense and syntax error checking on current document": "Use T-SQL intellisense and syntax error checking on current document",
   "Disable intellisense and syntax error checking on current document": "Disable intellisense and syntax error checking on current document",
+  "Possible non-SQL Server SQL file detected. Automatically disable Intellisense for all non-SQL Server SQL files?": "Possible non-SQL Server SQL file detected. Automatically disable Intellisense for all non-SQL Server SQL files?",
```
aasimkhan30, line 668: Maybe `Non-SQL Server SQL file detected. Disable IntelliSense for such files?`

cssuh:
Can you add a short description to each gif? just so I know what you're trying to show

cssuh (approved):

Benjin:
Prompting to disable IntelliSense after one detected keyword feels aggressive, but maybe that's just because your demo vid types the non-MSSQL keyword as the very first word.  Let's leave it for now and see what feedback we get.

src/controllers/connectionManager.ts
```diff
@@ -421,6 +429,64 @@ export default class ConnectionManager {
...
+                        SqlToolsServerClient.instance,
+                        event.ownerUri,
+                        Constants.noneProviderName,
+                        this._statusView,
+                    );
+
+                    sendActionEvent(
+                        TelemetryViews.QueryEditor,
+                        TelemetryActions.DisableLanguageServiceForNonTSqlFiles,
+                    );
```
Benjin, line 468: You'll want to capture telemetry for both responses.

package.json
```diff
@@ -1450,6 +1450,11 @@
           "default": 2097152,
           "description": "%mssql.query.maxXmlCharsToStore%"
         },
+        "mssql.autoDisableNonTSqlLanguageService": {
+          "type": "number",
```
Benjin, line 1454: I think this makes more sense as a boolean with "undefined" permissible:

https://code.visualstudio.com/api/references/contribution-points#Configuration-example

```
        "typescript.tsdk": {
          "type": ["string", "null"],
          "default": null,
          "description": "Specifies the folder path containing the tsserver and lib*.d.ts files to use."
        }
```

src/constants/locConstants.ts
```diff
@@ -561,6 +561,9 @@ export let flavorDescriptionMssql = l10n.t(
 export let flavorDescriptionNone = l10n.t(
     "Disable intellisense and syntax error checking on current document",
 );
+export let autoDisableNonTSqlLanguageServicePrompt = l10n.t(
+    "Non-SQL Server SQL file detected. Disable IntelliSense for such files?",
```
Benjin, line 565: "for such files" feels a bit funky.  General rule of thumb is that docs and UI strings should be written at a 5th grade level so that non-fluent speakers aren't confused.

Maybe "Non-SQL Server syntax detected.  Would you like to disable IntelliSense for all detected non-SQL Server scripts?"

laurennat: this is what carlos suggested

.vscode/launch.json
```diff
@@ -46,7 +46,7 @@
 				// Uncomment this to use a specified version of STS, see
 				// https://github.com/microsoft/vscode-mssql/blob/main/DEVELOPMENT.md#using-mssql_sqltoolsservice-environment-variable
 				// for more details
-				// "MSSQL_SQLTOOLSSERVICE": "<Path to STS>"
+				"MSSQL_SQLTOOLSSERVICE": "C:/Repos/sqltoolsservice/src/Microsoft.SqlTools.ServiceLayer/bin/Release/net8.0/win-x64/publish/"
```
Benjin, line 49: revert this

src/constants/locConstants.ts
Benjin: Find by me, then!

src/controllers/connectionManager.ts
```diff
@@ -421,6 +426,77 @@ export default class ConnectionManager {
...
+            if (autoDisable === false) {
+                return;
+            } else if (autoDisable) {
+                changeLanguageServiceForFile(
+                    SqlToolsServerClient.instance,
+                    event.ownerUri,
+                    Constants.noneProviderName,
+                    this._statusView,
+                );
+            } else {
```
Benjin, line 447: some comments helping readers through this flow would be nice

// autoDisable set to false, so do nothing

// autoDisable set to true, so change language service for the file to NONE

// autoDisable setting not set, so ask the user if they want to use it

Benjin (changes requested):

aasimkhan30 (approved):

Benjin (approved):

Benjin:
Still needs STS bump (pipeline in progress).  PR looks ready to go once that's added.

PR #18558, opened 2025-01-16
Adding a prompt to file a bug on Github if the user rates MSSQL poorly

Benjin:
After the feedback form is submitted, users who rate the extension poorly (NPS < 7 or NSAT dissatisfied) are asked if they'd like to file a github issue.  Any comments they included in the feedback prompt are prepopulated in the github issue.

Feedback prompt:
image

Prepopulated Github issue:
image

### DIFF
--- a/src/constants/locConstants.ts
+++ b/src/constants/locConstants.ts
@@ -775,6 +775,10 @@ export class UserSurvey {
             args: [featureName],
             comment: ["{0} is the feature name"],
         });
+    public static fileABugPrompt = l10n.t(
+        "Are you encountering a bug that you would like to submit more details about?",
+    );
+    public static submitBug = l10n.t("Submit a bug");
 }
 
 export class Webview {
--- a/src/constants/locConstants.ts
+++ b/src/constants/locConstants.ts
@@ -775,10 +775,10 @@ export class UserSurvey {
             args: [featureName],
             comment: ["{0} is the feature name"],
         });
-    public static fileABugPrompt = l10n.t(
-        "Are you encountering a bug that you would like to submit more details about?",
+    public static fileAnIssuePrompt = l10n.t(
+        "Encountering a problem?  Share the details with us by opening a GitHub issue so we can improve!",
     );
-    public static submitBug = l10n.t("Submit a bug");
+    public static submitIssue = l10n.t("Submit an issue");
 }
 
 export class Webview {
--- a/src/nps/userSurvey.ts
+++ b/src/nps/userSurvey.ts
@@ -6,6 +6,7 @@
 import * as constants from "../constants/constants";
 import * as locConstants from "../constants/locConstants";
 import * as vscode from "vscode";
+import * as os from "os";
 
 import {
     Answers,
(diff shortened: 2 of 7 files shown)


src/nps/userSurvey.ts
```diff
@@ -38,83 +39,24 @@ export class UserSurvey {
...
-                    });
-                });
-                sendSurveyTelemetry("nps", answers);
-                await globalState.update(IS_CANDIDATE_KEY, false);
-                await globalState.update(SKIP_VERSION_KEY, extensionVersion);
+                await this.launchSurvey("nps", state);
+
+                // TODO: revert before PR
+                //await globalState.update(IS_CANDIDATE_KEY, false);
+                //await globalState.update(SKIP_VERSION_KEY, extensionVersion);
```
cssuh, line 59: Do you need to revert this?

src/nps/userSurvey.ts
```diff
@@ -164,9 +107,65 @@ export class UserSurvey {
                 resolve({});
             });
         });
-        sendSurveyTelemetry(surveyId, answers);
+
+        // TODO: revert before PR
+        console.log(surveyId);
+        console.log(answers);
```
cssuh, line 113: revert these too

src/nps/userSurvey.ts
```diff
@@ -164,9 +107,65 @@ export class UserSurvey {
...
+
+        //sendSurveyTelemetry(surveyId, answers);
```
cssuh, line 115: uncomment?

cssuh:
Do you have a screenshot of the popup you can add to the PR?

src/nps/userSurvey.ts
```diff
@@ -38,83 +39,24 @@ export class UserSurvey {
...
-                    this._webviewController.onCancel(() => {
-                        resolve({});
-                    });
-                });
-                sendSurveyTelemetry("nps", answers);
-                await globalState.update(IS_CANDIDATE_KEY, false);
-                await globalState.update(SKIP_VERSION_KEY, extensionVersion);
+                await this.launchSurvey("nps", state);
+
+                // TODO: revert before PR
```
aasimkhan30, line 57: Just fyi @Benjin , there is a vscode command "User feedback" that launches the feedback form bypassing these checks

Benjin: I tried the `vscode., and all it does is prepopulate the github form itself (unless I missed an option?).  If I used that, then the user's flow would be:

Benjin:
@cssuh

> Do you have a screenshot of the popup you can add to the PR?

The `showInformationMessage()` popup?  It's just a regular vscode toast, so I didn't think it was interesting, but this is what it looks like:

image

src/constants/locConstants.ts
```diff
@@ -775,6 +775,10 @@ export class UserSurvey {
             args: [featureName],
             comment: ["{0} is the feature name"],
         });
+    public static fileABugPrompt = l10n.t(
+        "Are you encountering a bug that you would like to submit more details about?",
```
Benjin, line 779: This phrasing is okay, but I don't love it; open to suggestions.

src/nps/userSurvey.ts
Benjin: Yup; thanks for the reminder!

src/constants/locConstants.ts
aasimkhan30: Maybe adding more context? Honestly don't have a strong preference between existing and this.

`Encountered an issue? Share details by opening a GitHub issue so we can improve!`

kburtram: yeah, I was thinking the same thing.  Maybe issue is better than bug, in case the concern/feedback isn't technically a bug.

src/nps/userSurvey.ts
```diff
@@ -250,6 +273,32 @@ class UserSurveyWebviewController extends ReactWebviewPanelController<
...
+
+**Actual Behavior:**
+
+
+----
+|Software|Version|
+|--|--|
+|MSSQL Extension|${extensionVersion}|
+|VS Code|${vscode.version}|
+|OS|${os.type()} ${os.release()}|`;
```
kburtram, line 299: I wonder if it's worth adding a tag here, like "from-feedback-form".  So we can see what issues are coming in from this channel?  Not critical, since we need to look at all the issues.  But could be interesting to see which bugs are tied to NPS data.  Though I guess we would need to think through privacy concerns since we'd have two quasi-correlated input streams.

src/nps/userSurvey.ts
```diff
@@ -228,7 +225,33 @@ class UserSurveyWebviewController extends ReactWebviewPanelController<
...
+                (payload.answers.nsat as number) < 2 /* NSAT dissatisfied */
+            ) {
+                const response = await vscode.window.showInformationMessage(
+                    locConstants.UserSurvey.fileABugPrompt,
+                    locConstants.UserSurvey.submitBug,
+                    locConstants.Common.cancel,
+                );
+
+                if (response === locConstants.UserSurvey.submitBug) {
+                    const encodedIssueBody = encodeURIComponent(
```
kburtram, line 242: should we log a telemetry event here to understand if anyone is using this option?

src/nps/userSurvey.ts
Benjin: Scratch my comment; you were referring to the MSSQL command you added!

src/constants/locConstants.ts
Benjin: I like that phrasing better!

src/nps/userSurvey.ts
Benjin: I've updated it to automatically include the "user-filed" tag (as well as "Triage: needed").

Yup, let's definitely think through the privacy aspect.  I'll spin up a thread.

src/nps/userSurvey.ts
Benjin: added

aasimkhan30 (approved):

src/nps/userSurvey.ts
Benjin: Based on the email thread, both streams are GDPR compliant, so together they're supposed to be okay.  Users have full control over what they submit as a github issue.

PR #18564, opened 2025-01-17
Add schema compare contracts and endpoints

lewis-sanchez:
This PR adds the missing contracts and endpoints for schema compare.

### DIFF
--- a/src/services/schemaCompareService.ts
+++ b/src/services/schemaCompareService.ts
@@ -10,10 +10,161 @@ import * as mssql from "vscode-mssql";
 export class SchemaCompareService implements mssql.ISchemaCompareService {
     constructor(private _client: SqlToolsServiceClient) {}
 
+    public schemaCompare(
+        operationId: string,
+        sourceEndpointInfo: mssql.SchemaCompareEndpointInfo,
+        targetEndpointInfo: mssql.SchemaCompareEndpointInfo,
+        taskExecutionMode: mssql.TaskExecutionMode,
+        deploymentOptions: mssql.DeploymentOptions,
+    ): Thenable<mssql.SchemaCompareResult> {
+        const params: mssql.SchemaCompareParams = {
+            operationId: operationId,
+            sourceEndpointInfo: sourceEndpointInfo,
+            targetEndpointInfo: targetEndpointInfo,
+            taskExecutionMode: taskExecutionMode,
+            deploymentOptions: deploymentOptions,
+        };
+
+        return this._client.sendRequest(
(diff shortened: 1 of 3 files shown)


src/services/schemaCompareService.ts
```diff
@@ -10,11 +10,162 @@ import * as mssql from "vscode-mssql";
...
+            deploymentOptions: deploymentOptions,
+        };
+
+        return this._client.sendRequest(
+            schemaCompareContracts.SchemaCompareRequest.type,
+            params,
+        );
+    }
+
+    public schemaCompareGenerateScript(
```
Benjin, line 34: I think we can improve upon ADS here by removing the "schemaCompare" from "schemaCompareGenerateScript" and all the others that are more verbose like this.  It's already in the SchemaCompareService, so that should be obvious enough.

typings/vscode-mssql.d.ts
```diff
@@ -1155,12 +1163,149 @@ declare module 'vscode-mssql' {
 		createStreamingJobTsql: string;
 	}
 
+	export const enum SchemaCompareEndpointType {
+		Database = 0,
+		Dacpac = 1,
+		Project = 2,
+		// must be kept in-sync with SchemaCompareEndpointType in SQL Tools Service
+		// located at \src\Microsoft.SqlTools.ServiceLayer\SchemaCompare\Contracts\SchemaCompareRequest.cs
```
Benjin, line 1171: Can you also update the respective comment in STS to include this as a place to be kept in sync?

typings/vscode-mssql.d.ts
```diff
@@ -1155,12 +1163,149 @@ declare module 'vscode-mssql' {
...
+	}
+
+	export interface ConnectionInfo {
```
Benjin, line 1174: We already a class in connectionManager.ts called ConnectionInfo.  Can you rename this to "SchemaCompareConnectionInfo" to make sure we avoid any confusion?

Benjin (approved):

typings/vscode-mssql.d.ts
lewis-sanchez: Yes

typings/vscode-mssql.d.ts
lewis-sanchez: I'll change the name to avoid confusion.

typings/vscode-mssql.d.ts
lewis-sanchez: @Benjin, the PR for the respective STS comment: https://github.com/microsoft/sqltoolsservice/pull/2441

Issue #18514, opened 2024-12-16
Applying a filter to a column affects all result sets containing a column with that name


Benjin:
Repro:
1. Create two tables that have a shared column name, and insert data:
```
CREATE TABLE FirstTable (name nvarchar(50))
INSERT INTO FirstTable values ('Albert')
INSERT INTO FirstTable values ('Benjamin')

CREATE TABLE SecondTable (name nvarchar(50))
INSERT INTO SecondTable values ('Benjamin')
INSERT INTO SecondTable values ('Charlie')
```

2. Select from both tables:
```
SELECT * FROM FirstTable
SELECT * FROM SecondTable
```

3. Apply a filter on the `name` column of the first result set

Expected:
* only the first result set's values are filtered

Reality:
* the `name` column is filtered in both result sets

https://github.com/user-attachments/assets/27ca95a5-0339-40cc-be44-19b945e9c59f

PR #18571, opened 2025-01-22
Use grid id to track filters for each grid

cssuh:
Use grid id to track filters for each grid so when we apply the filters from the state, they are applied to the correct grid.

Before:
filterFixBefore

After:
filterFixAfter

Addresses #18514 
(will close after port request)

### DIFF
--- a/src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
+++ b/src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
@@ -61,12 +61,11 @@ export class HeaderFilter<T extends Slick.SlickData> {
 
     private _eventManager = new EventManager();
     private queryResultState: QueryResultState;
-    private gridId: string;
 
     constructor(
         theme: ColorThemeKind,
         queryResultState: QueryResultState,
-        gridId: string,
+        private gridId: string,
     ) {
         this.queryResultState = queryResultState;
         this.theme = theme;
@@ -468,7 +467,7 @@ export class HeaderFilter<T extends Slick.SlickData> {
                 sorted: this.columnDef.sorted,
             };
         }
-        let record: GridFilters = {
+        const record: GridFilters = {
             gridId: this.gridId,
             columnFilters: columnFilterState,
         };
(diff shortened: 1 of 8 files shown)


src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -461,10 +468,14 @@ export class HeaderFilter<T extends Slick.SlickData> {
                 sorted: this.columnDef.sorted,
             };
         }
-        this.updateState(columnFilterState);
+        let record: GridFilters = {
```
Benjin, line 471: can be `const`, no?

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -60,10 +61,16 @@ export class HeaderFilter<T extends Slick.SlickData> {
 
     private _eventManager = new EventManager();
     private queryResultState: QueryResultState;
+    private gridId: string;
 
-    constructor(theme: ColorThemeKind, queryResultState: QueryResultState) {
+    constructor(
+        theme: ColorThemeKind,
+        queryResultState: QueryResultState,
+        gridId: string,
```
Benjin, line 69: you can also do `constructor(private gridId: string)` as shorthand for all these constructor changes

src/reactviews/pages/QueryResult/table/table.ts
```diff
@@ -213,10 +220,11 @@ export class Table<T extends Slick.SlickData> implements IThemable {
         this.columns.forEach((column) => {
             if (column.field) {
                 const filters =
-                    this.queryResultState.state.filterState[column.field];
+                    this.queryResultState.state.filterState[this.gridId];
                 if (filters) {
                     (<FilterableColumn<T>>column).filterValues =
-                        filters.filterValues;
+                        //@ts-ignore
```
Benjin, line 226: what didn't it like?

Benjin (approved):

lewis-sanchez (approved):

src/reactviews/pages/QueryResult/table/table.ts
cssuh: It complained that I was using a string to index my array

Issue #18052, opened 2024-09-18
Object explorer filter throws an error when we ok is pressed without any filters.


aasimkhan30:
Steps to Reproduce:

1. Open OE filter.
2. Do not make any edits and press OK
3. console throws an error
<img width="785" alt="image" src="https://github.com/user-attachments/assets/61b8a44c-4b6b-45aa-8a96-c36f4cb8f732">

PR #18573, opened 2025-01-22
Fixing telemetry event for expand and fixing clear all in filters

### LINKED ISSUE #18572: Clearing filters from filter dialog doesn't clear filters..
- MSSQL Extension Version:
- VSCode Version:
- OS Version:

Steps to Reproduce:

1. Apply OE filters to a node.

Image

2. Open filter dialog and click clear all and submit.

Image

3. Filters are not cleared. 

Image

aasimkhan30:
This PR fixes #18572 
1. This PR fixes the clear all button in filter dialog.
2. It makes sure that we are setting up the parent node to expand result map even when we are refreshing the node. Previously this wasn't set so when we tried to access parent node in expand results due to a refresh, we were getting undefined error causing #18052
3. Additionally, this PR updates the telemetry event for expand operations to include the node's subtype instead of the type. Subtypes provide more detailed information—for example, the "Tables" and "Views" folders are both of type "Folder," but their subtypes are "Tables" and "Views," respectively.

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -379,6 +379,17 @@ export class ObjectExplorerService {
                     ),
                 );
                 self._treeNodeToChildrenMap.set(parentNode, children);
+                sendActionEvent(
+                    TelemetryViews.ObjectExplorer,
+                    TelemetryActions.ExpandNode,
+                    {
+                        nodeType: parentNode?.context?.subType ?? "",
+                        isErrored: (!!result.errorMessage).toString(),
+                    },
+                    {
+                        nodeCount: result?.nodes.length ?? 0,
+                    },
+                );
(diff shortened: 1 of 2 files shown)


src/objectExplorer/objectExplorerService.ts
```diff
@@ -956,6 +956,7 @@ export class ObjectExplorerService {
             RefreshRequest.type,
             refreshParams,
         );
+        this._expandParamsToTreeNodeInfoMap.set(refreshParams, node);
```
Benjin, line 959: So reverting the original "fix" and this like is the proper fix instead?

Benjin (approved):

src/objectExplorer/objectExplorerService.ts
aasimkhan30: Yup, this is the proper fix for the issue.

Benjin:
Approved by Ken via shiproom chat

PR #18576, opened 2025-01-23
Revert "Use grid id to track filters for each grid (#18571)"

cssuh:
This reverts commit b979652d84bc3440d9bbdb66ccc6bef4e1bf4dd2.


aasimkhan30 (approved):
Why are we reverting?

Benjin (approved):

Benjin:
@aasimkhan30

The code looked fine, but I found some significant issues while testing that Chris is working on now.

PR #18580, opened 2025-01-23
Adding schema visualizer to the extension.

aasimkhan30:
Entry point:
image

image
image
image


src/sharedInterfaces/schemaDesigner.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export enum OnAction {
+    CASCADE = "0",
+    NO_ACTION = "1",
+    SET_NULL = "2",
+    SET_DEFAULT = "3",
+}
+
+export interface GetSchemaModelRequestParams {
+    ConnectionUri: string;
```
kburtram, line 92: inconsistent casing here?

src/sharedInterfaces/schemaDesigner.ts
```diff
@@ -0,0 +1,102 @@
...
+export interface ISchema {
+    entities: IEntity[];
+    relationships: IRelationship[];
+}
+
+export interface IEntity {
+    /**
+     * Name of the entity
+     */
+    name: string;
```
kburtram, line 15: for entities we should consider where to use a URI vs. schema/name?  Like if we wanted to include multiple DBs in the future, or non-table types.  This is probably fine, but something to think about.  We had this problem with the Metadata lookup stuff in ADS, and later wished we had used SFI Paths or something like that for more flexible/deterministic entity referencing.

src/reactviews/pages/SchemaDesigner/schemaDesigner.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+import { useContext, useEffect } from "react";
+import { SchemaDesignerContext } from "./schemaDesignerStateProvider";
+import * as azdataGraph from "azdataGraph";
+import "azdataGraph/dist/index.css";
+import "azdataGraph/src/css/common.css";
+import "azdataGraph/src/css/explorer.css";
+import "./schemaDesigner.css";
+import {
+    addTableIcon,
```
kburtram, line 14: off-topic, but do you like the code formatting like this?  I think our "pretty" linting rules makes the code look like lisp-style or something, like where the column width is so short and it takes a ton of rows?

kburtram (approved):
Please make sure the feature doesn't show up in the UI at all until the March release, unless perhaps an undocumented feature switch is set.

.vscode/launch.json
```diff
@@ -1,103 +0,0 @@
-// A launch configuration that compiles the extension and then opens it inside a new window
```
caohai, line 1: @aasimkhan30 Is it intentional to remove `launch.json`?

Issue #18521, opened 2024-12-17
Default ConnectTimeout option should be 30 instead of 15


kburtram:
Steps to Reproduce:

1. Add a connection using Connection Dialog and don't change any advanced options

Expected: Settings basically match what's in ADS, specifically Connect Timeout should default to 30 seconds.

Actual: Timeout defaults to 15 seconds, which is too short in many scenarios.  The lower timeouts cause increased failures in environments with slow network, or on instances with high load.  I'd probably remove "applicationName" from the settings too, since the application can set this at runtime, assume the user doesn't want to override it, but that's not really a problem the same way as the short timeout.

For example, here's a 'default' connection snippet:

```JSON
  {
      "authenticationType": "SqlLogin",
      "connectTimeout": 15,
      "applicationName": "vscode-mssql",
      "profileName": "",
      "server": "localhost",
      "user": "sa",
      "password": "",
      "database": "",
      "trustServerCertificate": true
  }
```

Benjin:
Both the `applicationName` and `connectTimeout` default changes were explicitly requested by @croblesm in https://github.com/microsoft/vscode-mssql/issues/18152.

Happy to bump the `connectTimeout` back to 30, but I do remember users explicitly talking about how they prefer a default value for `applicationName` to be set in their tools.

Benjin:
I'm using this issue for the default timeout length, but created https://github.com/microsoft/vscode-mssql/issues/18592 to clean up the config entries later.  I'll fold that into my connection management updates.

Benjin:
Decision is to revert timeout at 30 seconds, but to look into flexing the number of retries.

PR #18593, opened 2025-01-24
Updating connect timeout from 15 to 30 seconds

Benjin:
Addresses https://github.com/microsoft/vscode-mssql/issues/18521


lewis-sanchez (approved):

kburtram (approved):
thanks, 15 sec is too slow for lower end Azure SQL DB SKU.  Instead of lower this value, let's look at that the "total connect timeout" with all the reliable connection retry stuff.  Maybe we can reduce the number of connect retry attempts of something like that.  But yeah, messing with this value is a bit of a balancing act.

aasimkhan30 (approved):

Benjin:
Closing this because this value is getting set several different places, and the better solution would be to consolidate.

PR #18608, opened 2025-01-29
Hide query editor action buttons when in diff editor

### LINKED ISSUE #18559: Don't show Editor Actions in Diff Editor
When looking at the git diff of a SQL script, the editor actions that are added by this extension (Execute Query, Reveal Query Results Panel, Estimated Plan, etc.) bump off the more relevant editor actions when viewing a file diff (Previous Change, Next Change, Swap Left and Right Side).

If I am viewing two versions of a SQL query, having a single execute button is a little ambiguous and problematic.

Can these editor actions be hidden from the diff editor context?

caohai:
This PR fixes #18559 
Currently we show query editor action buttons in the diff editor which is not relevant. This bumps the diff-related actions into `...`. This PR adds an `isInDiffEditor` check during the action button contribution to hide them when the editor is in diff view. This change works for diff views originated from both git diff and the manual `Select for Compare` operation.

Before:
image

After:
Screenshot 2025-01-29 at 12 12 13

VSCode doc for `isInDiffEditor` flag:
https://code.visualstudio.com/api/references/when-clause-contexts#:~:text=editorLangId%20%3D%3D%20typescript%22.-,isInDiffEditor,-The%20active%20editor

### DIFF
--- a/package.json
+++ b/package.json
@@ -307,47 +307,47 @@
       "editor/title": [
         {
           "command": "mssql.runQuery",
-          "when": "editorLangId == sql",
+          "when": "editorLangId == sql && !isInDiffEditor",
           "group": "navigation@1"
         },
         {
           "command": "mssql.cancelQuery",
-          "when": "editorLangId == sql && resourcePath in mssql.runningQueries",
+          "when": "editorLangId == sql && !isInDiffEditor && resourcePath in mssql.runningQueries",
           "group": "navigation@2"
         },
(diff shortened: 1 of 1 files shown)


lewis-sanchez (approved):

cssuh (approved):

Benjin (approved):

kburtram (approved):
lgtm

Issue #18540, opened 2024-12-20
Add a logger to react webviews and save logs to file


caohai:
Currently all logs from react webviews are printed in dev console, we need to add a logger and save the logs to file.

PR #18613, opened 2025-01-31
Adding logging infra to webview framework

Benjin:
Fixes https://github.com/microsoft/vscode-mssql/issues/18540

### DIFF
--- a/src/controllers/reactWebviewBaseController.ts
+++ b/src/controllers/reactWebviewBaseController.ts
@@ -11,6 +11,7 @@ import {
     TelemetryViews,
 } from "../sharedInterfaces/telemetry";
 import {
+    LogEvent,
     WebviewTelemetryActionEvent,
     WebviewTelemetryErrorEvent,
 } from "../sharedInterfaces/webview";
@@ -21,6 +22,8 @@ import {
 } from "../telemetry/telemetry";
 
 import { getNonce } from "../utils/utils";
+import { Logger } from "../models/logger";
+import VscodeWrapper from "./vscodeWrapper";
 
 /**
  * ReactWebviewBaseController is a class that manages a vscode.Webview and provides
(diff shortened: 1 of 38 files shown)


src/controllers/reactWebviewBaseController.ts
```diff
@@ -126,9 +133,20 @@ export abstract class ReactWebviewBaseController<State, Reducers>
...
         private _initialData: State,
-    ) {}
+    ) {
+        if (!vscodeWrapper) {
+            vscodeWrapper = new VscodeWrapper();
+        }
+
+        if (!ReactWebviewBaseController._channel) {
+            ReactWebviewBaseController._channel =
+                vscodeWrapper.createOutputChannel("MSSQL WebViews"); // TODO: use name of webview?
```
aasimkhan30, line 146: Do we need to have a different output channel for webview? 

I feel it will be easy to debug if we can see all the logs in one output channel. Otherwise we will have to keep switching between them if we are debugging events that are across extension and webviews

Maybe we can add a prefix and colors to differentiate between event sources.

aasimkhan30 (approved):

lewis-sanchez (approved):

src/controllers/reactWebviewBaseController.ts
Benjin: Good call.  I'll reuse that output channel and use prefixes to differentiate between webviews.

src/reactviews/common/vscodeWebviewProvider.tsx
```diff
@@ -42,6 +43,22 @@ export interface VscodeWebviewContext<State, Reducers> {
...
+    //isLocalizationLoaded: boolean; // TODO: this appears to be unused; confirm with Aasim if it can be removed
+    /**
+     * State of the webview.
+     */
+    state: TState;
+    /**
+     * Theme of the webview.
+     */
+    themeKind: ColorThemeKind;
+    log(message: string, level?: LoggerLevel): void;
```
Benjin, line 59: add telemetry props

PR #18619, opened 2025-02-05
Add Schema Compare Controller Logic

lewis-sanchez:
This PR adds controller logic for Schema Compare.

The bulk of the logic here is for the controller with minimal UI logic to create this placeholder page:
image


gulpfile.js
```diff
@@ -1,4 +1,4 @@
-﻿const gulp = require('gulp');
```
Benjin, line 1: weird... an ascii conversion says there are three hidden characters here that got removed.  Probably a good thing.

package.json
```diff
@@ -422,6 +422,14 @@
           "when": "view == objectExplorer && config.mssql.enableRichExperiences && viewItem =~ /\\bfilterable=true\\b.*\\bhasFilters=true\\b/",
           "group": "inline@2"
         },
+        {
+          "command": "mssql.schemaCompare",
+          "when": "view == objectExplorer && config.mssql.enableRichExperiences && viewItem =~ /\\btype=(disconnectedServer|Server|Database)\\b/"
+        },
+        {
+          "command": "mssql.schemaCompare",
+          "when": "view == dataworkspace.views.main && viewItem =~ /^(databaseProject.itemType.project|databaseProject.itemType.legacyProject)$/"
```
Benjin, line 431: check `enableRichExperiences` here too

package.json
```diff
@@ -524,6 +532,10 @@
           "command": "mssql.objectExplorerNewQuery",
           "when": "view == objectExplorer && viewItem =~ /\\btype=(disconnectedServer|Server|Database)\\b/"
         },
+        {
+          "command": "mssql.schemaCompare",
+          "when": "view == objectExplorer && config.mssql.enableRichExperiences && viewItem =~ /\\btype=(disconnectedServer|Server|Database)\\b/"
```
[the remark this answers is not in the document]
Benjin: I think it's fine for now, but we'll eventually GA the current preview experiences and then would want to move this to behind the regular preview flag, but open to other thoughts.

src/reactviews/pages/SchemaCompare/SchemaCompare.tsx
```diff
@@ -0,0 +1,12 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export const SchemaComparePage = () => {
+    return (
+        <div>
+            <h1>Schema Compare Page</h1>
```
Benjin, line 9: Maybe add a comment here that this is a WIP/initial check-in

src/reactviews/pages/SchemaCompare/SchemaCompareStateProvider.tsx
```diff
@@ -0,0 +1,149 @@
...
+                        "schemaCompareGenerateScript",
+                        {
+                            operationId: operationId,
+                            targetServerName: targetServerName,
+                            targetDatabaseName: targetDatabaseName,
+                            taskExecutionMode: taskExecutionMode,
+                        },
+                    );
+                },
+                schemaComparePublishDatabaseChanges: function (
```
Benjin, line 65: since these are all part of the schemaCompare controller/provider, we probably don't need to prefix everything with "schemaCompare"

src/reactviews/pages/SchemaCompare/SchemaCompareStateProvider.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    operationId: string,
+                    targetServerName: string,
+                    targetDatabaseName: string,
+                    taskExecutionMode: mssql.TaskExecutionMode,
+                ): void {
+                    webViewState?.extensionRpc.action(
+                        "schemaComparePublishDatabaseChanges",
```
Benjin, line 72: was publishing changes to a dacpac excluded for any particular reason?

Benjin, line 72: the two (or three if you add dacpac) publish targets can probably be combined, just with adding an enum for targetType.  it'd be cleaner to split these up in the controller based on the enum, IMO

src/reactviews/pages/SchemaCompare/SchemaCompareStateProvider.tsx
```diff
@@ -0,0 +1,149 @@
...
+                schemaCompareIncludeExcludeNode: function (
+                    operationId: string,
+                    diffEntry: mssql.DiffEntry,
+                    includeRequest: boolean,
+                    taskExecutionMode: mssql.TaskExecutionMode,
+                ): void {
+                    webViewState?.extensionRpc.action(
+                        "schemaCompareIncludeExcludeNode",
+                        {
+                            operationId: operationId,
```
Benjin, line 106: I wonder if operationId is something that the React layer actually needs to know, or if the controller can hold onto that since each schema compare window will have its own controller instance.

src/schemaCompare/schemaCompareUtils.ts
```diff
@@ -0,0 +1,124 @@
...
+        payload.operationId,
+        payload.diffEntry,
+        payload.includeRequest,
+        payload.taskExecutionMode,
+    );
+
+    return result;
+}
+
+export async function openScmp(
```
Benjin, line 88: add docstrings for these functions since they're exported.  For the scmp ones, maybe include what the scmp file is.  for cancel, say what it's cancelling

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -0,0 +1,320 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// import * as azdata from "azdata";
```
Benjin, line 6: remove if not needed

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -0,0 +1,320 @@
...
+                cancelResultStatus: undefined,
+            },
+            {
+                title: title,
+                viewColumn: vscode.ViewColumn.Active,
+                iconPath: {
+                    dark: vscode.Uri.joinPath(
+                        context.extensionUri,
+                        "media",
+                        "schemaCompare_dark.svg", // lewissanchez TODO - update icon for edit data
```
Benjin, line 66: ...edit data?

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -0,0 +1,320 @@
...
+        void this.start(node);
+        this.registerRpcHandlers();
+    }
+
+    // schema compare can get started with four contexts for the source:
+    // 1. undefined
+    // 2. connection profile
+    // 3. dacpac
+    // 4. project
+    public async start(
```
Benjin, line 90: convert to docstring

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -0,0 +1,320 @@
...
+        sourceContext: any,
+        targetContext: mssql.SchemaCompareEndpointInfo = undefined,
+        comparisonResult: mssql.SchemaCompareResult = undefined,
+    ): Promise<void> {
```
Benjin, line 94: I'd break this up into some helper methods for each sourceContext type check so it'll be easier to read and test

src/reactviews/pages/SchemaCompare/SchemaCompareStateProvider.tsx
lewis-sanchez: I didn't see any contract in STS that directly called out dacpac's. The two publish contracts I came across was for publishing databases and projects.

lewis-sanchez: The same went for the endpoints there was a total of 9 endpoints that I came across.

src/reactviews/pages/SchemaCompare/SchemaCompareStateProvider.tsx
lewis-sanchez: I think the `operationId` can stay out of the react layer and stay with the controller too.

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -0,0 +1,323 @@
...
+     * @param comparisonResult Result of a previous comparison, if available.
+     */
+    public async start(
+        sourceContext: any,
+        comparisonResult: mssql.SchemaCompareResult = undefined,
+    ): Promise<void> {
+        let source: mssql.SchemaCompareEndpointInfo;
+
+        let connectionProfile: IConnectionProfile | undefined = sourceContext
+            ? <IConnectionProfile>sourceContext.ConnectionInfo
```
Benjin, line 109: We should use the new `myVar as IConnectionProfile` syntax over the old `<IConnectionProfile>myVar` to avoid conflicts with JSX/TSX syntax in case the code moves around.

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -0,0 +1,323 @@
...
+
+    private registerRpcHandlers(): void {
+        this.registerReducer("compare", async (state, payload) => {
+            const result = await compare(
+                this.operationId,
+                payload,
+                this.schemaCompareService,
+            );
+
+            return { ...state, schemaCompareResult: result };
```
Benjin, line 244: why not just set `state.schemaCompareResult = result` then return it?   Same below if applicable

test/unit/schemaCompareWebViewController.test.ts
```diff
@@ -0,0 +1,1046 @@
...
+    test("controller - initialize title - is 'Schema Compare'", () => {
+        assert.deepStrictEqual(
+            controller.panel.title,
+            schemaCompareWebViewTitle,
+            "Webview Title should match",
+        );
+    });
+
+    test("compare reducer - when called - runs once", async () => {
+        const compareResult: mssql.SchemaCompareResult = {
```
Benjin, line 250: Can you name this something with the word "mock" in them so that it's apparent what role they're supposed to play in the test?

test/unit/schemaCompareWebViewController.test.ts
```diff
@@ -0,0 +1,1046 @@
...
+            publishDatabaseChangesStub.firstCall.args,
+            [operationId, payload, mockSchemaCompareService.object],
+            "publishDatabaseChanges should be called with correct arguments",
+        );
+
+        publishDatabaseChangesStub.restore();
+    });
+
+    test("publishDatabaseChanges reducer - when called - returns expected result", async () => {
+        const publishDatabaseResult = {
```
Benjin, line 490: Same comment, but with "expected"

test/unit/schemaCompareWebViewController.test.ts
```diff
@@ -0,0 +1,1046 @@
...
+
+    test("controller - initialize title - is 'Schema Compare'", () => {
+        assert.deepStrictEqual(
+            controller.panel.title,
+            schemaCompareWebViewTitle,
+            "Webview Title should match",
+        );
+    });
+
+    test("compare reducer - when called - runs once", async () => {
```
Benjin, line 249: Is there a reason to split apart the "runs once", "with correct arguments", and "returns expected results" tests?  I think those should usually be three assertions within the same test case.

lewis-sanchez: It's a personal preference of mine to break things apart like that, but I can shrink them down if you'd like.

src/schemaCompare/schemaCompareWebViewController.ts
lewis-sanchez: That works too. I'll change it.

test/unit/schemaCompareWebViewController.test.ts
lewis-sanchez: Also, I think it's easier to maintain these kinds of tests because when changes break the tests, determining what went wrong is practically immediate.

Benjin (approved):

PR #18647, opened 2025-02-11
Webview cleanup

Benjin:
I noticed some inconsistencies in how we were building webviews, and brought everything into alignment:

* contexts are all named `context`, states are all named `state`; before, we were mixing the names which led to some confusing code snippets where a variable named `state` was not actually the state.
* contextProp interfaces and providerProp interfaces are now consistently named across all webviews; similarly, we were calling some things that weren't state "state"
* removing unnecessary null propagation and non-null-assertion operators; instead, we should be checking that `context` is defined and returning if not.
  * We should only use `myVar!.something` if the type checker is unable to figure out that it's not null/undefined on its own.  Usually, we should add checks to help the type checker along.  Using a non-null assertion operator is a yellow flag that something may be wrong.
* Promoted instances where we had an explicit `provider` property in the context so that all the RPC functions for the webview are accessible directly.
  * This one isn't fixing an issue, just bringing things into alignment.  I could've also done the opposite, and "demoted" the RPC definitions where they were directly under the `context` object.  Happy to go either way, but we should keep consistent to make future work and debugging simpler.  Any opinions?


aasimkhan30 (approved):

aasimkhan30:
Looks good to me. Thanks for cleaning the code!

lewis-sanchez (approved):

Issue #17234, opened 2022-02-10
AAD Sign in not works on mssql but works on azure tools plugin (with proxy)


esauaguilar-kr:
- MSSQL Extension Version: 1.12.0
- VSCode Version: 1.64.1
- OS Version: 11.6.2

Note: I use proxy...
The sign in not works with the mssql plugin but works with azure tools

Steps to Reproduce:

1- Click Add Connection
2- Enter the server (AZURE_SERVER_NAME.database.windows.net)
<img width="667" alt="1" src="https://user-images.githubusercontent.com/94638011/153357973-ed90685f-42a0-417f-abc7-bca79de6182f.png">
3- Press Enter to select the default table
<img width="630" alt="2" src="https://user-images.githubusercontent.com/94638011/153357977-2284b06a-c3a1-4713-9ba1-f2b59077b03c.png">
   Options listed:
* SQL Login
* Integrated
* Azure Active Directory - Universal w/ MFA Support <- select this
<img width="639" alt="3" src="https://user-images.githubusercontent.com/94638011/153357979-6d684f52-6f38-4f5a-a21f-edbc3db6116c.png">

4- click on Add account, and press enter to optional step (set local profile name)
<img width="639" alt="4" src="https://user-images.githubusercontent.com/94638011/153357981-5eabfb60-b781-4d74-a941-22b42e4449a9.png">
<img width="605" alt="5" src="https://user-images.githubusercontent.com/94638011/153357982-0d486219-030d-40f1-b626-8356c92b7660.png">

5- Opens Sign In page on a Web browser and you Sign in
6- Get stuck and still loading, and never finishes to add the account
<img width="1792" alt="6" src="https://user-images.githubusercontent.com/94638011/153442343-c2097b22-cefd-4d78-9589-03e298c7b1ac.png">

Notes:
The Sign-in with Azure tools plugin works, but mssql is better for the things which I need to do.

Azure tools plugin

1- in Azure tools plugin options y select sign in to azure
<img width="474" alt="az1" src="https://user-images.githubusercontent.com/94638011/153360173-48e64904-d6f1-46b4-ba33-6a0eabcdec81.png">

2- Web browser opens, and select the account to sign in
<img width="1792" alt="Screen Shot 2022-02-10 at 1 34 37" src="https://user-images.githubusercontent.com/94638011/153360370-6096b03c-75e3-4249-a799-93c196a0360e.png">

3- It Works
[...]

PR #18679, opened 2025-02-14
Fix AAD connectivity issues with proxied networks

lewis-sanchez:
This PR fixes https://github.com/microsoft/vscode-mssql/issues/17234

The PR works by using the user provided proxy endpoint and creating the necessary http(s) over http(s) tunnel to connect to the tenants endpoint to complete adding an AAD account.

As seen in the screenshot, the HTTP CONNECT method is being used to establish a network tunnel between the MSSQL extension and https://management.azure.com/tenants?api-version=2019-11-01 to allow the AAD request to complete successfully when the network environment is proxied.
image

### DIFF
--- a/src/azure/msal/msalAzureAuth.ts
+++ b/src/azure/msal/msalAzureAuth.ts
@@ -31,11 +31,29 @@ import { AzureAuthError } from "../azureAuthError";
 import * as Constants from "../constants";
 import axios, { AxiosResponse, AxiosRequestConfig } from "axios";
 import { ErrorResponseBody } from "@azure/arm-subscriptions";
+import * as tunnel from "tunnel";
+import * as http from "http";
+import * as https from "https";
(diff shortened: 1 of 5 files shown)


src/azure/msal/msalAzureAuth.ts
```diff
@@ -362,7 +380,7 @@ export abstract class MsalAzureAuth {
     }
 
     private async makeGetRequest<T>(
-        uri: string,
+        requestUrl: string,
```
Benjin, line 383: This is a port of the same code that worked in ADS, right?

Benjin (approved):

src/azure/msal/msalAzureAuth.ts
lewis-sanchez: That's correct.

PR #18684, opened 2025-02-14
Refactoring connection management

Benjin:
Fixes https://github.com/microsoft/vscode-mssql/issues/18849

* Lays the groundwork for connection grouping
* Fixes an issue where similar connections could trample each other in the Object Explorer
* cleans up a bunch of duplicate or poorly-named code in connection management

Before this change, there was a mix of keys being used, some of which were not guaranteed to be unique. Examples include:

STS-provided session ID (constructed using all properties)
vscode-mssql-generated session ID that seemed like it was supposed to be the same as the STS-provided one, but wasn't always (constructed using four primary connection properties)
OE nodePath

When trying to identify the correct tree node to update, it would store a node label mapping keyed on the server (not guaranteed to be unique among all connections), then later select a saved connection based on the matching server name; if you had multiple connections to the same server, it could end up selecting the wrong one (depending on ordering, IIRC), and expanding the wrong server in the tree (with the right underlying connection, IIRC).

After this change: everything is keyed on the STS sessionID and I'm hoping to remove the concept of a vscode sessionID entirely. It no longer matters how similar a connection is, if any property differs, then it's treated as a separate connection at both layers.

There are still some other cases I need to follow up with (e.g. what gets displayed if there are identical profile names or server + db combos), but this is an important first step.

Before:

https://github.com/user-attachments/assets/4f9f2242-b9df-47dd-9631-537d8b95a13b

After:

https://github.com/user-attachments/assets/8821adda-3619-46d8-a127-24bfe40ead51


src/objectExplorer/objectExplorerService.ts
```diff
@@ -123,46 +127,26 @@ export class ObjectExplorerService {
...
-                        if (
-                            connection.label !==
-                            connection.connectionCreds.server
-                        ) {
-                            nodeLabel = connection.label;
-                        }
-                        break;
-                    }
-                }
+                // let savedConnections =
```
Benjin, line 137: clean up

src/objectExplorer/objectExplorerService.ts
```diff
@@ -679,32 +672,38 @@ export class ObjectExplorerService {
...
                     )
                 ) {
                     // show password prompt if SQL Login and password isn't saved
-                    let password = connectionCredentials.password;
+                    let password = connectionProfile.password;
                     if (Utils.isEmpty(password)) {
                         // if password isn't saved
                         if (
-                            !(<IConnectionProfile>connectionCredentials)
+                            !(<IConnectionProfile>connectionProfile)
```
Benjin, line 706: update cast

src/objectExplorer/objectExplorerService.ts
```diff
@@ -872,40 +869,30 @@ export class ObjectExplorerService {
         isDisconnect: boolean = false,
     ): Promise<void> {
         await this.closeSession(node);
-        const nodeUri = ObjectExplorerUtils.getNodeUri(node);
+        const nodeUri = this.getNodeIdentifier(node);
```
Benjin, line 872: update `ObjectExplorerUtils.getNodeUri()`?

src/connectionconfig/connectionconfig.ts
```diff
@@ -31,6 +38,51 @@ export class ConnectionConfig implements IConnectionConfig {
...
+        // Connection profiles
+        const profiles: IConnectionProfile[] = this.getProfilesFromSettings();
+
+        // ensure each profile has an id
+        profiles.forEach((profile) => {
+            ConnectionProfile.addIdIfMissing(profile);
+        });
+
+        // ensure each profile is in a group
+        profiles.forEach((profile) => {
```
lewis-sanchez, line 73: You could probably just use the previous forEach to also ensure that each profile is in a group.

lewis-sanchez (approved):

caohai:
> where similar connections could trample each other in the Object Explorer

Do we have an issue for this? Would love to understand how "similar" two connections need to be to trigger this isse, maybe a before/after screenshot with the fix?

src/models/contracts/objectExplorer/getSessionIdRequest.ts
```diff
@@ -0,0 +1,29 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { RequestType } from "vscode-languageclient";
+import { ConnectionDetails } from "vscode-mssql";
+
+// Create session request message callback declaration
+export namespace GetSessionIdRequest {
```
caohai, line 10: These are not currenly in use but will be needed in the future, right?

src/objectExplorer/objectExplorerService.ts
```diff
@@ -718,72 +717,70 @@ export class ObjectExplorerService {
...
-            if ((connectionCredentials as IConnectionProfile).profileName) {
-                this._nodePathToNodeLabelMap.set(
-                    // using the server name as the key because that's what the rest of the OE service expects.
-                    // TODO: this service should be refactored to use something guaranteed to be unique across all connections,
-                    // but that likely involves a larger refactor of connection management.
-                    connectionCredentials.server,
-                    (connectionCredentials as IConnectionProfile).profileName,
+            if ((connectionProfile as IConnectionProfile).profileName) {
+                this._sessionIdToNodeLabelMap.set(
+                    sessionIdResponse.sessionId,
```
caohai, line 782: This part appears to be one of the important changes. At a high level I can see why we want to use sessionId instead of server name but would love to see more detailed explanation of the before/after workfow. (Not necessarily in this PR)

caohai (approved):
LGTM overall, I'd say we might need do more manual testings next week just to make sure this change won't break any existing OE experience.

Benjin:
@caohai 
> I'd say we might need do more manual testings next week just to make sure this change won't break any existing OE experience.

Agreed.  I really wanted to get this in on the earlier side of a release cycle, but too many things were competing with my time when I wanted to get this PR out.  The right call could be to hold this PR after the release branch splits so we can decide whether to take it as a port.

src/objectExplorer/objectExplorerService.ts
Benjin: Yup, swapping everything to use the sessionId as the sole key was the main change that solidified some unexpected behavior in the OE.

Before this change, there was a mix of keys being used, some of which were not guaranteed to be unique.  Examples include:
* STS-provided session ID (constructed using all properties)
* vscode-mssql-generated session ID that seemed like it was supposed to be the same as the STS-provided one, but wasn't always (constructed using four primary connection properties)
* OE nodePath

When trying to identify the correct tree node to update, it would store a node label mapping keyed on the server (not guaranteed to be unique among all connections), then later select a saved connection based on the matching server name; if you had multiple connections to the same server, it could end up selecting the wrong one (depending on ordering, IIRC), and expanding the wrong server in the tree (with the right underlying connection, IIRC).

After this change: everything is keyed on the STS sessionID and I'm hoping to remove the concept of a vscode sessionID entirely.  It no longer matters how similar a connection is, if any property differs, then it's treated as a separate connection at both layers.

There are still some other cases I need to follow up with (e.g. what gets displayed if there are identical profile names or server + db combos), but this is an important first step.

Benjin: I'll add recordings of before/after the change

caohai: Thanks for the detailed explanation!

src/models/contracts/objectExplorer/getSessionIdRequest.ts
```diff
@@ -0,0 +1,29 @@
...
+        void
+    >("objectexplorer/getsessionid");
+}
+
+/**
+ * Contains a sessionId to be used when requesting
+ * expansion of nodes
+ */
+export class GetSessionIdResponse {
+    /**
```
aasimkhan30, line 24: I meant this comment.

src/objectExplorer/objectExplorerService.ts
aasimkhan30: Thanks for the detailed explanation. 

Can you update the js doc for this interface with some explanation? Something like this maybe
```
/**
 * A unique session ID used for all requests within a connection sub-tree. 
 * All child nodes must use the same session ID when sending expand requests.
 */
```
https://github.com/microsoft/vscode-mssql/blob/694c6db2df7411e1f7262ac7af2a38f5368b2e89/src/models/contracts/objectExplorer/createSessionRequest.ts#L26-L32

https://github.com/microsoft/vscode-mssql/blob/57f780a9e54745ac4c8eab67df8f5fb068d4c8c2/src/models/contracts/objectExplorer/createSessionRequest.ts#L48-L51

aasimkhan30 (approved):
Thanks for the code cleanup. Left a comment about a jsdoc comment.

src/objectExplorer/objectExplorerService.ts
Benjin: @caohai before/after recordings now added, with detailed (and consistent) repro information in the bug: https://github.com/microsoft/vscode-mssql/issues/18849

PR #18694, opened 2025-02-15
Adding VS Code walkthroughs

Benjin:
Adding two walkthroughs, "Getting Started" and "Next Steps".  As we add new features with more UX complexity, we can consider doing feature-specific walkthroughs.

Note: I've added automatic completions where already possible (e.g. a command being executed indicating that the user clicked a button), but didn't add new no-op commands just to track walkthrough progress.  Users can always manually indicate that they've finished a step.

image

image

image


package.json
```diff
@@ -1,1637 +1,1748 @@
...
+            }
         },
-        "mssql.objectExplorer.expandTimeout": {
-          "type": "number",
-          "default": 45,
-          "minimum": 1,
-          "description": "%mssql.objectExplorer.expandTimeout%"
-        }
-      }
+        "walkthroughs": [
```
lewis-sanchez, line 1644: Besides for this walkthroughs section, are the rest of the changes formatting changes?

Benjin: Oi, yes.  Lemme remove these formatting changes.  They shouldn't be in this PR.

package.json
```diff
@@ -524,10 +524,6 @@
           "command": "mssql.objectExplorerNewQuery",
           "when": "view == objectExplorer && viewItem =~ /\\btype=(disconnectedServer|Server|Database)\\b/"
         },
-        {
```
Benjin, line 527: revert this

lewis-sanchez:
Besides for the last PR check that is failing, everything else looks good.

Benjin:
VSIX size jumped by a 1MB because of the screenshots for the walkthrough.  I scaled them down a decent bit already, but can go further if desired.

croblesm (changes requested):
Hi @Benjin, everything seems fine. However, could you please update the “Get **S**tarted” text to “Get started”?

I’ve noticed this text in several places, so I didn’t want to bother you with multiple comments. A simple find/replace should do the trick.

croblesm (approved):
All good, thanks!

Issue #18620, opened 2025-02-05
Copy/Paste does not retain values copied when data has been filtered in the Result Pane


buckz1lla:
MSSQL Extension Version: 1.28.0
VSCode Version: 1.95.1
OS Version: Windows 11 Enterprise, 10.0.22631 Build 22631

Steps to Reproduce:

1. Execute a query
2. Filter the results in the Results Pane
3. Copy the data into a worksheet/notepad
4. Notice that it pastes the data from the rows as if they were still in the order of the original query, instead of what you copied from the filtered result set

cssuh:
Hi @buckz1lla, thanks for filing an issue.  I'll look into this and start working on a fix for our February release.

scottleach:
Found this in VSCode Version 1.97.2 as well. All my system versions other than that are identical to buckz1lla's. I thought I was losing my mind when I was trying to copy query result values in a filtered column and the wrong things were pasting.

Issue #18516, opened 2024-12-17
Copy doesn't work as expected after sorting in Query Results (Preview) window


LienBosmans:
- MSSQL Extension Version: v1.26.0
- VSCode Version: 1.95.0 (system setup)
- OS Version: Windows_NT x64 10.0.22631

Steps to Reproduce:

1. Run below SQL query to generate some example data.
`with example_data as (
    select 'I want to copy this' as first_column, 1 as second_column
    union all
    select 'I accidently copy this' as first_column, 2 as second_column
)
select * from example_data order by second_column desc
`

2. In the QUERY RESULT (PREVIEW) window, change the sort order by sorting the second column ascending (instead of descending as specified by the SQL query)

3. Select the first row ("I want to copy this") and copy it. (Right click + "Copy" or Ctrl+C)

4. Paste the result in a text editor. ("I accidently copy this")

PR #18695, opened 2025-02-15
Fix copy behavior for sorted and/or filtered grid

cssuh:
Correctly copies the grid values for a grid that has been sorted and/or filtered.

Before, we were using a getRows call to STS which fetches the data based on the initial query.  Once we sort or filter any data in the grid (which triggers isDataInMemory flag = true), we should use the dataProvider to tell us what data the user wants to copy, since it knows the context of the sort or filter states.  

Fixes https://github.com/microsoft/vscode-mssql/issues/18516
Fixes https://github.com/microsoft/vscode-mssql/issues/18620

### DIFF
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -713,6 +713,13 @@ export default class QueryRunner {
         await this.writeStringToClipboard(copyString);
     }
 
+    /**
+     * Construct the row mappings, which contain the row data and selection data and are used to construct the copy string
+     * @param data
+     * @param range
+     * @param rowIdToSelectionMap
+     * @param rowIdToRowMap
+     */
     private getRowMappings(
         data: DbCellValue[][],
         range: ISlickRange,
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -660,7 +660,7 @@ export default class QueryRunner {
         }
         await p;
 
-        this.constructCopyString(
+        copyString = this.constructCopyString(
             copyString,
             rowIdToRowMap,
             rowIdToSelectionMap,
@@ -714,7 +714,7 @@ export default class QueryRunner {
         }
         await p;
 
-        this.constructCopyString(
+        copyString = this.constructCopyString(
             copyString,
             rowIdToRowMap,
             rowIdToSelectionMap,
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -634,8 +634,6 @@ export default class QueryRunner {
         // create a mapping of the ranges to get promises
         let tasks = selection.map((range) => {
             return async () => {
-                //TODO: instead of using get rows, we should use the data provider to fetch the data and get the sorted rows
-                //TODO: need to check if data is in memory or not from the data provider so we know if it's sorted
                 const result = await this.getRows(
                     range.fromRow,
                     range.toRow - range.fromRow + 1,
@@ -723,14 +721,12 @@ export default class QueryRunner {
     }
 
     public async sendToClipboard(
-        //TODO: what type is data?
         data: DbCellValue[][],
         batchId: number,
         resultId: number,
         selection: ISlickRange[],
         headersFlag,
     ) {
-        //TODO: filter array data based on the columns selected
         let copyString = "";
         if (headersFlag) {
             copyString = this.addHeaders(
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -616,33 +616,14 @@ export default class QueryRunner {
     ): Promise<void> {
         let copyString = "";
 
-        // add the column headers
         if (this.shouldIncludeHeaders(includeHeaders)) {
-            let firstCol: number;
-            let lastCol: number;
-            for (let range of selection) {
-                if (firstCol === undefined || range.fromCell < firstCol) {
-                    firstCol = range.fromCell;
-                }
-                if (lastCol === undefined || range.toCell > lastCol) {
-                    lastCol = range.toCell;
-                }
-            }
-            let columnRange: ISlickRange = {
-                fromCell: firstCol,
-                toCell: lastCol,
-                fromRow: undefined,
-                toRow: undefined,
-            };
-            let columnHeaders = this.getColumnHeaders(
+            copyString = this.addHeaders(
+                copyString,
                 batchId,
                 resultId,
-                columnRange,
+                selection,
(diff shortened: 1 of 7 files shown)


src/controllers/queryRunner.ts
```diff
@@ -736,6 +717,148 @@ export default class QueryRunner {
...
+            copyString = copyString.substring(
+                0,
+                copyString.length - os.EOL.length,
+            );
+        }
+
+        await this.writeStringToClipboard(copyString);
+    }
+
+    public addHeaders(
```
caohai, line 828: Can you add doc to this and other public methods? You might also want to use a better name for this method as it also returns  the copy str as the return value but `addHeaders` doesn't indicate that.

src/controllers/queryRunner.ts
```diff
@@ -736,6 +717,148 @@ export default class QueryRunner {
             );
         }
 
+        await this.writeStringToClipboard(copyString);
+    }
+
+    public async sendToClipboard(
```
caohai, line 723: This function appears to be doing much more than `sendToClipboard`, can we get a better name for this?

src/controllers/queryRunner.ts
```diff
@@ -736,6 +717,148 @@ export default class QueryRunner {
...
+        data: DbCellValue[][],
+        batchId: number,
+        resultId: number,
+        selection: ISlickRange[],
+        headersFlag,
+    ) {
+        let copyString = "";
```
caohai, line 730: Is this entire function something new you created in this PR? IIRC some of the code here duplicates with the existing copyResults function, could we re-use (possibily by extracting out some common pieces) instead of duplicating them?

caohai:
Can you explain the following two things at a high level in the PR?
1. Root cause of the issues
2. The fix you implemented in this PR

Benjin:
Looks like this PR only addresses copying, but the same problem occurs for all copy/save targets.

src/controllers/queryRunner.ts
```diff
@@ -679,6 +660,74 @@ export default class QueryRunner {
...
+                copyString,
+                batchId,
+                resultId,
+                selection,
+            );
+        }
+
+        // create a mapping of rows to selections
+        let rowIdToSelectionMap = new Map<number, ISlickRange[]>();
+        let rowIdToRowMap = new Map<number, DbCellValue[]>();
```
Benjin, line 691: A selection potentially spans multiple rows, right?  So will this map have multiple rowId entries pointing to the same ISlickRange?

src/controllers/queryRunner.ts
```diff
@@ -679,6 +660,74 @@ export default class QueryRunner {
...
+                        let rowSelection = rowIdToSelectionMap.get(rowNumber);
+                        rowSelection.push(range);
+                    } else {
+                        rowIdToSelectionMap.set(rowNumber, [range]);
+                    }
+                    rowIdToRowMap.set(rowNumber, row);
+                    count += 1;
+                }
+            };
+        });
```
Benjin, line 710: Can you add more comments here?

* what do the promises represent?
* what are the rowId mappings being used to calculate?

src/controllers/queryRunner.ts
```diff
@@ -679,6 +660,74 @@ export default class QueryRunner {
...
+                selection,
+            );
+        }
+
+        // create a mapping of rows to selections
+        let rowIdToSelectionMap = new Map<number, ISlickRange[]>();
+        let rowIdToRowMap = new Map<number, DbCellValue[]>();
+
+        // create a mapping of the ranges to get promises
+        let tasks = selection.map((range) => {
```
Benjin, line 694: Looks like this code is largely the same as above; can you refactor to dedupe

cssuh: this is the only instance of this, are you perhaps looking at a stale version?

src/controllers/queryRunner.ts
cssuh: Yes that's correct

caohai:
I gave the PR branch a quick test and can confirm copying is fixed with sorted/filtered grid.

One more question: I noticed the filter state is kept during context switch but the sorting state is not, is this intentional or do we want to fix it later?

> Looks like this PR only addresses copying, but the same problem occurs for all copy/save targets. Maybe we need a more fundamental fix here.

Can confirm save is still broken, we might want to create a seperate issue for tracking.

src/models/sqlOutputContentProvider.ts
```diff
@@ -151,6 +151,26 @@ export class SqlOutputContentProvider {
...
 
+    public sendToClipboard(
+        uri: string,
+        data: Array<any>,
+        batchId: number,
+        resultId: number,
+        selection: ISlickRange[],
+        headersFlag: boolean,
+    ): void {
+        console.log(data);
```
caohai, line 162: Do we want to keep this logging line?

cssuh: I'll remove this

cssuh:
> I gave the PR branch a quick test and can confirm copying is fixed with sorted/filtered grid.
> 
> One more question: I noticed the filter state is kept during context switch but the sorting state is not, is this intentional or do we want to fix it later?
> 
> > Looks like this PR only addresses copying, but the same problem occurs for all copy/save targets. Maybe we need a more fundamental fix here.
> 
> Can confirm save is still broken, we might want to create a seperate issue for tracking.

Yeah, we want to keep the sort state as well, I will add an issue to track this as well as an issue for the save targets

src/reactviews/pages/QueryResult/table/plugins/contextMenu.plugin.ts
```diff
@@ -116,24 +118,90 @@ export class ContextMenu<T extends Slick.SlickData> {
...
+                        {
+                            uri: this.uri,
+                            batchId: this.resultSetSummary.batchId,
+                            resultId: this.resultSetSummary.id,
+                            selection: selection,
+                        },
+                    );
+                }
 
-                console.log("Copy with Headers action triggered");
```
caohai, line 136: Is deleting this line intentional? I see you kept the other logs for these events.
 I guess we need some sort of logger at some point....

caohai (approved):

src/reactviews/pages/QueryResult/table/plugins/contextMenu.plugin.ts
cssuh: Yeah, I'll clean up these in a separate PR.

PR #18699, opened 2025-02-17
Fix error handling in designer changes preview

aasimkhan30:
Previously designer preview error were not surfaced properly. This PR fixes that behavior. Happens in very rare edge cases where the validation happens once the user clicks the publish button. 

Previously:
image

Now:
image

### DIFF
--- a/src/reactviews/pages/TableDesigner/designerChangesPreviewButton.tsx
+++ b/src/reactviews/pages/TableDesigner/designerChangesPreviewButton.tsx
@@ -53,8 +53,9 @@ const useStyles = makeStyles({
     errorIcon: {
         fontSize: "100px",
         opacity: 0.5,
+        color: "var(--vscode-errorForeground)",
     },
-    retryButton: {
+    dialogFooterButtons: {
         marginTop: "10px",
     },
     markdownContainer: {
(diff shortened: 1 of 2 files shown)


src/reactviews/pages/TableDesigner/designerChangesPreviewButton.tsx
```diff
@@ -211,11 +211,20 @@ export const DesignerChangesPreviewButton = () => {
         <>
             <DialogContent className={classes.dialogContent}>
                 <ErrorCircleRegular className={classes.errorIcon} />
-                <div>{locConstants.tableDesigner.errorLoadingPreview}</div>
+                <div>
+                    {designerContext.state.generatePreviewReportResult
+                        ?.schemaValidationError ??
```
Benjin, line 216: Would be nice to have the error formatted onto multiple lines.  Are we losing newlines when it comes from STS?

Benjin (approved):

src/reactviews/pages/TableDesigner/designerChangesPreviewButton.tsx
aasimkhan30: Yea, I will have to investigate more to see where this error is surfacing from. Probably will require fix in DACFX/STS. 
At least this PR corrects the behavior and matches it with ADS. 
alt text

aasimkhan30: image

Issue #18577, opened 2025-01-23
Using new tab renders too many grids on first run


cssuh:
- MSSQL Extension Version: 1.28.0 

Steps to Reproduce (not consistently reproduceable):

1. Change setting to "Open Query Results In Tab By Default" = true
2. run a query generating multiple result grids

Expected: 2 grids displayed

Actual: 2nd grid has another grid overlapping:

![Image](https://github.com/user-attachments/assets/48ce00e5-c155-4128-a31c-4cc8da67970b)

Note - From testing previous versions, I believe this issue was introduced in 1.27.0

PR #18717, opened 2025-02-19
[Port Request] remove refreshKey, which was causing multiple grid rendering issue

cssuh:
Removes the refresh key which is no longer needed, and was causing a multiple grid rendering issue

Ports https://github.com/microsoft/vscode-mssql/pull/18710

Fixes https://github.com/microsoft/vscode-mssql/issues/18577


aasimkhan30 (approved):

Benjin:
@cssuh @kburtram Looks like this port went to the wrong branch (1.28 instead of 1.29). This is definitely one we want to take in our Feb release!

Issue #18014, opened 2024-08-28
Resuming a paused serverless DB hangs indefinitely


kburtram:
Steps to Reproduce:

1. Connect to an Azure SQL server that contains a paused serverless DB
2. Expand the DB node to resume the DB

Expected: Some detail to resume the server and then it connects successfully.  Or in the worse-case the expansion times-out with an appropriate error message and then succeeds on retry.

Actual: Expansion hangs indefinitely with a spinner icon and doesn't generate an error in the UI or console log.  Trying to disconnect the server doesn't work.  VS Code needs to be restarted to get the connection back into a work state as far as I could tell.

![image](https://github.com/user-attachments/assets/740279c0-4d8b-4632-8ead-c7a00d655ce9)

Benjin:
AB#3456212

kburtram:
I'm still seeing this one today.  But also noticed that while the serverless instance is in a busted state, other available servers (like localhost in the below screen) are blocked too.  VS Code needs to be restarted to recover into a usage state.  We should prioritize getting this one fixed as it's blocking a core connectivity scenario.

![Image](https://github.com/user-attachments/assets/b4e08230-e637-4294-af43-83d4f4b946cb)

Benjin:
I was planning on addressing this one as part of the larger connection management changes, but lemme see if I can get the fix in without all those.

Benjin:
Investigation notes:

For the tree being blocked/stuck:

If a child node is in the process of loading (attempting to connect to a serverless DB that's disabled) and then a root node refresh is triggered (as a result of another database successfully connecting), then the root refresh is held until after the long-running child has completed (or is stuck forever if that never completes).  I have a fix for that, but it currently results in another issue where said root node is not refreshed, so it still appears disconnected despite being otherwise operable.  I've tried a few things, but no luck yet.

For the serverless db timeout:

I'm getting the expected timeout (45s) from STS when simulating trying to connect to a paused serverless DB.  I'll try against a real paused DB to see if the behavior's any different.  For the simulated timeout, it looks like it should be a simple fix: the code assumes that any failure where your connection has a password is due to the creds being wrong, which is untrue.

PR #18816, opened 2025-02-21
Adding handling for expand node failures

Benjin:
Addresses #18014

When connecting to a serverless database's master db, it connects immediately.  If the underlying serverless db is paused, then expanding the database node can time out in STS, which sends a notification to MSSQL.  Unsuccessful expansions were not being handled at all; this PR adds basic handling there.


Benjin:
Adding a regression test still...

aasimkhan30 (approved):

Issue #18923, opened 2025-03-11
Set focus to messages tab if user cancels query


cssuh:
Steps to Reproduce:

1. Run a query
2. Cancel the query

Focus will be set to results tab instead of the messages tab

Issue #18702, opened 2025-02-18
Query Result defaults to show an empty result set when an error occurs, instead of the error in the Messages tab


caskeyje:
When a query is executed that encounters an error, the Query Results tab defaults to show an empty result set. There is no readily visible indicator that an error in the query has occurred. The error message is, output in the Messages sub-tab, however. Even when the Messages sub-tab is selected, the empty result is shown when re-running the error producing query.

vscode-mssql should default to displaying query errors to the user and/or give some other visual indicator that an error has occurred. 

SSMS automatically shows the messages tab when a query runs the produces an error.

![Image](https://github.com/user-attachments/assets/bc2b5c2c-057e-4dbe-b51e-dade5cbad933)

![Image](https://github.com/user-attachments/assets/9dffbda0-d57a-4507-bada-e10d8b1ed6a3)

PR #18932, opened 2025-03-13
Set messages tab active if there is a query execution error

cssuh:
Before: When you run `SELECT 1/0`, the empty results pane is shown, instead of the messages tab with the divide by zero error

After: We show the messages tab by default if there is an error with the query execution.

(also sets view to messages tab if the user cancels the query)

Fixes https://github.com/microsoft/vscode-mssql/issues/18702
Fixes https://github.com/microsoft/vscode-mssql/issues/18923


src/models/sqlOutputContentProvider.ts
```diff
@@ -597,14 +597,21 @@ export class SqlOutputContentProvider {
                                     ),
                                 isError: false, // Elapsed time messages are never displayed as errors
                             });
-                        const tabState =
-                            Object.keys(
-                                this._queryResultWebviewController.getQueryResultState(
-                                    uri,
-                                ).resultSetSummaries,
-                            ).length > 0
```
caohai, line 605: If there's an error, wouldn't the length of `resultSetSummaries` be `0` leading to `tabState` set to `Messages`?

cssuh: It seems like that is not the case, for example if I run this query:
```
SELECT 1/0
```

currently it gets set to the Results tab showing this (even though there is an error message in the messages pane):

Screenshot 2025-03-13 130251
Screenshot 2025-03-13 130505


I also think it makes sense to explicitly check for errors rather than rely on the resultSetSummary length

caohai: ah I see, in this case resultSummaries would have a length of 1, makes sense.

caohai (approved):

Issue #18854, opened 2025-02-26
Duplicate connections in the Connections view


kburtram:
Repro:

1. Set up your user connections such that they don't have the `id` and `groupId` properties.
    This can be done either by manually deleting these props from the connections in  User `settings.json`, or by deleting and recreating the connections in MSSQL v1.29 or earlier
2. Check out a version of the code after [this commit](https://github.com/microsoft/vscode-mssql/commit/5e1d743b36873db1a2a4ef1942e369aea26daf75) and build
3. Launch the MSSQL extension

Expected:
* Connections in User `settings.json` have new `id` and `groupId` properties
* Correct number of connections displayed in OE

Reality:
* Connections are duplicated in OE (but if you restart VS Code or MSSQL, the issue disappears and doesn't return)

---

I'm seeing duplicate connections in OE.  The package.json looks fine.

![Image](https://github.com/user-attachments/assets/4f707f24-c1af-447f-b4a0-56bae4410e52)

bhubert:
I had the same issue because i created the connection before enabling previews feature and restarting vs code.

I activated the features, restarted vs code then recreated the connection and no more duplicates

bhubert:
In fact i manage to reproduce it when using different connection by workspace for example one workspace with connection for dev env and one workspace for prod env.

if the two workspaces are opened connection appear duplicated

![Image](https://github.com/user-attachments/assets/1dc9ea64-3efb-494e-8fab-edbae3b44116)

caohai:
Reported by more internal and external users
https://github.com/microsoft/vscode-mssql/issues/18901#issuecomment-2717197548

croblesm:
> I'm seeing duplicate connections in OE. The package.json looks fine. There's no specific repro beyond having these connections in the settings file.
> 
> ![Image](https://github.com/user-attachments/assets/4f707f24-c1af-447f-b4a0-56bae4410e52)

same here. I'm using VS Code Insiders and the `settings.json` file looks correct.

![Image](https://github.com/user-attachments/assets/212c6029-14fb-4294-811b-6faaf84a3606)

Benjin:
@bhubert thank you for identifying the repro!  I'll work on getting this fixed for our next release.

Benjin:
Turns out that we have a couple different issues, all manifesting in the same way (duplicate connections appearing in OE).

[...]

PR #18972, opened 2025-03-20
Fixing issue where OE nodes would appear the first time time connections were upgraded with IDs and groups

Benjin:
Addresses https://github.com/microsoft/vscode-mssql/issues/18854

Investigation found two bugs occuring with two interweaving flows:

1) Duplciate connections in OE immediately after upgrading:

The core bug was that when ConnectionConfig would detect connections that needed upgrading (with new `id` and `groupId` properties) in the User `settings.json` and save those, it would trigger the `onDidChangeConfiguration` event which refreshes the OE with the updated connections.  Because this would occur before the OE had actually loaded, it adds new root nodes intead.  When OE eventually loads, it reads the connections from config, and adds the connection nodes to the root node list (because it assumed it was the first to get there).

Fix: replace the list of root nodes when OE loads instead of appending to it.

2) When OE would do its later loading, it would read the connections from config before the ConnectionStore had finished initializing, so those connections wouldn't have the new IDs populated.  Because the duplicate OE nodes were triggered as a result of the onDidChangeConfiguration event, those _did_ have the IDs, so the result was that one of the dupes had the new ID and the other didn't.

Fix: await the deferred initialization flag before reading the connections from config.

### DIFF
--- a/.vscode-test.mjs
+++ b/.vscode-test.mjs
@@ -7,7 +7,7 @@ export default testCli.defineConfig([
 		version: 'insiders',
 		mocha: {
 			ui: 'tdd',
-			timeout: 5000
+			timeout: 6000
 		}
 	}
 ])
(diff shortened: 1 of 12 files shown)


lewis-sanchez:
Code looks good, but looks like there's a type issue in connectionUI.test.ts that is breaking the checks.

lewis-sanchez (approved):

aasimkhan30 (approved):

PR #18984, opened 2025-03-23
Making sure tree node connection info is not accidentally modified by consumers. 

aasimkhan30:
Several features rely on `treeNode.connectionInfo` to create or reuse existing connections. However, when this object is modified directly, it can lead to unintended side effects — such as breaking future connection reuse or connection lookup logic which breaks apis like `getUriForConnection()`

This PR introduces the following improvements to make connection management safer and more intentional:

1. **Makes `connectionInfo` return a shallow copy**  
   Exposes a shallow copy of the connectionInfo to prevent accidental modifications by callers.

2. **Adds `updateConnectionInfo` method**  
   Provides a deliberate, controlled way to update the internal connection info.
   
3. **Removes `connectionInfo` setter**
   Removes connectionInfo setter so that callers cannot accidentally update the connectionInfo.

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -448,7 +448,7 @@ export default class MainController implements vscode.Disposable {
         executeScript: boolean = false,
     ): Promise<void> {
         const nodeUri = ObjectExplorerUtils.getNodeUri(node);
-        let connectionCreds = Object.assign({}, node.connectionInfo);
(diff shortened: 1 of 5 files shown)


kburtram (approved):
We might be able to add a couple more tests that target this functionality a little more directly?

lewis-sanchez (approved):

Issue #18616, opened 2025-02-04
creating connection from connection string also adds undefined connection


dzsquared:
- MSSQL Extension Version: 
1.28.0

VSCode Version: 1.96.4
Commit: cd4ee3b1c348a13bafd8f9ad8060705f6d4b9cba
Date: 2025-01-16T00:16:19.038Z
Electron: 32.2.6
ElectronBuildId: 10629634
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.3.0

Steps to Reproduce:

1. Create new connection using connection string option for "Server=localhost,1433;Initial Catalog=master;User Id=sa;Password=SomePassword;TrustServerCertificate=true"
2. Based on my testing, this occurs for other local ports and different passwords - connection succeeds but also adds an "undefined" connection in the object explorer
3. reloading the window from command palette causes the undefined connection to disappear

![Image](https://github.com/user-attachments/assets/1e8c354e-9af9-4804-941b-f3d81d83a78f)

edit: root cause for this in the repro is that I leave the connection name blank

cedx:
More or less the same issue here: when I open a connection, the connection seems to be duplicated.
The first connection is OK, while the second one does nothing.
If I remove the second connection... it also removes the first one!

![image](https://github.com/user-attachments/assets/08641cc8-4313-4210-a06f-69f64043111e)

Also using MSSQL extension v1.28.0.

Benjin:
I'm validating a fix for this one, but I think it'll land in the March release.  Thanks for reporting!

kburtram:
This should be fixed in latest release candidate build.

PR #19000, opened 2025-03-25
Fixing smattering of connection management bugs

Benjin:
Addresses bugs:
* Adding a connection string creates an "undefined <undefined> (undefined)" Object Explorer entry (https://github.com/microsoft/vscode-mssql/issues/18616)
* Adding a connection via connection string in Connection Dialog results in a password in the config file, and the OE label (for a connection without a profileName) includes that (https://github.com/microsoft/vscode-mssql/issues/19005)

With this change, all new connections are serialized the same way, regardless of input method (same as Azure Data Studio), which also allows connectionString connections to be more readily parsed for sanitized default OE connection labels.

### DIFF
--- a/src/connectionconfig/connectionDialogWebviewController.ts
+++ b/src/connectionconfig/connectionDialogWebviewController.ts
@@ -640,6 +640,7 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
                     this.state.selectedInputMode ===
                     ConnectionInputMode.ConnectionString
                 ) {
+                    // convert connection string to an IConnectionProfile object
                     const connDetails =
                         await this._mainController.connectionManager.buildConnectionDetails(
                             this.state.connectionProfile.connectionString,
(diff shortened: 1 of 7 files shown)


src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -634,8 +636,31 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
 
         try {
             try {
+                if (
+                    this.state.selectedInputMode ===
```
kburtram, line 640: I wonder if we should adjust the code formatting rules?  This 4 line if statement seems less readable than the normal ` line way we'd normally write this.

src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -734,6 +766,26 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
         return state;
     }
 
+    private removeDefaultProperties(
+        connProfile: IConnectionProfile,
+    ): IConnectionProfile {
+        // TODO: ideally this compares against the default values acquired from a source of truth (e.g. STS),
+        // so that it can clean up more than just undefined properties.
```
kburtram, line 773: perhaps the name of the method should better reflect the default, i.e., removeUndefinedProperties or something like that?

kburtram (approved):

src/connectionconfig/connectionDialogWebviewController.ts
Benjin: 100% agree that Prettier is wrapping too narrowly.  I'll send a PR to update the width.

lewis-sanchez (approved):

PR #19019, opened 2025-03-27
Schema Designer Editor

aasimkhan30:
Basic diagram
image
Table Editor
image
Foreign Key Editor
image
Filtering 
image
Drag and drop foreign key creation
image
Export diagram
image
Publish changes
image
View Create As script
image

Theme Support:
Light:
image

HC:
image
image


kburtram (approved):

aasimkhan30:
The PR is currently failing due to an 11% increase in the webview bundle size, caused by the addition of the React Flow library. However, this shouldn't affect the performance of other webviews, as each webview is bundled separately and only includes the libraries it directly depends on.

PR #19025, opened 2025-03-28
Adding unit tests for the cross-extension API

Benjin:


src/controllers/connectionManager.ts
```diff
@@ -245,12 +245,19 @@ export default class ConnectionManager {
     }
 
     /**
-     * Get the connection view.
+     * Exposed for testing purposes
```
aasimkhan30, line 248: Maybe just type cast to any to use the method in tests? Exposing a setter runs into the risk of accidental modifications.

aasimkhan30, line 248: Also wasn't the getter already exposed? Maybe revert back the comment for the getter.

.vscode-test.mjs
```diff
@@ -7,7 +7,7 @@ export default testCli.defineConfig([
 		version: 'insiders',
 		mocha: {
 			ui: 'tdd',
-			timeout: 6000
+			timeout: 60_000
```
aasimkhan30, line 10: Did you mean 60000 here? Is it some special syntax? 
Didn't see anything related to `_` here
https://mochajs.org/api/mocha#timeout

aasimkhan30 (approved):

PR #19026, opened 2025-03-28
Updating Prettier rules and applying to all .ts, .tsx, and .json files

Benjin:
**There are no (intentional) logic changes in this PR.  If you see something, say something!**

Two rule changes in this PR:
* increase the `printWidth` from its default 80 characters to 100.  This decreases the aggressiveness of the line wrapping, and IMO generally makes the code far more readable.
* set `bracketSameLine` to `true` to prevent the awkward dangling closing angle-bracket in TSX syntax.

Take a quick scroll through and lemme know if you think this change is overall an improvement, or made things worse.


kburtram (approved):
Oh thank goodness!  I really don't like the way the code reads with all the lines like 40 columns wide, way less readable IMO, and just looks unusual to me.  Now if we can fix the rule putting all the '>' on lines by themself, that would be nice too. ;-)

Issue #18864, opened 2025-02-26
Feature Request: Include live "Time Elapsed" in the Status Bar


djones-breuer:
**Request:** 
Update the total execution time indicator to function like the ADS "Time Elapsed" Status Bar feature. 

**Details:** 
In ADS, the Status Bar includes `Time Elapsed`, a live timer that begins as soon as in-focus query begins execution. 

The vscode-mssql Extension's additions to the Status Bar does include a total execution time indicator for the last-run query, but this is not a live counter like the ADS time elapsed is. It only updates after a given query completes execution. 

**Screenshots**

![Image](https://github.com/user-attachments/assets/354b34e7-11aa-4f32-b1e7-2588df29a330)

![Image](https://github.com/user-attachments/assets/07bd3d69-504c-4771-96ce-30099df46ba0)

jagdishrajan:
Agreed, this is a critically important feature now that you have deprecated ADS.

PR #19052, opened 2025-04-01
Add query execution status timer

cssuh:
Adds a timer to display the seconds elapsed for a running query.
statusTimer

Fixes https://github.com/microsoft/vscode-mssql/issues/18864

### DIFF
--- a/src/views/statusView.ts
+++ b/src/views/statusView.ts
@@ -417,14 +417,13 @@ export default class StatusView implements vscode.Disposable {
         let milliseconds = 0;
         bar.queryTimer = setInterval(() => {
             milliseconds += 1000;
-            //convert milliseconds to display time
-            const timeString = self.formatMilliseconds(milliseconds);
+            const timeString = self.formatMillisecondsToTimeString(milliseconds);
             statusBarItem.text = statusText + " " + timeString;
             self.showStatusBarItem(fileUri, statusBarItem);
         }, 1000);
     }
 
-    private formatMilliseconds(milliseconds: number): string {
+    private formatMillisecondsToTimeString(milliseconds: number): string {
         const minutes = Math.floor(milliseconds / 60000);
         const seconds = ((milliseconds % 60000) / 1000).toFixed(0);
         return minutes + ":" + (parseInt(seconds) < 10 ? "0" : "") + seconds;


src/views/statusView.ts
```diff
@@ -406,19 +413,20 @@ export default class StatusView implements vscode.Disposable {
...
-                index = 0;
-            }
-
-            let progressTick = progressTicks[index];
-            statusBarItem.text = statusText + " " + progressTick;
+        let milliseconds = 0;
+        bar.queryTimer = setInterval(() => {
+            milliseconds += 1000;
+            //convert milliseconds to display time
+            const timeString = self.formatMilliseconds(milliseconds);
```
lewis-sanchez, line 421: Rename the method to `displayFormattedMilliseconds`, so that you can also get rid of the comment above this call.

lewis-sanchez (approved):

src/views/statusView.ts
```diff
@@ -406,19 +413,19 @@ export default class StatusView implements vscode.Disposable {
...
+        bar.queryTimer = setInterval(() => {
+            milliseconds += 1000;
+            const timeString = self.formatMillisecondsToTimeString(milliseconds);
+            statusBarItem.text = statusText + " " + timeString;
             self.showStatusBarItem(fileUri, statusBarItem);
-        }, 200);
+        }, 1000);
+    }
+
+    private formatMillisecondsToTimeString(milliseconds: number): string {
```
kburtram, line 426: you could do the below for queries taking over an hour, but it's probably not needed.

```TypeScript
private formatMillisecondsToTimeString(milliseconds: number): string {
    const hours = Math.floor(milliseconds / 3600000);
    const minutes = Math.floor((milliseconds % 3600000) / 60000);
    const seconds = Math.floor((milliseconds % 60000) / 1000);

    // Pad minutes/seconds for uniform formatting (e.g. "08:05")
    const mm = minutes.toString().padStart(2, "0");
    const ss = seconds.toString().padStart(2, "0");

    if (hours === 0) {
        // no need to show hours if it's 0
        return `${mm}:${ss}`;
    } else {
        // show hours if we have any
        const hh = hours.toString().padStart(2, "0");
        return `${hh}:${mm}:${ss}`;
    }
}
```

kburtram (approved):

PR #19085, opened 2025-04-04
Add custom ESLint rule to prevent importing from reactviews directory in extension. 

aasimkhan30:
Importing from the reactviews directory can break our VSIX packages, as it may introduce dependencies that aren't compatible with the Node environment. These issues typically don't surface during compile time, making them especially problematic. Errors only appear at runtime when activating the extension, after installing the VSIX.

image

After using this custom rule, I will get lint errors and I cannot even push stuff in main because my commit hooks will prevent me from doing that. 

image


caohai (dismissed):

caohai:
Looks like we already have a bunch of "violations" across features, do we think it's necessary to fix all of them?

aasimkhan30:
Hey @caohai, I just fixed the issue. 
This bug is bad as it only affects VSIX builds and doesn’t repro in dev environments. So it can be a while before we catch it. 
I moved everything to shared interfaces to keep things consistent across the codebase.

caohai (dismissed):

Benjin (approved):

lewis-sanchez (approved):

Issue #19108, opened 2025-04-09
Query Result - Right clicking on JSON cell still opens the value in new editor


caohai:
Steps to Reproduce:

1. Run Query:
```SQL
select (
    select 
    'John' AS name,
    25 AS age
    FOR JSON PATH, WITHOUT_ARRAY_WRAPPER
) AS PersonJson

```
2. Right click on the result cell and the cell value still opens up
<img width="395" alt="Image" src="https://github.com/user-attachments/assets/cbdcd71d-7519-4086-909f-29037a60e47f" />

Let me know if you need a gif.

laurenastrid1:
expected behavior: 

![Image](https://github.com/user-attachments/assets/c226e124-235a-431d-976d-7d01653973cc)

Issue #18998, opened 2025-03-24
Query plan SQL text overlays the toolbar


kburtram:
Steps to Reproduce:

1. Get a query plan for a large block of SQL text

Expected: Query text does not obscure the toolbar

Actual: Query text overlays the toolbar buttons

![Image](https://github.com/user-attachments/assets/bba0720e-f267-4abc-9562-df79614195a2)

croblesm:
+1 —I noticed the same problem while working on a demo today.

yooakim:
Same issue for me as can be seen here:

<img width="812" alt="Image" src="https://github.com/user-attachments/assets/dca54e2a-b099-4982-92f4-6c6899ca2505" />

I'm using VS Code version as follows:

```
Version: 1.98.2 (user setup)
Commit: ddc367ed5c8936efe395cffeec279b04ffd7db78
Date: 2025-03-12T13:32:45.399Z
Electron: 34.2.0
ElectronBuildId: 11161602
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Windows_NT x64 10.0.26100
```

Issue #18889, opened 2025-03-02
Result pane should show Average, Count, and Sum for selection just like Azure Data Studio


jagdishrajan:
Update vsccode-mssql to show Average, Count, and Sum for a selection of rows or columns in the result pane, just like Azure Data Studio.  This is important to make vscode-mssql usable now that MSFT has deprecated ADS.

Issue #18539, opened 2024-12-20
Query Result - Rendering Exceptions


caohai:
<img width="812" alt="Image" src="https://github.com/user-attachments/assets/24d6d546-6258-45b7-b7ff-f1e57b0907a8" />

PR #19129, opened 2025-04-11
Fix Query Plan and Query Result GA Bugs

laurenastrid1:
Fixes https://github.com/microsoft/vscode-mssql/issues/18889
Fixes https://github.com/microsoft/vscode-mssql/issues/18539
Fixes https://github.com/microsoft/vscode-mssql/issues/19108
Fixes https://github.com/microsoft/vscode-mssql/issues/18998

Here's a gif of the selection stats in action: 
statsStatusView

### DIFF
--- a/src/queryResult/queryResultWebViewController.ts
+++ b/src/queryResult/queryResultWebViewController.ts
@@ -36,6 +36,12 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
         new Map<string, QueryResultWebviewPanelController>();
     private _sqlOutputContentProvider: SqlOutputContentProvider;
     private _correlationId: string = randomUUID();
+    private _selectionSummaryStatusBarItem: vscode.StatusBarItem =
+        vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Right, 2);
+    private readonly nonNumericStatsLength = 3;
+    private readonly averageIndex = 0;
+    private readonly countIndex = 1;
+    private readonly sumIndex = 6;
     public actualPlanStatuses: string[] = [];
 
     constructor(
@@ -373,5 +379,26 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
             });
         });
         return total;
+    }
+
+    public updateSelectionSummaryStatusItem(updatedText: string) {
+        if (updatedText === "") {
+            this._selectionSummaryStatusBarItem.text = ``;
+            this._selectionSummaryStatusBarItem.hide();
+        } else {
+            const stats = updatedText.split("  ");
+            if (stats.length != this.nonNumericStatsLength) {
+                const shownStats = [
+                    stats[this.averageIndex],
+                    stats[this.countIndex],
+                    stats[this.sumIndex],
+                ];
+                this._selectionSummaryStatusBarItem.text = shownStats.join("  ");
+            } else {
+                this._selectionSummaryStatusBarItem.text = `${updatedText}`;
+            }
+            this._selectionSummaryStatusBarItem.tooltip = `${updatedText}`;
+            this._selectionSummaryStatusBarItem.show();
+        }
     }
 }
--- a/src/queryResult/queryResultWebViewController.ts
+++ b/src/queryResult/queryResultWebViewController.ts
@@ -23,6 +23,7 @@ import {
     recordLength,
     registerCommonRequestHandlers,
 } from "./utils";
+import { QueryResult } from "../constants/locConstants";
 
 export class QueryResultWebviewController extends ReactWebviewViewController<
     qr.QueryResultWebviewState,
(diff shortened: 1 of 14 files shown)


src/reactviews/pages/QueryResult/queryResultPane.tsx
```diff
@@ -375,7 +375,11 @@ export const QueryResultPane = () => {
             const batchId = parseInt(batchIdStr);
             for (const resultIdStr in state?.resultSetSummaries[batchId] ?? {}) {
                 const resultId = parseInt(resultIdStr);
-                grids.push(renderGrid(batchId, resultId, count));
+                grids.push(
+                    <React.Fragment key={`grid-${batchId}-${resultId}`}>
```
caohai, line 379: Is this for fixing the rendering error msg in console?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -374,4 +380,25 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
         });
         return total;
     }
+
+    public updateSelectionSummaryStatusItem(updatedText: string) {
+        if (updatedText === "") {
+            this._selectionSummaryStatusBarItem.text = ``;
```
Benjin, line 386: No need to use formatting quotes here if there's no formatting.

src/queryResult/queryResultWebViewController.ts
```diff
@@ -374,4 +380,25 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+            const stats = updatedText.split("  ");
+            if (stats.length != this.nonNumericStatsLength) {
+                const shownStats = [
+                    stats[this.averageIndex],
+                    stats[this.countIndex],
+                    stats[this.sumIndex],
+                ];
+                this._selectionSummaryStatusBarItem.text = shownStats.join("  ");
+            } else {
+                this._selectionSummaryStatusBarItem.text = `${updatedText}`;
```
Benjin, line 398: Any reason not to just use a normal variable assignment?

src/reactviews/pages/QueryResult/table/plugins/cellSelectionModel.plugin.ts
```diff
@@ -419,6 +427,70 @@ export class CellSelectionModel<T extends Slick.SlickData>
...
+            let sum = 0;
+            let min = Infinity;
+            let max = -Infinity;
+
+            for (let row = firstRange.fromRow; row <= firstRange.toRow; row++) {
+                for (let col = firstRange.fromCell; col <= firstRange.toCell; col++) {
+                    const cell = this.grid.getCellNode(row, col);
+                    if (!cell) continue;
+                    const value = cell.innerText;
+                    const numValue = Number(value);
```
Benjin, line 453: this can go after the NULL check

src/reactviews/pages/QueryResult/table/plugins/cellSelectionModel.plugin.ts
```diff
@@ -419,6 +427,70 @@ export class CellSelectionModel<T extends Slick.SlickData>
...
+                );
+            } else {
+                summary = locConstants.queryResult.nonNumericSelectionSummary(
+                    count,
+                    distinctCount,
+                    nullCount,
+                );
+            }
+        }
+        await this.webViewState.extensionRpc.call("setSelectionSummary", {
```
Benjin, line 489: Why is this formatting this as a string, just to parse the pieces back out again in order to reformat them?  Seems like changing `summary` to be an interface with each stat as a prop would be cleaner.

src/reactviews/pages/QueryResult/queryResultPane.tsx
laurenastrid1: yes!

src/reactviews/pages/QueryResult/table/plugins/cellSelectionModel.plugin.ts
```diff
@@ -419,6 +427,76 @@ export class CellSelectionModel<T extends Slick.SlickData>
         }
     }
 
+    private async setSelectionSummaryText(isSelection?: boolean) {
```
Benjin, line 430: Please add a test for this function!  If you extract this to a helper method that takes grid as an argument, should be easy.

src/reactviews/pages/QueryResult/table/table.ts
```diff
@@ -266,7 +266,8 @@ export class Table<T extends Slick.SlickData> implements IThemable {
         slickEvent.subscribe((e: Slick.EventData) => {
             const originalEvent = (e as JQuery.TriggeredEvent).originalEvent;
             const cell = this._grid.getCellFromEvent(originalEvent!);
-            if (cell) {
+            // If event is left click
+            if (cell && originalEvent instanceof MouseEvent && originalEvent.button === 0) {
```
Benjin, line 270: Better would be to defined an enum somewhere that defines MouseButton.Left = 0 so the code is self-documenting

Benjin (approved):

Benjin (approved):

jagdishrajan:
The Average, Count, Sum on the status bar only work with a mouse selected set of cells.  It doesn't work with column selection or row selection.  Also, the initial count (xxx rows affected) only shows on the status bar after running a SQL query, but doesn't work after running a stored procedures that yields one or more query results.

PR #19130, opened 2025-04-11
Get the correct scmp connection to open scmp files for database endpoint types

lewis-sanchez:
Before:
Opening an scmp file is not able to load source and target correctly because it is not able to find the connection URI and leaves "source" and "target" empty;
image
image

After:
Opening an scmp file loads source and target correctly because it is able to find the connection URI correctly:
image
image

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -329,6 +329,15 @@ export default class ConnectionManager {
         return undefined;
     }
 
+    public getUriForScmpConnection(connection: IConnectionInfo): string {
+        for (let uri of Object.keys(this._connections)) {
+            if (Utils.isSameScmpConnection(this._connections[uri].credentials, connection)) {
+                return uri;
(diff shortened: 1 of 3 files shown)


aasimkhan30 (approved):

Benjin:
I came in a little too late for this, so dropping comments for posterity @lewis-sanchez :

I don't think that `connectionManager` and `utils` should have scmp-specific functions in them, especially when they're so similar to existing functions in the same spots.

What about this instead:
* update `getUriForConnection()` to take an optional custom comparison function
* move `isSameScmpConnection()` to the schemaCompare controller, and pass that in
* Then since isSameScmpConnection() is now with the other scmp code, we can be a little more detailed with the docstring, i.e. mentioning that it picks the first _similar_ connection.

I suppose in the future, we'll want to simply open a new connection with the endpoint info in the scmp file, but we're nowhere close to there yet.

PR #19139, opened 2025-04-14
Removing freeform question from feedback prompt

Benjin:
Holding this PR pending guidance in this area.


Benjin:
@kburtram We can also remove (or disable) the feedback prompt in its entirety instead, if you think that makes more sense for where we are currently.

lewis-sanchez (approved):

aasimkhan30 (approved):

Benjin:
Closing as this is resolved on the backend.

PR #19142, opened 2025-04-14
Using Ids to match is same connection profile.

### LINKED ISSUE #19091: Duplicate connection entry created when connecting to saved connection
Steps to Reproduce:

*  No definitive reproduction steps found yet.

*  Occurred during normal usage, specifically when connecting to a saved connection.

*  STS was under debug mode at the time. Observed a recoverable SQL client exception in logs, but unclear if related.

Expected Behavior: Connecting to an existing saved connection does not create duplicate connection entries.

Actual Behavior: A duplicate connection entry appeared. Notably, the duplicated connection included the server version info, which wasn't observed in previous duplicate connection bugs. (Uncertain if displaying server version is new behavior or simply unnoticed previously.)

Image

aasimkhan30:
Fixes #19091

This PR resolves an issue where connections created using a connection string and the old connection prompt flow were not being correctly reconciled in the Object Explorer (OE).

Initially, we stored the disconnected node's profile like this:
```json
{
  "connectionString": "{lookup-key}"
}
```

But once the connection was established, the node's profile changed to:
```json
{
  "connectionString": "{actual-connection}"
}
```

This mismatch caused OE to treat them as different connections, so instead of replacing the disconnected node with the connected one, it just added a new connected node leading to duplicates.

This PR ensures the profiles match correctly so OE can update the existing node instead of duplicating it.

### DIFF
--- a/src/models/utils.ts
+++ b/src/models/utils.ts
@@ -261,6 +261,10 @@ export function isSameProfile(
     currentProfile: IConnectionProfile,
     expectedProfile: IConnectionProfile,
 ): boolean {
+    if (currentProfile.id && expectedProfile.id) {
+        // If connection profile has an id, use that to compare connections
+        return currentProfile.id === expectedProfile.id;
+    }
     if (currentProfile === undefined) {
         return false;
     }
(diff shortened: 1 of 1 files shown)


lewis-sanchez (approved):

src/models/utils.ts
```diff
@@ -308,6 +312,13 @@ export function isSameConnectionInfo(
     conn: IConnectionInfo,
     expectedConn: IConnectionInfo,
 ): boolean {
+    // If connection info has an id, use that to compare connections
+    const connId = (<IConnectionProfile>conn).id;
```
Benjin, line 316: Nit: good practice to use `as` instead of `<>` for casting just due to JSX/TSX portability.

Benjin (approved):

Benjin (approved):

lewis-sanchez (approved):

PR #19150, opened 2025-04-15
Adding separate icon for db based connections.

aasimkhan30:
### Improved Icon for Database-Level Connection Nodes

Now that we support database-specific features like **"Chat with Database"** and **"Schema Designer"**, it makes sense to visually distinguish **database-level connection nodes** from server-level ones.

These database nodes come with a different context menu and behavior, so we’ve introduced a **new icon** to reflect that difference.

---

### Before:
All connection nodes used the same icon, regardless of whether they were server-based or database-based:  
Before

---

### Now:
Database-level connections (both connected and disconnected) now use a new, distinct icon:  
After

This makes it easier for users to recognize nodes that support database-scoped features and know what to expect from their context menus.


aasimkhan30:
@benjin, I guess it doesn't work on purely connection string based profiles. However, since we are phasing it out, I don't see the point of adding additional code to support that scenario.

lewis-sanchez (approved):

Issue #19160, opened 2025-04-15
selecting an item in schema compare differences doesn't persist any indicator which item I selected


dzsquared:
Version: 1.99.2
Commit: 4949701c880d4bdb949e3c0e6b400288da7f474b
Date: 2025-04-10T01:21:25.295Z
Electron: 34.3.2
ElectronBuildId: 11161073
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

Identifier
ms-mssql.mssql
Version
1.31.0
Last Updated
2025-04-15, 10:24:20
Source
VSIX

Steps to Reproduce:

1. compare a source and target database, especially ones with a large number of differences
2. select any difference, move your cursor away from the difference - now you can't tell which one you selected unless you read the script and look for the name again in the list

![Image](https://github.com/user-attachments/assets/2de93252-a2d4-45d5-a15b-8c7a57040164)

PR #19183, opened 2025-04-15
Select first diff after comparison completes automatically

### LINKED ISSUE #19146: Majority of schema compare dashboard is taken up by black screen
Majority of the compare screen is taken up by a black screen, this is unavoidable, and you cannot get rid of it. This does not happen in the ADS experience. Solution would be to not have anything blocking the dashboard initially, unless a user clicks on an object. 

Screenshot of the blocked view: 

Image

### LINKED ISSUE #19159: Initial comparison has large blank area under list of changes
Design issue.  Possible solutions:
* this large area should have a message like "Select a change to see the SQL differences"
* large area should start completely collapsed until the user selects a change.
* automatically select the first item to display the delta for

Probably 1 or 3.

Image

lewis-sanchez:
This PR fixes #19146 
This PR fixes #19159
This PR fixes https://github.com/microsoft/vscode-mssql/issues/19160

This PR selects the first diff after a comparison to avoid showing an empty space below the list of schema differences after a comparison is made and highlights the first diff item in the diff list.
image

Selecting a different list item will change which diff item is highlighted:
image

### DIFF
--- a/src/reactviews/pages/SchemaCompare/SchemaCompare.tsx
+++ b/src/reactviews/pages/SchemaCompare/SchemaCompare.tsx
@@ -64,7 +64,12 @@ export const SchemaComparePage = () => {
 
             {showMessage() && <Message />}
 
-            {!showMessage() && <SchemaDifferences onDiffSelected={handleDiffSelected} />}
+            {!showMessage() && (
+                <SchemaDifferences
+                    selectedDiffId={selectedDiffId}
+                    onDiffSelected={handleDiffSelected}
+                />
+            )}
 
             {!showMessage() && <CompareDiffEditor selectedDiffId={selectedDiffId} />}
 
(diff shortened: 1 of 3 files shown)


Benjin:
Also, I'm noticing that there isn't any "currently selected" indication when a diff _is_ selected to be shown in the script diff view.  We definitely need that.

image

src/reactviews/pages/SchemaCompare/SchemaCompare.tsx
```diff
@@ -66,9 +66,7 @@ export const SchemaComparePage = () => {
 
             {!showMessage() && <SchemaDifferences onDiffSelected={handleDiffSelected} />}
```
Benjin, line 67: I'd expect that the current diff selection also gets piped through to the list of schema differences so that row can be highlighted.

lewis-sanchez:
> Also, I'm noticing that there isn't any "currently selected" indication when a diff _is_ selected to be shown in the script diff view. We definitely need that.
> 
> image

We have an issue for that here: https://github.com/microsoft/vscode-mssql/issues/19160

I'll bundle the fix for 19160 in here too since they're closely related.

aasimkhan30 (approved):

PR #19188, opened 2025-04-16
Add Resizing to schema compare diff list

### LINKED ISSUE #19145: Unable to minimize compare panel
When you run a compare the panel takes up a majority of the screen and the user is not able to minimize the panel at all. The experience in ADS allowed you to do this so user will see this as a downgrade in VS Code. Also, because of this you can only view a list of 4-5 objects at a time.  

Screenshot of panel in VS Code and compared to ADS:

Image

Image

### LINKED ISSUE #19181: Schema compare - have a way to vertically resize the diff section
There should be a splitter so that the diff section can be resized, so users can see more of the script diff they want and they can see more of the item comparison if they want.

lewis-sanchez:
This PR fixes #19145
This PR fixes #19181

This PR adds functionality to resize the list of schema compare differences and the diff editor
 
image

image

### DIFF
--- a/src/reactviews/hooks/useResizable.ts
+++ b/src/reactviews/hooks/useResizable.ts
@@ -1,0 +1,92 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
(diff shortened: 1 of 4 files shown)


Benjin:
Can you combine https://github.com/microsoft/vscode-mssql/pull/19155 into this PR so that we can see the related/adjacent changes working together?

lewis-sanchez:
The merge conflicts in this PR got confusing to work with, so I opened a new PR: https://github.com/microsoft/vscode-mssql/pull/19205

PR #19195, opened 2025-04-16
Correct logic to disable buttons in Schema Compare panel

### LINKED ISSUE #19170: Schema Compare - Only Stop button should be enabled when a comparison is happening
When a comparison is happening, only the "Stop" button should be enabled. All the other ones should not be enabled, since they should only be able to be changed before or after a comparison happens. Like "Generate Script" should only be enabled after the comparison has completed
Image

### LINKED ISSUE #19177: Clicking "generate script" on comparison with all changes excluded always displays "succeeded" message, but doesn't produce script

lewis-sanchez:
This PR fixes #19170
This PR fixes #19177

When a schema comparison is in progress, all buttons except for "stop" are disabled
disable_buttons_with_running_comparison

If all differences are excluded, the generate script button is disabled.
NEW: The apply button is also fixed to address @tauseefsiddique10's comment in the discussion.
image

Including one or more diffs will re-enable the generate script button
including_a_single_diff_entry_will_enable_generate_script_btn

A comparison with no found differences will not have the generate script button enabled.
no_differences_in_a_comparison_will_disable_generate_script_btn

### DIFF
--- a/src/reactviews/pages/SchemaCompare/components/CompareActionBar.tsx
+++ b/src/reactviews/pages/SchemaCompare/components/CompareActionBar.tsx
@@ -87,6 +87,14 @@ const CompareActionBar = (props: Props) => {
         );
     };
 
+    const hasIncludedDiffs = (): boolean => {
+        const includedDiffs = context.state.schemaCompareResult.differences.filter(
+            (diff) => diff.included,
+        );
+
+        return includedDiffs.length > 0;
+    };
+
     const disableGenerateScriptButton = (): boolean => {
         if (
             !(
@@ -106,10 +114,8 @@ const CompareActionBar = (props: Props) => {
             return true;
         }
 
-        const includedDiffs = context.state.schemaCompareResult.differences.filter(
-            (diff) => diff.included,
-        );
-        if (includedDiffs.length === 0) {
+        const isIncludingDiffs = hasIncludedDiffs();
+        if (!isIncludingDiffs) {
             return true;
         }
 
@@ -123,6 +129,11 @@ const CompareActionBar = (props: Props) => {
             context.state.schemaCompareResult.differences.length > 0 &&
             Number(context.state.targetEndpointInfo.endpointType) !== 1 // Dacpac lewissanchez todo: Figure out how to move away from these magic numbers for enums
         ) {
+            const isIncludingDiffs = hasIncludedDiffs();
+            if (!isIncludingDiffs) {
+                return true;
+            }
+
             return false;
         }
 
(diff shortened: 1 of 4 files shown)


src/reactviews/pages/SchemaCompare/components/CompareActionBar.tsx
```diff
@@ -87,6 +87,35 @@ const CompareActionBar = (props: Props) => {
         );
     };
 
+    const disableGenerateScriptButton = (): boolean => {
+        if (
+            !(
+                (
+                    context.state.targetEndpointInfo &&
+                    Number(context.state.targetEndpointInfo.endpointType) === 0
```
Benjin, line 95: Better to use the SchemaCompareEndpointType enum instead of a hardcoded number with a comment.

Benjin (approved):

src/reactviews/pages/SchemaCompare/components/CompareActionBar.tsx
```diff
@@ -87,13 +88,53 @@ const CompareActionBar = (props: Props) => {
         );
     };
 
+    const hasIncludedDiffs = (): boolean => {
+        const includedDiffs = context.state.schemaCompareResult.differences.filter(
```
laurenastrid1, line 92: Try this instead:

return context.state.schemaCompareResult.differences.some(
        (diff) => diff.included,
);

instead of making an entire new list, we can just check for any existing diff that is included

laurenastrid1, line 92: Try this instead:

return context.state.schemaCompareResult.differences.some(
(diff) => diff.included,
);

instead of making an entire new list through filter, we can just check for any existing diff that is included

src/reactviews/pages/SchemaCompare/components/CompareActionBar.tsx
```diff
@@ -87,13 +88,53 @@ const CompareActionBar = (props: Props) => {
...
             context.state.schemaCompareResult &&
             context.state.schemaCompareResult.differences &&
             context.state.schemaCompareResult.differences.length > 0 &&
-            Number(context.state.targetEndpointInfo.endpointType) !== 1 // Dacpac lewissanchez todo: Figure out how to move away from these magic numbers for enums
+            Number(context.state.targetEndpointInfo.endpointType) !==
+                SchemaCompareEndpointType.Dacpac
         ) {
+            const isIncludingDiffs = hasIncludedDiffs();
+            if (!isIncludingDiffs) {
+                return true;
```
laurenastrid1, line 135: return !hasIncludedDiffs()

Benjin (approved):

PR #19197, opened 2025-04-17
Adding conn-id to OE session id and also when creating new connections.

aasimkhan30:
As a part of switching everything to connection ids, I am adding connection id to OE session id and connection id when we create a new connection in connection dialog. 

This updates the STS version which also brings in Dacfx 170. 

Note about PR: 
I see that a lot of code is duplicated across different places, Ideally we should consolidate and centralize all connection and OE related code. However, those fixes will be made for the next release. For now, I am just adding stopgap fixes.

### DIFF
--- a/src/connectionconfig/connectionDialogWebviewController.ts
+++ b/src/connectionconfig/connectionDialogWebviewController.ts
@@ -64,7 +64,6 @@ import { generateConnectionComponents, groupAdvancedOptions } from "./formCompon
 import { FormWebviewController } from "../forms/formWebviewController";
 import { ConnectionCredentials } from "../models/connectionCredentials";
 import { Deferred } from "../protocol";
-import { generateGuid } from "../models/utils";
 
 export class ConnectionDialogWebviewController extends FormWebviewController<
     IConnectionDialogProfile,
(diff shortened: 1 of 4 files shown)


src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -655,6 +656,10 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
                 cleanedConnection as any,
             );
 
+            if (!cleanedConnection.id) {
```
Benjin, line 659: Can you use `ConnectionProfile.addIdIfMissing(profile);` so we can more easily track everywhere that IDs are getting added?  Might help with debugging later.

Benjin: Actually, [this PR](https://github.com/microsoft/vscode-mssql/pull/19137/files) that I currently have out fixes the same issue, but in a more consolidated way (in `connectionStore.saveProfile()`).  I think that approach is better here.

src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -64,6 +64,7 @@ import { generateConnectionComponents, groupAdvancedOptions } from "./formCompon
 import { FormWebviewController } from "../forms/formWebviewController";
 import { ConnectionCredentials } from "../models/connectionCredentials";
 import { Deferred } from "../protocol";
+import { generateGuid } from "../models/utils";
```
Benjin, line 67: no longer needed

Benjin (approved):

Benjin (approved):

Issue #18507, opened 2024-12-12
Fix select multiple ranges in query result grid


cssuh:
Steps to Reproduce:

1. Run a query
2. Click on a cell in the query result grid
3. Try to control + click to select another cell in the query result grid

Expected:
Able to select multiple cells

Actual:
Context menu opens (on Mac, need to test Windows as well)

PR #19227, opened 2025-04-22
Add command + click to select multiple ranges on mac

cssuh:
Adds command + click functionality to mac so users can select multiple ranges in the query results grid. (this works via Ctrl + click on windows)

Fixes https://github.com/microsoft/vscode-mssql/issues/18507

### DIFF
--- a/src/reactviews/pages/QueryResult/table/plugins/cellRangeSelector.ts
+++ b/src/reactviews/pages/QueryResult/table/plugins/cellRangeSelector.ts
@@ -3,6 +3,11 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import {
+    QueryResultWebviewState,
+    QueryResultReducers,
+} from "../../../../../sharedInterfaces/queryResult";
+import { VscodeWebviewContext } from "../../../../common/vscodeWebviewProvider";
 import { mixin } from "../objects";
 
 const defaultOptions: ICellRangeSelectorOptions = {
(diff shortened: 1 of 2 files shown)


src/reactviews/pages/QueryResult/table/plugins/cellSelectionModel.plugin.ts
```diff
@@ -58,16 +59,20 @@ export class CellSelectionModel<T extends Slick.SlickData>
...
+                    },
                 },
-            });
+                this.webViewState,
+            );
         }
     }
 
-    public init(grid: Slick.Grid<T>) {
+    public async init(grid: Slick.Grid<T>) {
```
Benjin, line 73: Making this async without having something that checks for initialization to be complete (awaiting this) can lead to unexpected behavior.  Does this need to be checked somewhere?

src/reactviews/pages/QueryResult/table/plugins/cellRangeSelector.ts
```diff
@@ -43,20 +48,25 @@ export class CellRangeSelector<T extends Slick.SlickData> implements ICellRangeS
...
+    private platform: string | undefined;
 
     public onBeforeCellRangeSelected = new Slick.Event<Slick.Cell>();
     public onCellRangeSelected = new Slick.Event<Slick.Range>();
     public onAppendCellRangeSelected = new Slick.Event<Slick.Range>();
 
-    constructor(private options: ICellRangeSelectorOptions) {
+    constructor(
+        private options: ICellRangeSelectorOptions,
+        private webViewState: VscodeWebviewContext<QueryResultWebviewState, QueryResultReducers>,
```
Benjin, line 59: why not just make platform part of the state and pass that all the way through instead of adding RPC calls and introducing complexity?

Also, dumb question, but can `process.platform` not be accessed directly in the react layer?

Benjin (approved):

PR #19233, opened 2025-04-22
Fixing duplicate connection showing up in OE

aasimkhan30:
Issue: #19091:

Making sure we consistently generate label for connection nodes and we only use connection ids to replace disconnected node with connected node.


caohai (approved):
I verified the fix with my saved connections and can confirm the duplicate nodes are now gone.

Benjin (approved):

PR #19252, opened 2025-04-23
Minor Query Results fixes

Benjin:
* Localized "Query Results" pane title
* Localized "Sort Ascending" and "Show Filter" text in query results pane
* Updating Sort tooltip state according to the current sort mode (previously was always "Sort Ascending")

image


cssuh (approved):

kburtram (approved):

aasimkhan30 (approved):

src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -31,10 +31,7 @@ export interface CommandEventArgs<T extends Slick.SlickData> {
     command: SortDirection;
 }
 
-const ShowFilterText = locConstants.queryResult.showFilter;
```
Benjin, line 34: this was saving the value from locConstants the very first time this loaded, and never again.   Since this would run before loc had loaded, these would always be the English strings.

PR #19275, opened 2025-04-25
Preserve results when user saves file (changing URI)

### LINKED ISSUE #18568: Query Results disappear when saving sql file
- MSSQL Extension Version: 1.27.0

Steps to Reproduce:

1. Open a new file 
2. Run a query
3. Select Save As... and save your file
4. Query results pane displays the "No results" view

I suspect this is due to the URI for the file changing, but we should find some way to preserve the query results when the URI changes from saving a file.

cssuh:
[GIF in progress]

Adds handling to transfer the URI mapping for a query editor & it's result set when the user saves the file.  

Note: I had to increase the `untitledSaveTimeThreshold` to 50 (ms) to capture the corresponding close & save event correctly because I never saw this value lower than 10 ms in my testing.

Fixes #18568

### DIFF
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -126,8 +126,8 @@ export const outputContentTypeShowWarning = "showWarning";
 export const outputServiceLocalhost = "http://localhost:";
 export const msgContentProviderSqlOutputHtml = "dist/html/sqlOutput.ejs";
 export const contentProviderMinFile = "dist/js/app.min.js";
-export const untitledSaveTimeThreshold = 10.0;
-export const renamedOpenTimeThreshold = 10.0;
+export const untitledSaveTimeThreshold = 40.0;
+export const renamedOpenTimeThreshold = 200.0;
 export const timeToWaitForLanguageModeChange = 10000.0;
 export const macOpenSslHelpLink =
     "https://github.com/Microsoft/vscode-mssql/wiki/OpenSSL-Configuration";
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -126,8 +126,8 @@ export const outputContentTypeShowWarning = "showWarning";
 export const outputServiceLocalhost = "http://localhost:";
 export const msgContentProviderSqlOutputHtml = "dist/html/sqlOutput.ejs";
 export const contentProviderMinFile = "dist/js/app.min.js";
-export const untitledSaveTimeThreshold = 40.0;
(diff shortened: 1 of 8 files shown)


src/constants/constants.ts
```diff
@@ -126,7 +126,7 @@ export const outputContentTypeShowWarning = "showWarning";
 export const outputServiceLocalhost = "http://localhost:";
 export const msgContentProviderSqlOutputHtml = "dist/html/sqlOutput.ejs";
 export const contentProviderMinFile = "dist/js/app.min.js";
-export const untitledSaveTimeThreshold = 10.0;
+export const untitledSaveTimeThreshold = 50.0;
```
Benjin, line 129: In your testing, how high did the gap between events tend to be?  10ms to 50ms is a big jump.

cssuh: Just ran it 5 times in a row, here are the results:
Mac:
12 ms
14 ms
13 ms
8 ms
11 ms


Windows:
26 ms
29 ms
30 ms
41 ms
32 ms

src/controllers/mainController.ts
```diff
@@ -1783,16 +1784,54 @@ export default class MainController implements vscode.Disposable {
...
+            if (state) {
+                state.uri = this._lastSavedUri;
+                this._queryResultWebviewController.postNotification(
+                    DefaultWebviewNotifications.updateState,
+                    state,
+                );
+                this._queryResultWebviewController.setQueryResultState(this._lastSavedUri, state);
+                this._queryResultWebviewController.deleteQueryResultState(closedDocumentUri);
+            }
+            this._outputContentProvider.onUntitledFileSaved(closedDocumentUri, this._lastSavedUri);
```
Benjin, line 1801: Looks like this logic is mostly duplicated below (only difference is the new URI).  Can you move it to a shared function?

src/controllers/mainController.ts
```diff
@@ -1783,16 +1784,54 @@ export default class MainController implements vscode.Disposable {
...
+            this._lastSavedTimer.getDuration() < Constants.untitledSaveTimeThreshold
         ) {
             // File was renamed and connection will be transfered
             await this._connectionMgr.transferFileConnection(
                 closedDocumentUri,
                 this._lastOpenedUri,
             );
+
+            // Call STS to update URI
+            this._outputContentProvider.updateQueryRunnerUri(
```
Benjin, line 1814: Does the unsaved -> saved case need this?

Issue #18865, opened 2025-02-27
Feature Request: Color Coding for DB Connections -- Clearer Indication of what db you are working in.


ehartford2:
I work in many db's and environments throughout the day and currenlty, the only spot I can see which db a query is operating on is in a tiny little box down in the lower right hand corner.

With ADS I was able to group the database by environment, db type and color code the tabs so it was clear when I was operating on Product data.

winzig:
This is the #1 request I have and basically the only reason I can't use this extension currently, it's too dangerous otherwise.

swaschitz:
Having this as a feature would be great.  The other methods also are not helpful because the server + database name area is experiencing truncation as I have just reported in #19485 leading to serious danger in running against the wrong database.  Ability for my developers to accidentally run against the wrong database is the major limiting factor for our team.

PureKrome:
Excellent request. I also asked for this in [ADS a year ago](https://github.com/microsoft/azuredatastudio/issues/25692), way before the retirement announcement.

Such an important VISUAL indicator.

croblesm:
Hi @ehartford2 for your suggestion. This feature was implemented in the [v1.34.0 release](https://github.com/microsoft/vscode-mssql/releases/tag/1.34.0).

PR #19276, opened 2025-04-26
Add max length configuration for connection info in status bar

bathetrade:
Allows configuring the display width of the connection info in the status bar.

This makes it possible to extend the width and see the entire server name, database name, and user name without needing to hover over the item with the mouse. This in turn reduces the likelihood that a query is issued to the wrong database and may be more familiar for SSMS users where such information is always displayed in the title bar.


bathetrade:

kburtram:
@Benjin what do you think of this one?  It looks fine to me.  Though maybe the default should be -1, or 100.  Or maybe 50 is enough for most people?

bathetrade:
@kburtram The default width before this PR was 50 (Constants.maxDisplayedStatusTextLength), so I just left it at that. The servers, databases, and usernames I regularly work with however aren't particularly long and are truncated enough that a good amount of the database name is hidden, so I imagine increasing the default would be a welcome change.

src/models/connectionInfo.ts
```diff
@@ -149,9 +150,19 @@ export function getConnectionDisplayString(creds: IConnectionInfo): string {
     let user: string = getUserNameOrDomainLogin(creds);
     text = appendIfNotEmpty(text, user);
 
+    let statusBarConnectionInfoMaxLength: number = vscode.workspace
+        .getConfiguration(Constants.extensionConfigSectionName)
+        .get(Constants.configStatusBarConnectionInfoMaxLength);
+
+    statusBarConnectionInfoMaxLength ||= Constants.maxDisplayedStatusTextLength;
```
Benjin, line 157: If the config value has a default, this'll never be undefined, so you can just drop this line and delete `maxDisplayedStatusTextLength` from `Constants`.

Benjin (approved):

Benjin:
@kburtram agreed; this is a nice addition!  I left one minor comment, then I'd be happy to take it.

src/models/connectionInfo.ts
bathetrade: Done. I bumped the default up to 100 as well.

Benjin:
Sorry for taking a while to circle back to this; @bathetrade, can you resolve the conflicts?  Then I'll be ready to merge it in :)

bathetrade:
@Benjin I resolved the merge conflicts and made a couple more changes:

1. Removed nesting of `connectionInfo` test suites since the "getConnectionDisplayString" suite wasn't running or showing up in the test explorer for me
2. Moved VS Code configuration loading out of `getConnectionDisplayString` to make it easier to test and more flexible for callers
3. Updated the README to reflect the new setting
4. Changed the default length to -1 instead of 100 in light of #19485 and #18865

While testing, I ran into the following issues (unrelated to this PR)
1. Debug with "Launch Extension (With Other Extensions Disabled)"
2. Open a new file and change the type to SQL
3. Press `ctrl+shift+c` to open the connection picker and select a profile
4. image
5. Open the extension settings from the extension's cogwheel in the sidebar
6. Change a setting (I used the new status bar setting, but others seem to have the same behavior)
7. Click somewhere outside of the setting area to commit the change
8. image

I tested off `main` and was getting the same behavior but wanted to mention it anyway in case it's not yet being tracked.

Another issue is that the display string is now using icon tokens such as "$(database)". If you specify a max length that lands in the middle of such a token, the user will see something like "myServer: $(data ..." in the status bar. I didn't have time to fix it but would be happy to at some point if deemed worthwhile.

swaschitz:
As an update since this feature/change doesn't appear done, with the release of 1.33.0, the default behavior has actually gotten worse.  It now truncates earlier.  

Image

Benjin (approved):

PR #19289, opened 2025-04-29
Fixing parallel node expansion.

aasimkhan30:
Returning a temporary loading node to keep the UI responsive in the tree. It is necessary to do this in order to allow multiple refresh of the tree. 

How the UI looks for long running expansions.

https://github.com/user-attachments/assets/efa00ee2-6733-458b-9827-2acab20cf1d2

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -25,7 +25,11 @@ import {
     CloseSessionResponse,
 } from "../models/contracts/objectExplorer/closeSessionRequest";
 import { TreeNodeInfo } from "./nodes/treeNodeInfo";
-import { AuthenticationTypes, IConnectionProfile } from "../models/interfaces";
+import {
+    AuthenticationTypes,
+    IConnectionProfile,
+    IConnectionProfileWithSource,
+} from "../models/interfaces";
 import * as LocalizedConstants from "../constants/locConstants";
 import { AddConnectionTreeNode } from "./nodes/addConnectionTreeNode";
 import { AccountSignInTreeNode } from "./nodes/accountSignInTreeNode";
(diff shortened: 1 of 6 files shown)


src/objectExplorer/objectExplorerService.ts
```diff
@@ -370,7 +388,21 @@ export class ObjectExplorerService {
                 return this._treeNodeToChildrenMap.get(element);
             }
         }
-        return this.getOrCreateNodeChildrenWithSession(element);
+        /**
+         * If no children are cached, return a temporary loading node to keep the UI responsive
+         * and trigger the async call to fetch real children.
+         * This node will be replaced once the data is retrieved and the tree is refreshed.
+         * Without this, tree expansion is queued—so if multiple connections are expanding,
```
Benjin, line 395: Nit: confusing wording.  Maybe:
> Tree expansion is queued, so without this if multiple...

src/objectExplorer/objectExplorerService.ts
```diff
@@ -302,6 +306,20 @@ export class ObjectExplorerService {
...
         let savedConnections = await this._connectionManager.connectionStore.readAllConnections();
+        // Remove any connections that have duplicated IDs
+        if (savedConnections.length > 0) {
+            const uniqueConnections = new Map<string, IConnectionProfileWithSource>();
+            for (const conn of savedConnections) {
+                if (!uniqueConnections.has(conn.id)) {
+                    uniqueConnections.set(conn.id, conn);
+                } else {
+                    this._logger.verbose(
+                        `Duplicate connection ID found: ${conn.id}. Removing duplicate connection.`,
```
Benjin, line 317: Is it "removing" the duplicate connection or "ignoring"?

src/objectExplorer/objectExplorerService.ts
```diff
@@ -302,6 +306,20 @@ export class ObjectExplorerService {
...
+                    );
+                }
+            }
+            savedConnections = Array.from(uniqueConnections.values());
```
Benjin, line 321: I feel like this logic should be in the readAllConnections() method for universality instead of in OE.  What do you think?

Benjin (approved):

src/objectExplorer/objectExplorerService.ts
aasimkhan30: Yup, I agree. This is just a sanity check for OE  so we don't run into any duplicate id issues. Ideally, we should have this in `readAllConnections` with an error message surfaced to the users.

Benjin (approved):

Issue #19243, opened 2025-04-23
Canceling subscription filter in Connection Dialog unnecessarily refreshes subscription list


Benjin:
Repro:
1. Open Connection Dialog
2. Select "Browse Azure"
3. Wait for servers to load.
4. Click subscription filter button
5. Cancel the selection prompt

Expected:
* Prompt closes, nothing refreshed because nothing changed

Reality:
* Prompt closes, list of servers is refreshed unnecesssarily.

PR #19296, opened 2025-04-30
Fixing bug where cancelling azure subscription filter reloads filters

Benjin:
Resolves https://github.com/microsoft/vscode-mssql/issues/19243


src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -291,9 +291,9 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
         });
 
         this.registerReducer("filterAzureSubscriptions", async (state) => {
-            await promptForAzureSubscriptionFilter(state);
-            await this.loadAllAzureServers(state);
-
+            if (await promptForAzureSubscriptionFilter(state)) {
+                await this.loadAllAzureServers(state);
+            }
```
aasimkhan30, line 296: Maybe add a return statement in finally? What happens if the functions error out?

aasimkhan30, line 296: Maybe add a return statement in finally? What happens if the functions error out?

aasimkhan30 (changes requested):

src/connectionconfig/connectionDialogWebviewController.ts
Benjin: Both functions that get called are fully wrapped in try-catches, but I'll add that here just in case something slips through.

aasimkhan30 (approved):

aasimkhan30 (approved):

Issue #19104, opened 2025-04-09
Missing unsaved changes prompt


croblesm:
Suppose I spend a considerable amount of time designing multiple tables and defining their relationships using Schema Designer. If I accidentally close all VS Code windows (or the window crashes), all my work is lost because I never receive a warning or prompt indicating that there are unsaved changes pending to be applied or saved.

This results in a poor user experience and also frustration.

The Schema Designer experience should implement unsaved changes prompt (or dirty state detection) to warn users before closing the window or navigating away when there are unapplied or unsaved changes. This could be similar to a “close guard” or exit confirmation dialog commonly found in many design tools and editors.

https://github.com/user-attachments/assets/ffaefb31-64e2-4c5e-b02e-4d51748e4b90

Suggested prompt:
**You have unsaved changes in your schema. Are you sure you want to exit without saving?**

cc @yoleichen

PR #19306, opened 2025-04-30
Fixed Schema Designer Save Prompt Bug

laurenastrid1:
Fixes https://github.com/microsoft/vscode-mssql/issues/19104

schemaDesignerSave


package.json
```diff
@@ -636,6 +636,10 @@
                 {
                     "command": "mssql.editConnection",
                     "when": "config.mssql.enableRichExperiences"
+                },
```
aasimkhan30, line 639: Why are we adding this command here?

src/schemaDesigner/schemaDesignerWebviewController.ts
```diff
@@ -75,11 +80,23 @@ export class SchemaDesignerWebviewController extends ReactWebviewPanelController
         });
 
         this.registerRequestHandler("initializeSchemaDesigner", async () => {
-            const sessionResponse = await this.schemaDesignerService.createSession({
-                connectionString: this.connectionString,
-                accessToken: this.accessToken,
-                databaseName: this.databaseName,
-            });
+            let sessionResponse: SchemaDesigner.CreateSessionResponse;
```
aasimkhan30, line 83: It seems we are only storing state after we perform certain sts based actions? 

I guess we should create a new request handler just for syncing state. Since this won't capture some user actions like moving tables/ editing tables.

package.json
laurenastrid1: before, you could access the command through the command prompt with no OE node selected, which would make it error out

aasimkhan30 (approved):

PR #19328, opened 2025-05-02
Fixing multiple_matching_tokens in Entra Auth.

### LINKED ISSUE #19286: Sign in issues for Azure SQL databases with Entra
**Area - Connection**

## Description 
When connecting to Azure SQL DB or Fabric SQL DB, with Entra ID i run into error to connect since there are too many tokens 

## Steps to repro for Azure SQL 
1. Open MSSQL Ext
2. Select browse azure and provide DB details 
3. Select Entra ID auth 
4. Select sign in button to authenticate
5. Select connect button 
6. Fails with error shown below 

Image

**Error**

Image

Once the extension is in this state it fails for any other databases you connect unless cache is cleared. The experience here can be better to avoid asking for auth when user clicks on Sign in and connect and rather do this on connect itself

aasimkhan30:
This PR fixes #19286 

This fix is inspired by
https://github.com/microsoft/azuredatastudio/pull/23441

### DIFF
--- a/src/azure/constants.ts
+++ b/src/azure/constants.ts
@@ -86,7 +86,7 @@ export const mdsUserAccountNotReceived = "User account not received.";
  */
 export const noAccountInSilentRequestError = "no_account_in_silent_request";
 /**
- * multiple_matching_tokens error can occur in scenarios when users try to run ADS as different users, reference issue:
+ * multiple_matching_tokens error can occur in scenarios when users try to run vscode-mssql as different users, reference issue:
  * https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/5134
  * Error message: multiple_matching_tokens The cache contains multiple tokens satisfying the requirements.
  * Call AcquireToken again providing more requirements such as authority or account.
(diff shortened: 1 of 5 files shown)


src/azure/constants.ts
```diff
@@ -85,6 +85,13 @@ export const mdsUserAccountNotReceived = "User account not received.";
  * Thrown by TS layer, indicates user account hint not provided. We request user to reauthenticate when this error occurs.
  */
 export const noAccountInSilentRequestError = "no_account_in_silent_request";
+/**
+ * multiple_matching_tokens error can occur in scenarios when users try to run ADS as different users, reference issue:
```
Benjin, line 89: > run ADS as different users

should the comment get updated for MSSQL?

src/azure/msal/msalAzureController.ts
```diff
@@ -201,6 +202,20 @@ export class MsalAzureController extends AzureController {
                 } catch (ex) {
                     this._vscodeWrapper.showErrorMessage(ex);
                 }
+            }
+            if (getErrorMessage(ex).includes(AzureConstants.multiple_matching_tokens_error)) {
+                const response = await this._vscodeWrapper.showErrorMessage(
+                    LocalizedConstants.ConnectionDialog.multipleMatchingTokensError(
```
Benjin, line 208: If `account?.displayInfo?` ends up being undefined, we should have another message that doesn't add the account info so it looks cleaner.

PR #19341, opened 2025-05-05
New Feature: Get connection credentials by file uri

MikhailProfile:
New feature:

Since Azure data studio is retired soon, I'm trying to recreate SearchEverywhere extension for VS Code (https://github.com/MikhailProfile/SearchEverywhere)

In scenario of using MSSQL Api it's useful to have opportunity to get current connection for opened document. 
Could we add this feature?

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -920,6 +920,19 @@ export default class ConnectionManager {
     }
 
     /**
+     * Retrieves Connection Info by file uri
+     * @param fileUri The URI of the file
+     * @returns Connection info
+     */
+    public getCredentialsByFileUri(fileUri: string): IConnectionInfo | undefined {
+        if (this.isConnected(fileUri)) {
+            return this._connections[fileUri].credentials;
+        }
+
+        return undefined;
+    }
+
+    /**
      * Retrieves the list of databases for the connection specified by the given URI.
      * @param connectionUri The URI of the connection to list the databases for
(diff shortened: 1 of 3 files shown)


Benjin (changes requested):
@croblesm, do you have any thoughts on whether this is something we want to add?  Could leave it for next month as well.

Marking this as "request changes" so we don't accidentally merge it.

src/controllers/connectionManager.ts
```diff
@@ -919,6 +919,19 @@ export default class ConnectionManager {
...
     }
 
+    /**
+     * Retrieves Connection Info by file uri
+     * @param fileUri The URI of the file
+     * @returns Connection info
+     */
+    public getCredentialsByFileUri(fileUri: string): IConnectionInfo | undefined {
+        if (this.isConnected(fileUri)) {
+            return this._connections[fileUri].credentials;
```
kburtram, line 929: if we're going to share connection credentials across extensions we probably need to get explicit consent from the user here, such as popping up a confirmation notification with "Yes", "No", and remember selection.  Similar to when an extension requests the copilot LLM the first time.

MikhailProfile: Child extension depends on SQL Server (mssql) extension. When user installs it, he is agree to share data between em. But I could be wrong here and we need to have one more "ok" from the user.
image

aasimkhan30:
Closing as discussed in https://github.com/microsoft/vscode-mssql/issues/19411.

PR #19343, opened 2025-05-05
Telemetry for MSSQL Chat Handler

lewis-sanchez:
This PR adds telemetry for the chat handler.

### DIFF
--- a/src/chat/chatAgentRequestHandler.ts
+++ b/src/chat/chatAgentRequestHandler.ts
@@ -111,7 +111,7 @@ export const createSqlAgentRequestHandler = (
             }
 
             const activity = startActivity(
-                TelemetryViews.SqlCopilot,
+                TelemetryViews.MssqlCopilot,
                 TelemetryActions.StartConversation,
                 conversationUri,
             );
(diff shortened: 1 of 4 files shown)


src/chat/chatAgentRequestHandler.ts
```diff
@@ -108,12 +110,20 @@ export const createSqlAgentRequestHandler = (
                 return { metadata: { command: "" } };
             }
 
+            const activity = startActivity(
```
caohai, line 113: The `startConversation` here is an STS call and represents some 'internal' conversation. For tracking GHCP conversation, we should add telemetry at the top of the handler.

src/chat/chatAgentRequestHandler.ts
```diff
@@ -51,6 +51,12 @@ export const createSqlAgentRequestHandler = (
             `Starting new chat conversation: conversion '${conversationUri}' with connection '${connectionUri}'`,
         );
 
+        const activity = startActivity(
+            TelemetryViews.MssqlCopilot,
+            TelemetryActions.StartConversation,
+            conversationUri,
```
caohai, line 57: We might want to remove this info.

caohai (approved):

src/chat/chatAgentRequestHandler.ts
```diff
@@ -49,6 +51,12 @@ export const createSqlAgentRequestHandler = (
...
```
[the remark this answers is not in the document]
lewis-sanchez: I'm thinking a GUID will work better. Every first chat is going to have a URI, like "conversation1" otherwise which would make it hard to analyze.

PR #19344, opened 2025-05-05
Removing (Preview) tag from features for May GA

Benjin:
Removes Preview tag from Connection Dialog, Table Designer, Object Explorer Filtering, Query Results, and Query Plan.
Removes `mssql.enableNewQueryResultFeature` config entry

Adds unregistered/hidden `mssql.useLegacyConnectionExperience` and `mssql.useLegacyQueryResultExperience` flags for instances where users need to selectively revert those features until they're removed entirely.

Connection:
image
image
image

Table Designer:
image
image
image

OE Filter:
image
image
image
image

Query Results:
image

Query Plan:
image
image
image
image
image

With Legacy Flags:
image
> None for legacy query results


aasimkhan30 (approved):

package.json
```diff
@@ -1103,6 +1095,20 @@
                     "description": "%mssql.enableRichExperiencesDoNotShowPrompt.description%",
                     "scope": "application"
                 },
+                "mssql.useLegacyConnectionExperience": {
```
Benjin, line 1098: I had to add these for the playwright tests to work.  Figuring out why was a bit of a journey:

* The playwright tests run VS Code with a temporary user profile.
* When VS Code runs with a different user profile, only settings that are scoped to the application are read/used.  Anything else is ignored.
* Settings not registered in `package.json` are defaulted to the user scope.
* Therefore, "hidden"/unregisted flags (like I was using for the legacy connection UI) aren't able to be used with the playwright setup.

My solution is to add them with the application scope, but mark them as deprecated so that they don't show up in the settings UI or JSON autocomplete.

package.nls.json
```diff
@@ -1,208 +1,210 @@
...
+    "mssql.filterNode": "Filter",
+    "mssql.clearFilters": "Clear Filters",
+    "mssql.openExecutionPlanFile": "Open Execution Plan File",
+    "mssql.enableExperimentalFeatures.description": "Enables experimental features in the MSSQL extension. The features are not production-ready and may have bugs or issues. Restart Visual Studio Code after changing this setting.",
+    "mssql.enableRichExperiences": "Enable Modern Features",
+    "mssql.enableRichExperiences.description": "Enables UI-based features in the MSSQL extension for richer and more powerful features. Restart Visual Studio Code after changing this setting.",
+    "mssql.enableRichExperiencesDoNotShowPrompt.description": "Do not show prompts to enable UI-based features",
+    "mssql.useLegacyConnectionExperience.description": "Use the legacy connection experience instead of the modern experience.",
+    "mssql.useLegacyQueryResultExperience.description": "Use the legacy query result experience instead of the modern experience.",
+    "mssql.legacySetting.deprecationMessage": "This setting will be removed in a future release.",
```
Benjin, line 179: These are the only changes here.

aasimkhan30 (approved):

PR #19345, opened 2025-05-05
Adding checks for missing critical properties (serverName and id)

Benjin:
Fixes https://github.com/microsoft/vscode-mssql/issues/19277


src/connectionconfig/connectionconfig.ts
```diff
@@ -166,25 +166,50 @@ export class ConnectionConfig implements IConnectionConfig {
         if (getWorkspaceConnections) {
             // Read from workspace settings
             let workspaceProfiles = this.getProfilesFromSettings(false);
+
+            // We don't currently auto-update connections in workspace/workspace folder config,
+            // so alert the user if any of those are missing their ID property that they need manual updating.
+            workspaceProfiles = workspaceProfiles.filter((profile) => {
+                if (!profile.id) {
+                    this._vscodeWrapper.showErrorMessage(
```
aasimkhan30, line 174: Do you think we should limit it to just one notification? It’s mostly a nit, but showing multiple could clutter the launch experience—especially if the user has several workspace connections without IDs.

As a stretch idea (not for this release), we could add a “Show Skipped Connections” button that opens a text doc listing the skipped connections along with their workspace URIs.

aasimkhan30 (approved):

src/connectionconfig/connectionconfig.ts
Benjin: Yeah, good thought.  I'll update to combine all of them into one notification.

kburtram (approved):

aasimkhan30 (approved):

PR #19357, opened 2025-05-06
Object Explorer Filter Tests

laurenastrid1:
## PR for OE Filter Tests

There were multiple changes to get this working:

- Moved some OE Filter code around to make it more testable — mainly pulling utility code out of the React component and into the utils file.

- Fixed the SQL Server container setup in the pipeline. Not sure if it was a setup issue, but the query execution and connection e2e tests couldn’t connect to the server. The pipeline now spins up a Docker container for it.

- Added a flag to disable the new connection dialog for testing — in the pipeline e2e tests, trying to connect using this dialog was causing hanging.

- Fixed the old query execution test — it wasn’t actually checking if the query ran successfully.

- Refactored the mergeCoverage script to support multiple test suites.

Screenshot 2025-05-05 201904
image


test/resources/launchDir/User/settings.json
```diff
@@ -1,5 +1,6 @@
 {
     "mssql.openQueryResultsInTabByDefaultDoNotShowPrompt": false,
-    "mssql.enableRichExperiences": false,
+    "mssql.enableRichExperiences": true,
+    "mssql.enableNewConnectionFeature": false,
```
aasimkhan30, line 4: We are not enabling new connection features?

Benjin: err, what _is_ `enableNewConnectionFeature`?

.gitignore
```diff
@@ -30,4 +30,15 @@ webviews-metafile.json
 # Ignore all l10n files except the default one (bundle.l10n.json)
 /localization/l10n/bundle.l10n.*.json
 .eslintcache
-.nyc_output
\ No newline at end of file
+.nyc_output
+/test/resources/launchDir/*
```
aasimkhan30, line 34: why not put these info in an 'artifact' dir and ignore the entire dir? Seems like a pain to keep updating.

test/unit/objectExplorerFilter.test.ts
```diff
@@ -0,0 +1,413 @@
...
+
+suite("ObjectExplorerFilterUtils", () => {
+    let sandbox: sinon.SinonSandbox;
+    let firstValueEmptyError: any;
+    let secondValueEmptyError: any;
+    let firstValueLessThanSecondError: any;
+
+    setup(() => {
+        sandbox = sinon.createSandbox();
+        ObjectExplorerFilterUtils.initializeStrings({
```
aasimkhan30, line 159: You can stub l10.t and just return return the passed string instead of doing this. Does that not work?

test/e2e/objectExplorerFilter.spec.ts
```diff
@@ -0,0 +1,273 @@
...
+        await new Promise((resolve) => setTimeout(resolve, 1 * 1000));
+        serverName = getServerName();
+        databaseName = getDatabaseName();
+        authType = getAuthenticationType();
+        userName = getUserName();
+        password = getPassword();
+        savePassword = getSavePassword();
+        profileName = getProfileName();
+
+        await addDatabaseConnection(
```
aasimkhan30, line 51: This is going to be a common scenario for smoke tests. Maybe we try extracting it out as a helper method.

test/e2e/objectExplorerFilter.spec.ts
```diff
@@ -0,0 +1,273 @@
...
+            databaseName,
+            authType,
+            userName,
+            password,
+            savePassword,
+            profileName,
+        );
+
+        // click filter button
+        await vsCodePage
```
aasimkhan30, line 63: How about we use settings to add connectionProfile to make it UI agnostic? @Benjin any better ideas?

test/e2e/utils/testHelpers.ts
```diff
@@ -62,6 +65,72 @@ export async function addDatabaseConnection(
...
+    databaseName: string,
+    authType: string,
+    userName: string,
+    password: string,
+    savePassword: string,
+    profileName: string,
+): Promise<void> {
+    if (authType !== "SQL Login") return;
+
+    console.log("Connection Profile: ", serverName, databaseName, authType, userName, profileName);
```
aasimkhan30, line 83: Maybe we use settings to add connection? That will make it UI agnostic. @Benjin any better ideas?

package.json
```diff
@@ -1040,6 +1040,12 @@
                     "description": "%mssql.openQueryResultsInTabByDefaultDoNotShowPrompt.description%",
                     "scope": "application"
                 },
+                "mssql.enableNewConnectionFeature": {
```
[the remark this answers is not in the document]
Benjin: I already have #19344 that sets up the post-GA behavior and flags, so let's not do anything with those in this PR until after that one goes it.  The flag "direction" in #19344 makes more sense for what and how things are gated.

Benjin (changes requested):

.github/workflows/build-and-test.yml
```diff
@@ -152,26 +152,35 @@ jobs:
...
-          echo "PASSWORD=$(LC_ALL=C tr -dc 'A-Za-z0-9!?%=' < /dev/urandom | head -c 16)$(LC_ALL=C tr -dc '@^#' < /dev/urandom | head -c 1)" >> $GITHUB_ENV
-          sudo cp build/mssql.init /etc/init.d/mssql-server
-          sudo chmod +x /etc/init.d/mssql-server
-          curl https://packages.microsoft.com/keys/microsoft.asc | sudo tee /etc/apt/trusted.gpg.d/microsoft.asc
-          sudo add-apt-repository "$(wget -qO- https://packages.microsoft.com/config/ubuntu/22.04/mssql-server-2022.list)"
-          sudo apt-get update
-          sudo apt-get install -y mssql-server
-          sudo ACCEPT_EULA='Y' MSSQL_PID='Developer' MSSQL_SA_PASSWORD=${{ env.PASSWORD }} /opt/mssql/bin/mssql-conf setup || true
-          sudo service mssql-server start
+          export SA_PASSWORD=$(LC_ALL=C tr -dc 'A-Za-z0-9!?%=' < /dev/urandom | head -c 16)$(LC_ALL=C tr -dc '@^#' < /dev/urandom | head -c 1)
```
aasimkhan30, line 155: Any reason to move this to docker?

.gitignore
laurenastrid1: I need to track /test/resources/launchDir/User/settings.json, but everything else unter the launchDir directory is ignored.

.github/workflows/build-and-test.yml
laurenastrid1: the tests weren't connecting properly to the server; i don't know if it was something to do with the setup or not, but i changed it to docker, and it started working.

aasimkhan30: I fixed it in this PR. Seems like they have added a new language options and we just needed to provide that. Can you please revert this back?  Thanks!
https://github.com/microsoft/vscode-mssql/pull/19384

Benjin (approved):

aasimkhan30:
@laurenastrid1 , do you know why the smoke test step is passing when we have tests that are failing?

https://github.com/microsoft/vscode-mssql/actions/runs/14981316773/job/42085878114?pr=19357#step:18:382

aasimkhan30 (changes requested):
Please check why the smoke test failure is not reported.

Benjin (approved):

Issue #18736, opened 2025-02-21
add savePassword to connect Uri


dzsquared:
Launching an existing connection profile from the URI `vscode://ms-mssql.mssql/connect` where the password was previously saved results in an odd experience - the connection dialog opens with the saved password in place but the "save password" checkbox is now false so if you hit connect without "rechecking" it you loose that saved password. 

This isn't exactly a bug (maybe?), but it means incoming connections to existing servers result in users losing the password.

to repro:
- create a valid uri for a SQL auth connection similar to ``
- launch the uri, enter the password and click save password before connecting
- disconnect from the server
- re-launch the uri, (the password is already present in the dialog this time) and don't click save password (because it's already saved, right?)
- connection fails asking for the password even though it was previously saved

https://github.com/microsoft/vscode-mssql/blob/d0c76b10f1610dfc4524440bc3bdf60fd4b4cada/src/mssqlProtocolHandler.ts#L80

the protocol handler uses IConnectionInfo and tacks on profileName, but savePassword is in IConnectionProfile

https://github.com/microsoft/vscode-mssql/blob/d0c76b10f1610dfc4524440bc3bdf60fd4b4cada/src/models/interfaces.ts#L59

The repeated profile connection is a pattern from the SQL containers Docker Desktop extension:

![Image](https://github.com/user-attachments/assets/3df8c701-84d9-4d64-9479-5dcaea04e6bc)

currently adding the mssql extension command handler to the container extension, but anticipating user complaints about losing the password.

PR #19364, opened 2025-05-07
Adding password handling to connect URI handler

Benjin:
Fixes https://github.com/microsoft/vscode-mssql/issues/18736

Though this yields some other questions about the behavior here: what should happen when a user launches the same URI multiple times?  The first time, it's new and get saves.  What does the second launch do?

Some options:
1. Launch connection dialog, and...
    1. Always make a new connection
    2. Assume I'm meaning to edit the first existing connection it finds
    3. Only assume it's editing the matching connection if there's only one that matches
    4. ...?
2. Connect directly to the existing connection (not launch the connection dialog)
    1. If there are multiple, prompt user for which?

I think I'm inclined toward 2 (with 2.1 for the multi-match scenario).

### DIFF
--- a/src/connectionconfig/connectionDialogWebviewController.ts
+++ b/src/connectionconfig/connectionDialogWebviewController.ts
@@ -867,12 +867,15 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
     ): Promise<IConnectionDialogProfile> {
         // Load the password if it's saved
         if (Utils.isEmpty(connection.connectionString)) {
-            const password =
-                await this._mainController.connectionManager.connectionStore.lookupPassword(
-                    connection,
-                    false /* isConnectionString */,
-                );
-            connection.password = password;
+            if (!connection.password) {
+                // look up password in credential store if one isn't already set
+                const password =
+                    await this._mainController.connectionManager.connectionStore.lookupPassword(
+                        connection,
+                        false /* isConnectionString */,
+                    );
+                connection.password = password;
+            }
         } else {
             this.logger.logDebug(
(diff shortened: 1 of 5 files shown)


aasimkhan30 (approved):

aasimkhan30:
Right now, it looks like we’re creating a new connection profile each time. We might need to revert to the old logic where we compare profiles based on their properties and, if matches are found, show a quick pick so users can choose from existing connections. If there’s only one match, we can skip the quick pick altogether.

I can imagine people using azure or fabric portal to open the extension and connect to a database often. It would be frustrating if we kept adding new connections to their settings every time.

Benjin:
@aasimkhan30 

> Right now, it looks like we’re creating a new connection profile each time. We might need to revert to the old logic where we compare profiles based on their properties and, if matches are found, show a quick pick so users can choose from existing connections. If there’s only one match, we can skip the quick pick altogether.

Because the new connection doesn't yet have an ID assigned, it falls back to the property comparison, matches on the properties, and replaces the existing connection (not adding a new one each time).  This isn't ideal because it will trample any properties in the one it's replacing that aren't part of the incoming new connection, but it's still an improvement on the password-saving behavior.

My proposed behavior is this:
* the `/connect` handler takes optional args, and looks for a matching connection (on properties because it doesn't yet have an ID):
    * No args supplied: open the connection dialog
    * No match: open the connection dialog
    * One match: connect immediately, skip the connection dialog
    * Multiple matches: show a quickpick with existing saved connections for selection
* the `/openConnectionDialog` handler takes no args, and opens the connection dialog

PR #19367, opened 2025-05-07
Fixing trust server certificate notification for OE and query editor. 

aasimkhan30:
This PR fixes two issues related to the Trust server certificate prompt:

1. Object Explorer:
When the user fixes a connection profile after the prompt, we weren’t updating the internal node profile, causing an infinite loop of Trust server certificate prompts.

2. Query Editor:
Opening a disconnected editor and connecting to a profile that requires Trust server certificate (but has it disabled) causes the connection flow to repeat once and ultimately fail. This then triggers an OE session with the same invalid profile, resulting in problem 1.

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -1125,7 +1125,7 @@ export default class ConnectionManager {
                 const wasCreated = await addFirewallRuleController.dialogResult;
 
                 if (wasCreated === true /** dialog closed is undefined */) {
-                    await this.connect(fileUri, connection.credentials);
+                    return await this.connect(fileUri, connection.credentials);
                 } else {
                     return false;
                 }
@@ -1135,7 +1135,7 @@ export default class ConnectionManager {
                     connectionCreds as IConnectionProfile,
                 );
                 if (updatedConnection) {
-                    await this.connect(fileUri, updatedConnection);
+                    return await this.connect(fileUri, updatedConnection);
                 } else {
                     return false;
                 }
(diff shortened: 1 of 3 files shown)


src/controllers/connectionManager.ts
```diff
@@ -1115,6 +1129,16 @@ export default class ConnectionManager {
                 } else {
                     return false;
                 }
+            } else if (connection.errorNumber === Constants.errorSSLCertificateValidationFailed) {
+                const updatedConnection = await this.handleSSLError(
+                    fileUri,
+                    connectionCreds as IConnectionProfile,
+                );
+                if (updatedConnection) {
+                    await this.connect(fileUri, updatedConnection);
```
Benjin, line 1138: `return await this.connect()`, right?  And likely the same on line 1128 above, I think.

src/controllers/mainController.ts
```diff
@@ -784,8 +785,17 @@ export default class MainController implements vscode.Disposable {
...
         connectionCredentials?: IConnectionInfo,
     ): Promise<TreeNodeInfo> {
-        const sessionCreationResult =
-            await this._objectExplorerProvider.createSession(connectionCredentials);
+        let retry = true;
+        // There can be many reasons for the session creation to fail, so we will retry until we get a successful result or the user cancels the operation.
+        let sessionCreationResult: CreateSessionResult = undefined;
+        while (retry) {
+            retry = false;
+            sessionCreationResult =
```
Benjin, line 793: Looks like this may return `undefind` as well.  Probably can't/shouldn't retry on failure for that.

Benjin (approved):

Issue #18324, opened 2024-10-28
[Rich Experiences] White Screen When Moving Query Results


tmonoceros:
- MSSQL Extension Version: 1.25.0
- VSCode Version: 1.94.2
- OS Version: Windows 10 10.0.19045 Build 19045

Steps to Reproduce:

1. Enable "rich experiences" in settings.json `  "mssql.enableRichExperiences": true`
2. Execute any query

My results show up in the default "Query results" pane as expected (actual data removed from the screenshot)

![Image](https://github.com/user-attachments/assets/87c09ae4-480a-4efc-921d-4157431b60fc)

However, if I try to move the query results window to another pane, it goes completely white. Nothing I can do - not rerunning the query, not moving it back - seems to fix this. 

![Image](https://github.com/user-attachments/assets/b8d0241a-0aec-459d-a7fd-f0bba05e4548)

caohai:
@tmonoceros Thanks for creating the issue! 
> move the query results window to another pane

Are you doing this via these options?
![Image](https://github.com/user-attachments/assets/ff672b73-b118-42de-a363-0838e571ca15)
![Image](https://github.com/user-attachments/assets/0b79a520-99a9-4581-b5cf-91c3ed283c51)

Here's what it looks like if I move the panel area to the right (tested on Mac and Windows)
![Image](https://github.com/user-attachments/assets/f224bd5f-16c8-4d35-a325-32143e14a759)

tmonoceros:
I was clicking and dragging to move the window. Here's a .gif showing this in action

![Image](https://github.com/user-attachments/assets/7a0b70b5-965f-4273-8727-0ed7f9160c0e)

tmonoceros:
I don't have a strong opinion on the exact fix right now. The fact that you're allowed to pop the query results out, but doing so breaks them and forces you to disable and reenable rich experiences to fix it, is a pretty clear bug that can actually be pretty disruptive. But it might just be muscle memory that causes me to prefer popping the query results out - that's how it works without Rich Experiences enabled, so that's what I'm used to. 

I wouldn't complain (yet, anyway) if the fix was to just prevent the query results from being popped out. But if there's no reason to disallow popping the query results out (I don't know enough about the difference between editor panels and the rest panel to have an opinion there), seems like a straightforward win to me.

caohai:
[...]

PR #19369, opened 2025-05-07
Remove dispose

cssuh:
Removing this dispose allows users to drag the query results view into different sidebars.

webviewMove

Fixes https://github.com/microsoft/vscode-mssql/issues/18324

### DIFF
--- a/src/controllers/reactWebviewViewController.ts
+++ b/src/controllers/reactWebviewViewController.ts
@@ -76,7 +76,6 @@ export class ReactWebviewViewController<State, Reducers>
 
             localResourceRoots: [this._context.extensionUri],
         };
-        this._webviewView.onDidDispose(() => {});
 
         this._webviewView.webview.html = this._getHtmlTemplate();
         this.registerDisposable(
--- a/src/controllers/reactWebviewViewController.ts
+++ b/src/controllers/reactWebviewViewController.ts
@@ -76,9 +76,7 @@ export class ReactWebviewViewController<State, Reducers>
 
             localResourceRoots: [this._context.extensionUri],
         };
-        this._webviewView.onDidDispose(() => {
-            this.dispose();
-        });
+        this._webviewView.onDidDispose(() => {});
 
         this._webviewView.webview.html = this._getHtmlTemplate();
         this.registerDisposable(


src/controllers/reactWebviewViewController.ts
```diff
@@ -76,9 +76,7 @@ export class ReactWebviewViewController<State, Reducers>
 
             localResourceRoots: [this._context.extensionUri],
         };
-        this._webviewView.onDidDispose(() => {
-            this.dispose();
-        });
+        this._webviewView.onDidDispose(() => {});
```
aasimkhan30, line 79: Let's remove this listener since it's a noop

Benjin:
Can you add a video to the PR of you dragging query results to a new pane and back?

aasimkhan30 (commented):
Does it now retain context when dragged? I'm guessing your other fix with initial state fixes that

src/controllers/reactWebviewViewController.ts
```diff
@@ -76,9 +76,7 @@ export class ReactWebviewViewController<State, Reducers>
 
             localResourceRoots: [this._context.extensionUri],
         };
-        this._webviewView.onDidDispose(() => {
-            this.dispose();
```
caohai, line 80: Is `onDidDispose` also fired during webviewview drag?
image

cssuh: Yes, it gets fired during a drag event

cssuh:
> Does it now retain context when dragged? I'm guessing your other fix with initial state fixes that

Yes it keeps the context

src/controllers/reactWebviewViewController.ts
caohai: Wow that's unexpected but I guess that's just how vscode implemented this, the fix LGTM and I think not being able to clean up resource for the use case I underlined in the screenshot is acceptable.

caohai:
@cssuh Does this fix work if you don't have Chat window opened?

cssuh:
> @cssuh Does this fix work if you don't have Chat window opened?

Without the chat window opened, there is no anchor sidebar for me to drag it to.  But I can also drag it to the left sidebar where we have the connections list and it displays correctly

caohai:
> > @cssuh Does this fix work if you don't have Chat window opened?
> 
> Without the chat window opened, there is no anchor sidebar for me to drag it to. But I can also drag it to the left sidebar where we have the connections list and it displays correctly

@cssuh Can you take a look at this reply? https://github.com/microsoft/vscode-mssql/issues/18324#issuecomment-2444475350 The user appeared to be able to drag and drop without an anchor.

cssuh:
> > > @cssuh Does this fix work if you don't have Chat window opened?
> > 
> > 
> > Without the chat window opened, there is no anchor sidebar for me to drag it to. But I can also drag it to the left sidebar where we have the connections list and it displays correctly
> 
> @cssuh Can you take a look at this reply? [#18324 (comment)](https://github.com/microsoft/vscode-mssql/issues/18324#issuecomment-2444475350) The user appeared to be able to drag and drop without an anchor.

Yeah that is the same sidebar where my chat window is, it pops up if you drag it there.  I removed my chat window and it works still

caohai (approved):

lewis-sanchez (approved):

PR #19375, opened 2025-05-08
Fixing husky and loc files string ordering.

aasimkhan30:
Reverting husky back to old version and fixing loc string ordering which is causing loc pipeline failure.


caohai (approved):

Benjin:
Merging to get pipeline unblocked.

Benjin:
@aasimkhan30 how'd you ID this as the issue?

PR #19387, opened 2025-05-08
Add inMemoryDataProcessingThreshold for filtering/sorting

cssuh:
Adds a threshold to disable filtering/sorting for large data sets.  The default is set at 5000 rows.  If a user tries to filter/sort for data sets larger than this amount, we will display this toast:
Screenshot 2025-05-08 at 4 09 36 PM

Also cleaned up some code that was no longer needed.

### DIFF
--- a/src/reactviews/pages/QueryResult/resultGrid.tsx
+++ b/src/reactviews/pages/QueryResult/resultGrid.tsx
@@ -38,8 +38,6 @@ declare global {
     }
 }
 
-const DEFAULT_IN_MEMORY_DATA_PROCESSING_THRESHOLD = 5000;
-
 export interface ResultGridProps {
     loadFunc: (offset: number, count: number) => Thenable<any[]>;
     resultSetSummary?: ResultSetSummary;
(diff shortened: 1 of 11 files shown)


src/reactviews/pages/QueryResult/resultGrid.tsx
```diff
@@ -125,6 +127,15 @@ const ResultGrid = forwardRef<ResultGridHandle, ResultGridProps>((props: ResultG
         const setupState = async () => {
             await table.setupFilterState();
             await table.restoreColumnWidths();
+            if (context.state.inMemoryDataProcessingThreshold) {
```
Benjin, line 130: This should handle the case where a user sets it to a negative number or to 0.  Perhaps both of those are treated as unlimited?

cssuh: I added a minimum value of 1 to the option value, I don't think it makes sense to allow 0 or negative numbers

Benjin:
Looks good to me, but I think it'd be good to run this change by @croblesm since it's a behavioral change rather than a bugfix.

Benjin (approved):

Benjin:
Also, if I manually set this to -1 (or another invalid value), does the config just expose the default 5000?

lewis-sanchez (approved):

cssuh:
@Benjin do you mean manually set this to -1 using the JSON? Then it will just take that value.  But if users try to change the setting using the settings UI, they will see this:
Screenshot 2025-05-09 at 3 14 02 PM

kburtram (approved):

croblesm:
> Looks good to me, but I think it'd be good to run this change by @kburtram and @croblesm since it's a behavioral change rather than a bugfix.

I'm good with this change. In fact, it makes sense to add to setting to ensure a good user experience (performance-wise).

PR #19394, opened 2025-05-09
Fetching password for connection profile when opening  "New Query" editor from disconnected OE node.

aasimkhan30:
Fetching passwords for connections if the profile doesn't contain it.


caohai:
I verified the fix on my saved connections and can confirm it fixes the connection err when I click "New Query" on a disconnected server/db.

caohai (approved):

Issue #18556, opened 2025-01-15
New Query should use existing connection


match-gabeflores:
- VSCode Version: 1.27

Steps to Reproduce:

1. If I have an existing query window open that is connected to a server
2. create a New Query from the command palette
3. it should open a new window that is connected to the existing server as the previous query window. This is the behavior on sql server management studio. 

current behavior - it opens a new window and then prompts for the server name (an extra tedious step , esp if there are dozens of servers)

![Image](https://github.com/user-attachments/assets/d37847a8-9938-4559-aa37-104f9a2dee22)

jsteinberg613:
Please add this! I know that you have hear all these complaints before, but when you force users to switch tools, I really wish that someone would take note of these little things that really add up to serious annoyance. Much of my work is ad-hoc sql support, and I open dozens of new queries a day, having to go through extra steps to get a query connected, that were not required in the past, is a serious bummer

djones-breuer:
Note: this was also a feature in ADS. This is a critically important feature now that you have deprecated ADS

PR #19395, opened 2025-05-10
Adding automatic connnection selection for additional New Query scenarios

Benjin:
Fixes https://github.com/microsoft/vscode-mssql/issues/18556

Adds automatic connection selection for two more scenarios.  In priority order:
1. User right-clicked on an OE node and selected "New Query": use that node's connection profile
2. **NEW:** User triggered "New Query" from command palette and the active document has a connection: copy that to the new document
3. **NEW:** User triggered "New Query" from command palette while they have a connected OE node selected: use that node's connection profile
4. User triggered "New Query" from command palette and there's no reasonable context: prompt for connection to use

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -1866,6 +1866,18 @@ export default class MainController implements vscode.Disposable {
             // from the object explorer context menu
             const editor = await this._untitledSqlDocumentService.newQuery(content);
             const uri = editor.document.uri.toString(true);
+
+            // if the command came from the command palette, add connection from the selected OE node
+            if (!node) {
+                // grab the selected node from the object explorer, if any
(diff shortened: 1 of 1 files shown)


src/controllers/mainController.ts
```diff
@@ -1858,72 +1858,128 @@ export default class MainController implements vscode.Disposable {
         await vscode.env.openExternal(vscode.Uri.parse(Constants.gettingStartedGuideLink));
     }
 
+    private async newQueryFromProfile(
```
Benjin, line 1861: `newQueryFromProfiles()` and `newQueryFromPrompt()` are mostly just extractions of the existing logic to helper functions for readability.

The only change should be where the `createObjectExplorerSession` flag is used.

kburtram (approved):

Issue #19363, opened 2025-05-06
[GitHub Copilot - Bug]: Consider logging channel for copilot messages


kburtram:
### Description

Consider logging channel for copilot messages per feedback on https://github.com/microsoft/vscode-mssql/pull/19283

### Steps to Reproduce

Code review feedback

### Affected Area

- [ ] Chat and inline suggestions
- [ ] Code generation
- [ ] Schema Explorer & Designer
- [ ] Smart Query Builder
- [ ] Query Optimizer Assistant
- [ ] Business Logic Explainer
- [ ] Security Analyzer
- [ ] Localization & Formatting Helper
- [ ] Testing and Mocking Data Generator
- [ ] Other (please describe below)

### Confirmation

- [x] I have searched existing issues and couldn’t find a match
- [ ] I want to work on this issue

PR #19398, opened 2025-05-10
Add logging channel and more logging to chat agent handler

lewis-sanchez:
This PR fixes https://github.com/microsoft/vscode-mssql/issues/19363

### DIFF
--- a/src/chat/chatAgentRequestHandler.ts
+++ b/src/chat/chatAgentRequestHandler.ts
@@ -63,7 +63,7 @@ export const createSqlAgentRequestHandler = (
         const logger = getLogger();
         let conversationUri = getNextConversationUri();
         let connectionUri = vscodeWrapper.activeTextEditorUri;
-
+        logger.info("In handler");
         logger.logDebug(
             `Starting new chat conversation: conversion '${conversationUri}' with connection '${connectionUri}'`,
         );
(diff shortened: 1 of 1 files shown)


src/chat/chatAgentRequestHandler.ts
```diff
@@ -46,17 +47,24 @@ export const createSqlAgentRequestHandler = (
...
         stream: vscode.ChatResponseStream,
         token: vscode.CancellationToken,
     ): Promise<ISqlChatResult> => {
         const correlationId = Utils.generateGuid();
+        const logger = getLogger();
         let conversationUri = getNextConversationUri();
         let connectionUri = vscodeWrapper.activeTextEditorUri;
-
-        Utils.logDebug(
+        logger.info("In handler");
```
Benjin, line 66: This "entered/exited XYZ function" can probably be verbose- instead of info-level logging.  Same for the other spots

Benjin (approved):

src/chat/chatAgentRequestHandler.ts
```diff
@@ -46,17 +47,24 @@ export const createSqlAgentRequestHandler = (
         return () => `conversationUri${idCounter++}`;
     })();
 
+    const getLogger = (() => {
```
kburtram, line 50: Is it best practice to create a different logging channel per feature area?  This seems like total over kill.  I'd expect the extension to have 1 or 2, or maybe 3.  But it seems like we're creating these for more smaller feature areas in recent PRs.  Is there some guidance from the vscode team this is a good pattern to follow?  I prefer to have my logs consolidated generally to be able to spot errors that cross subsystem boundaries (up to a point, or course)

kburtram (commented):
Not sure if we need this?  Verses logging to the primary extension output channel.

src/chat/chatAgentRequestHandler.ts
lewis-sanchez: @Benjin, do you have additional info on why it was suggested to use a MssqlCopilot logger in addition to copilot specific messages being easier to filter?

kburtram:
My take on this PR is that we should probably send most the logging to a single channel, perhaps for very specific logging we might want to isolated it, but I don't think we should create a different logging channel for each feature area.  Unless VS Code is recommending extensions do that.

src/chat/chatAgentRequestHandler.ts
Benjin: This call reuses the same "MSSQL" output channel as the rest of the extension; the second parameter is a string tag that gets included in each message so that you can optionally filter your messages further by searching for `[tag]` in the MSSQL channel:

image

Should be easier to both view the full picture and isolate specific areas in the logs.

Benjin:
@kburtram 

Unless I'm misunderstanding, I think we're already doing that; everything's going to the `MSSQL` channel
image
image

kburtram (approved):

src/chat/chatAgentRequestHandler.ts
kburtram: Thanks!  Prefix label on message in the main MSSQL channel makes sense to me.

Benjin (approved):

Issue #18837, opened 2025-02-24
Cannot cancel long running Query.


nycgavin:
- MSSQL Extension Version: 1.28.0
- VSCode Version: 1.97.2
- OS Version: Mac sonoma 14.5
- M1 chip
- MS Sql server 2019

Steps to Reproduce:

1. run this query on a new tab: Select * from ... join  [Linked Server] --this query runs for a long time
2. the query starts running, but the Query History says "No Queries Available" and the cancel button is hidden
3. The bottom of VS code says Executing Query. (P.S. if you click on the disconnect button at this point, the bottom says canceling query, but it will continue spinning for the next x minutes)
4. if you open another query tab and try to execute a different query, it opens a prompt and lets you choose the connection, but after you choose it, it just says testing profile on the bottom right corner but doesn't do anything.
5. When the first query finished executing (after 5 minutes, another error shows up that says: mssql: A query is already running for this editor session. Please cancel this query or wait for its completion.)

nycgavin:
This extension is full of bugs, and I don't understand why Microsoft wants to retire Azure Data Studio, this is so bad.

match-gabeflores:
I've also started to encounter this occasionally. I have to restart VSCode as a solution

match-gabeflores:
hi @cssuh 

I don't know exactly when this occurs - it seems sporadic. I had a large result set and I'm not sure if it was still populating when i think i cancelled it (not sure here). the query successfully cancelled but it says "executing query" erroneously. 

1. What happens is the status bar will say "`Executing query`" forever and there's no Cancel button visible. 
![Image](https://github.com/user-attachments/assets/5bcde98c-d15d-4dee-b4ef-c94525c05a2e)

2. if i do the keyboard shortcut for cancel query, status bar says "`cancelling query`" forever.


![Image](https://github.com/user-attachments/assets/d2c5cd99-23da-4264-b526-ac2e5f237a6e)

When I actually look at what's going on in the server, the query isn't actually running. 

so it's left in a state where it says it's executing query or cancelling query forever and I can't run a new query. Worse, if open a new query window, it doesn't even connect to the server (just says `connecting `forever). 

On latest version of mssql on Cursor IDE (but don't think this is a cursor issue)

match-gabeflores:
Related bug from 2022. they also mention restarting is the main solution. They're on VPN. I'm also always on VPN. Not sure if that helps. 
https://github.com/microsoft/vscode-mssql/issues/17433

[...]

PR #19407, opened 2025-05-12
Prevent query runner deletion after editor close & reopen

cssuh:
Currently, when we close a query editor, the associated query runner gets flagged for deletion, with a timer of 30 minutes.  This flag does not get reset if we re-open the same query editor and run a query, which can lead to deletion of the query runner in the middle of a long-running query (due to the 30 minute timer).  

I have added a conditional which will reset the `flaggedForDeletion` property if we are re-using an existing query runner.

Fixes https://github.com/microsoft/vscode-mssql/issues/18837

### DIFF
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -45,6 +45,7 @@ import * as os from "os";
 import { Deferred } from "../protocol";
 import { sendActionEvent } from "../telemetry/telemetry";
 import { TelemetryActions, TelemetryViews } from "../sharedInterfaces/telemetry";
+import { Logger } from "../models/logger";
 
 export interface IResultSet {
     columns: string[];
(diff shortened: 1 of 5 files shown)


src/controllers/queryRunner.ts
```diff
@@ -91,6 +93,13 @@ export default class QueryRunner {
             this._vscodeWrapper = new VscodeWrapper();
         }
 
+        // Setup Logger
+
+        let channel = this._vscodeWrapper.createOutputChannel(
```
kburtram, line 98: do we need a separate logging channel?  I'm wonder if it's best practice to create new output channels per feature area?

Benjin: We should be using vscodeWrapper.outputChannel, which will reuse the existing MSSQL one.  Also, ideally adding a logging prefix to make logs easier to filter through: `Logger.create(channel, "Querys")` or maybe "QueryRunner"

aasimkhan30:
closing as I am handling this issue in a different way.

PR #19461, opened 2025-05-20
Adding code lens suggestion for sql connection. 

aasimkhan30:
## Description

This adds codelense items for sql files that mentions the active connection and the active database. Clicking on active connection let you change the server connection and the database lets you change active database for the connection. 
image

On a disconnected sql file, the codelense item lets user connect to a server. 

image

 The reason for adding this is to make the connected server more visible to the user. 

https://github.com/user-attachments/assets/8e36527d-0ee5-494f-8641-9740a4d17908

Settings toggle for this feature:
image
 

This also fixes unnecessary string trim we do for quick pick items. 
Before:
image
Now:
image

We are still trimming the string in status view
image

## Code Changes Checklist

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -67,6 +67,7 @@ import { ConnectionNode } from "../objectExplorer/nodes/connectionNode";
 import { CopilotService } from "../services/copilotService";
 import * as Prompts from "../chat/prompts";
 import { CreateSessionResult } from "../objectExplorer/objectExplorerService";
+import { SqlCodeLensProvider } from "../queryResult/sqlCodeLensProvider";
 
 /**
  * The main controller class that initializes the extension
@@ -276,6 +277,14 @@ export default class MainController implements vscode.Disposable {
             this._event.on(Constants.cmdDisableActualPlan, () => {
                 this.onToggleActualPlan(false);
             });
+
+            this._context.subscriptions.push(
+                vscode.languages.registerCodeLensProvider(
+                    { language: "sql" },
+                    new SqlCodeLensProvider(this._connectionMgr),
+                ),
+            );
+
             this.initializeObjectExplorer();
 
             this.registerCommandWithArgs(Constants.cmdConnectObjectExplorerProfile);
--- a/src/queryResult/sqlCodeLensProvider.ts
+++ b/src/queryResult/sqlCodeLensProvider.ts
@@ -7,7 +7,7 @@ import * as vscode from "vscode";
 import * as Constants from "../constants/constants";
 import ConnectionManager from "../controllers/connectionManager";
 import { QueryEditor } from "../constants/locConstants";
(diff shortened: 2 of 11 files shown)


Benjin (approved):

croblesm:
This is fantastic @aasimkhan30! I was about to work on a series of new UX features. This one was on my list. 
To me, this works great, but I would also like to have @jdesai-24 opinion from the UI/UX side.

src/controllers/mainController.ts
```diff
@@ -276,6 +277,14 @@ export default class MainController implements vscode.Disposable {
             this._event.on(Constants.cmdDisableActualPlan, () => {
                 this.onToggleActualPlan(false);
             });
+
+            this._context.subscriptions.push(
+                vscode.languages.registerCodeLensProvider(
+                    { language: "sql" },
```
caohai, line 283: Does this mean CodeLens will also show up for non-TSQL queries as long as the editor is set to `sql`? This might get annoying if a user is working on non-TSQL queries with MSSQL extension installed.
IIRC we added the ability to disable intellisense for non-MSSQL editor (by detecting error #), not sure if something similar can be done with this.
https://github.com/microsoft/vscode-mssql/pull/18557

aasimkhan30: Yeah, for now. However, we are having the same problem for the toolbar and status view where we show mssql options for non-tsql files. 

We probably need to improve that option to turn off all those things.

caohai: makes sense, we need to "smartly" enable MSSQL features when needed. IMO this specific change in the editor can be more invasive than for example in the toolbar but I'm fine with having a separate issue to track all these as a whole.

caohai:
I see in the video you have "Connect to SQL Server" and in code you have "Connect to MSSQL", I assume it's just outdated video, right? The latter makes more sense to me.

kburtram (approved):
I wonder if we want an option to disable this, like we do with the other language service options.  Or you think almost everyone will want this on?

src/queryResult/sqlCodeLensProvider.ts
```diff
@@ -0,0 +1,52 @@
...
+            .get<boolean>(Constants.configShowActiveConnectionAsCodeLensSuggestion);
+        if (!shouldShowActiveConnection) {
+            return [];
+        }
+        const connection = this._connectionManager.getConnectionInfo(document.uri.toString());
+        if (connection) {
+            return [
+                new vscode.CodeLens(new vscode.Range(0, 0, 0, 0), {
+                    title: getConnectionServerName(connection.credentials),
+                    command: Constants.cmdConnect,
```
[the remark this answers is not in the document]
aasimkhan30: One is to change the connection and the other one is to change the database. Copying the behavior from PGSQL extension.

caohai (approved):

PR #19502, opened 2025-05-28
Local Containers

laurenastrid1:
# Pull Request Tmplate – vscode-mssql

### Scenarios

- User tries to make a new Docker container, and Docker is not installed. The extension prompts the user to install Docker.
dockerNotInstalledError

- User tries to make a new Docker container, and Docker is not started. The extension then attempts to start Docker.
startDockerForUser

- User tries to make a new Docker container, but something is wrong with the engine (e.g., Windows containers, no user privilege to run Docker engine commands on Linux, Rosetta not enabled on Mac). On Windows platforms, the extension will prompt the user to switch to linux containers; if they accept the prompt, then the extension will switch to linux containers for them. On Linux and Mac, the extension will notify the user what the problem with the engine is.
promptUserForLinuxEngineSwitch

- User makes a new Docker container without a custom container name or port. The extension will use the default name of the container, which will be sql_server_container_{number}, and use the default port for container, which will be the next available port beginning at 1433
createDefaultContainer

- User makes a new Docker container with a custom container name and port. If the user tries to use a port/container name already in use, they will be prompted for a unique input.
createCustomContainer

- User adds a Docker container through the connection dialog, and the extension detects that it's a Docker container.
autoDetectDockerContainer

- User starts Docker container through the extension.
startContainer

- User stops Docker container through the extension.
stopContainer

- User deletes Docker container through the extension.
deleteContainre

Test Coverage: 
image

## Code Changes Checklist


src/constants/locConstants.ts
```diff
@@ -475,6 +476,7 @@ export let autoDisableNonTSqlLanguageServicePrompt = l10n.t(
     "Non-SQL Server SQL file detected. Disable IntelliSense for such files?",
 );
 export let msgAddConnection = l10n.t("Add Connection");
+export let msgAddLocalContainerConnection = l10n.t("Add Local Container Connection");
```
Benjin, line 479: Move this with the rest of the ContainerDeployment strings

src/constants/locConstants.ts
```diff
@@ -732,6 +734,142 @@ export class QueryResult {
...
+            message: "Failed to delete {0}.",
+            args: [name],
+            comment: ["Failed to delete {0}."],
+        });
+    public static selectImage = l10n.t("Select image");
+    public static selectImageTooltip = l10n.t("Select the SQL Server Container Image");
+    public static sqlServer2025Image = l10n.t("SQL Server 2025 - latest");
+    public static sqlServer2022Image = l10n.t("SQL Server 2022 - latest");
+    public static sqlServer2019Image = l10n.t("SQL Server 2019 - latest");
+    public static sqlServer2017Image = l10n.t("SQL Server 2017 - latest");
```
Benjin, line 786: Can these be pulled dynamically from the image registry?

src/reactviews/pages/ContainerDeployment/icons/EasyManagement_dark.svg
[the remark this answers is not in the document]
laurenastrid1: Janki sent them to me. I think they're from the postgres team?

src/controllers/mainController.ts
```diff
@@ -1205,6 +1221,109 @@ export default class MainController implements vscode.Disposable {
...
+        );
+        // Delete container command
+        this._context.subscriptions.push(
+            vscode.commands.registerCommand(
+                Constants.cmdDeleteContainer,
+                async (node: TreeNodeInfo) => {
+                    if (!node) return;
+
+                    const confirmation = await vscode.window.showInformationMessage(
+                        LocalizedConstants.Common.areYouSureYouWantTo("delete the container"),
```
aasimkhan30, line 1300: This is still not localized. Ideally this should be a separate string entirely as the translators won't translate "delete the container" part.

croblesm:
Hi @laurenastrid1, this feature looks great!
I'm just adding have some comments related to text, content, etc.

****************

I updated the "welcome page", including new text and updated links:
[Container](https://www.figma.com/design/NLsa0kpc5tevFiJLG5iLMr/Local-SQL-container?node-id=0-1)

****************

> User tries to make a new Docker container, and Docker is not installed. The extension prompts the user to install Docker.

In this scenario, the user is blocked because Docker is not installed. So it would be great adding the linkfor them to download/install Docker. This is the link: https://docs.docker.com/engine/install
image

****************
> User makes a new Docker container without a custom container name or port. The extension will use the default name of the container, which will be sql_server_container_{number}, and use the default port for container, which will be the next available port beginning at 1433

In this scenario, the EULA link needs to be updated to the latest EULA (SQL 2025), this is the link:
https://go.microsoft.com/fwlink/?LinkId=746388

image

****************

> User adds a Docker container through the connection dialog, and the extension detects that it's a Docker container.

In this scenario, I would suggest adding more clarification for the user that the container is starting, so sharing live feedback in OE will make it better. 

You might want to transition from `Starting container`, to `Connecting to container` instead of just showing the text as `Loading`.
image

****************

> User stops Docker container through the extension.

In this scenario, I would suggest adding more clarification for the user that the container is stopping, so sharing live feedback in OE will make it better. 

You might want to show `Stopping container` instead of just showing the text as `Loading`.
image

> User deletes Docker container through the extension.

The current delete experience is confusing because the user is prompted twice—once to delete the container, and again to delete the connection. This makes the flow feel fragmented and unclear.

image
image

I would suggest combining both confirmations into a single dialog, clearly explaining that both the container and its connection will be removed.

Here is an example of the text we can use for the Toast notification / modal dialog:
*Are you sure you want to delete this container?
This will remove both the container and its connection from VS Code.*

image

src/reactviews/common/forms/form.component.tsx
```diff
@@ -152,15 +152,27 @@ export const FormField = <
...
-                          }
-                        : component.label
+                    // The html here shouldn't need to be sanitized, and should be safe
+                    // because it's only ever set by forms internal to the extension
+                    component.tooltip ? (
+                        {
+                            children: (_: unknown, slotProps: LabelProps) => (
+                                <InfoLabel {...slotProps} info={component.tooltip}>
+                                    <span
+                                        dangerouslySetInnerHTML={{
```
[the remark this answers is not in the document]
laurenastrid1: I needed a form item to be a link: 
image

croblesm:
Hi @laurenastrid1,

I created a simplified version of the *welcome page*:
<a href="https://www.figma.com/design/NLsa0kpc5tevFiJLG5iLMr/Local-SQL-container?node-id=96-1861" target="_blank" rel="noopener noreferrer">
  image
</a>

We can sync up with @jdesai-24 for the icons :)

src/reactviews/pages/ContainerDeployment/stepCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+    const [showFullErrorText, setShowFullErrorText] = useState(false);
+
+    // If this passes, container deployment state is guaranteed
+    // to be defined, so we can reference it as non-null
+    if (!state) {
+        return undefined;
+    }
+
+    const getStatusIcon = () => {
+        if (step.loadState === ApiStatus.Loaded) {
```
aasimkhan30, line 87: How about adding a 'Not Started' State. The UX gets a little confusing with all the steps in loading state at once.

croblesm:
We discussed this change on the call yesterday. Let's simplify the entry point text to "Create local SQL container":
image

caohai:
A few questions on the prerequisite checks page, 
1. looks like we are checking all 3 prerequisite of them concurrently. Should we do them sequentially / pretend to do them sequentially and fail at the first error?
2. Can we auto-expand the first error during the checks?

caohai:
> The extension will try to fix the problem, or prompt user to fix the problem

Can you list all the problems we try to fix/prompt in the PR description?

> for example, on windows, it will change the engine to use linux containers if it doesn't already.

looks like the PR description is outdated, I remember you mentioned this behavior has been updated to only prompt the user.

laurenastrid1:
> A few questions on the prerequisite checks page,
> 
> 1. looks like we are checking all 3 prerequisite of them concurrently. Should we do them sequentially / pretend to do them sequentially and fail at the first error?
> 2. Can we auto-expand the first error during the checks?

1. It is checking sequentially! It checks which step is the next step that isn't started yet, and runs that. 
2. Just implemented! Check out the first gif in the pr :)

media/newContainer_dark.svg
aasimkhan30, line 1: NIT: These icons look more like the server icon. Maybe we can use better icons.  Something like this with +sign. 
eeab (1)

src/containerDeployment/dockerUtils.ts
```diff
@@ -0,0 +1,688 @@
...
+import * as path from "path";
+import { FormItemOptions, FormItemValidationState } from "../sharedInterfaces/form";
+import { getErrorMessage } from "../utils/utils";
+import { Logger } from "../models/logger";
+
+/**
+ * The maximum length of the error text to display in the UI.
+ * If the error text exceeds this length, it will be truncated.
+ */
+const MAX_ERROR_TEXT_LENGTH = 300;
```
aasimkhan30: Where is the number coming from? You should try to use CSS styling to truncate text with ellipsis. 

```
text-overflow: ellipsis;
```

src/containerDeployment/dockerUtils.ts
```diff
@@ -0,0 +1,688 @@
...
+    if (result.success) {
+        steps[currentStep].loadState = ApiStatus.Loaded;
+    } else {
+        steps[currentStep].loadState = ApiStatus.Error;
+        steps[currentStep].errorMessage = result.error;
+        steps[currentStep].fullErrorText = truncateAndSanitizeErrorTextIfNeeded(
+            result.fullErrorText,
+        );
+        for (let i = currentStep + 1; i < steps.length; i++) {
+            steps[i].loadState = ApiStatus.Error;
```
aasimkhan30: Why do we set the future steps to error. Shouldn't we just keep them NOT started?

src/containerDeployment/dockerUtils.ts
```diff
@@ -0,0 +1,688 @@
...
+    }
+    return steps;
+}
+
+/**
+ * Truncates the error text if it exceeds the maximum length and sanitizes sensitive info.
+ */
+export function truncateAndSanitizeErrorTextIfNeeded(errorText: string): string {
+    // Sanitize
+    const sanitized = errorText.replace(/(SA_PASSWORD=)([^ \n]+)/gi, '$1******"');
```
aasimkhan30: Since you are displaying this in react. Stick to css instead of truncation. It is not accessible this way.

src/containerDeployment/dockerUtils.ts
```diff
@@ -0,0 +1,688 @@
...
+
+    // Truncate if necessary
+    if (sanitized.length > MAX_ERROR_TEXT_LENGTH) {
+        return `${sanitized.substring(0, MAX_ERROR_TEXT_LENGTH)}...`;
+    }
+
+    return sanitized;
+}
+
+/**
```
aasimkhan30, line 147: NIT: add link to the password policy.

aasimkhan30: https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-docker-container-configure?view=sql-server-ver17&pivots=cs1-bash#mount-a-host-directory-as-data-volume

src/containerDeployment/dockerUtils.ts
```diff
@@ -0,0 +1,688 @@
...
+ * Checks if the SQL Server password meets the complexity requirements.
+ * If the password is valid, it returns the validation message, which is an empty string.
+ * If the password is invalid, it returns an error message.
+ */
+export function validateSqlServerPassword(password: string): string {
+    if (password.length < 8) {
```
aasimkhan30: NIT: Also check the upper limit for password. It is 128

src/containerDeployment/dockerUtils.ts
```diff
@@ -0,0 +1,688 @@
...
+
+    return "";
+}
+
+/**
+ * Checks if the provided connection name is valid and not a duplicate connection name within
+ * mssql.connections
+ */
+export function validateConnectionName(connectionName: string): boolean {
+    const connections = vscode.workspace
```
aasimkhan30: I think we don't enforce this rule in connections. Don't know if we should do that here or it keep it consistent. @Benjin.

aasimkhan30: Especially with connection groups. I can imagine people having multiple connections with same name.

src/containerDeployment/dockerUtils.ts
```diff
@@ -0,0 +1,688 @@
...
+        .getConfiguration(extensionName)
+        .get(connectionsArrayName, []);
```
aasimkhan30: NIT: Use connection manager to read connection. This is a little prone to breaking. It also doesn't get global connections. So if someone has connections in the workspace you don't read global values.

src/containerDeployment/dockerUtils.ts
```diff
@@ -0,0 +1,688 @@
...
+}
+
+//#region Docker Command Implementations
+
+/**
+ * Helper function to execute a command in the shell and return the output.
+ */
+async function execCommand(command: string): Promise<string> {
+    return new Promise((resolve, reject) => {
+        exec(command, (error, stdout) => {
```
aasimkhan30, line 181: Never used this before. Please make sure it cleans out the child process.

src/containerDeployment/dockerUtils.ts
```diff
@@ -0,0 +1,688 @@
...
+ * Finds an available port for a new Docker container, starting from the specified port.
+ * It checks the currently running containers and their exposed ports to find an unused port.
+ */
+export async function findAvailablePort(startPort: number): Promise<number> {
+    try {
+        const stdout = await execCommand(COMMANDS.GET_CONTAINERS);
+        const containerIds = stdout.split("\n").filter(Boolean);
+        if (!containerIds.length) return startPort;
+
+        const usedPorts = await getUsedPortsFromContainers(containerIds);
```
aasimkhan30, line 587: NIT: what if the port is used by something other than docker.

src/containerDeployment/containerDeploymentWebviewController.ts
```diff
@@ -0,0 +1,442 @@
...
+            trustServerCertificate: true,
+            server: "",
+            database: "",
+            user: "SA",
+            password: "",
+            applicationName: connectionApplicationName,
+            authenticationType: sqlAuthentication,
+            savePassword: false,
+            containerName: "",
+            version: "2025",
```
aasimkhan30: Is there a reason why we are serializing this information in connection profile?

src/containerDeployment/containerDeploymentWebviewController.ts
```diff
@@ -0,0 +1,442 @@
...
+            emptyPasswordInput: false,
+            azureAuthType: undefined,
+            accountStore: undefined,
+        };
+
+        sendActionEvent(TelemetryViews.ContainerDeployment, TelemetryActions.CreateSQLContainer, {
+            version: dockerProfile.version,
+        });
+
+        this.mainController.connectionManager.connectionUI
```
aasimkhan30: Just use standard async await instead of .then/.catch

src/controllers/mainController.ts
```diff
@@ -1205,6 +1215,126 @@ export default class MainController implements vscode.Disposable {
...
+        this._context.subscriptions.push(
+            vscode.commands.registerCommand(
+                Constants.cmdStartContainer,
+                async (node: TreeNodeInfo) => {
+                    if (!node) return;
+                    try {
+                        // doing it this way instead of directly calling startContainer
+                        // allows for the object explorer item loading UI to show
+                        this._objectExplorerProvider.deleteChildrenCache(node);
+                        await this._objectExplorerProvider.setNodeLoading(node);
```
aasimkhan30, line 1228: NIT: Maybe this code should have been added in OE service so that we can keep track of everything in one place.

src/objectExplorer/nodes/treeNodeInfo.ts
```diff
@@ -66,6 +68,13 @@ export class TreeNodeInfo extends vscode.TreeItem implements ITreeNodeInfo {
         } else {
             this.iconPath = ObjectExplorerUtils.iconPath(this.nodeType);
         }
+        if (connectionProfile && connectionProfile.containerName) {
```
aasimkhan30: Since this only applied to connection nodes, this should be present in ConnectionNode class

aasimkhan30 (approved):
While there are many changes made to the code, this looks good to me for the initial checkin. Please test and make sure everything is working before doing the merge.

aasimkhan30 (approved):

laurenastrid1:
@aasimkhan30 merged it in for me because of my github permissions bug

Issue #19477, opened 2025-05-23
[Feature Request]: Keep same connection when opening a file


richardtallent-erm:
### Feature Description

When I click a `.sql` file in the EXPLORER panel, I want the new tab that is opened to auto-connect to the same database as the previously active tab.

### Problem and Motivation

I feel like I spend half my life now reconnecting to the same database as I open SQL files, make changes, and execute them.

This feature would be in line with ADS and MSSQLMS.

This is related to, but not the same as, #1769 and #788. All express the same general idea that the user should be able to move between tabs and have a persistent database connection rather than having to break their workflow to constantly select and wait for connections to the database they are working in.

### Related Area

- [ ] Connection dialog
- [x] Query editor
- [ ] IntelliSense/auto-completion
- [ ] Query results panel
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Confirmation

- [x] I have searched existing feature requests and couldn't find a match
- [ ] I want to help implement this feature

PR #19503, opened 2025-05-28
Transferring active connection when opening a new SQL file

Benjin:

## Description

Fixes https://github.com/microsoft/vscode-mssql/issues/19477

Automatically transfers connection from the active SQL file to a newly-opened SQL file.

## Code Changes Checklist


aasimkhan30 (approved):

richardtallent-erm:
I'm glad to see this here, since it seems to be related to my issue (#19477)! However, the way this code seems to be set up (and I freely admit ignorance of the VSC codebase), it's unclear what would happen if a new tab were opened while a query is pending or running in the existing tab.

It's fine to transfer connections when there is no active query work going on. But if I have a query pending or running in Tab 1 and I open an SQL file creating Tab 2 (or just open a new query tab), I would want my query to continue running in Tab 1 and show its results there, but I would also want Tab 2 to auto-connect to the same database fluently. Ideally in a way that doesn't make me pause my work to wait for it to connect or update IntelliSense in order to run the new query. And certainly in a way that doesn't require me to manually choose the connection again (the current state).

aasimkhan30:
@richardtallent-erm ,
The way this code works, it doesn’t reuse the exact same connection between files. Instead, it gets the connection credentials and establishes a new connection for the new file. 

As a result:
	•	Each file is connected independently.
	•	Ongoing queries in the original file won’t be interrupted.
	•	Disconnecting one file won’t affect the other; the second file remains connected.

aasimkhan30 (approved):

src/controllers/connectionManager.ts
```diff
@@ -1514,7 +1514,11 @@ export default class ConnectionManager {
         }
     }
 
-    public async transferFileConnection(oldFileUri: string, newFileUri: string): Promise<void> {
+    public async transferFileConnection(
```
aasimkhan30: Can you add a doc comment and maybe update the name to cloneConnectionForFile or duplicateFileConnection?

aasimkhan30 (approved):

aasimkhan30 (approved):

aasimkhan30 (approved):

PR #19508, opened 2025-05-29
Add connection labels to Copilot chat response

caohai:
## Description
This PR adds connection status labels to the chat response in Copilot chat.
image

One more change I made was to add an extra connection check for the disconnected case, previously we were only checking the existence of the connection URI, now we also check if the connection object exists.

## Code Changes Checklist

### DIFF
--- a/localization/l10n/bundle.l10n.json
+++ b/localization/l10n/bundle.l10n.json
@@ -1229,6 +1229,13 @@
   "No model found.": "No model found.",
   "No tools to process.": "No tools to process.",
   "You are not connected to any database.": "You are not connected to any database.",
+  "Connected to {0},{1}/{0} is the server name{1} is the database name": {
+    "message": "Connected to {0},{1}",
+    "comment": [
+      "{0} is the server name",
+      "{1} is the database name"
+    ]
+  },
   "Using {0} ({1}).../{0} is the model name{1} is whether the model can send requests": {
     "message": "Using {0} ({1})...",
     "comment": [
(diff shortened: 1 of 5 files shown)


kburtram (approved):

test/unit/chatAgentRequestHandler.test.ts
```diff
@@ -137,10 +156,19 @@ suite("Chat Agent Request Handler Tests", () => {
                 })(),
             );
 
-        // Mock Language Model API
+        // Had to create a real object instead of using TypeMoq for the response object
```
caohai, line 159: @lewis-sanchez I was not able to get the `mockLanguageModelChatResponse`, looks like it's a broken proxy that cannot be resolved, any idea?
Here's a MVP demmo that can isolate the issue.
```typescript
    test("MVP Demo: TypeMoq proxy issue", async () => {
        // Use the SAME mock object from setup
        await Promise.resolve(mockLanguageModelChatResponse.object);
        expect(true).to.be.true; // Simple assertion
    });

```

As a workaround I had to create an actual object...

Benjin (approved):

croblesm:
Hi @caohai!

This enhancement looks great. However, many customers have very long server / database names. Therefore, adding a line break for each property would be even more clear:
image

caohai:
> Hi @caohai!
> 
> This enhancement looks great. However, many customers have very long server / database names. Therefore, adding a line break for each property would be even more clear: image

Updated!

PR #19552, opened 2025-06-10
Add agent mode tools

caohai:
## Description
This PR adds 4 tools for VSCode GitHub Copilot agent mode. `list_servers`, `connect`, `disconnet`, `show_schema`. It does this by contributing to VS Code [LanguageModelTools](https://code.visualstudio.com/api/extension-guides/tools) that enable the VS Code agents to perform actions.
These tools can also be manually invoked using `#` prefix in both Ask Mode and Agent Mode.

## Schema Designer related changes:
The `show_schema` tool opens the schema designer using a connection uri, this is not something previously supported so I added an extra parameter to related code path to support such use case while minimizing the changes to existing code. In the long run we might want to separate schema designer from OE completely but I don't think that's something we should do in ask mode.

Demo:

https://github.com/user-attachments/assets/b426ca7b-09ae-4403-a95f-be1d819cc62d

## Connect Flow:

1. **Connect via Saved Profile**
   - User can connect by referencing a saved connection profile by name
   - The LLM calls `mssql_list_servers` to validate the profile exists
   - The `mssql_connect` tool uses the provided `profileId` to establish the connection using all saved parameters

2. **Connect via Server/Database Specification**
   - **Exact Match Found**:
     - LLM calls `mssql_list_servers` to identify a matching profile with both server and database
     - LLM passes this information to `mssql_connect`
     - The connect tool uses the identified profile for connection

   - **Server Match Only**:
     - LLM calls `mssql_list_servers` and finds a matching server but different database
     - LLM passes server match and requested database to `mssql_connect`
     - The tool uses the matching server profile but substitutes the requested database
     - If connection fails, an error is reported
     - Future enhancement: Implement `mssql_list_databases` to improve database selection
   - **No Server Match**:
     - If no saved profile matches the specified server, the tool reports an error
     - Future enhancement: Open the connection dialog to create a new profile

A few examples:
1. Connect using profile name
 image
2. Connect using server/db
image
3.  Connect using db name with typo
image
4. Connect using server/db, db name is included in any saved connections
image

 

## Chat variables
image

## Code Changes Checklist


src/copilot/tools/connectTool.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    constructor(private connectionManager: ConnectionManager) {
+        super();
+    }
+
+    async call(
+        options: vscode.LanguageModelToolInvocationOptions<ConnectToolParams>,
+        _token: vscode.CancellationToken,
+    ) {
+        // TODO: should we connect via the connection profile name or server/db name?
```
Benjin: Who sets the input here/how?   The only guaranteed unique property is profile.id, so if this is machine-set, then maybe that can be used.  In order of confidence, I'd say it's probably id > profile name > server + db.

src/copilot/tools/connectTool.ts
```diff
@@ -0,0 +1,85 @@
...
+        const { serverName, database } = options.input;
+        // Fetch all profiles and find the requested one
+        const profiles = await this.connectionManager.connectionStore.readAllConnections();
+        const profile = profiles.find((p) => p.server === serverName);
```
Benjin: Maybe we should instead have a util function that's something like `findConnectionWithProperties` that'll select the best-matching profile with the supplied id/profileName/server/database

src/copilot/tools/connectTool.ts
```diff
@@ -0,0 +1,85 @@
...
+        let message: string;
+        try {
+            success = await this.connectionManager.connect(connectionId, {
+                ...profile,
+                database: database,
+            });
+            message = success ? loc.connectToolSuccessMessage : loc.connectToolFailMessage;
+        } catch (err) {
+            success = false;
+            message = err instanceof Error ? err.message : String(err);
```
Benjin: `utils.getErrorMessage()`

src/copilot/tools/connectTool.ts
caohai: The agent sets the input here (by combing the info from user input + result from `list_servers` call)
I'm actually working on changes for exactly this (add profile id to the input and prioritize profile id if present) but I need to confirm with Carlos about the whole connection workflow.

src/copilot/tools/toolBase.ts
```diff
@@ -0,0 +1,46 @@
...
+import { startActivity } from "../../telemetry/telemetry";
+
+export abstract class ToolBase<T> implements vscode.LanguageModelTool<T> {
+    toolName: string;
+
+    async invoke(
+        options: vscode.LanguageModelToolInvocationOptions<T>,
+        _token: vscode.CancellationToken,
+    ) {
+        const telemetryActivity = startActivity(
```
Benjin, line 17: There are already tool invocations happening before these were added, right?  Do we also capture telemetry for those tool calls?

src/schemaDesigner/schemaDesignerWebviewController.ts
```diff
@@ -188,7 +189,20 @@ export class SchemaDesignerWebviewController extends ReactWebviewPanelController
                         sessionId: this._sessionId,
                     });
                     // Open the document in the editor with the connection
-                    void this.mainController.onNewQuery(this.treeNode, result?.script);
+                    if (this.treeNode) {
```
Benjin: Is this change related to the agent-mode tools?

src/copilot/tools/connectTool.ts
caohai: Here's connect workflow I'm thinking:
1. A user can connect by referring to a saved connection profile. In this case we directly use that saved profile by telling agent to grab the profile id.
2. A user can connect by referring to a server/db combo.
   - If we can find a saved connection with the exact combo (find first match), use that. 
   -  If we can find a saved connection with the same server, but not the database mentioned by the user, we attempt to connect using that saved connection profile with the db name plugged. If connection failed, report error / list available databases [list_databsaes NOT IMPLEMENTED YET]
   - [NOT IMPLEMENTED YET] If we cannot find a saved connection with the same server, report error & prompt to open connection dialog.

Benjin: Yeah, that flow sounds good!

src/copilot/tools/toolBase.ts
caohai: Those are tools provided by Scriptoria and they are only used in chat mode. They are not natively exposed to vscode yet.
 We will be looking into if we want to bring them to vsode agent mode (using the same API in the PR) later.

src/schemaDesigner/schemaDesignerWebviewController.ts
caohai: Yes, we want to support `show_schema` tool in agent mode and that uses schema designer.
Currently schema designer is coupled with OE (using treeNode), treeNode is used for two purposes:
1. connectionStr gen for the backend service
2. opening the publish script in a connected query editor (by calling `NewQuery` on a treeNode)

This PR adds an extra code path that takes in a connectionUri and doesn't rely on OE. This change here handles opening the publish script in a new editor and connect to the server/db using connectionUri.

src/copilot/tools/connectTool.ts
caohai: agreed, will do that once I confirm the flow with Carlos, the input should be profileId/server/database, we can tell LLM to give us profileId if possible and prioritize that. I might not take profileName as it's not always available and will be redundant info if LLM can get the id right already.

src/schemaDesigner/schemaDesignerWebviewController.ts
```diff
@@ -31,6 +31,7 @@ export class SchemaDesignerWebviewController extends ReactWebviewPanelController
         private databaseName: string,
         private treeNode: TreeNodeInfo,
```
aasimkhan30: Can you make this optional and update the doc comment?

aasimkhan30:
The changes look good to me. 
NIT: We can make the tree node param optional for schema designer but not necessary for getting this PR in.

src/copilot/tools/connectTool.ts
caohai: updated

src/copilot/tools/connectTool.ts
caohai: updated

src/copilot/tools/connectTool.ts
caohai: I've updated the prompts and tool implementation to reflect this workflow. I've added this same flow info and a few screenshots to the PR description.

src/controllers/connectionManager.ts
```diff
@@ -1249,25 +1249,15 @@ export default class ConnectionManager {
...
-     * @returns true if connection is successful, false otherwise
+     * Handles password-based credential authentication by prompting for password if needed.
+     * This method checks if a password is required and prompts the user if it's not saved or available.
+     *
+     * @param connectionCreds The connection credentials to process
+     * @returns Promise that resolves to true if password handling was successful, false if user cancelled
      */
-    public async connect(
-        fileUri: string,
+    public async handlePasswordBasedCredentials(
```
caohai, line 1258: Not sure why git decides to show the diff this way. The change here is to extract some code for adding pwd to the cred that Aasim added a few weeks ago into a helper function and this function is later used by connectTool.

caohai: For example I have localhost master saved, now I want to connect to localhost TestDB2, I'll re-use the master profile, get the pwd, plug in the TestDB2 name, attempt the connection and see if the connection succeeds or not.

aasimkhan30 (approved):

PR #19607, opened 2025-06-14
Allowing container checks to fail gracefully

Benjin:

## Description

Allows connection attempt to proceed if the attempt to check container status throws an error instead of a definitive confirmation that the container is running.  Previously, if a connection had a `containerName` property but docker was not accessible, all connection attempts would fail, both in OE and the connection dialog.

## Code Changes Checklist


aasimkhan30:
Based on the code, if the docker commands fail to run we try a new connection with the profile and then expose a sql client error, right?

Benjin:
@aasimkhan30 if the docker commands fail to execute, it will proceed as though it's not a docker connection, so:
* server is actually running (and we just failed to detect it): succeeds
* server is not actually running: fails with SqlClient error as though the server doesn't exist.

Sound good?

Issue #19365, opened 2025-05-07
SCreen reader a11y: schema comparison grid keyboard accessibility not ideal


LeonarddeR:
- MSSQL Extension Version: 1.31.1
- VSCode Version: 1.99.3
- OS Version: 1Windows 11 24H2

## Steps to Reproduce:

1. Perform a schema compare
2. Tab to the comparison table

## Expected results:
1. Pressing tab again moves you out of the table
2. Arrow keys allow you to move through the table, e.g. up/down arrow moves between rows, left/right arrow moves between the several columns

## Actual results:
1. Pressing tab moves through the several check boxes in the table to enable/disable items. I assume tabindex for these checkboxes should be set to -1 explicitly to avoid this
2. Keyboard navigation with up/down arrows is either not available yet or not reported by screen reader NVDA

Issue #19182, opened 2025-04-15
Schema compare - comparison results checkboxes not consistently lined up


kisantia:
Sometimes the "Select all" checkbox is not lined up with the checkboxes below
![Image](https://github.com/user-attachments/assets/78ef4c0d-fbe3-47c5-8a83-367d6aa63deb)

Issue #19178, opened 2025-04-15
Schema compare - checkbox overlaps with long source name


kisantia:
If the source name is long enough, then the checkbox covers it
![Image](https://github.com/user-attachments/assets/52a720f9-bf47-43d1-9b73-f2d6b5d6361c)

PR #19641, opened 2025-06-23
Improve Schema Compare Differences List

lewis-sanchez:

## Description
I'm aware of the new rule surrounding a single PR fixing a single issue, but by replacing the table components with the datagrid component, I was able to fix these issues with a single change.

This PR fixes https://github.com/microsoft/vscode-mssql/issues/19365
This PR fixes https://github.com/microsoft/vscode-mssql/issues/19178
This PR fixes https://github.com/microsoft/vscode-mssql/issues/19182

This PR replaces the Table Fluent UI component that was previously used to list all schema comparison differences with the DataGrid Fluent UI component. The DataGrid component features long text truncation, features a better layout for consistent checkbox alignment, and built in keyboard accessibility.

### Improved keyboard navigation
keyboard-accessibility

### Checkbox alignment:

Before:
Previously when the schema comparison differences list was small, a scrollbar wouldn't appear which would cause table body rows to shift to the right and get out of alignment.
image

But if the differences list contained a scroll bar then rows would appear correctly aligned:
image

After:
With the changes in this PR that alignment issue with checkboxes is fixed as seen here with a small differences list:
image

And here with a differences list that contains a scroll bar: 
image

### Long source/target names no longer overlap with checkboxes.

Before:
Long text overlaps with checkboxes in the schema compare differences.
image

After:
Long names are truncated, so that they don't overlap with the checkboxes like they did previously:
image

### Compacted row height:
image

### Fixed spinner alignment issue:
In the gif, the spinner appears slightly to the left of the checkboxes, but that is now fixed as seen here:
image

## Code Changes Checklist

### DIFF
--- a/src/reactviews/pages/SchemaCompare/components/SchemaDifferences.tsx
+++ b/src/reactviews/pages/SchemaCompare/components/SchemaDifferences.tsx
@@ -14,7 +14,6 @@ import {
     DataGridHeaderCell,
     Text,
     TableColumnSizingOptions,
-    mergeClasses,
 } from "@fluentui/react-components";
 import {
     DataGridBody,
@@ -81,9 +80,6 @@ const useStyles = makeStyles({
     },
     dataGridHeader: {
         backgroundColor: "var(--vscode-keybindingTable-headerBackground)",
-    },
-    dataGridRowHeight: {
-        height: "30px !important",
     },
 });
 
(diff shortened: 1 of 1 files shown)


aasimkhan30:
quick suggestions based on the gif. 

1. Can you make sure that loading spinner is aligned with the checkboxes. 
2. Can you add a different background color to the table header? (--vscode-keybindingTable-headerBackground variable can be used for that)
3. If possible make the table header sticky so that only the rows are moved and the header remains at the top.
4. Maybe make the rows compact (reduce height) so that we can see more rows in the same screen real estate.

aasimkhan30 (approved):

Issue #19158, opened 2025-04-15
Authentication type dropdown hover not readable in light theme


kisantia:
- MSSQL Extension Version: mssql-1.31.0 (from bugbash)
- VSCode Version: 1.99.2
- OS Version: Windows 11

Steps to Reproduce:

1. Use a light theme, like Light+
2. Open Connection dialog
3. Click on Authentication dropdown and hover over the dropdown items with mouse. The background of the hovered over item is dark, making the text difficult to read and not accessible

![Image](https://github.com/user-attachments/assets/56d371fb-40d9-4e18-a369-e8055f6b0cd1)

kisantia:
Also happens when hovering over the "..." and "Compare" buttons in schema compare

PR #19655, opened 2025-06-25
Fix dropdown option stylings for lighter themes.

lewis-sanchez:

## Description

This PR fixes https://github.com/microsoft/vscode-mssql/issues/19158

This PR corrects the hover style for drop down options. As seen in the screenshots I've verified that drop downs are appearing consistently across feature areas to ensure that UI regressions are not introduced by also checking the connection dialog in addition to schema compare dialog.

### Before:
The hover style for dropdowns was hard to read in lighter themes, but looked fine with darker themes as seen here:
image
image
image
image

### After:
The hover style for drop down options is much more readable with the styling adjustments:
image
image
image
image

## Code Changes Checklist

### DIFF
--- a/src/reactviews/common/theme.ts
+++ b/src/reactviews/common/theme.ts
@@ -18,6 +18,9 @@ export const webviewTheme: fluentui.Theme = {
     colorBrandBackground2Hover: "var(--vscode-button-hoverBackground)",
     colorBrandForeground2: "var(--vscode-list-inactiveSelectionForeground)",
     colorNeutralForegroundOnBrand: "var(--vscode-button-foreground)",
+    /**
+     * Background color for a dropdown option that is hovered over.
+     */
     colorNeutralBackground1Hover: "var(--vscode-dropdown-background)",
     colorNeutralForeground1Hover: "var(--vscode-editor-foreground)",
     colorNeutralBackgroundInverted: "var(--vscode-editor-selectionBackground)",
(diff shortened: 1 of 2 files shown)


src/reactviews/index.css
```diff
@@ -43,6 +43,12 @@ button[role="combobox"] .placeholder {
     color: var(--vscode-input-placeholderForeground) !important;
 }
 
+.fui-Option:hover,
```
Benjin: I think the correct fix ought to be in `theme.ts`, rather than here.  Fixing the CSS directly rather than the Fluent theme means that other components that are still using the wrong theme variable remain dark:

image

Benjin: Playing around, I think you actually want:
```
    colorNeutralBackground1Hover: "var(--vscode-dropdown-background)",
    colorNeutralBackground1Pressed: "var(--vscode-dropdown-background)",
```

lewis-sanchez: Thank you, and this suggestion is working better too.

src/reactviews/common/theme.ts
```diff
@@ -18,13 +18,13 @@ export const webviewTheme: fluentui.Theme = {
     colorBrandBackground2Hover: "var(--vscode-button-hoverBackground)",
     colorBrandForeground2: "var(--vscode-list-inactiveSelectionForeground)",
     colorNeutralForegroundOnBrand: "var(--vscode-button-foreground)",
-    colorNeutralBackground1Hover: "var(--vscode-button-secondaryHoverBackground)",
+    colorNeutralBackground1Hover: "var(--vscode-dropdown-background)",
```
aasimkhan30, line 24: can you please add comments for this variable?

aasimkhan30 (approved):

PR #19670, opened 2025-06-27
Fix result grid losing state when switching query editors. 

### LINKED ISSUE #18666: Results view scrolls back to top when switching tabs
Type: <b>Feature Request</b>

If you have query results displayed in the results window and they are more than a page long then the window jumps back to the top result if you leave the document tab and come back. The easiest way to replicate this is to have a query that returns JSON in a column.

1. Run the query.
2. Scroll to the middle or end of the results in the results window.
3. Click on the JSON column to open it in a new editor window.
4. Close the new window and return to the SQL query.

The results window has jumped back to the top of the results. It should remain unchanged such that you are still looking at the row you previously clicked on.

This has a detrimental impact on functionality:
- You have to keep track of which row(s) you're looking at if you switch away from the editor and back.
- If the resultset is large then you have to continually scroll back to your previous location if you are examining columns in another window.

Extension version: 1.28.0
VS Code version: Code 1.97.1 (e249dada235c2083c83813bd65b7f4707fb97b76, 2025-02-10T13:13:58.153Z)
OS version: Windows_NT x64 10.0.26100
Modes:

aasimkhan30:

## Description

This PR fixes #18666 

## Code Changes Checklist

### DIFF
--- a/src/reactviews/pages/QueryResult/resultGrid.tsx
+++ b/src/reactviews/pages/QueryResult/resultGrid.tsx
@@ -125,6 +125,7 @@ const ResultGrid = forwardRef<ResultGridHandle, ResultGridProps>((props: ResultG
         const setupState = async () => {
             await table.setupFilterState();
             await table.restoreColumnWidths();
+            await table.setupScrollPosition();
             table.headerFilter.enabled =
                 table.grid.getDataLength() < context.state.inMemoryDataProcessingThreshold!;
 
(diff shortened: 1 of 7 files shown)


src/reactviews/pages/QueryResult/resultGrid.tsx
```diff
@@ -125,6 +125,7 @@ const ResultGrid = forwardRef<ResultGridHandle, ResultGridProps>((props: ResultG
         const setupState = async () => {
             await table.setupFilterState();
             await table.restoreColumnWidths();
+            await table.setupScrollPosition();
```
Benjin, line 128: which of these are dependent on each other?  I wonder if we can `Promise.all()` any of them to parallelize processing.

aasimkhan30: I think setupFilterState and column Width are needed to be done before so that we get the right scroll position.

caohai (approved):

PR #19686, opened 2025-07-03
Adding support for multiple Azure accounts

Benjin:
## Description

Adding support for multiple Azure accounts in the connection dialog and firewall rule management dialog.  Also, miscellaneous UX improvements related to azure accounts.

Dropdown with sign-in option in Firewall rule dialog:
image

Subscription filter groups subscriptions by account:
image

Displays helpful message if user cancels sign-in:
image

Sign-in link next to "Browse Azure" can be used to sign into additional accounts:
image

Message banner is now dismissible:
image

## Code Changes Checklist


caohai:
I tested this branch with Benjin and here are a few findings/comments:

1. With this change there are two subscription level filters, we might want to merge these two and have one filter with multi-selection support. The other fields (resource group etc) already work as expected if we only use filter 1.
image
2. It will be better to have some sort of info about which accounts are logged in. This requires UX/PM input.
3. The `Add Firewall Rule` dialog currently resets all fields to default value after the "Add firewall rule" button is clicked. This is just a UI bug, the functionality is working as expected.

I'll take a closer look at code changes tomorrow.

src/configurations/config.json
```diff
@@ -1,7 +1,7 @@
 {
     "service": {
         "downloadUrl": "https://github.com/Microsoft/sqltoolsservice/releases/download/{#version#}/microsoft.sqltools.servicelayer-{#fileName#}",
-        "version": "5.0.20250616.1",
+        "version": "5.0.20250702.1",
```
caohai, line 4: Can you add a screenshot / list of the changes in STS with this vbump?

src/reactviews/pages/ConnectionDialog/azureBrowsePage.tsx
```diff
@@ -192,200 +224,225 @@ export const AzureBrowsePage = () => {
...
-                        // TODO: swap out subscription ID parsing for an AzureSubscriptionInfo
-                        const openParen = sub.indexOf("(");
+                                if (sub === undefined) {
+                                    return;
+                                }
 
-                        if (openParen === -1) {
-                            return;
-                        }
+                                // TODO: swap out subscription ID parsing for an AzureSubscriptionInfo
```
caohai: Will this get updated with this PR or a separate PR?

src/utils/utils.ts
```diff
@@ -93,3 +93,16 @@ export async function listAllIterator<T>(iterator: PagedAsyncIterableIterator<T>
 
     return resources;
 }
+
+/** Groups items by a specified key */
+export function groupBy<K, V>(values: V[], key: keyof V): Map<K, V[]> {
+    // TODO: replace with Object.groupBy once ES2024 is supported
```
caohai, line 99: Will this TODO get implemented?

Benjin: I just moved this code from elsewhere.  Since it requires bumping the ES target, I'll do that in a separate PR so it's easier to roll back if needed.

src/reactviews/pages/ConnectionDialog/azureBrowsePage.tsx
Benjin: I'll delete this TODO; I think it's no longer possible without upgrading the custom SearchableCombobox component, which is out of scope for this PR.

src/configurations/config.json
Benjin: image

Benjin:
@caohai 

> I tested this branch with Benjin and here are a few findings/comments:
> 
> 1. With this change there are two subscription level filters, we might want to merge these two and have one filter with multi-selection support. The other fields (resource group etc) already work as expected if we only use filter 1.
>    image
> 2. It will be better to have some sort of info about which accounts are logged in. This requires UX/PM input.
> 3. The `Add Firewall Rule` dialog currently resets all fields to default value after the "Add firewall rule" button is clicked. This is just a UI bug, the functionality is working as expected.

I'll need UX imput for (1), but have an idea for how to improve (2).  I'll make items to track [(2)](https://github.com/microsoft/vscode-mssql/issues/19719) and [(3)](https://github.com/microsoft/vscode-mssql/issues/19718), and send follow-up PRs.

PR #19715, opened 2025-07-08
Added clearer telemetry to local containers workflow

laurenastrid1:

## Description
Added clearer telemetry in local containers for different actions, and added error telemetry for local containers

## Code Changes Checklist

### DIFF
--- a/test/unit/containerDeploymentWebviewController.test.ts
+++ b/test/unit/containerDeploymentWebviewController.test.ts
@@ -18,7 +18,6 @@ import {
     ContainerDeploymentWebviewState,
     DockerStepOrder,
 } from "../../src/sharedInterfaces/containerDeploymentInterfaces";
-import * as telemetry from "../../src/telemetry/telemetry";
 import { AddLocalContainerConnectionTreeNode } from "../../src/containerDeployment/addLocalContainerConnectionTreeNode";
 import { ConnectionUI } from "../../src/views/connectionUI";
 import { stubTelemetry } from "./utils";


src/containerDeployment/containerDeploymentWebviewController.ts
```diff
@@ -147,6 +166,21 @@ export class ContainerDeploymentWebviewController extends FormWebviewController<
...
+                sendErrorEvent(
+                    TelemetryViews.ContainerDeployment,
+                    TelemetryActions.RunDockerStep,
+                    new Error(currentStep.errorMessage),
+                    true, // includeErrorMessage
+                    undefined, // errorCode
+                    undefined, // errorType
+                    {
+                        dockerStep: cd.DockerStepOrder[currentStepNumber],
+                        fullErrorMessage: currentStep.fullErrorText || "",
```
Benjin: is this necessary with includeErrorMessage: true on line 176?  Check to see what's already getting emitted

src/containerDeployment/dockerUtils.ts
```diff
@@ -473,6 +494,13 @@ export async function checkIfContainerIsReadyForConnections(
                 if (readyLine) {
                     clearInterval(interval);
                     dockerLogger.appendLine(`${containerName} is ready for connections!`);
+                    sendActionEvent(
+                        TelemetryViews.ContainerDeployment,
+                        TelemetryActions.StartContainer,
+                        {
+                            timeToStartInMs: `${Date.now() - start}`,
```
Benjin: stuff that you might want to math over (like this) should part of the "measures" bag (the next parameter, iirc) rather than the "properties" bag

src/containerDeployment/dockerUtils.ts
```diff
@@ -494,12 +522,19 @@ export async function checkIfContainerIsReadyForConnections(
...
         await execCommand(COMMANDS.DELETE_CONTAINER(containerName));
+        sendActionEvent(TelemetryViews.ContainerDeployment, TelemetryActions.DeleteContainer);
         return true;
-    } catch {
+    } catch (e) {
+        sendErrorEvent(
+            TelemetryViews.ContainerDeployment,
+            TelemetryActions.DeleteContainer,
+            e,
+            true, // includeErrorMessage
```
Benjin: for everywhere that you've set `includeErrorMessage: true`, please confirm that the nature of errors here aren't likely to have PII (local paths that might include the user's name, passwords, etc).  Obviously, we don't know every possible error message that might emit, but just getting a sense of the most likely one(s) would be fine.

test/unit/containerDeploymentWebviewController.test.ts
```diff
@@ -311,6 +311,9 @@ suite("ContainerDeploymentWebviewController", () => {
 
     test("completeDockerStep reducer updates step status and handles success/failure", async () => {
         const addContainerConnectionStub = sinon.stub(controller as any, "addContainerConnection");
+        // Stub telemetry method
+        const sendActionEventStub = sinon.stub(telemetry, "sendActionEvent");
+        const sendErrorEventStub = sinon.stub(telemetry, "sendErrorEvent");
```
Benjin: There's a helper that you could use instead to consolidate behavior:
`const { sendErrorEvent } = stubTelemetry(sandbox);`

test/unit/containerDeploymentWebviewController.test.ts
```diff
@@ -311,6 +311,9 @@ suite("ContainerDeploymentWebviewController", () => {
 
     test("completeDockerStep reducer updates step status and handles success/failure", async () => {
         const addContainerConnectionStub = sinon.stub(controller as any, "addContainerConnection");
```
Benjin, line 313: For another PR, but we generally want to use `sinon.sandbox` instead; set it up in the `setup()` method, and restore it in the `teardown()`.  Every stub that's created in the sandbox gets restored when `sandbox.restore()` is called, so it saves the effort of having to remember the different references.

See `test\unit\addFirewallRuleWebviewController.test.ts` for an example

Benjin (approved):

src/containerDeployment/dockerUtils.ts
laurenastrid1: it's hard to know whether the error will contain PII or not, because it depends on the user's system and environment. but for now i've changed it to be false for include error message so that we don't have pii problems

laurenastrid1:
@whoever approves this can you merge this pull request? thanks!

PR #19723, opened 2025-07-09
Added local container connection group unit tests

laurenastrid1:

## Description
Added tests for the local container connection group reducers, and also added null checks for some telemetry emission for safety

## Code Changes Checklist

### DIFF
--- a/test/unit/containerDeploymentWebviewController.test.ts
+++ b/test/unit/containerDeploymentWebviewController.test.ts
@@ -289,8 +289,11 @@ suite("ContainerDeploymentWebviewController", () => {
     });
 
     test("Test formAction reducer", async () => {
-        const validateProfileSpy = sinon.spy(controller as any, "validateDockerConnectionProfile");
-        const updateStateSpy = sinon.spy(controller as any, "updateState");
+        const validateProfileSpy = sandbox.spy(
+            controller as any,
+            "validateDockerConnectionProfile",
+        );
+        const updateStateSpy = sandbox.spy(controller as any, "updateState");
 
         const callState = controller["state"];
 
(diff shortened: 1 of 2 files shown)


laurenastrid1:
@ whoever approves this, please merge it as well due to my git permissions bug

test/unit/containerDeploymentWebviewController.test.ts
```diff
@@ -501,6 +502,64 @@ suite("ContainerDeploymentWebviewController", () => {
...
+            "Should include error message",
+        );
+
+        createConnectionGroupStub.resolves({ id: "test-group-id", name: "Test Group" });
+        result = await controller["_reducerHandlers"].get("createConnectionGroup")(callState, {
+            connectionGroupSpec: {
+                name: "Test Group",
+            },
+        });
+        assert.ok(
```
Benjin: IMO, better to call `resetHistory()` on the stub in between the two cases instead of checking `calledOnce` then `calledTwice` since it makes it more clear what the expected outcome is for each individual case.

test/unit/containerDeploymentWebviewController.test.ts
```diff
@@ -501,6 +502,64 @@ suite("ContainerDeploymentWebviewController", () => {
...
     });
 
+    test("Test createConnectionGroup reducer", async () => {
+        const createConnectionGroupStub = sinon.stub(
+            ConnectionGroupWebviewController,
+            "createConnectionGroup",
+        );
+
+        createConnectionGroupStub.resolves("Error creating group");
+        let callState = (controller as any).state;
```
Benjin: Does `controller["state"]` work here?  That'd have the advantages of 1) getting compile-time validation that the property exists, and 2) typing callState correctly.

test/unit/containerDeploymentWebviewController.test.ts
```diff
@@ -406,13 +413,13 @@ suite("ContainerDeploymentWebviewController", () => {
                 DockerStepOrder.connectToContainer
             ].errorMessage.includes("dev-profile"),
         );
-        assert.ok(sendErrorEvent.calledTwice, "sendErrorEvent should be called twice");
+        assert.ok(sendErrorEvent.calledOnce, "sendErrorEvent should be called twice");
 
         addContainerConnectionStub.restore();
```
Benjin: restore call shouldn't be necessary since the stub is created in the same test.  this is for when you need to "unstub" something for later execution, but in this case, it's all cleaned up.

Same for several other restores in this file.

aasimkhan30 (approved):

PR #19765, opened 2025-07-15
Adding vsix based smoke tests

aasimkhan30:

## Description

Adding 2 playwright tests that test the packaged vsix file by installing it in a test version of vscode.

1. Loads OE. This test also makes sure STS is installed correctly.
2. Loads Connection Dialog Webview to check if the web views are loading properly. 

I tried breaking the vsix by changing actual dependencies to dev-dependencies. 
The smoke test step did fail at vsix.

image

Since code coverage doesn't work with packaged vsix (js files are in the package and can't be instrumented properly). I haven't converted all tests to run using the packaged vsix.

## Code Changes Checklist


caohai:
~~Is this the run with the smoke test failure? The error message appears to be irrelevant to extension activation.~~
https://github.com/microsoft/vscode-mssql/actions/runs/16298523350/job/46026676160#step:21:47
nvm found the detailed error messages

Benjin:
> Since code coverage doesn't work with packaged vsix (js files are in the package and can't be instrumented properly). I haven't converted all tests to run using the packaged vsix.

Might be nice to add this as a comment to the VSIX tests: that _only_ the VSIX-specific tests should be run with that flag, and everything else should be run directly off of the raw build output.

Benjin:
> Is this the run with the smoke test failure? The error message appears to be irrelevant to extension activation. https://github.com/microsoft/vscode-mssql/actions/runs/16298523350/job/46026676160#step:21:47

@laurenastrid1 I remember that you were going to look into having the smoketest step itself fail if there were test failures instead of the smoketest code coverage step.  Did you have a chance to circle back to that?

test/e2e/vsix.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    test.beforeEach(async () => {
+        // Launch with new UI off
+        const { electronApp, page } = await launchVsCodeWithMssqlExtension({
+            useVsix: true,
+        });
+        vsCodeApp = electronApp;
+        vsCodePage = page;
+    });
+
+    // Test if extension activates correctly
```
caohai, line 30: IIRC the extension icon shows up upon installation and does not rely on the activation to be successful, at least that's what I saw with the RC build yesterday...

Benjin: Yeah, anything that comes directly from the package.json manifest (with no additional computation) is visible immediately.

test/e2e/utils/launchVscodeWithMsSqlExt.ts
```diff
@@ -3,51 +3,111 @@
...
+
+    if (isSelected !== "true") {
+        const sqlServerTabElement = sqlServerTabContainer.locator("a");
+        await sqlServerTabElement.waitFor({ state: "visible", timeout: 30 * 1000 });
+        await sqlServerTabElement.click();
+    }
 
-    // Wait for extension to load
+    // Wait for Object Explorer to load
     const objectExplorerProviderElement = page
```
aasimkhan30, line 121: @caohai. It seems like this part of code takes care of it. Don't know why that count exists in the test. I got this test directly from https://github.com/microsoft/vscode-mssql/blob/896cd5af7902b0747370a58132cf3626c2f3ecc4/test/e2e/activityBar.spec.ts

caohai: ah I see the `launchVsCodeWithMssqlExtension`  call in `test.beforeEach` now, makes sense, I was wondering why this test would fail and this explains it.
Yeah, the `count` check is not checking anything useful here, we can probably just remove this section. The `launchVsCodeWithMssqlExtension` call + connection dialog check should be sufficient.

Benjin: In that case, maybe the test itself should check the "Add Connection" button (which only shows up after the extension has loaded and not found any saved connections) instead of just getting the activity bar item?

Both are redundant with the setup code, but only the "Add Connection" button is actually relevant to activation state.

aasimkhan30: @benjin, I have renamed the tests. 

The first test just checks activation.
The second test checks for add connection button and subsequent webview load.

aasimkhan30:
@laurenastrid1, 

1. can you please fix the pipeline steps to upload screenshots on test failure? It seems that step is skipped if there is failure in tests. It defeats the purpose of having the step in the first place.
2. Also, can you please move the test runtime artifacts to their own folder (use `artifacts` instead of `resource` and add it to .gitignore? Currently the runtime artifacts pollute the source control tree. 
image

PR #19782, opened 2025-07-16
Fixing firewall rule dialog account selection reset upon submission

Benjin:
## Description

Because arrays were used as the triggers for the account refresh/selection effect and the references were changing upon state refresh, the dialog was automatically re-selecting first account in the list.  While this didn't impact the firewall rule creation at all, it's odd to see for users.

## Code Changes Checklist

- [n/a] New or updated **unit tests** added


Benjin:
@caohai Turned out that it _did_ repro on Mac; only the account selection was reset getting reset unexpectedly, which wasn't part of my original validation.

Issue #19581, opened 2025-06-11
[Bug]: Error when editing profile name of a connection inside a group


croblesm:
### Description

When attempting to edit the profile name of a saved connection that belongs to a Connection Group, the MSSQL extension throws an error and fails to update the name. This only occurs for connections within a group—connections outside groups work as expected.

Error message:
`Cannot resolve tree item for element XYZ from extension ms-mssql.mssql`

https://github.com/user-attachments/assets/0b1e7204-1a3e-4cd5-ab2a-a04680bfe12d

Expected behavior:
The profile name should update successfully without errors, regardless of whether the connection is grouped.

### Steps to Reproduce

1. Right-click the connection and choose Edit.
2. Change the Profile name and click OK.
3. Observe that an error message appears, and the name is not updated.

### Affected Area

- [x] Connection dialog
- [ ] Query editor
- [ ] IntelliSense/auto-completion
- [ ] Query results panel
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

### If you selected "Other", please describe the affected area

Connection Groups

### Environment Information

VS Code version: 1.101.0-insider
MSSQL Extension Version: 1.32.0
OS: macOS 15.5
SQL Server Version: 2025
Connection Type: Local container

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #19785, opened 2025-07-16
Fixing issue where renaming an open connection within a folder on macOS fails

Benjin:
## Description

Addresses https://github.com/microsoft/vscode-mssql/issues/19581

There's a race condition between events triggered by config writes and object explorer refreshes that only repros on macOS.  The work-around is to submit the save and OE refresh a second time after failure.

To address this issue more completely, we should bring over ADS's system of config update events that are driven in just one direction (config store event emissions) instead of VS Code's current design of triggering on config file changes (which we can't control the ordering of relative to the config write itself) and back-calculating the OE updates that need to occur.

## Code Changes Checklist

### DIFF
--- a/src/connectionconfig/connectionDialogWebviewController.ts
+++ b/src/connectionconfig/connectionDialogWebviewController.ts
@@ -75,6 +75,7 @@ import {
 } from "../controllers/connectionGroupWebviewController";
 import { populateAzureAccountInfo } from "../controllers/addFirewallRuleWebviewController";
 import { MssqlVSCodeAzureSubscriptionProvider } from "../azure/MssqlVSCodeAzureSubscriptionProvider";
+import { TreeNodeInfo } from "../objectExplorer/nodes/treeNodeInfo";
 
 export class ConnectionDialogWebviewController extends FormWebviewController<
     IConnectionDialogProfile,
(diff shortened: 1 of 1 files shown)


src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -822,21 +823,40 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
...
+            let node = await saveConnectionAndCreateSession(this);
 
             this.state.connectionStatus = ApiStatus.Loaded;
 
-            await this._mainController.objectExplorerTree.reveal(node, {
-                focus: true,
-                select: true,
-                expand: true,
-            });
+            try {
```
aasimkhan30, line 845: Should we add an event in the OE provider that is fired whenever we refresh a tree item. We can listen to the event and only continue when we refresh for the right parent is done. 

I also like your idea of making config writes from the extension unidirectional.

Benjin: ObjectExplorerService already has `_onDidChangeTreeData(node)` that triggers upon refresh, but I think leveraging that for this particular scenario might result in more event spaghetti.  My hope is that driving the config write eventing from the bottom-up would untangle much of that.

PR #19812, opened 2025-07-20
Removing gulp from the extension building process 

aasimkhan30:

## Code Changes Checklist


localization/README.md
```diff
@@ -77,18 +77,10 @@ const test = t(`Test loc`);
...
 2. This task reads the translated XLIFF files and generates localized JSON files:
    - `bundle.l10n.{locale}.json` files are saved in the `localization\l10n` folder.
    - `package.nls.{locale}.json` files are saved in the root folder.
 
 ### Step 4: Using the Localization Files
-- The extension uses these generated JSON files to localize the strings in the user interface (UI). The root of the l10n is provided in the `package.json` file with `l10n` key. For now it is set to `localization/l10n`.
-
-## To test localization.
-
-1. Run `yarn gulp ext:generate-pseudo-loc` to generate pseudo localization files.
```
Benjin, line 90: No new equivalent?

package.json
```diff
@@ -50,20 +50,25 @@
...
         "test": "node ./out/test/unit/runTest.js",
-        "package": "vsce package",
+        "package": "node scripts/package-extension.js",
         "prepare": "husky",
         "lint-staged": "lint-staged --quiet",
         "precommit": "run-p lint-staged localization",
-        "clean-package": "git clean -xfd && yarn install && yarn build && yarn gulp package:online",
         "instrument": "nyc instrument ./out/src ./out/src --in-place",
-        "e2eTest": "npx playwright test; npx nyc report --reporter=cobertura --reporter=text-summary --reporter=html --report-dir=\"coverage/coverage-e2e\" --temp-dir=\"coverage/reactviews-coverage\" --include=\"src/reactviews/pages/**/*.tsx\""
+        "e2eTest": "npx playwright test ; npx nyc report --reporter=cobertura --reporter=text-summary --reporter=html --report-dir=\"coverage/coverage-e2e\" --temp-dir=\"coverage/reactviews-coverage\" --include=\"src/reactviews/pages/**/*.tsx\""
```
Benjin, line 72: accidental space?

scripts/copy-assets.js
```diff
@@ -0,0 +1,130 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+const fs = require('fs').promises;
+const path = require('path');
+const logger = require('./terminal-logger');
+
+const COPY_TASKS = [
```
Benjin, line 10: When does something need to be added to this list?  I.e. `/media/*` and `/src/reactviews/media/*` aren't here.  Can you add a comment explaining what gets included vs. excluded?

scripts/copy-assets.js
```diff
@@ -0,0 +1,130 @@
...
+    name: 'Configuration Files',
+    source: 'src/configurations/config.json',
+    destination: 'out/src/config.json'
+  }
+];
+
+/**
+ * Copy files from source to destination
+ */
+async function copyFiles(src, dest, filter = null) {
```
Benjin, line 40: Please add a docstring describing behavior for `filter` (no filter =all files, filter = moved if true)

scripts/copy-assets.js
```diff
@@ -0,0 +1,130 @@
...
+  let count = 0;
+
+  try {
+    const srcStats = await fs.stat(src);
+
+    if (srcStats.isFile()) {
+      if (!filter || filter(src)) {
```
Benjin: if a file is explicitly specified, why evaluate the filter?  I would expect an explicit call to override a filter.

scripts/extract-localization.js
```diff
@@ -0,0 +1,128 @@
...
+};
+
+if (require.main === module) {
+  extractLocalizationStrings()
+    .then(() => {
+      logger.success("🎉 Script completed successfully!");
+      process.exit(0);
+    })
+    .catch((error) => {
+      logger.error(`💥 Script failed: ${error.message}`);
```
Benjin: We should probably (eventually) be consistent in emoji usage for different states.  I'm seeing both ✨ and 🎉 for successes, <nothing> and ⚠️ for warning, and <nothing> and 💥 for errors.  It'd just make it easier to scan/search the output for status.

We don't need do anything for this PR; realistically, this'll be something we'll just tweak with usage until the output looks "right".

scripts/package-extension.js
```diff
@@ -0,0 +1,266 @@
...
+Usage:
+  node package-extension.js [mode]
+
+Modes:
+  --online     Package for online distribution (downloads service at runtime)
+  --offline    Package for offline distribution (includes service for all platforms)
+  --help       Show this help message
+
+Examples:
+  node package-extension.js --online    # Create online package
```
Benjin, line 209: Update this to indicate that online is the default mode:

`node package-extension.js [--online]  # Create online package.  Default behavior if none specified`

test/e2e/README.md
```diff
@@ -30,13 +30,13 @@ From the root of the repo, install all of the build dependencies:
...
 ```shell
-gulp build
+yarn build
 ```
 
 or watch for changes in the tests by running:
 
 ```shell
-gulp watch-tests
+yarn watch
```
Benjin, line 39: does this need to be run from within the tests subfolder folder, or does the overall watcher now include /tests/?

scripts/extract-localization.js
aasimkhan30: Removing these icons as the prefix icon is already handled by logger class.

test/e2e/README.md
aasimkhan30: The root-level watch task now compiles both the extension code and the tests. Not sure why they were split in the Gulp file. `ext:compile-tests` was never run on its own, and the code and tests are pretty tightly coupled anyway.

package.json
aasimkhan30: on windows it was considering `test;` as a single command.

localization/README.md
aasimkhan30: Decided to not port it as it wasn't being used.

scripts/copy-assets.js
aasimkhan30: I am going to do a follow up PR where everything is moved in /media/* and will not be copied to `out`. The code will change to just refer things from `extension-root/media/file-name`

/src/reactviews/media/* is a bit different as those files get bundled by esbuild which handles the paths remapping.

Benjin: Sounds good.  Just be sure to add this context to the applicable readme files!

PR #19845, opened 2025-07-26
Bundling extension with esbuild

aasimkhan30:

## Description

## Code Changes Checklist


scripts/copy-assets.js
```diff
@@ -8,26 +8,11 @@ const path = require("path");
 const logger = require("./terminal-logger");
 
 const COPY_TASKS = [
-    {
```
aasimkhan30, line 11: We don't have to copy these resources as it is moved to extension root.

src/languageservice/decompressProvider.ts
```diff
@@ -4,29 +4,104 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as DecompressTar from "tar";
-import * as DecompressZip from "decompress-zip";
```
aasimkhan30, line 7: decompress-zip is an old package that was having trouble with esbuild. I have replaced it with yauzl which is a new actively maintained un-zip lib.

typings/vscode-mssql.d.ts
```diff
@@ -1235,11 +1195,7 @@ declare module "vscode-mssql" {
         getLocations(session: IAzureAccountSession): Promise<azure.subscription.Location[]>;
     }
 
-    export const enum TaskExecutionMode {
```
aasimkhan30, line 1238: removing const enums from typings files as eslint ignores those and doesn't bundle it in the out file.

package.json
```diff
@@ -153,15 +155,14 @@
         "@microsoft/ads-extension-telemetry": "^3.0.2",
         "@microsoft/vscode-azext-azureauth": "^4.1.1",
         "axios": "^1.8.2",
-        "decompress-zip": "^0.2.2",
```
aasimkhan30, line 156: Replace with yauzl.

package.json
```diff
@@ -153,15 +155,14 @@
...
         "dotenv": "^16.4.5",
         "error-ex": "^1.3.0",
         "figures": "^1.4.0",
-        "find-remove": "1.2.1",
+        "find-remove": "5.1.1",
```
aasimkhan30, line 161: Upgraded to latest to work with esbuild

package.json
```diff
@@ -153,15 +155,14 @@
...
         "has-proto": "^1.2.0",
         "jquery": "^3.4.1",
         "lockfile": "1.0.4",
-        "opener": "1.4.2",
+        "opener": "1.5.2",
```
aasimkhan30, line 165: Upgraded to latest to work with esbuild

.vscodeignore
```diff
@@ -36,4 +36,6 @@ eslint
 eslint.config.mjs
 KERBEROS_HELP.md
 prettier.config.mjs
-scripts
\ No newline at end of file
+scripts
+out
+node_modules/**
```
aasimkhan30, line 41: Removing out and node_modules as we don't need those in bundled vsix.

.gitignore
```diff
@@ -31,3 +31,4 @@ webviews-metafile.json
 .eslintcache
 .nyc_output
 samples/connection-sharing-sample/dist
+dist
```
aasimkhan30: out contains tsc compiled code for tests where as dist contains actual bundled code that runs.

aasimkhan30:
Will fix web view bundle size comment in a separate PR to avoid checks failing.

.gitignore
Benjin: Good note to add as a comment in the file

scripts/build.js
```diff
@@ -0,0 +1,13 @@
+const { execSync } = require('child_process');
+
+const isProd = process.argv.includes('--prod');
+const prodArg = isProd ? 'prod' : '';
+
+try {
+  execSync('yarn build:prepare', { stdio: 'inherit' });
+  execSync('yarn build:extension', { stdio: 'inherit' });
+  execSync(`yarn build:extension-bundle --${prodArg}`, { stdio: 'inherit' });
```
Benjin: non-prod means this'll have just `--`, which is a bit odd (but not problematic).  Why not have `const prodArg = isProd? '--prod' : '';` instead?

scripts/package-extension.js
```diff
@@ -10,8 +10,8 @@ const del = require('del');
 const logger = require('./terminal-logger');
 
 const args = process.argv.slice(2);
-const isOnline = args.includes('--online');
-const isOffline = args.includes('--offline');
+let isOnline = args.includes('--online');
+let isOffline = args.includes('--offline');
```
Benjin: Did these need to change?

src/models/utils.ts
```diff
@@ -5,7 +5,7 @@
 
 import * as os from "os";
 import * as path from "path";
-import * as findRemoveSync from "find-remove";
+const findRemoveSync = require("find-remove");
```
Benjin: why the change in import style?

src/reactviews/pages/SchemaCompare/SchemaCompareStateProvider.tsx
```diff
@@ -151,8 +151,8 @@ const SchemaCompareStateProvider: React.FC<SchemaCompareStateProviderProps> = ({
                 },
                 publishProjectChanges: function (
                     targetProjectPath: string,
-                    targetFolderStructure: mssql.ExtractTarget,
-                    taskExecutionMode: mssql.TaskExecutionMode,
+                    targetFolderStructure: sc.ExtractTarget,
```
Benjin, line 154: Can you remind me again why the enums had to move from the types file to the sharedInterfaces file - something about esbuild not liking `export const enum`?  This'll make it more difficult if anyone wants to build on top of MSSQL/STS, and it's been nice to have all STS contract types in the same spot, IMO.

test/e2e/executionPlan.spec.ts
```diff
@@ -129,7 +129,7 @@ test.describe("MSSQL Extension - Query Plan", async () => {
         const customZoomInput = iframe.locator("#customZoomInputBox");
         await expect(customZoomInput).toBeVisible();
 
-        await customZoomInput.fill("25");
+        await customZoomInput.fill((currentZoom - 5).toString());
```
Benjin, line 132: Unexpected change; why was this necessary?

test/unit/extension.test.ts
```diff
@@ -105,7 +105,7 @@ suite("Extension API Tests", () => {
         );
     });
 
-    test("connect", async () => {
+    test.skip("connect", async () => {
```
Benjin, line 113: Add comments for why these are skipped.  Are they obsolete, or just more work to get passing with these changes (and will be lit up soon)?

typings/vscode-mssql.d.ts
Benjin: Instead of moving these out, could a const object work instead?

```
export const TaskExecutionMode = {
  Execute: 0,
  Script: 1,
  ExecuteAndScript: 2,
} as const;
type TaskExecutionMode = typeof TaskExecutionMode[keyof typeof TaskExecutionMode];
```

Benjin (approved):

scripts/package-extension.js
aasimkhan30: We actually override the variable later 
https://github.com/microsoft/vscode-mssql/blob/784d314302a6c9ae9f8ec90ac2aa87c44deac6c8/scripts/package-extension.js#L230

test/e2e/executionPlan.spec.ts
aasimkhan30: I was running this on my local devbox and the initial zoom was always a little different depending on the diagram and the screen size. (Since we try to fit all nodes in the screen). Therefore, setting the zoom to a constant value meant it can be bigger or smaller than current zoom. To fix that flakiness, I changed the test to always set the zoom to be lesser than the current zoom. In that way we can check the custom zoom functionality.

test/unit/extension.test.ts
aasimkhan30: Added a comment about it. 

```
/**
     * Since the runtime for the extension is using the esbuild version, we cannot correctly mock or
     * spy on the connect method of the MainController as the tests are executed on the tsc version.
     * For this reason, we skip this test for now.
     */
 ```

Benjin:
Can you update the PR description to explain the swap to esbuild?

typings/vscode-mssql.d.ts
Benjin: It doesn't. :( Classes (and enums) are durable through the tsc phase, which is why interfaces are okay here, but classes and enums aren't.

src/reactviews/pages/SchemaCompare/SchemaCompareStateProvider.tsx
Benjin: Discussed offline (and captured in another PR thread)

Benjin (dismissed):

Benjin (approved):

Issue #18912, opened 2025-03-07
New Feature Request - "Default" Connections


yorek:
Having to select the connection against which to run a query after opening a `.sql` file is quite repetitive, It would be great if I could set a default connection for my project. For example, I could have a file "mssql.json" in the ".vscode" folder where I can set my default connection that will be used automatically and without ask when I click or the "Run" button. Then there could be another "Run" button, called "Run with ad-hoc connection" that would work is it is today, asking me which connection to use before running the query.

Here's an idea:

![Image](https://github.com/user-attachments/assets/73c6892c-8078-4d71-8917-c52a196917ba)

Galaeci:
It could even reuse the active connection with the current file, so no need to have extra files or settings. 
Another helpful feature would be to reuse the same database when changing connections if the connection profile is set to default and not to a particular database. Instead of going to master (in most cases) default should pick the database from the current connection profile and if that is not available use the default for the login. This would greatly improve the usability when working across multiple environments: dev, test, prod

Galaeci:
No need for a separate json file, right? This could be done as part of a workspace file. Still having this setting is not incompatible with the other behavior that I was advocating for. Also, re-using the active connection works better for projects where you are using multiple databases at the same time.
The important thing here is to limit the need to define a connection on each file open or creation. Azure Data Studio already supported some of this behavior where a new sql query would reuse the connection of the active one. That was a huge time saver for heavy db users.

yorek:
Yes, I guess it would make sense to use the workspace file if that is possible.

kburtram:
Please consider the details from https://github.com/microsoft/vscode-mssql/issues/18498 when resolving this issue.  Specifically, the "default connection" mechanism should consider connection settings from the environment variables, if applicable.

PR #19870, opened 2025-08-02
Added support for Default connection (#18912)

yorek:
## Description

Added support for a workspace default connection as requested in #18912. The default connection is saved in the `defaultConnectionName` setting:

```json
{
    "mssql.defaultConnectionName": "YourConnectionProfileName"
}
```

## Code Changes Checklist


sggoose:
Been hoping this could be a feature! Would you be able to modify this to always set the default connection to the last manually connected profile in a workspace? This would most accurately reflect the default connection behavior of SSMS in my opinion.

yorek:
> Been hoping this could be a feature! Would you be able to modify this to always set the default connection to the last manually connected profile in a workspace? This would most accurately reflect the default connection behavior of SSMS in my opinion.

That could be set as an option that can be turned on or off. Together with a right-click menu on a connection to make it as default. Let's have this first improvement done and then we can do this other two in the next, so that we move swiftly forward

yorek:
I have added the context menu item to set a connection as default for the Workspace, to make the UX better

Benjin:
An interesting thought; we'll discuss taking this.  Not knowing if we will end up pulling this in, I'll still leave some PR comments to make it more merge-ready if we do.

package.json
```diff
@@ -1224,6 +1235,11 @@
                     "description": "%mssql.maxRecentConnections%",
                     "scope": "window"
                 },
+                "mssql.defaultConnectionName": {
+                    "type": "string",
+                    "description": "Name of the default connection profile to use when executing queries without an active connection. This should match the name of a saved connection profile.",
```
Benjin: the default connection should be the ID (a GUID), not the profileName.  Saved connections are not guaranteed to have a user-specified profile name.

`mssql.defaultConnectionId` would therefore be a better config value.

src/controllers/connectionManager.ts
```diff
@@ -1051,6 +1051,125 @@ export default class ConnectionManager {
...
+    ): Promise<IConnectionInfo | undefined> {
+        // Get configuration from the workspace
+        const config = this.vscodeWrapper.getConfiguration(
+            Constants.extensionConfigSectionName,
+            fileUri,
+        );
+        const defaultConnectionName = config.get<string>(Constants.configDefaultConnectionName);
+
+        // Debug logging
+        this.vscodeWrapper.logToOutputChannel(
```
Benjin: Better to use `this._logger` where available so that the output messages can be appropriately tagged in the output pane.

src/controllers/connectionManager.ts
```diff
@@ -1051,6 +1051,125 @@ export default class ConnectionManager {
...
+            LocalizedConstants.msgDefaultConnectionNameFromConfig(defaultConnectionName),
+        );
+
+        if (!defaultConnectionName) {
+            this.vscodeWrapper.logToOutputChannel(
+                LocalizedConstants.msgNoDefaultConnectionNameConfigured,
+            );
+            return undefined;
+        }
```
Benjin: Cleaner output would be for this to be checked before the previous prints.

src/controllers/connectionManager.ts
```diff
@@ -1051,6 +1051,125 @@ export default class ConnectionManager {
...
+
+        // Get saved connections to find the one with the matching name
+        const connectionProfileList = await this._connectionStore.getPickListItems();
```
Benjin: `connectionConfig.getConnectionById()` would be a more direct way to look this up once you switch to the ID as the identity.

src/controllers/mainController.ts
```diff
@@ -2068,10 +2115,24 @@ export default class MainController implements vscode.Disposable {
...
+        Utils.logDebug(`checkIsReadyToExecuteQuery called for ${uri}`);
+
+        if (this._connectionMgr.isConnected(uri)) {
+            Utils.logDebug(`Connection already established in ${uri}`);
+            return true;
+        }
+
+        Utils.logDebug(`Not connected, trying default connection`);
+        // Try to connect with default connection from configuration
+        const defaultConnectionResult = await this._connectionMgr.connectWithDefaultConnection(uri);
```
Benjin, line 2129: Does this work with other the automatic connection selection mechanisms we already have?  i.e. if you have a connection selected in the Object Explorer, that connection is selected; if you already have a connected SQL file and open another one, that connection is copied to the new file.

package.json
```diff
@@ -443,6 +443,11 @@
                     "when": "view == objectExplorer && !config.mssql.useLegacyConnectionExperience && viewItem =~ /\\btype=(disconnectedServer|Server)\\b/",
                     "group": "1_MSSQL_serverAndDbManagement@2"
                 },
+                {
+                    "command": "mssql.setAsWorkspaceDefault",
```
kburtram, line 450: I'm not sure this command is common enough to justify adding it to the Server node's context menu in Object Explorer. Since there are already quite a few server/database commands, adding more could clutter the menu. An alternative UX might be to expose this via the command palette, using a quick pick menu to select the server.

Benjin: We could also reorganize the our menu items now that VS Code supports contributing submenus.  We may want to do that eventually anyway as we bring in more functionality from ADS.

kburtram: Yes, we’ll likely need to reorganize our context menus as we introduce more commands. That said, we should continue to keep the menus focused on the most frequently used actions, and move less common commands to the Command Palette to reduce clutter and keep the UI clean.

package.json
yorek: while I understand the ratio behind this request, using an ID make the `settings.json` less human-friendly. If I want to see what is the default connection by opening `settings.json` and then I just see an ID, that doesn't help much. Any idea on how to make this better while still human friendly? Are names unique or they can be duplicated?

src/controllers/mainController.ts
yorek: At the moment if I have a connection selected in the object explorer I'm still asked to pick a connection when I try to run a query, if the opened .sql files isn't connected to a server yet. The default connection will be used only if the .sql doesn't have an active connection already

package.json
yorek: May I can make a change to have the default connection highlighted in bold to improve the UI

yorek: Apparently bold is not possible, so I added the "(default)" description.

kburtram:
Based on recent discussions, there are two primary concerns with this change.
First, it introduces a context menu to the connection node for all users, but it’s unclear how common this operation will be. The recommendation is to consider exposing it through the Command Palette instead.
Second, there is ongoing and planned work around default connections in the product. It’s not yet clear how this setting fits within that broader context or whether it may override other, more relevant default connection logic in some scenarios. We’ll need to review the overall “default connection” experience to ensure this aligns with the larger connectivity specification.

Given these concerns, we won’t be able to merge this PR as-is. Please reach out to me or @croblesm if you’d like to discuss how best to address them. Closing this PR for now as part of the repo cleanup of older inactive items, but we can reopen it once the issues are resolved.

Issue #18992, opened 2025-03-24
Recent connections is filling in <default> with master


kburtram:
Steps to Reproduce:

1. Connect to a server using default db and save it

Expected: Both saved and recent connection are to the "default" database

Actual: Recent gets populated with the actual current database, then using "default".  Since the default database can change, this is not actually the same thing and isn't what was specified in the user dialog.

![Image](https://github.com/user-attachments/assets/a3bad7d9-41c3-4fc4-820e-b005bed010b3)

allancascante:
Closing as work was merged in: https://github.com/microsoft/vscode-mssql/pull/19934

PR #19883, opened 2025-08-05
Fix for bug: 18992

allancascante:
Fix for #18992 

When unnamed connections made without a database name made it to the recent connections list, they show the database they connected to on the time of connection, while default database should be shown. 

Existing Recent Connections:

image

With the Change to display default database:

image

## Code Changes Checklist


allancascante:
@microsoft-github-policy-service agree company="Microsoft"

Benjin:
Best to use a more descriptive title.  Once this is merged, we won't want to look up what the specific bug is.  Maybe something like `Fixes bug where default database is replaced by actual database in MRU connections`

src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -87,6 +87,10 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
 
     public readonly initialized: Deferred<void> = new Deferred<void>();
 
+    /// <summary>
+    /// The main options for the connection dialog. The names here correspond to the properties in the connection profile.
+    /// If a name is added here, it is going to be rendered in the main options section of the connection dialog.
+    /// </summary>
```
Benjin, line 93: This is a C#-style docstring.  In Typescript, type `/**` and it should autocomplete.  If not, take a look at other examples in the codebase.

typings/vscode-mssql.d.ts
```diff
@@ -446,6 +446,11 @@ declare module "vscode-mssql" {
          * connection is not hosted by a container
          */
         containerName: string | undefined;
+
+        /**
+         * True if the connections doesn't use a database name, false if it does.
+         */
+        defaultDatabase: boolean | undefined;
```
Benjin, line 453: My concern with adding a separate property is that it could get out of sync with other dependent properties (namely `database`) and that could lead to confusing behavior.  E.g., what happens if I set `defaultDatabase` to true, but also set `database` to a specific database?

In theory, checking if  `database` is `undefined` or `""` in the right places should do the right thing.

Benjin:
Where does the MRU connection get saved?  Seems to me that the issue is that the MRU connection is getting saved with the actual database when it shouldn't be, not that we aren't tracking whether a database has been set (because we already are, via the `database` property).

typings/vscode-mssql.d.ts
allancascante: I added the code in [here](https://github.com/microsoft/vscode-mssql/pull/19883/files/745c678cb38f4f9839fe00e696a53112b0bcf776#diff-aeaa81c514c4cf62e0ce4f8311c6e991804b83f291765b1966b71b09ac20c8b3R276) to update the value as the store saves the connections.

src/connectionconfig/connectionDialogWebviewController.ts
allancascante: done

typings/vscode-mssql.d.ts
Benjin: That looks like the spot where it saves the connection to the config JSON, not where it's getting saved to the MRU list (see `addRecentlyUsed()` in the same file).  Is there a reason that we need to persist defaultDatabase as a separate property instead of just checking the `database` prop?

allancascante: When I was debugging this, I found that addRecentlyUsed() is called from tryAddMruConnection in connectionManager, and this happens when a saved connection successfully connects. The connection object used in the process of connecting, comes from the saved list and includes the defaultDatabase field with its initialized value (true or false). While the object also contains the actual database it connected to, the defaultDatabase flag indicates whether the database name was not explicitly set in the original connection configuration—this is what we consider a "default" database. That distinction isn’t always inferable from the database property alone as the database field itself is set by the JRPC connect call during the connection process; this alone is the reason for the problem.

allancascante: In short Create Connection -> Saved Connection -> Connect -> tryAddMruConnection in that flow when saving a connection, if database name is null or empty -> defaultDatabase = true. AFAIK that's the only flow for connections to make into recent connections.

PR #19885, opened 2025-08-05
Improvement (not fix) for Bug #18837:  Query Cancellation and Disposal Now Handled Correctly

samgusick:

## Description

Issue #18837. 

- Ensures consistent state reset and cleanup after errors during query cancellation and disposal. 
- Removes references to non-existent runner unregistration. 
- Enhances feedback and event emission on failures to prevent orphaned promises and incomplete executions. 
- Multiple times a day I need to restart VS Code because this issue randomly breaks the editor and query execution, making it one of the biggest turn offs to using the extension currently. This should make considerable progress towards fixing the issue.

## Code Changes Checklist

### DIFF
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -160,41 +160,10 @@ export default class QueryRunner {
                 cancelParams,
             );
         } catch (error) {
-            // Ensure state is reset on error
-            this._isExecuting = false;
-            this._hasCompleted = true;
-            this.removeRunningQuery();
-            // Removed call to unregisterRunner (does not exist)
-            const promise = this._uriToQueryPromiseMap.get(this._ownerUri);
-            if (promise) {
-                promise.reject(error);
-                this._uriToQueryPromiseMap.delete(this._ownerUri);
-            }
-            this.eventEmitter.emit(
-                "complete",
-                Utils.parseNumAsTimeString(this._totalElapsedMilliseconds),
-                true,
-            );
-            this._statusView.executedQuery(this._ownerUri);
-            this._vscodeWrapper.showErrorMessage("Cancel failed: " + getErrorMessage(error));
+            this._handleCancelDisposeCleanup("Cancel failed: ", error);
             return;
         }
-        // Always reset state after cancel
-        this._isExecuting = false;
-        this._hasCompleted = true;
-        this.removeRunningQuery();
-        // Removed call to unregisterRunner (does not exist)
-        const promise = this._uriToQueryPromiseMap.get(this._ownerUri);
-        if (promise) {
-            promise.resolve();
-            this._uriToQueryPromiseMap.delete(this._ownerUri);
-        }
-        this.eventEmitter.emit(
-            "complete",
-            Utils.parseNumAsTimeString(this._totalElapsedMilliseconds),
-            true,
-        );
-        this._statusView.executedQuery(this._ownerUri);
+        this._handleCancelDisposeCleanup();
         return queryCancelResult;
     }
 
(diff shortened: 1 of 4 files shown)


samgusick:

src/controllers/queryRunner.ts
```diff
@@ -152,10 +152,48 @@ export default class QueryRunner {
...
+            }
+            this.eventEmitter.emit(
+                "complete",
+                Utils.parseNumAsTimeString(this._totalElapsedMilliseconds),
+                true,
+            );
+            this._statusView.executedQuery(this._ownerUri);
+            this._vscodeWrapper.showErrorMessage("Cancel failed: " + error.message);
+            return;
+        }
```
aasimkhan30, line 168: Can we create a new method and move all the repeated code to that method?

Something like:

`completeExecution(error?: Error)`

src/controllers/queryRunner.ts
```diff
@@ -152,10 +152,48 @@ export default class QueryRunner {
...
+                promise.reject(error);
+                this._uriToQueryPromiseMap.delete(this._ownerUri);
+            }
+            this.eventEmitter.emit(
+                "complete",
+                Utils.parseNumAsTimeString(this._totalElapsedMilliseconds),
+                true,
+            );
+            this._statusView.executedQuery(this._ownerUri);
+            this._vscodeWrapper.showErrorMessage("Cancel failed: " + error.message);
```
aasimkhan30: use getErrorMessage from https://github.com/microsoft/vscode-mssql/blob/main/src/utils/utils.ts

src/controllers/queryRunner.ts
```diff
@@ -237,7 +281,7 @@ export default class QueryRunner {
         this._totalElapsedMilliseconds = 0;
         this._statusView.executingQuery(this.uri);
 
-        let onSuccess = (result) => {
+        let onSuccess = (_result: unknown) => {
```
aasimkhan30, line 257: Remove the unused var 

```
 let onSuccess = () => {

**Comment by aasimkhan30** [comment]
@samgusick , thanks for fixing the issues with cancel and dispose. The core logic looks good just needs some code cleanup.

File: src/controllers/queryRunner.ts [hunk 4]
```diff
@@ -248,15 +292,40 @@ export default class QueryRunner {
...
+            }
+            this.eventEmitter.emit(
+                "complete",
+                Utils.parseNumAsTimeString(this._totalElapsedMilliseconds),
+                true,
+            );
             // TODO: localize
-            this._vscodeWrapper.showErrorMessage("Execution failed: " + error.message);
+            let errorMsg = error instanceof Error ? error.message : String(error);
+            this._vscodeWrapper.showErrorMessage("Execution failed: " + errorMsg);
```
Benjin, line 286: While you're here, could you localize this string?

src/controllers/queryRunner.ts
samgusick: Should I use it anytime an error is passed to this method?

aasimkhan30: We use it to safely extract a display message from an error.

src/controllers/queryRunner.ts
```diff
@@ -95,6 +98,8 @@ export default class QueryRunner {
         this._isExecuting = false;
         this._totalElapsedMilliseconds = 0;
         this._hasCompleted = false;
+        this._cancelCleanupPrefix = vscode.l10n.t("Cancel failed: ");
```
aasimkhan30: Could you move this string to [locConstants.ts](https://github.com/microsoft/vscode-mssql/blob/dc925df6a410872b7182fb183043fde53ae1df8f/src/constants/locConstants.ts#L1487)](https://github.com/microsoft/vscode-mssql/blob/dc925df6a410872b7182fb183043fde53ae1df8f/src/constants/locConstants.ts#L1487)?


Also, please use `l10n.t` with arguments, like this:

```ts
public static queryCancelFailed(errorMessage: string) {
    return l10n.t({
        message: "Cancel failed: {0}",
        args: [errorMessage],
        comment: ["{0} is the error message"]
    });
}
```

samgusick:
After more extensive testing this doesn't seem to remove the issue entirely, though it appears to happen noticeably less frequently. If this gets merged, I'll continue to see if there's anything else that could be done to mitigate the issue.

aasimkhan30:
@samgusick , can you please let me know the specific repro steps for the issue? Also if it doesn't fix the issue I am going to edit your description to only mention the issue instead of 

`Fix for Bug https://github.com/microsoft/vscode-mssql/issues/18837.`

As it will close the original issue.

samgusick:
@aasimkhan30 I haven't figured out an exact way to reproduce the issue, it just starts happening after I've been running queries for a few hours. Seems random almost.

aasimkhan30 (approved):

PR #19916, opened 2025-08-08
New setting for connection groups to collapse/expand on load by default in the UI

samgusick:

## Description

This pull request implements a new MSSQL setting that, when enabled, makes the connection groups load expanded in the UI by default (which is the only option currently) but disabling the setting makes all connection groups instead load collapsed in the UI by default.

As someone who uses connection groups to organize databases from different projects, there are many connections I don't need to see while working in a particular workspace. It's a minor nuisance, but I have to collapse all of the connection groups I don't need everytime I open vs code, so I figured having the option to load all of the connection groups collapsed would mean I only ever have to expand the 1 or 2 connection groups I need. 

Additionally, when many connections and connection groups exist, they take up space in the editor window.

## Code Changes Checklist

### DIFF
--- a/src/objectExplorer/objectExplorerService.ts
+++ b/src/objectExplorer/objectExplorerService.ts
@@ -412,9 +412,19 @@ export class ObjectExplorerService {
         const newConnectionGroupNodes = new Map<string, ConnectionGroupNode>();
         const newConnectionNodes = new Map<string, ConnectionNode>();
 
+        // Read user setting for default expansion state
+        const config = this._vscodeWrapper.getConfiguration("mssql");
+        const expandGroupsByDefault = config.get<boolean>(
+            "connectionGroupsExpandedByDefault",
+            true,
+        );
+        const defaultCollapsibleState = expandGroupsByDefault
+            ? vscode.TreeItemCollapsibleState.Expanded
+            : vscode.TreeItemCollapsibleState.Collapsed;
+
         // Add all group nodes from settings first
         for (const group of serverGroups) {
-            const groupNode = new ConnectionGroupNode(group);
+            const groupNode = new ConnectionGroupNode(group, defaultCollapsibleState);
 
             if (this._connectionGroupNodes.has(group.id)) {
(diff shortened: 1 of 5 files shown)


src/objectExplorer/objectExplorerService.ts
```diff
@@ -412,9 +412,19 @@ export class ObjectExplorerService {
         const newConnectionGroupNodes = new Map<string, ConnectionGroupNode>();
         const newConnectionNodes = new Map<string, ConnectionNode>();
 
+        // Read user setting for default expansion state
+        const config = this._vscodeWrapper.getConfiguration(Constants.extensionConfigSectionName);
+        const expandGroupsByDefault = config.get<boolean>(
+            "connectionGroupsExpandedByDefault",
```
aasimkhan30: Can you move this constants.ts file?
https://github.com/microsoft/vscode-mssql/blob/main/src/constants/constants.ts.
For example: 
https://github.com/microsoft/vscode-mssql/blob/5cc8621043de3e876d09f52a950f3e4c8c766f77/src/constants/constants.ts#L237

aasimkhan30 (changes requested):

aasimkhan30:
Overall, the changes look good to me except for the hardcoded config name string. Assigning to @Benjin since he owns this area.

samgusick:
Having trouble understanding why these unit tests are failing now
image
image

localization/LCL/de/vscode-mssql.xlf.lcl
```diff
@@ -9589,6 +9589,15 @@
         </Str>
         <Disp Icon="Str" />
       </Item>
+      <Item ItemId=";package.mssql.schemaDesigner.enableExpandCollapseButtons" ItemType="0" PsrId="308" Leaf="true">
```
Benjin: Merge the latest from main; these localization deltas should disappear.

Benjin (approved):

samgusick:
Created a new pull request with all corrections and tests passing to simplify the branch history and make file change comparison easy and more clear.

PR #19925, opened 2025-08-08
Fixing password handling logic and adding temp session password for unsaved password conns. 

aasimkhan30:
## Description
1. Using conn ids to store and fetch password. To reduce conflicts between profiles that have very similar creds 
2. Migrating old ids to new ids when a conn is successfully established using old ids
3. Using a temp session cache to store passwords for connections with no saved password(savePassword = false) to prevent the extension asking passwords for those connections when they have been succesfully connected before. This is cleared after vscode exit. Fixes 

Removes
## Code Changes Checklist

### DIFF
--- a/src/models/connectionStore.ts
+++ b/src/models/connectionStore.ts
@@ -33,6 +33,11 @@ import { Deferred } from "../protocol";
  * @class ConnectionStore
  */
 export class ConnectionStore {
+    /**
+     * Map to store password for connections where savePassword is false for the session.
+     * For connections with savePassword = true, the password will be saved to the credential
+     * store.
+     */
     private _sessionPasswords: Map<string, string> = new Map();
     constructor(
         private _context: vscode.ExtensionContext,
@@ -329,15 +334,6 @@ export class ConnectionStore {
     }
 
     /**
-     * Clear session password for a connection
-     * @param connectionCredentials Connection credentials
-     */
-    public clearSessionPassword(connectionCredentials: IConnectionInfo): void {
-        const sessionKey = this.getSessionPasswordKey(connectionCredentials);
-        this._sessionPasswords.delete(sessionKey);
-    }
-
-    /**
      * Generate a session key for password storage
      * @param connectionCredentials Connection credentials
      * @returns Session storage key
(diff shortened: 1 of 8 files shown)


src/models/connectionStore.ts
```diff
@@ -222,25 +261,62 @@ export class ConnectionStore {
...
+
+        // First, try to get password from session storage for non-saved passwords
+        if (!profile.savePassword && ConnectionStore.shouldSavePasswordUntilRestart) {
+            const sessionPassword = this._sessionPasswords.get(credentialId);
+            if (sessionPassword) {
+                return sessionPassword;
+            }
+        }
+
+        // Read from credential store with new format id
```
Benjin: "If not found, read from..." (it's nice for comments to read like they're branches to logic instead of sequential steps when there's a likely early bailout).

src/models/connectionStore.ts
```diff
@@ -222,25 +261,62 @@ export class ConnectionStore {
     }
 
     /**
-     * Lookup credential store
+     * Lookup credential store with migration support
```
Benjin: describe what the migration is so that future-us can better decide when it's okay to remove migration support.

src/models/connectionStore.ts
```diff
@@ -103,10 +126,26 @@ export class ConnectionStore {
...
+        ConnectionStore.pushIfNonEmpty(profileId, ConnectionStore.CRED_PROFILE_PREFIX, cred);
+        ConnectionStore.pushIfNonEmpty(itemType, ConnectionStore.CRED_ITEMTYPE_PREFIX, cred);
+        return cred.join(ConnectionStore.CRED_SEPARATOR);
+    }
+
     /**
      * Creates a formatted credential usable for uniquely identifying a SQL Connection.
      * This string can be decoded but is not optimized for this.
-     * @static
+     * @deprecated
```
Benjin: add deprecation note indicating that callers should use `formatCredentialIdFromProfileId` instead

src/models/connectionStore.ts
```diff
@@ -448,34 +521,39 @@ export class ConnectionStore {
...
-            } else {
-                resolve(true);
+
+        if (Utils.isEmpty(password)) {
+            return true;
+        }
+
+        const profile = conn as IConnectionProfile;
+
+        let legacyCredentialId = ConnectionStore.formatCredentialId(
```
Benjin, line 550: Probably should capture some telemetry about whether the old credential key found something so that we can inform a decision about when to drop support for the old one.

Benjin (approved):

src/models/connectionStore.ts
aasimkhan30: Updating the comment as previously I was migrating here but now it only happens if there was a successful connection.

PR #19929, opened 2025-08-11
Not copying connections for sql files opened from internal extension …

aasimkhan30:
## Description

A few releases ago, we introduced a feature that automatically copied the SQL connection from the last active file to any newly opened SQL file. This saved users from repeatedly connecting new files to the same SQL Server they were already working with.
[https://github.com/microsoft/vscode-mssql/pull/19503](https://github.com/microsoft/vscode-mssql/pull/19503)

However, this caused problems for SQL files opened within the extension itself. Many of these files are already connected when opened. For example, in Object Explorer (OE), scripting an object opens an untitled editor with the corresponding CREATE/UPDATE/DELETE script and uses the object’s parent connection. The new feature would overwrite this with the last active connection, which could be different, resulting in errors when the script referenced objects that didn’t exist in that server.

In some cases, the issue was even worse: the editor would be connected to two different servers simultaneously
1. one from the scripting logic 
2. another from the “copy last active connection” logic

creating a race condition over which connection would be active.

While the example above uses scripting, this bug also affects other untitled SQL editors opened from features like Table Designer, Schema Designer, and more.

Fixes: #19930

This PR adds a temporary workaround that tracks a set of URIs where the “copy last active connection” feature should be skipped. If a URI is in this set, the last active connection will not be applied to it.

## Code Changes Checklist

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -907,7 +907,7 @@ export default class MainController implements vscode.Disposable {
         // Handle case where SQL file is the 1st opened document
         const activeTextEditor = this._vscodeWrapper.activeTextEditor;
         if (activeTextEditor && this._vscodeWrapper.isEditingSqlFile) {
-            this.onDidOpenTextDocument(activeTextEditor.document);
+            await this.onDidOpenTextDocument(activeTextEditor.document);
         }
         await this.sanitizeConnectionProfiles();
         await this.loadTokenCache();
@@ -2553,14 +2553,21 @@ export default class MainController implements vscode.Disposable {
      * Called by VS Code when a text document is opened. Checks if a SQL file was opened
      * to enable features of our extension for the document.
      */
-    public onDidOpenTextDocument(doc: vscode.TextDocument): void {
+    public async onDidOpenTextDocument(doc: vscode.TextDocument): Promise<void> {
         if (this._connectionMgr === undefined) {
             // Avoid processing events before initialization is complete
             return;
         }
         this._connectionMgr.onDidOpenTextDocument(doc);
 
-        if (this._previousActiveDocument && doc.languageId === Constants.languageId) {
+        const isOpenedFromUntitledService =
+            await this._untitledSqlDocumentService.isUriOpenedFromService(doc.uri.toString(true));
+
+        if (
+            this._previousActiveDocument &&
+            doc.languageId === Constants.languageId &&
+            !isOpenedFromUntitledService
+        ) {
             void this._connectionMgr.copyConnectionToFile(
                 this._previousActiveDocument.uri.toString(true),
                 doc.uri.toString(true),
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -2363,7 +2363,7 @@ export default class MainController implements vscode.Disposable {
         const currentDocUri = vscode.window.activeTextEditor
             ? vscode.window.activeTextEditor.document.uri.toString(true)
             : undefined;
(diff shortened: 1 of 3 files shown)


src/controllers/mainController.ts
```diff
@@ -2553,14 +2553,22 @@ export default class MainController implements vscode.Disposable {
...
+    public async onDidOpenTextDocument(doc: vscode.TextDocument): Promise<void> {
         if (this._connectionMgr === undefined) {
             // Avoid processing events before initialization is complete
             return;
         }
         this._connectionMgr.onDidOpenTextDocument(doc);
 
-        if (this._previousActiveDocument && doc.languageId === Constants.languageId) {
+        await this.untitledSqlDocumentService.waitForAllOperations();
+        const isOpenedFromUntitledService =
```
Benjin: instead of tracking this way, what if the untitledDocumentService sets the target connection in the map?  Then this function could check the map to retrieve the intended connection, and only transfer if none was found.

I _think_ that would avoid the funky `waitForAllOperations()` call.

Benjin: Never mind; still would hit the same race conditions.

src/controllers/mainController.ts
```diff
@@ -2553,14 +2553,22 @@ export default class MainController implements vscode.Disposable {
...
+            await this._untitledSqlDocumentService.isUriTrackedByService(doc.uri.toString(true));
+
+        if (
+            this._previousActiveDocument &&
+            doc.languageId === Constants.languageId &&
+            !isOpenedFromUntitledService
```
Benjin: Add comment explaining why it matters if it's tracked by untitled service

src/controllers/mainController.ts
```diff
@@ -2553,14 +2553,22 @@ export default class MainController implements vscode.Disposable {
...
-    public onDidOpenTextDocument(doc: vscode.TextDocument): void {
+    public async onDidOpenTextDocument(doc: vscode.TextDocument): Promise<void> {
         if (this._connectionMgr === undefined) {
             // Avoid processing events before initialization is complete
             return;
         }
         this._connectionMgr.onDidOpenTextDocument(doc);
 
-        if (this._previousActiveDocument && doc.languageId === Constants.languageId) {
+        await this.untitledSqlDocumentService.waitForAllOperations();
```
Benjin: comment explaining what ops need to be awaited

src/controllers/untitledSqlDocumentService.ts
```diff
@@ -9,22 +9,78 @@ import VscodeWrapper from "./vscodeWrapper";
...
-        // Open an untitled document. So the  file doesn't have to exist in disk
-        let doc = await this.vscodeWrapper.openMsSqlTextDocument(content);
-        // Show the new untitled document in the editor's first tab and change the focus to it.
+    public async newQuery(content?: string, trackUri: boolean = true): Promise<vscode.TextEditor> {
+        // Create a unique key for this operation to handle potential duplicates
+        const operationKey = `${Date.now()}-${Math.random()}`;
+
+        try {
+            const operationPromise = this.createDocument(content, trackUri);
+            this._pendingOperations.set(operationKey, operationPromise);
```
Benjin: Is there any risk of `onDidOpenTextDocument` (and therefore `waitForAllOperations()` getting executed before this is added as a pending operation?  Maybe `createDocument()` should be sync instead of async and should  have its logic wrapped in a Promise that's returned immediately.

lewis-sanchez (approved):

PR #19934, opened 2025-08-11
Setting database to it original value when saving (fix for bug 18992)

allancascante:
## Description

Database value is being updated after a successful connection; however, this is making the items saved to MRU list (recent connections) have a different value than the original saved connections. So it displays the database name it was last connected to; while the value should be <default>. 

SQL Server doesn't have a <default> database, and the actual database it connects to when the database name is not supplied might change, between connections attempts. So recent connections should not save a particular value as it might change later.

Fix for #18992 

## Code Changes Checklist

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -847,7 +847,7 @@ export default class ConnectionManager {
             try {
                 // if the connection was originally for an empty database, the database should be empty when being saved to match the original connection
                 if (saveAsDefaultDatabase) {
-                    connectionToSave.database = undefined;
+                    connectionToSave = { ...connectionToSave, database: undefined };
                 }
                 await this._connectionStore.addRecentlyUsed(connectionToSave);
                 connection.connectHandler(true);
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -580,6 +580,8 @@ export default class ConnectionManager {
 
             let mruConnection: IConnectionInfo = <any>{};
 
+            const saveDefaultDatabase = !connection?.credentials?.database; // connection used empty database for the original connection
+
             if (Utils.isNotEmpty(result.connectionId)) {
                 // Use the original connection information to save the MRU connection.
                 // for connections that a database is not provided, the database information will be updated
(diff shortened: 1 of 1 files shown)


src/controllers/connectionManager.ts
```diff
@@ -643,7 +645,7 @@ export default class ConnectionManager {
                 await self.handleConnectionErrors(fileUri, connection, result);
             }
 
-            await self.tryAddMruConnection(connection, mruConnection);
+            await self.tryAddMruConnection(connection, mruConnection, saveDefaultDatabase);
```
Benjin: It appears to me that the `mruConnection` is only used in the saving path, so I think it should be save to simply remove line 614 (`mruConnection.database = result.connectionSummary.databaseName;`) with no need to add the extra flag.  I don't think we should add an extra flag (with extra cases to handle in the future) unless we have a good idea of why we might need it.

In this case, `mruConnection` appears to only be passed to the `tryAddMruConnection()` method, and within there, it's only used when saved, so it should be safe unless I've missed some side-effect.

src/controllers/connectionManager.ts
```diff
@@ -838,10 +840,15 @@ export default class ConnectionManager {
     private async tryAddMruConnection(
         connection: ConnectionInfo,
         newConnection: IConnectionInfo,
+        saveAsDefaultDatabase: boolean = false,
```
Benjin: while you're in this method, can you rename the `newConnection` param to `mruConnection` and add docstrings that explain what the difference is between the two parameters?

src/controllers/connectionManager.ts
```diff
@@ -580,6 +580,8 @@ export default class ConnectionManager {
 
             let mruConnection: IConnectionInfo = <any>{};
 
+            const saveDefaultDatabase = !connection?.credentials?.database; // connection used empty database for the original connection
+
             if (Utils.isNotEmpty(result.connectionId)) {
                 // Use the original connection information to save the MRU connection.
```
Benjin: this comment should also probably be removed because we're now intentionally tracking saved connections and recent connections separately

Benjin:
Looks like one of the unit tests needs updating with the new behavior.  Lemme know if you need help running them!

src/controllers/connectionManager.ts
allancascante: in that case we can remove the mruConnection as a whole, connection. credentials will have the same values, so we can just remove the variable and the extra parameter in the 'tryAddMruConnection', we can just save credentials as it is in connection being passed over to the method.

Benjin (approved):

PR #19952, opened 2025-08-13
New setting for connection groups to collapse/expand on load by default in the UI (v2.0)

samgusick:

## Description

This pull request implements a new MSSQL setting that, when enabled, makes the connection groups load expanded in the UI by default (which is the only option currently) but disabling the setting makes all connection groups instead load collapsed in the UI by default.

As someone who uses connection groups to organize databases from different projects, there are many connections I don't need to see while working in a particular workspace. It's a minor nuisance, but I have to collapse all of the connection groups I don't need everytime I open vs code, so I figured having the option to load all of the connection groups collapsed already would help me and a few others.

Additionally, when many connections and connection groups exist, they take up space in the editor window.

## Code Changes Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1928,7 +1928,7 @@
                 },
                 "mssql.objectExplorer.connectionGroupsCollapsedOnStartup": {
                     "type": "boolean",
-                    "description": "Connection Groups Collapsed on Startup",
+                    "description": "%mssql.objectExplorer.connectionGroupsCollapsedOnStartupDescription%",
                     "default": false,
                     "scope": "window"
                 },
--- a/package.json
+++ b/package.json
@@ -1926,6 +1926,12 @@
                     "description": "%mssql.objectExplorer.groupBySchema%",
                     "default": false
                 },
+                "mssql.objectExplorer.connectionGroupsCollapsedOnStartup": {
+                    "type": "boolean",
+                    "description": "Connection Groups Collapsed on Startup",
+                    "default": false,
+                    "scope": "window"
+                },
                 "mssql.objectExplorer.expandTimeout": {
                     "type": "number",
                     "default": 45,
--- a/package.json
+++ b/package.json
@@ -1926,9 +1926,9 @@
                     "description": "%mssql.objectExplorer.groupBySchema%",
                     "default": false
                 },
-                "mssql.objectExplorer.connectionGroupsCollapsedOnStartup": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1896,6 +1896,12 @@
                     "description": "%mssql.objectExplorer.groupBySchema%",
                     "default": false
                 },
+                "mssql.objectExplorer.connectionGroupsCollapsedOnStartup": {
+                    "type": "boolean",
+                    "description": "Connection Groups Collapsed on Startup",
```
aasimkhan30: Please localize the string. Add the loc string to package.nls.json and refer it here by id. 
`%mssql.objectExplorer.connectionGroupsCollapsedOnStartupDescription%'

aasimkhan30: Maybe change the descirption to `Collapse all connection groups in Object Explorer on startup instead of showing them expanded.`

package.nls.json
```diff
@@ -167,6 +167,7 @@
     "mssql.chooseAuthMethod": "Chooses which Authentication method to use",
     "mssql.authCodeGrant.description": "Prompts users to sign in using their browser.",
     "mssql.deviceCode.description": "Allows users to sign in to input-constrained devices.",
+    "mssql.objectExplorer.connectionGroupsCollapsedOnStartupDescription": "When enabled, all connection groups in the Object Explorer will collapse on startup instead of showing them expanded.",
```
Benjin:
```suggestion
    "mssql.objectExplorer.connectionGroupsCollapsedOnStartupDescription": "When enabled, connection groups will be collapsed instead of expanded at startup.",
```

package.json
```diff
@@ -1896,6 +1896,12 @@
                     "description": "%mssql.objectExplorer.groupBySchema%",
                     "default": false
                 },
+                "mssql.objectExplorer.connectionGroupsCollapsedOnStartup": {
```
Benjin: Small change to keep naming style consistent.  Be sure to update the description var name to match this as well!

```suggestion
                "mssql.objectExplorer.collapseConnectionGroupsOnStartup": {
```

Benjin:
Changes look good!  I left a couple of minor comments, and would be happy to take this once you've pushed those changes.

Thanks for your contribution!

samgusick:
@Benjin I rebased the branch on main, build and test should now succeed.

Issue #19839, opened 2025-07-25
[Bug]: Azure Account and Tenant Not Loading After Successful MFA Sign-In in MSSQL Extension


croblesm:
### Description

When connecting to an Azure SQL Database using the MSSQL extension for VS Code, the user successfully completes the Microsoft Entra ID (MFA) sign-in flow. However, the **Azure Account** and **Tenant ID** dropdowns remain empty, making it impossible to proceed. 

Expected behavior: After sign-in, the Azure account should populate the dropdown, and associated tenants should be listed.

Actual behavior: Despite successful sign-in, the dropdowns stay blank. 
The developer console logs repeated errors:
```
Error loading Azure accounts: Cannot read properties of undefined (reading 'displayName')
Error loading Azure tenants: Cannot read properties of undefined (reading 'tenants')
```
**Connection Dialog**
<img width="412" height="456" alt="Image" src="https://github.com/user-attachments/assets/4d82ba81-d8ed-4bc1-8e3b-86a59bb56e18" />

**Error message from developer console**
<img width="695" height="379" alt="Image" src="https://github.com/user-attachments/assets/c9920b0a-efae-4684-a504-8bd902147056" />

**Error messages from MSSQL log**
<img width="671" height="167" alt="Image" src="https://github.com/user-attachments/assets/1065be46-72ae-4689-bacc-affb8ec741ed" />

cc @kendalvandyke — Please feel free to add more details in case I missed something. Thanks for reporting this issue!

### Steps to Reproduce

1. Open the MSSQL extension connection dialog.
2. Select **Browse Azure** as input type.
3. Choose a subscription and resource group.
4. Set **Authentication Type** to `Microsoft Entra ID - Universal with MFA support`.
5. Click **Sign In** and complete the MFA flow.
6. Observe that the **Azure Account** and **Tenant ID** dropdowns remain empty.
7. Open the Developer Tools (F1 > Developer: Toggle Developer Tools) and check the console logs for errors.

### Affected Area

- [x] Connection dialog
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local container
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

- VS Code Version: 1.102.2
- MSSQL Extension Versions: v1.32, v1.33, v1.34
[...]

PR #19959, opened 2025-08-14
Automatically prune invalid/corrupted Entra account entries in global state

Benjin:
## Description

Addresses https://github.com/microsoft/vscode-mssql/issues/19839

The saved Entra account list can potentially have an invalid entry written to it due to no checks being done on the shape of data at runtime.  Based on the corrupted entry I was able to source from someone hitting this issue, it may be related to the auth window getting cancelled.  This fix adds logic to clean up this entry upon launch, and adds validation of entries when they're added.

Sample corrupted entry:
```
{
    "azureAccount": [
        {
            "canceled": false
        }
    ]
}
```

Sample correct entry:
```
{
    "azureAccount": [
        {
            "key": {
                "providerId": "azure_publicCloud",
                "id": "<GUID>.<GUID>",
                "accountVersion": "2.0"
            },
            "name": "Your Name - yourName@microsoft.com",
            "displayInfo": {
                "accountType": "work_school",
                "userId": "<GUID>.<GUID>",
                "contextualDisplayName": "Microsoft Corp",
                "displayName": "Your Name - yourName@microsoft.com",
                "email": "yourName@microsoft.com",
                "name": "Your Name"
            },
            "properties": {
                "providerSettings": { ... },
                    "scopes": [ ... ]
                },
                "isMsAccount": false,
                "owningTenant": { ... },
                "tenants": [
                    {
                        "id": "<GUID>",
                        "displayName": "Microsoft",
                        "tenantCategory": "Home"
                    }
                ],
                "azureAuthType": 0
            },
            "isStale": false
        }
    ],
}
```

## Code Changes Checklist

### DIFF
--- a/src/azure/accountStore.ts
+++ b/src/azure/accountStore.ts
@@ -48,7 +48,9 @@ export class AccountStore {
      * Gets a specific Entra account from the global state cache
      * @param key Account key to look up by.  Recommended to be `IAccount.key.id`
      */
-    public getAccount(key: string): IAccount | undefined {
+    public async getAccount(key: string): Promise<IAccount | undefined> {
+        await this.initialized;
+
         let account: IAccount | undefined;
         let configValues = this._context.globalState.get<IAccount[]>(Constants.configAzureAccount);
         if (!configValues) {
@@ -95,7 +97,7 @@ export class AccountStore {
             return false;
         }
 
-        if (!account.key?.id || !account.displayInfo?.displayName) {
+        if (!this.isValidAccount(account)) {
             this._logger.error(
                 `Attemped to add incomplete Entra account: ${JSON.stringify(account)}`,
             );
@@ -109,8 +111,6 @@ export class AccountStore {
         // remove account if already present in store
         if (configValues.length > 0) {
             configValues = configValues.filter((val) => val.key.id !== account.key.id);
-        } else {
-            configValues = [];
         }
 
         configValues.unshift(account);
@@ -132,7 +132,7 @@ export class AccountStore {
 
         accounts = accounts.filter((val) => {
             try {
-                if (val.key?.id && val.displayInfo?.displayName) {
+                if (this.isValidAccount(val)) {
                     return true;
                 } else {
                     numRemoved++;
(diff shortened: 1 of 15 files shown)


src/azure/accountStore.ts
```diff
@@ -43,57 +68,104 @@ export class AccountStore {
...
+        await this.initialized;
+
+        let configValues = await this.getAccounts();
+
+        // remove account if already present in store
+        if (configValues.length > 0) {
+            configValues = configValues.filter((val) => val.key.id !== account.key.id);
         } else {
-            this._logger.error("Empty Azure Account cannot be added to account store.");
+            configValues = [];
```
kburtram: what does this do?  is this already a 0 len array here?

src/azure/accountStore.ts
```diff
@@ -43,57 +68,104 @@ export class AccountStore {
...
-                return true;
-            } else {
+    /**
+     * Removes all Entra accounts that are incomplete because they are missing either a key ID or a display information
+     */
+    public async pruneInvalidAccounts(): Promise<void> {
+        // Because this runs during initialization and the public CRUD calls check for initialization,
+        // this method must work directly with the globalState API.
+
+        let accounts =
```
aasimkhan30, line 129: Should we run this check on getAccount and addAccount and remove any references outside of this class?

aasimkhan30 (approved):

src/azure/accountStore.ts
Benjin: Looks like this was some previous (still unnecessary) logic that I forgot to clean up when moving things around.

src/azure/accountStore.ts
Benjin: `addAccount()` only runs after initialization (which executes this cleanup) completes, and has its own guard to ensure that it doesn't add any new bad entries.

I'll centralize the logic that confirms it's a good entry, and add the intialization check to `getAccount()` as well.

Since this happens at intialization, there shouldn't be a way for a "bad" reference to leak outside this class, right?

caohai (approved):

lewis-sanchez (approved):

aasimkhan30 (approved):

PR #19961, opened 2025-08-15
Connections & Connection Groups can now be defined at both the User and Workspace level

### LINKED ISSUE #19780: [Bug]: Connection groups cannot be defined in workspace
### Description

I have connections that are defined at the workspace level, for two use cases:

1. Containerized, where the container is in the workspace,
2. Remote, where all developers share common settings for our shared databases

So, in two different projects, I would like to define connections at the workspace level.

Currently, I can't get this to work, and I believe the problem is the addition of the new `mssql.connectionGroups` setting, which seems to be ignored at the workspace level.

### Steps to Reproduce

If I define a structure like this:

```
    "mssql.connectionGroups": [
        {
            "name": "Remote",
            "id": "2E7E979B-0158-4352-989F-79205680C156"
        }
    ],
    "mssql.connections": [
        {
            "server": "127.0.0.1,14002",
            "database": "CoreDevDcaldwell",
            "authenticationType": "SqlLogin",
            "user": "CoreDevDcaldwell",
            "password": "",
            "emptyPasswordInput": false,
            "savePassword": true,
            "profileName": "CoreDevDcaldwell",
            "encrypt": "Mandatory",
            "trustServerCertificate": true,
            "connectTimeout": 15,
            "commandTimeout": 30,
            "applicationName": "vscode-mssql",
            "groupId": "2E7E979B-0158-4352-989F-79205680C156",
            "id": "FE195C16-9F90-474D-BA7F-078804F74996"
        },
```

Then, even though the `mssql.connectionGroups` is defined right above it, and is not defined at the user level, I get messages in the log (with tracing `Verbose`) like:

```
[7:46:39 AM] [ConnectionConfig] [Warning]: Connection 'CoreDevDcaldwell' with ID 'FE195C16-9F90-474D-BA7F-078804F74996' has a group ID that does not exist (2E7E979B-0158-4352-989F-79205680C156) so it is being ignored.  Correct its group ID to keep using this connection.
```

### Affected Area

- [ ] Connection dialog
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local container
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

Cannot define connections at the workspace settinglevel, must be defined manually at the user level.

### Environment Information

VSCode 1.102.0
MSSQL extension 1.33.0

### Confirmation

samgusick:

## Description

This pull request fixes #19780

For those Interested in testing this branch, I've started doing releases [here](https://github.com/samgusick/vscode-mssql/releases)!

Problem: 
Connections and connection groups cannot be defined at the workspace level (at least not in a user friendly, safe way). Having connections specific to workspaces can help with organization, especially when there are many connections across different projects and only a handful need to be used in each workspace.

What this does:

WARNING: these changes fundamentally alter the extensions approach to connections and connection groups within the existing framework. Many tests will need to be rewritten, and I'll need to do more work to make sure ones that should remain still pass. Additionally, user and workspace settings will be modified, so be mindful when testing with this branch.

- Instead of all connections/connection groups falling under ROOT, there will now be a "Workspace Connections" and "User Connections" group defined, with the legacy connections and their existing hierarchy migrated automatically to User Connections after upgrade.
- New connections are defined under these two new groups, or any of their child user made connection groups which the user can select in the connection dialog
- connections and connection groups defined in the workspace connection group will only be remembered at the workspace level. functionality for the User connections remains the same.
- When creating new connection groups, there's now a dropdown to decide whether the connection group will be defined at the user or workspace level.

My TO DO:

- [ ] Right now workspace settings are being created in every single folder opened with vs code on boot. If there are no workspace settings, the user should have to confirm they want workspace settings before being able to add workspace connections
- [ ] User and Workspace Connection folders should use more user friendly icons, differentiating them from regular connection folders.
 
## Code Changes Checklist

### DIFF
--- a/mssql/localization/l10n/bundle.l10n.json
+++ b/mssql/localization/l10n/bundle.l10n.json
@@ -449,9 +449,9 @@
         "message": "({0} rows affected)",
         "comment": ["{0} is the number of rows affected"]
     },
-    "Result Set {0}/{0} is the index of the result set": {
-        "message": "Result Set {0}",
-        "comment": ["{0} is the index of the result set"]
+    "Result Set Batch {0} - Query {1}/{0} is the batch number{1} is the query number": {
+        "message": "Result Set Batch {0} - Query {1}",
+        "comment": ["{0} is the batch number", "{1} is the query number"]
     },
     "Loading text view...": "Loading text view...",
     "Loading results...": "Loading results...",
--- a/mssql/localization/xliff/vscode-mssql.xlf
+++ b/mssql/localization/xliff/vscode-mssql.xlf
@@ -3061,9 +3061,10 @@
     <trans-unit id="++CODE++26b4293018ea72418733109f2e245387d9dfcfe7d4083c0c1f29eb1155c98365">
       <source xml:lang="en">Restore properties pane</source>
     </trans-unit>
-    <trans-unit id="++CODE++5d7e64a2598658aaaf1674a1b6222a0f6ab5bb588831a51d0936879cbf88670c">
-      <source xml:lang="en">Result Set {0}</source>
-      <note>{0} is the index of the result set</note>
+    <trans-unit id="++CODE++a1519957e8b397de6037c8e8a9a1af37009bf6d89fa0f388b1e4b3dc76179e64">
+      <source xml:lang="en">Result Set Batch {0} - Query {1}</source>
+      <note>{0} is the batch number
+{1} is the query number</note>
     </trans-unit>
     <trans-unit id="++CODE++219c4a6c86a716e99d5dc9611b61e584deb11b57e972d87d4b773387ebc3cadd">
       <source xml:lang="en">Results</source>


Benjin:
I'd like to take a different approach, and am interested in your feedback.  What if, instead, the root ID is hardcoded as `ROOT_NODE_ID`?

There'd no longer be an explicit entry needed for the root in the JSON, and both workspace and user config entries could reference `ROOT_NODE_ID` without worry.  If a user wants to create a user-json group called "User Connections" or a workspace-json group called "Workspace Connections", they can do that, both of which would have `parentId: "ROOT_NODE_ID"`.  This would also allow for shared workspace files to define connections without having to worry about syncing ROOT guids as they currently would.

What do you think?

samgusick:
@Benjin I definitely like the flexibility that could offer, I'll give it a go. I think for that approach there will need to be some indicators in in the connections UI (visual/text based) distinguishing workspace connections from user ones. Or potentially splitting the connections window in half in some respect to create an even more clear separation, though I think there's more flexibility and less modifications necessary with the former.

samgusick:
@Benjin after some experimenting, I think I've decided on a hybrid approach of mine and yours. We can switch to a hard coded root for the purposes you described, but I'd still like to keep workspace and user connections in separate predefined groups. Allowing users to mix and match user/workspace connections and groups just seems a bit disorganized and could create a lot of unexpected and problematic behavior that's more trouble than it's worth accounting for.

samgusick:
@Benjin I think creating a hard coded root is out of scope for this pull request, though it shouldn't be very difficult with this new approach. That being said, all of my intended functionality is now implemented. While many tests still don't pass (and I need to rework some code to use constants and locConstants), I'd appreciate a review just to make sure I'm on the right track here and it's worth continuing on to fixing the unit tests.

samgusick:
@aasimkhan30 @Benjin I've been consistently rebasing this branch on your changes, I've started doing[ releases with this branch](https://github.com/samgusick/vscode-mssql/releases) for easier testing. Let me know any thoughts/what might need to be done beyond reworking tests to get this pull request merged.

Benjin:
Hey @samgusick, I think we're likely going to end up making some further changes in this space internally, so while we're still looking to resolve this issue, I'm not sure if we'll end up using this specific PR.  Because the final design or implementation details may end up being different from how this PR works, I worry it could cause more confusion to other users if you're posting your own releases.  Also, it's important to note that we can only support the official, first-party releases that we post on the official page.

Though I also want to reiterate that we do appreciate the passion and contributions, even in instances where we don't merge the PRs in. :)

samgusick:
@Benjin Apologies for any confusion, I'd understand how an issue like this would want to be dealt with internally, and how sharing my own builds could be potentially problematic. I plan to continue iterating on this build, as I use it in my day-to-day work for my job. But if your team decides to go in a different direction, no offense taken! I'll switch to using the stable releases once the issue is closed.

kburtram:
@Benjin could you please help provide any feedback on this PR to get it ready to merge for Dec release?  Or if this isn't aligned with plans for this area or with the feature design, let's provide those details.  Thanks!

parkcheolhong (approved):
수정해 줘서 감사합니다.

Benjin:
Hey @samgusick!  Thank you for all your work on this feature.  I took a different approach to adding Workspace connection support in [another PR](https://github.com/microsoft/vscode-mssql/pull/20853), but drew some ideas from your work here. ❤️ 

I ended up doing a few things:
1. swapping the virtual root node ID to `ROOT` like we talked about here, automatically converting
2. treating workspace connections and groups as equals to their global config counterparts.  They can parent under each other if desired, instead of forcing users' connections into a specific layout, but not blocking someone who wants to have all workspace config items under one folder (you'd just create a folder in your workspace that parents to ROOT and have everything under there).
3. changes to workspace or global connections/groups get written back to their specific location

If you're interested, you can build the latest from `main` and try it out!

There's some piping that should make the addition of workspace _folder_ support easier in the future if we hear from users that they want that as well.

One thing I didn't do (yet?) was updating the connection dialog and connection group UIs with a way to specify where they want something saved.  Gotta run that by our designers, but open to ideas/prototypes on the new code base.

PR #19994, opened 2025-08-21
Fix race conditions in webview view, open in tab and reveal panel 

aasimkhan30:
## Description

1. Extracted all webview controller logic from the main controller into **`SqlOutputContentProvider`**.
2. Moved actual plan status handling into **`SqlOutputContentProvider`**, which will eventually become the central class for managing query execution and results lifecycle.
3. Refactored webview state management to avoid conflicts when multiple queries run simultaneously, preventing results from being overwritten.

https://github.com/user-attachments/assets/f72788de-cf2f-4355-a3ff-36dd02189f2a

5. Fixed a bug where results were still loaded into the webview view even if a panel for the query was already open.
6. Updated the **Reveal button** behavior: instead of always loading results in the webview view, it now checks for an open panel and reveals it when available.
7. Improved handling of the **“hide webview in tab”** setting: previously results were loaded into both the panel and the webview view (then forcibly hidden). Now, when this setting is enabled, webview views are not shown at all.

## Code Changes Checklist

### DIFF
--- a/src/models/sqlOutputContentProvider.ts
+++ b/src/models/sqlOutputContentProvider.ts
@@ -695,6 +695,7 @@ export class SqlOutputContentProvider {
 
         if (openInNewTabConfig) {
             this._queryResultWebviewController.revealPanel(uri);
+            return;
         }
 
         const isContainedInWebviewView =


src/models/sqlOutputContentProvider.ts
```diff
@@ -587,11 +685,26 @@ export class SqlOutputContentProvider {
...
-        );
+    /**
+     * Reveals the results grid in either webview panel or webview view.
+     * @param uri
+     */
+    public revealQueryResultPanel(uri: string): void {
+        const openInNewTabConfig = isOpenQueryResultsInTabByDefaultEnabled();
+
+        if (openInNewTabConfig) {
+            this._queryResultWebviewController.revealPanel(uri);
```
caohai, line 697: do we want to return early after this line?

src/models/sqlOutputContentProvider.ts
```diff
@@ -587,11 +685,26 @@ export class SqlOutputContentProvider {
...
-    public getIsExecutionPlan(): boolean {
-        return (
-            (this._executionPlanOptions?.includeEstimatedExecutionPlanXml ?? false) ||
-            (this._executionPlanOptions?.includeActualExecutionPlanXml ?? false)
-        );
+    /**
+     * Reveals the results grid in either webview panel or webview view.
+     * @param uri
+     */
+    public revealQueryResultPanel(uri: string): void {
```
caohai: The doc makes sense but the word `QueryResultPanel` in the function name could be a bit confusing (largely caused by the API names from VS Code) but I couldn't think about a better name.
The bottom area (Terminal/Output) is called `Panel` and the Webview API also has `createWebviewPanel`.
Probably just `revealQueryResult`?

caohai (dismissed):
LGTM

caohai (approved):

PR #20020, opened 2025-08-25
Dev/benjin/add fabric helper support

Benjin:

## Description

Adding a Fabric Helper that handles Fabric REST API calls related to SQL databases and Analytics endpoints.  For these operations, a token will be acquired from the vscode account system, depending on the tenant specified in the call.

This change also moves all proxy-related HTTP methods into a shareable helper class.

## Code Changes Checklist

### DIFF
--- a/src/fabric/fabricHelper.ts
+++ b/src/fabric/fabricHelper.ts
@@ -1,0 +1,225 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from "vscode";
+import { FabricSqlDbInfo, IWorkspace, IFabricError } from "../sharedInterfaces/connectionDialog";
+import { HttpHelper } from "../http/httpHelper";
+
+export class FabricHelper {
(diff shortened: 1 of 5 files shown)


src/fabric/fabricHelper.ts
```diff
@@ -0,0 +1,225 @@
...
+        workspaceId: string,
+        tenantId: string,
+    ): Promise<IWorkspace> {
+        const response = await this.fetchFromFabric<IWorkspace>(
+            `workspaces/${workspaceId}`,
+            `getting Fabric workspace '${workspaceId}'`,
+            tenantId,
+        );
+
+        return response;
```
lewis-sanchez, line 34: Out of curiosity, why did you use response.value on line 21, but not here?

lewis-sanchez (approved):

src/fabric/fabricHelper.ts
Benjin: That's just the way the Fabric REST API returns data: [List Workspaces](https://learn.microsoft.com/en-us/rest/api/fabric/admin/workspaces/list-workspaces?tabs=HTTP) sticks the array under `.value`, but [GetWorkspace](https://learn.microsoft.com/en-us/rest/api/fabric/admin/workspaces/get-workspace?tabs=HTTP#examples) returns it directly.

PR #20028, opened 2025-08-26
Adding a new webview state provider to use only a slice of entire state. 

aasimkhan30:
Introduce a slice-aware state provider for the webviews. Components subscribe to just a selected subset of state, avoiding unnecessary re-renders when unrelated state fields change.

I haven't change any existing webviews as the changes will be big and will take away from reviewing and understanding this component. 

## Code Changes Checklist

### DIFF
--- a/src/reactviews/common/vscodeWebviewProvider2.tsx
+++ b/src/reactviews/common/vscodeWebviewProvider2.tsx
@@ -1,0 +1,194 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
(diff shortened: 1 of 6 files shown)


src/reactviews/common/vscodeWebviewProvider2.tsx
```diff
@@ -0,0 +1,194 @@
...
+                // Send load stats notification
+                await extensionRpc.sendNotification(LoadStatsNotification.type, {
+                    loadCompleteTimeStamp: Date.now(),
+                });
+
+                // Mark as initialized and emit state change
+                setIsInitialized(true);
+                emit();
+            } catch (error) {
+                console.error("Bootstrap failed:", error);
```
Benjin, line 153: When do the coreRPCs get loaded?  Can this be a proper logger call instead of an easy-to-miss console error?

PR #20078, opened 2025-09-04
Adding Schema compare run comparison command to hookup SchemaCompare UI for sqlproj update proj usage

ssreerama:

## Description
This pull request introduces a new command to immediately run a schema comparison in the MSSQL extension, improves command registration, and adds a corresponding unit test to ensure correct behavior. The changes are primarily focused on enhancing the extension's schema compare functionality and making it callable programmatically (such as from Azure Data Studio).

**Schema Compare Feature Additions:**

* Added a new command `mssql.schemaCompare.runComparison` to `package.json` activation events and command palette, enabling users and external tools to trigger a schema comparison directly. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R40) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R907-R911)
* Registered the new command in `mainController.ts`, which invokes the `onSchemaCompareRunComparison` method to launch the schema compare UI and immediately start the comparison. [[1]](diffhunk://#diff-6551749c251883d4354dde8db43c093919871c9f70d05488ad28d1d99926f6e3R53) [[2]](diffhunk://#diff-226c7e3c205fb1d29e63f899a152924e487bf890bb934ecf2542a739504d2a08R1520-R1532)
* Implemented the `onSchemaCompareRunComparison` method in `mainController.ts`, which initializes the schema compare webview and runs the comparison with provided source and target endpoints.

**Testing Improvements:**

* Added a unit test in `mainController.test.ts` to verify that executing the new command correctly calls the handler and passes through the source and target parameters as expected. [[1]](diffhunk://#diff-a8df059fd593e193a83a0b67902dd68a41b1cefaaf2dd214456ed2a88e8e4bd9R20) [[2]](diffhunk://#diff-a8df059fd593e193a83a0b67902dd68a41b1cefaaf2dd214456ed2a88e8e4bd9R418-R445)

## Code Changes Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     ],
     "activationEvents": [
         "onUri",
+        "onCommand:mssql.schemaCompare.runComparison",
         "onCommand:mssql.loadCompletionExtension"
     ],
     "main": "./dist/extension",
@@ -901,6 +902,11 @@
             {
                 "command": "mssql.schemaCompareOpenFromCommandPalette",
                 "title": "%mssql.schemaCompare%",
+                "category": "MS SQL"
+            },
+            {
+                "command": "mssql.schemaCompare.runComparison",
+                "title": "Run Schema Comparison",
                 "category": "MS SQL"
             },
             {
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -7,7 +7,7 @@ import * as events from "events";
 import * as fs from "fs";
 import * as path from "path";
 import * as vscode from "vscode";
-import { IConnectionInfo, IScriptingObject } from "vscode-mssql";
+import { IConnectionInfo, IScriptingObject, SchemaCompareEndpointInfo } from "vscode-mssql";
 import { AzureResourceController } from "../azure/azureResourceController";
 import * as Constants from "../constants/constants";
 import * as LocalizedConstants from "../constants/locConstants";
@@ -1517,6 +1517,19 @@ export default class MainController implements vscode.Disposable {
                 ),
             );
 
+            // Register command to launch schema compare and immediately run comparison (callable by ADS)
+            this._context.subscriptions.push(
+                vscode.commands.registerCommand(
+                    Constants.cmdSchemaCompareRunComparison,
+                    async (
+                        source?: SchemaCompareEndpointInfo,
+                        target?: SchemaCompareEndpointInfo,
+                    ) => {
+                        await this.onSchemaCompareRunComparison(source, target);
+                    },
+                ),
+            );
+
             this._context.subscriptions.push(
                 vscode.commands.registerCommand(
                     Constants.cmdEditConnection,
@@ -2609,6 +2622,27 @@ export default class MainController implements vscode.Disposable {
             result,
             SchemaCompare.Title,
         );
+
+        schemaCompareWebView.revealToForeground();
+    }
+
+    public async onSchemaCompareRunComparison(
+        source?: SchemaCompareEndpointInfo,
+        target?: SchemaCompareEndpointInfo,
+    ): Promise<void> {
+        const result = await this.schemaCompareService.schemaCompareGetDefaultOptions();
+        const schemaCompareWebView = new SchemaCompareWebViewController(
+            this._context,
+            this._vscodeWrapper,
+            undefined,
+            this.schemaCompareService,
+            this._connectionMgr,
+            result,
+            SchemaCompare.Title,
+        );
+
+        // Launch with runComparison=true to immediately start the comparison
+        await schemaCompareWebView.launch(source, target, true, undefined);
 
         schemaCompareWebView.revealToForeground();
     }
--- a/test/unit/mainController.test.ts
+++ b/test/unit/mainController.test.ts
@@ -17,6 +17,7 @@ import VscodeWrapper from "../../src/controllers/vscodeWrapper";
 import { TestExtensionContext } from "./stubs";
 import { activateExtension } from "./utils";
 import StatusView from "../../src/views/statusView";
+import { SchemaCompareEndpointInfo } from "vscode-mssql";
 
 suite("MainController Tests", function () {
     let document: vscode.TextDocument;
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -903,6 +904,11 @@
                 "title": "%mssql.schemaCompare%",
                 "category": "MS SQL"
             },
+            {
```
Benjin: This file can be reverted because we're now updating an existing command instead of adding a new one.

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -123,11 +125,20 @@ export class SchemaCompareWebViewController extends ReactWebviewPanelController<
             `SchemaCompareWebViewController created with operation ID: ${this.operationId} - OperationId: ${this.operationId}`,
         );
 
-        if (node && !this.isTreeNodeInfoType(node)) {
-            node = this.getFullSqlProjectsPathFromNode(node);
+        if (sourceNode && !this.isTreeNodeInfoType(sourceNode) && sourceNode.projectFilePath) {
```
Benjin: Can you add comments here to clarify what you're checking?

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -123,11 +125,20 @@ export class SchemaCompareWebViewController extends ReactWebviewPanelController<
...
+            sourceNode = this.getFullSqlProjectsPathFromNode(sourceNode);
+        }
+        if (
+            targetNode &&
+            !this.isTreeNodeInfoType(targetNode) &&
+            targetNode.projectFilePath &&
+            targetNode.treeDataProvider?.roots?.length > 0 &&
+            targetNode.treeDataProvider.roots[0]?.projectFileUri?.fsPath
```
Benjin: Comments here about what these checks are meant to do.  Why are there so many more conditions than above?  Maybe this condition should be moved into a helper function with a obvious/readable name.

package.json
ssreerama: I thought i removed it. reverting the changes now

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -209,7 +230,39 @@ export class SchemaCompareWebViewController extends ReactWebviewPanelController<
             this.logger.verbose(`No source context provided - OperationId: ${this.operationId}`);
         }
 
-        await this.launch(source, undefined, false, comparisonResult);
+        // Resolve target (same rules as source)
```
Benjin: If it's the same rules of source, this should be a helper method.  When you move this to a helper method, please clearly comment the different cases (types of input that sourceContext might be).

test/unit/mainController.test.ts
```diff
@@ -414,6 +415,49 @@ suite("MainController Tests", function () {
...
+            gotMaybeTarget = wrapped.target;
+            gotRunComparison = wrapped.runComparison ?? false;
+        }
+
+        assert.equal(called, true, "Expected onSchemaCompare to be called");
+        assert.deepStrictEqual(gotMaybeSource, src, "Expected source passed through to handler");
+        assert.deepStrictEqual(gotMaybeTarget, tgt, "Expected target passed through to handler");
+        assert.equal(gotRunComparison, false, "Expected runComparison to be false");
+        // restore original handler so the test doesn't leak state
+        (mainController as any).onSchemaCompare = originalHandler;
```
Benjin: this needs to be in a `finally{}` block.

test/unit/schemaCompareWebViewController.test.ts
```diff
@@ -327,6 +327,8 @@ suite("SchemaCompareWebViewController Tests", () => {
             mockContext,
             vscodeWrapper.object,
             treeNode,
+            undefined,
```
Benjin, line 330: Can you also add more tests to validate the targetContext resolution and that it runs the comparison if `runComparison: true`?  Mocking `this.launch()` and checking what that was called with should be good enough.

src/controllers/mainController.ts
```diff
@@ -1502,9 +1502,13 @@ export default class MainController implements vscode.Disposable {
         });
 
         if (this.isRichExperiencesEnabled) {
+            // Register the command as async and forward all arguments
             this._context.subscriptions.push(
-                vscode.commands.registerCommand(Constants.cmdSchemaCompare, async (node: any) =>
-                    this.onSchemaCompare(node),
+                vscode.commands.registerCommand(
+                    Constants.cmdSchemaCompare,
+                    async (...args: any[]) => {
```
aasimkhan30: Can you please add proper types here?

src/controllers/mainController.ts
```diff
@@ -2598,12 +2602,28 @@ export default class MainController implements vscode.Disposable {
         return await this.newQueryFromPrompt(newDocUri);
     }
 
-    public async onSchemaCompare(node?: any): Promise<void> {
+    public async onSchemaCompare(...args: any[]): Promise<void> {
+        let sourceNode: any;
```
aasimkhan30: Same here. Can we please add types.

src/controllers/mainController.ts
ssreerama: properties datatypes could be any of these

ConnectionNode | TreeNodeInfo | SchemaCompareEndpointInfo | boolean | undefined

ssreerama: Having these props types, I tested the vscode-mssql all schemacompare  entry points, need to test from sqlproj( running hard to install the vsix)
 
 update: sqlproj testing is also completed

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -61,7 +61,9 @@ export class SchemaCompareWebViewController extends ReactWebviewPanelController<
     constructor(
         context: vscode.ExtensionContext,
         vscodeWrapper: VscodeWrapper,
-        node: any,
+        sourceNode: any,
```
Benjin: Ideally, these would be defined types (or unions) instead of `any`

src/controllers/mainController.ts
```diff
@@ -2598,12 +2610,38 @@ export default class MainController implements vscode.Disposable {
...
+     * Accepts variable arguments, typically:
+     *   - [sourceNode, targetNode, runComparison] when invoked from update Project SC or programmatically,
+     *   - [sourceNode, undefined] when invoked from a project tree node/ server / database node,
+     *   - [] when invoked from the command palette.
+     * This method normalizes the arguments and launches the Schema Compare UI.
+     */
+    public async onSchemaCompare(
+        ...args: (ConnectionNode | TreeNodeInfo | SchemaCompareEndpointInfo | boolean | undefined)[]
+    ): Promise<void> {
+        let sourceNode = undefined;
```
Benjin: This arg parsing should be in the function above.  We want to marshall the args into known variables and types/unions as soon as possible.

src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -164,52 +169,80 @@ export class SchemaCompareWebViewController extends ReactWebviewPanelController<
         );
     }
 
+    /**
+     * Checks if the given node is a SQL project node with a valid project file path.
+     * Returns true if the node is not a TreeNodeInfo, has a projectFilePath property,
+     * and its treeDataProvider contains a root with a valid projectFileUri.fsPath.
+     */
+    private isSqlProjectNodeWithFilePath(node: any): boolean {
```
Benjin: This seems like it should be in the SQL Projects extension instead of MSSQL.  When the Project -> Schema Compare command is moved to the SqlProj extension, it will have the project path directly that can be passed from there.

src/schemaCompare/schemaCompareWebViewController.ts
Benjin: This can happen in a later PR.

src/controllers/mainController.ts
Benjin: This can happen in a later PR

src/schemaCompare/schemaCompareWebViewController.ts
ssreerama: True, since we are handling the SC command from sqlproj side, sending the source/target as just sqlproj file path would remove the unnecessary checks. refactored the code accordingly.

src/controllers/mainController.ts
ssreerama: I did these changes already in this PR

src/schemaCompare/schemaCompareWebViewController.ts
ssreerama: same here

Issue #20081, opened 2025-09-04
[Bug]: open connection dialog prompts for keychain access on mac


amthomas46:
### Description

I'm not sure if it's a bug or not, but I wouldn't expect to provide my mac token or allow it to open. For Azure it just opens browser.

### Steps to Reproduce

In 1.36.0, when I click new connection I am asked to provide access
<img width="467" height="227" alt="Image" src="https://github.com/user-attachments/assets/a31ba6e8-79a1-48b6-9f76-d9de84cce29d" />

Then when I sign into fabric I get this view
<img width="303" height="239" alt="Image" src="https://github.com/user-attachments/assets/fb908171-981e-491d-a343-996cf87a6a48" />

Then I was able to sign in, search workspaces, connect

<img width="391" height="238" alt="Image" src="https://github.com/user-attachments/assets/6ca493f6-415c-49e7-9a64-32359e65060e" />

### Affected Area

- [x] Connection dialog
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local container
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

1.36.0
macOs Sequoia
SQL DB in Fabric

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

amthomas46:
When I tried to create a new fabric database, I got something similar again

<img width="287" height="301" alt="Image" src="https://github.com/user-attachments/assets/cf692909-ddfe-4ea6-8b29-c0b98fade637" />

kburtram:
The authentication prompt to access the keychain is expected behavior, as it’s an operating system security mechanism for accessing encrypted storage on macOS. We’re planning to investigate a design change to use the VS Code native keystore instead, which would remove this prompt from SQL Tools Service and leverage the existing VS Code prompts.  But this would be in Oct or Nov.

If you’re seeing the keychain prompt more than once per connection, that may indicate a bug—we can investigate this further for the September milestone.

PR #20085, opened 2025-09-04
Switching to credential store for all OSes

aasimkhan30:

## Description

Previously we were using vscode secret store for only Linux. With this we will switch to using secret store for all OSes. 

Fixes: https://github.com/microsoft/vscode-mssql/issues/20081

## Code Changes Checklist

### DIFF
--- a/src/credentialstore/credentialstore.ts
+++ b/src/credentialstore/credentialstore.ts
@@ -82,14 +82,10 @@ export class CredentialStore implements ICredentialStore {
         return true;
     }
 
-    public async deleteCredential(credentialId: string): Promise<boolean> {
+    public async deleteCredential(credentialId: string): Promise<void> {
         let cred: Contracts.Credential = new Contracts.Credential();
         cred.credentialId = credentialId;
         await this._secretStorage.delete(credentialId);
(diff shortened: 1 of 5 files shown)


kburtram (dismissed):

lewis-sanchez (dismissed):

src/credentialstore/credentialstore.ts
```diff
@@ -34,33 +40,52 @@ export class CredentialStore implements ICredentialStore {
...
     }
 
     public async deleteCredential(credentialId: string): Promise<boolean> {
         let cred: Contracts.Credential = new Contracts.Credential();
         cred.credentialId = credentialId;
-        if (Utils.isLinux) {
-            await this._secretStorage.delete(credentialId);
-        }
+        await this._secretStorage.delete(credentialId);
         const success = await this._client!.sendRequest(
```
caohai: Is this `boolean` return value used by any caller? For migrated credentials / credentials saved using vscode secret store, this will return false, right?

aasimkhan30: Nice catch. Remove the boolean return as it is not used anywhere.

caohai (approved):

Issue #19342, opened 2025-05-05
Feature Request - Addition of setting to color the "Null" query result values


delaclqm:
In the Azure Data Studio, when the query results view showed a "Null" value in a column, that null value was colored with a purple background. Can that feature be added to the backlog here?

srivi-surya:
Yes Please

jasperkpi:
This is the absolute number one most requested in my company, the color of the `NULL` values as compared to a text `"NULL"` and, as small as it is, a reason to not migrate yet.

PR #20086, opened 2025-09-04
Added setting to allow setting the background color in the query results for Null values 

allancascante:

## Description

Users have requested a way to set the background color in results when the value is Null as described in this issue [19342](https://github.com/microsoft/vscode-mssql/issues/19342); to make it customizable a setting was added to allow entering a hex value in case users want to change the default color.

## Code Changes Checklist


package.json
```diff
@@ -1606,6 +1606,15 @@
                     "default": null,
                     "scope": "resource"
                 },
+                "mssql.resultsMissingValueColor": {
```
Benjin, line 1609: Instead of adding a config option, let's just choose a good color that we know has contrast from the vscode theme.  I'd be surprised if people really wanted full customization here, and we've had users complain that there are too many settings already. 🙃

aasimkhan30: Yeah, lets choose a color from vscode theme. Hardcoding color doesn't work across themes.

PR #20087, opened 2025-09-05
Updating protocol handler to find existing saved connections

Benjin:
## Description

This PR updates the protocol handler with new, more intentional behavior.

Previously:
* all invocations (regardless of whether it's `connect` or `openConnectionDialog`) would result in the connection dialog being opened

Now:
* `connect` will first attempt to find the closest matching connection based on server, database, auth info, and extra properties.  If it finds one, that connection is automatically opened; if not, the connection dialog is opened with the values pre-filled
* `openConnectionDialog` will _always_ open the connection dialog with the values pre-filled

This avoids the old situation where repeated invocations via protocol handler (e.g. Fabric portal or vscode-fabric extension) would always create a new connection.

## Code Changes Checklist


test/unit/mssqlProtocolHandler.test.ts
```diff
@@ -3,119 +3,259 @@
...
-            connInfo.connectionString,
-            "Server=myServerAddress;Database=myDataBase;User Id=myUsername;Password=myPassword;",
-        );
+        expect(openConnectionDialogStub).to.have.been.calledOnceWith(undefined);
     });
 
-    test("handleUri - with connect command and query - parses query and returns connection info object", async () => {
-        let uri =
-            "vscode://ms-mssql.mssql/connect?server=myServer&database=dbName&authenticationType=SqlLogin&connectTimeout=15&trustServerCertificate=true&user=testUser";
+    suite("Connect command", () => {
```
Benjin, line 86: add test to ensure that it chooses the _best_ match

Benjin, line 86: For follow-up: check with PM to see what the minimum match level should be for selecting a connection.

src/models/utils.ts
```diff
@@ -275,6 +275,174 @@ export function isSameProfile(
...
+            "user",
+            "password",
+            "accountId",
+            "profileName",
+            "connectionString",
+            "savePassword",
+            "id",
+        ]);
+
+        // TODO: this is broken and results in all property reads being undefined.
```
Benjin: TODO

aasimkhan30 (dismissed):

src/models/utils.ts
```diff
@@ -275,6 +275,174 @@ export function isSameProfile(
...
+     */
+    AllAvailableProps = 4,
+
+    /**
+     * Matches on connection ID GUID.  Other properties are assumed to be the same.
+     */
+    Id = 5,
+}
+
+export class ConnInfoMatcher {
```
Benjin: Rename to ConnectionMatcher

Benjin: Add tests

src/models/connectionStore.ts
```diff
@@ -727,6 +728,26 @@ export class ConnectionStore {
         return connResults;
     }
 
+    public async findMatchingProfile(
+        connProfile: IConnectionProfile,
+    ): Promise<{ profile: IConnectionProfile; score: MatchScore } | undefined> {
+        const savedConnections = await this.readAllConnections();
```
Benjin, line 734: add test

PR #20152, opened 2025-09-10
Change on style for NULL results 

allancascante:
## Description

Change for the Feature Request [19342](https://github.com/microsoft/vscode-mssql/issues/19342) this change will use theme colors for the background of the cells in the results from queries where the value is NULL. 

Currently it is displayed only using Italic to diferentiate the NULL values:
image

New Style will still have the italic, and add the background color and the corresponding text color (to avoid the theme default text not having the necessary contrast for it to be redable). In this example the JobTitle is a string with value 'NULL' versus OrganizationNode and OrganizationLevel having NULL values.

Not selected:
image

Selected and Focused:
image

Selected and not Focused:
image

## Code Changes Checklist


src/reactviews/media/slickgrid.css
```diff
@@ -111,6 +111,9 @@
 
 .slick-cell .grid-cell-value-container.missing-value {
     font-style: italic;
+    display: block;
+    background-color: var(--vscode-editor-inactiveSelectionBackground);
```
Benjin: Looks like this is just updating the styling for the text within the cell, but not the cell itself.  In the HTML, I'm seeing

```
<div tabindex="-1" role="gridcell" aria-colindex="4" aria-label="NULL" class="slick-cell l3 r3 cell-null">
  <span title="" style="" class="grid-cell-value-container missing-value">
    NULL
  </span>
</div>
```

so maybe `cell-null` is the better class to target.

src/reactviews/media/slickgrid.css
```diff
@@ -111,6 +111,9 @@
...
+    color: var(--vscode-editor-selectionForeground);
```
Benjin: Also, it looks like most styles are applied in [`style()` from table.ts](https://github.com/microsoft/vscode-mssql/blob/d8ed8efe5ed29a6e02b5c882f9882dc7cdd22cea/src/reactviews/pages/QueryResult/table/table.ts#L552) with the colors themselves defined [here](https://github.com/microsoft/vscode-mssql/blob/d8ed8efe5ed29a6e02b5c882f9882dc7cdd22cea/src/reactviews/pages/QueryResult/table/interfaces.ts#L71)

src/reactviews/media/slickgrid.css
allancascante: done

src/reactviews/media/slickgrid.css
allancascante: done

src/reactviews/pages/QueryResult/table/table.ts
```diff
@@ -558,6 +558,18 @@ export class Table<T extends Slick.SlickData> implements IThemable {
             );
         }
 
+        if (styles.nullCellBackground) {
```
Benjin, line 561: Nice, this looks much better.  Can you also add a different color for when NULL is selected so that users can still see the difference when they have a range selected?

image

allancascante: @Benjin just to confirm you want selection color on focus selected or also inactive selected too?

allancascante: I updated the description with the screenshots for having the two extra styles, and updated the code accordingly

Benjin: Maybe something less... angry?  NULL isn't an error, so we don't want to use a color that implies an error.  I think a darker shade of the "selected" color (currently that blue) would make sense.

Also, I noticed that when I mouse-over the values, they turn blue again.  I think it's okay to not do anything special for NULLs when the user mouses over.  There's already some logic that tints the entire row, and that should be good enough.

https://github.com/user-attachments/assets/ae9cfa90-e170-4d46-8ff4-5f09d9ae7687

allancascante: As we discussed in a previous attempt at this feature, I should use colors from the vscode theme; so the colors used are the from the styles available as described here:

https://code.visualstudio.com/api/references/theme-color

Would you confirm that is still the case? I used:

 vscode-editor-inactiveSelectionBackground
 vscode-editor-selectionForeground
 vscode-editor-wordHighlightStrongBackground
 vscode-editor-linkedEditingBackground

Would you also please confirm hover color needs to be changed, for both selected and not selected null cells? if so, in addition to the ones already added would you please confirm either the editor styles to use or hex values for the colors that must be used? Please bear in mind the use of colors from the theme would not allow for selecting "darker" shades from one used already. We will have to check the default theme and pick one of the available styles.

allancascante: Also I tried to pick text editor background colors from theme and foreground, in hope that whoever makes the theme will use colors with enough contrast to allow the NULL text to be readable.

allancascante: I have made some mods to have the colores needed as theme colors, though I haven't set any colors, we need to come up with some. These will cover majority of the styles for null cells (selected, not selected, hover for both and active). I know that having an extra setting was not the preferred option but using colors will not add a setting and will be hidden to users unless they really want to customize them.

Benjin (approved):

PR #20153, opened 2025-09-10
Reduce console noise

allancascante:
## Description

objectExplorerService methods that rely on nodes to be loaded, were being called before the service state was initialized. This led to the errors, added logic to initialize the class state to avoid calling objectExplorerService methods on invalid state. Now it will load the root connections and nodes even if the Object Explorer from the UI hasn't been loaded.

## Code Changes Checklist


src/connectionconfig/connectionconfig.ts
```diff
@@ -184,7 +184,7 @@ export class ConnectionConfig implements IConnectionConfig {
         groups = groups.filter((group) => group.name === ConnectionConfig.RootGroupName);
 
         if (groups.length === 0) {
-            this._logger.error(
+            this._logger.verbose(
```
Benjin: This should stay as an error; if it can't find a root group in the config, that's an actual problem.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -87,7 +87,7 @@ export class ObjectExplorerService {
         const rootId = this._connectionManager.connectionStore.rootGroupId;
 
         if (!this._connectionGroupNodes.has(rootId)) {
-            this._logger.error(
+            this._logger.verbose(
```
Benjin: The issue here is less that the logging level is wrong and more that the OE isn't loaded when this is getting called.  Given that the repro is that the user has not opened OE but _is_ connecting, we should drop a breakpoint here and see what's triggering this, then decide whether we want to swallow this error conditionally (maybe if the map is empty, or some initalization flag isn't set) or code some alternate path to avoid calling this entirely.

src/objectExplorer/objectExplorerService.ts
```diff
@@ -613,6 +613,11 @@ export class ObjectExplorerService {
         }
     }
 
+    public async initialize(): Promise<void> {
```
Benjin: Should this be getting called in the constructor?  Take a look at how connectionConfig.ts handles this:

```
public initialized: Deferred<void> = new Deferred<void>();


constructor() {
...
void initialize();
}

private async initialize(): Promise<void> {
//... perform async init tasks, then:
this.initialized.resolve();
}
```

allancascante: > Should this be getting called in the constructor? Take a look at how connectionConfig.ts handles this:
> 
> ```
> public initialized: Deferred<void> = new Deferred<void>();
> 
> 
> constructor() {
> ...
> void initialize();
> }
> 
> private async initialize(): Promise<void> {
> //... perform async init tasks, then:
> this.initialized.resolve();
> }
> ```

that is a possibility, but wouldn't that affect plugin load time if the user has too many 'root nodes'?

Benjin: I think the additional time would be pretty minimal.  Most of the slow startup time is getting SQL Tools Service started and reading from that, and getRootNodes is mostly just parsing some JSON.

Also, if we call this in the constructor asynchronously and have a deferred initialization flag, then it shouldn't block anything during plugin load time except what chooses to wait on the `initialized` flag.

Issue #19830, opened 2025-07-23
[Bug]: Add Connection nodes don't show up under an empty connection group


laurenastrid1:
### Description

Expected the Add new connection options to show up under empty connection groups, like they do when there is nothing present in the object explorer

<img width="254" height="211" alt="Image" src="https://github.com/user-attachments/assets/f52f6471-7737-43db-a258-94d4242bffe3" />

<img width="265" height="156" alt="Image" src="https://github.com/user-attachments/assets/4ab979e4-9e02-438f-b594-810ac91d2759" />

### Steps to Reproduce

1. Make a new connection group through the extension
2. Observe that the add new connection nodes don't show up.

### Affected Area

- [ ] Connection dialog
- [ ] Query results panel
- [ ] Query editor
- [x] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local container
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

VS Code Version: 1.33

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #20154, opened 2025-09-10
Object Explorer should show options under empty connections groups

allancascante:
## Description

Adding change to show Add Connection and New Deployment options under empty connections groups, this address [19830](https://github.com/microsoft/vscode-mssql/issues/19830)

Before the change:

image
 

Example after the fix:

Screenshot 2025-09-10 142338

Example on clicking deployment under a new connection group:

image

## Code Changes Checklist

### DIFF
--- a/src/connectionconfig/connectionDialogWebviewController.ts
+++ b/src/connectionconfig/connectionDialogWebviewController.ts
@@ -396,7 +396,8 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
         });
 
         this.registerReducer("openCreateConnectionGroupDialog", async (state) => {
-            return getDefaultConnectionGroupDialogProps(state) as ConnectionDialogWebviewState;
+            state.dialog = getDefaultConnectionGroupDialogProps();
+            return state;
         });
 
         this.registerReducer("closeDialog", async (state) => {


aasimkhan30:
Is this also pre-populating parent group in the dialogs? I think implementing this feature also means making sure that connection and conn group is created within the parent group of these two actions.

src/deployment/localContainersHelpers.ts
```diff
@@ -42,7 +43,7 @@ export async function initializeLocalContainersState(
         port: undefined,
         hostname: "",
         acceptEula: false,
-        groupId: groupOptions[0].value,
+        groupId: selectedGroupId || (groupOptions.length > 0 ? groupOptions[0].value : ""),
```
aasimkhan30: NIT: selectedGroupId || groupOptions[0]?.value || ""

src/controllers/mainController.ts
```diff
@@ -220,9 +220,17 @@ export default class MainController implements vscode.Disposable {
             this._event.on(Constants.cmdClearPooledConnections, async () => {
                 await this.onClearPooledConnections();
             });
-            this.registerCommand(Constants.cmdDeployNewDatabase);
-            this._event.on(Constants.cmdDeployNewDatabase, () => {
-                this.onDeployNewDatabase();
+            this.registerCommandWithArgs(Constants.cmdDeployNewDatabase);
+            this._event.on(Constants.cmdDeployNewDatabase, (args?: any) => {
+                let initialConnectionGroup: { id?: string } | undefined = undefined;
```
Benjin: I think this can just be `initialConnectionGroupId: string` instead, if all that' the rest of this is doing is extracting the ID.

src/deployment/deploymentWebviewController.ts
```diff
@@ -61,6 +62,10 @@ export class DeploymentWebviewController extends FormWebviewController<
                 ),
             },
         });
+        // If an initial connection group was provided, try to pre-populate the form state
+        if (initialConnectionGroup && initialConnectionGroup.id) {
+            (this.state.formState as any).groupId = initialConnectionGroup.id;
```
Benjin: This should go inside the `initialize()` function because it's initialization logic.

Benjin: this shouldn't need casting as `any` to access `groupId`

src/deployment/fabricProvisioningHelpers.ts
```diff
@@ -65,7 +66,7 @@ export async function initializeFabricProvisioningState(
 
     state.formState = {
         accountId: defaultAccountId,
-        groupId: groupOptions[0].value,
+        groupId: selectedGroupId || groupOptions[0]?.value || "",
```
Benjin: What case does the extra `|| ""` cover?  groupOptions should always have at least one (the root group)

src/deployment/localContainersHelpers.ts
```diff
@@ -42,7 +43,7 @@ export async function initializeLocalContainersState(
         port: undefined,
         hostname: "",
         acceptEula: false,
-        groupId: groupOptions[0].value,
+        groupId: selectedGroupId || groupOptions[0]?.value || "",
```
Benjin: same q here

src/controllers/connectionGroupWebviewController.ts
```diff
@@ -124,16 +124,19 @@ export function createConnectionGroupFromSpec(spec: ConnectionGroupState): IConn
...
         id: Utils.generateGuid(),
+        parentId: spec.parentId,
     };
 }
 
 /**
  * Shared function to get the default properties for the Create Connection Group dialog.
  */
-export function getDefaultConnectionGroupDialogProps(state) {
+export function getDefaultConnectionGroupDialogProps(state, parentId?: string) {
```
Benjin: Since this is just for getting the default props, probably should go to the callers and set the parentId on state there instead.

src/sharedInterfaces/connectionGroup.ts
```diff
@@ -41,6 +41,7 @@ export interface ConnectionGroupState {
     description?: string;
     color?: string;
     message?: string;
+    parentId?: string;
```
Benjin: Why does this need to be added for this showing the OE options?  The state should only need the info necessary for the UI, and the connection dialog UI doesn't (currently) have an option for setting the parent when creating a new connection group.

src/sharedInterfaces/connectionDialog.ts
```diff
@@ -79,14 +79,21 @@ export interface IAzureTenant {
     name: string;
 }
 
-export interface IDialogProps {
+export interface IDialogProps<T extends IDialogContextProps = IDialogContextProps> {
```
Benjin: Are the changes in this file needed?  It looks like it's only used to add a `parentId` property to the `ConnectionGroupState` interface, but since every new connection group always goes to the root (and changing that would require some UI changes in the New/Edit Connection Group dialog; out of scope for this PR if we wanted to change that), that `parentId` property doesn't get used for anything.

src/deployment/deploymentWebviewController.ts
```diff
@@ -138,17 +146,18 @@ export class DeploymentWebviewController extends FormWebviewController<
...
         });
 
         this.registerReducer("setConnectionGroupDialogState", async (state, payload) => {
             if (payload.shouldOpen) {
-                state = getDefaultConnectionGroupDialogProps(state) as DeploymentWebviewState;
+                state = getDefaultConnectionGroupDialogProps(state);
             } else {
                 state.dialog = undefined;
             }
+            state.dialog.props.parentId = state.formState.groupId;
```
Benjin: this is the only spot that `parentId` is used, but then it's just being stored, not read or set anywhere else.

src/deployment/deploymentWebviewController.ts
```diff
@@ -138,17 +146,18 @@ export class DeploymentWebviewController extends FormWebviewController<
...
-            state.deploymentTypeState.dialog = state.dialog;
+            state.deploymentTypeState.dialog = undefined;
 
             return state;
         });
 
         this.registerReducer("setConnectionGroupDialogState", async (state, payload) => {
             if (payload.shouldOpen) {
-                state = getDefaultConnectionGroupDialogProps(state) as DeploymentWebviewState;
+                state = getDefaultConnectionGroupDialogProps(state);
```
Benjin: This function taking and returning untyped values, which I don't like.  Can you clean that up while you're here?

I think changing the function to this:
```
export function getDefaultConnectionGroupDialogProps(): CreateConnectionGroupDialogProps {
    return {
        type: "createConnectionGroup",
        props: {} as ConnectionGroupState,
    };
}
```

and changing the call to `state.dialog = getDefaultConnectionGroupDialogProps()` would make a lot more sense.

Benjin (commented):
If the addition of `parentId` isn't actually necessary, then this PR can touch many fewer files, I think.

Benjin (approved):

Benjin:
Changes look good!  Build is failing because the `openCreateConnectionGroupDialog` reducer in connectionDialogWebViewController needs updating since the `getDefaultConnectionGroupDialogProps` signature changed.

PR #20155, opened 2025-09-10
Aasim/fix/internal connects

aasimkhan30:

## Description

Consolidating all new query creation in the extension. 

## Code Changes Checklist

### DIFF
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -40,7 +40,7 @@ import { SqlTasksService } from "../services/sqlTasksService";
 import StatusView from "../views/statusView";
 import { IConnectionGroup, IConnectionProfile, ISelectionData } from "./../models/interfaces";
 import ConnectionManager from "./connectionManager";
-import SqlDocumentService from "./sqlDocumentService";
+import SqlDocumentService, { ConnectionStrategy } from "./sqlDocumentService";
 import VscodeWrapper from "./vscodeWrapper";
 import { sendActionEvent, startActivity } from "../telemetry/telemetry";
 import { ActivityStatus, TelemetryActions, TelemetryViews } from "../sharedInterfaces/telemetry";
@@ -703,8 +703,8 @@ export default class MainController implements vscode.Disposable {
             let scriptingObject = this._scriptingService.getObjectFromNode(node);
             let title = `${scriptingObject.schema}.${scriptingObject.name}`;
             const editor = await this._sqlDocumentService.newQuery({
-                copyLastActiveConnection: false,
                 content: selectStatement,
+                connectionStrategy: ConnectionStrategy.CopyConnectionFromInfo,
                 connectionInfo: connectionCreds,
             });
             if (executeScript) {
@@ -1234,8 +1234,8 @@ export default class MainController implements vscode.Disposable {
             const script = await this._scriptingService.script(scriptingParams);
 
             await this._sqlDocumentService.newQuery({
-                copyLastActiveConnection: false,
                 content: script,
+                connectionStrategy: ConnectionStrategy.CopyConnectionFromInfo,
                 connectionInfo: connection?.credentials,
             });
         } catch (error) {
@@ -2510,7 +2510,9 @@ export default class MainController implements vscode.Disposable {
         }
         await this.sqlDocumentService.newQuery({
             content,
-            copyLastActiveConnection: connectionCreds ? false : true,
+            connectionStrategy: connectionCreds
+                ? ConnectionStrategy.CopyConnectionFromInfo
+                : ConnectionStrategy.PromptForConnection,
             connectionInfo: connectionCreds,
         });
 
--- a/src/controllers/mainController.ts
+++ b/src/controllers/mainController.ts
@@ -1412,16 +1412,6 @@ export default class MainController implements vscode.Disposable {
             ),
         );
 
-        // Search Objects in Database
-        this._context.subscriptions.push(
-            vscode.commands.registerCommand(
-                Constants.cmdSearchObjects,
-                async (node?: ConnectionNode) => {
-                    await this.onSearchObjects(node);
-                },
-            ),
-        );
-
         this.registerCommand(Constants.cmdConnectionGroupCreate);
         this._event.on(Constants.cmdConnectionGroupCreate, () => {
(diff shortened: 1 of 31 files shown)


src/controllers/sqlDocumentService.ts
```diff
@@ -311,3 +394,62 @@ export default class SqlDocumentService implements vscode.Disposable {
...
 }
+
+/**
+ * Connection strategy for new SQL documents.
+ */
+export enum ConnectionStrategy {
+    /**
+     * No connection will be established.
+     */
+    None = "none",
```
Benjin: Maybe this should be "DoNotConnect".  "None" sounds like it'd mean letting it choose a default.

src/controllers/mainController.ts
```diff
@@ -2557,52 +2495,50 @@ export default class MainController implements vscode.Disposable {
...
-                true, // createObjectExplorerSession
-            );
+            // Case 1: User right-clicked on an OE node and selected "New Query"
+            connectionCreds = node.connectionProfile;
+            nodeType = node.nodeType;
+        } else if (activeEditor) {
+            // Case 2: User triggered "New Query" from command palette and the active document has a connection
+            connectionCreds = undefined;
+            nodeType = "previousEditor";
+        } else if (this.objectExplorerTree.selection?.length > 0) {
```
Benjin: I realize this is old (my?) logic, but this should probably only do this case if there's just one OE node selected.  If there are multiple, we should probably treat is as though there's no selection because the user's intention is unclear.

src/controllers/mainController.ts
```diff
@@ -2557,52 +2495,50 @@ export default class MainController implements vscode.Disposable {
...
-            if (connectionInfo) {
-                return await this.newQueryFromProfile(
-                    newDocUri,
-                    connectionInfo.credentials,
-                    undefined, // sessionId
-                    "previousEditor",
-                    false, //createObjectExplorerSession
-                );
-            }
+        if (connectionCreds) {
```
Benjin, line 2513: Readability: try to leave blank lines after closing braces for blocks.

src/controllers/mainController.ts
```diff
@@ -2557,52 +2495,50 @@ export default class MainController implements vscode.Disposable {
...
-            return await this.newQueryFromProfile(
-                newDocUri,
-                selectedNode.connectionProfile,
-                selectedNode.sessionId,
-                selectedNode.nodeType,
-                true, // createObjectExplorerSession
-            );
-        }
+        await this.connectionManager.connectionStore.removeRecentlyUsed(
+            <IConnectionProfile>connectionCreds,
```
Benjin: `as` casting instead of `<>`

src/controllers/sqlDocumentService.ts
```diff
@@ -15,13 +15,16 @@ import * as Constants from "../constants/constants";
 import * as LocalizedConstants from "../constants/locConstants";
 import MainController from "./mainController";
 import * as vscodeMssql from "vscode-mssql";
+import { Deferred } from "../protocol";
+// no-op imports removed after execution flow simplification
```
Benjin: Should this comment be removed?  Seems to just describe an action during your dev workflow.

src/controllers/sqlDocumentService.ts
```diff
@@ -245,61 +244,145 @@ export default class SqlDocumentService implements vscode.Disposable {
...
+            const documentUriKey = getUriKey(editor.document.uri);
+            const connectionPromise = new Deferred<boolean>();
+
+            await this._connectionMgr.connect(
+                documentUriKey,
+                connectionConfig.connectionInfo,
+                connectionPromise,
+            );
+
+            await connectionPromise.promise;
```
Benjin: Does the resolved value (a boolean) need to be checked before calling `createObjectExplorerSession()`?

src/controllers/sqlDocumentService.ts
```diff
@@ -245,61 +244,145 @@ export default class SqlDocumentService implements vscode.Disposable {
...
+
+            await this._mainController.createObjectExplorerSession(connectionConfig.connectionInfo);
+        }
+
+        // Update status views
+        const documentUriKey = getUriKey(editor.document.uri);
```
Benjin: this is called multiple times; just save it before the if-statement above.

src/controllers/sqlDocumentService.ts
```diff
@@ -245,61 +244,145 @@ export default class SqlDocumentService implements vscode.Disposable {
...
+
+                const resolvedConnectionInfo = this._connectionMgr.getConnectionInfoFromUri(
+                    options.sourceUri,
+                );
+                return resolvedConnectionInfo
+                    ? {
+                          shouldConnect: true,
+                          connectionInfo: Utils.deepClone(resolvedConnectionInfo),
+                      }
+                    : { shouldConnect: false };
```
Benjin, line 369: does this represent an error case that should be reported?  Same for case below

src/queryHistory/queryHistoryProvider.ts
```diff
@@ -143,28 +144,29 @@ export class QueryHistoryProvider implements vscode.TreeDataProvider<any> {
...
+            const uri = getUriKey(editor.document.uri);
+            const title = path.basename(editor.document.fileName);
             const queryPromise = new Deferred<boolean>();
             await this._outputContentProvider.runQuery(
                 this._statusView,
                 uri,
                 undefined,
                 title,
-                {}, // empty execution plan options
+                {},
```
Benjin: add deleted comment back; we wanna comment all the "magic constants" since Typescript annoyingly doesn't have named parameters.

src/queryResult/queryResultWebviewPanelController.ts
```diff
@@ -78,6 +87,7 @@ export class QueryResultWebviewPanelController extends ReactWebviewPanelControll
     }
 
     public revealToForeground() {
+        this._viewColumn;
```
Benjin: err, does this do anything?

Benjin:
Can you elaborate on the changes (what and why) in the PR description?

src/queryResult/queryResultWebviewPanelController.ts
aasimkhan30: Had added console statement here previously, probably forgot about it.

PR #20167, opened 2025-09-13
Allowing users to quickly change connections 

### LINKED ISSUE #18867: Feature request: Easily switch between connections
In ADS when a script is connected to a database, you can easily click the "Change" button and select an existing connection to change to. -> 1.Click Change, 2.Click connection, 3.Click Connect

It requires more clicks and up-down mouse movements to do this in VSCode. -> 1.Click connection, 2.Click disconnect, 3.Are you sure want to disconnect?, 4.Disconnected, 5.Select new connection

aasimkhan30:
## Description
Fixes #18867

1. Added a **change connection** icon in the status bar.   Matches ADS change connection icon
image

2. Split the status bar into two buttons: one for changing the connection and another for changing the database.  

| Before | Now |
|--------|-----|
| image| image image |

## Code Changes Checklist

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -113,19 +113,3 @@ export function getEditorEOL(): string {
         ? os.EOL
         : vscode.workspace.getConfiguration("files").get<string>("eol");
 }
-
-/**
- * Trims the given display string to the specified maximum length, adding an ellipsis if trimmed.
- * @param displayString long display string
- * @param maxLength maximum length of the string, 0 for empty string, negative or undefined for no limit
- * @returns trimmed display string
- */
-export function trimDisplayString(displayString: string, maxLength: number): string {
-    let result = displayString;
-    if (maxLength === 0) {
-        result = "";
-    } else if (maxLength > 0 && result.length > maxLength) {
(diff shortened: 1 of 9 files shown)


src/utils/utils.ts
```diff
@@ -113,3 +113,19 @@ export function getEditorEOL(): string {
         ? os.EOL
         : vscode.workspace.getConfiguration("files").get<string>("eol");
 }
+
+/**
+ * Trims the given display string to the specified maximum length, adding an ellipsis if trimmed.
+ * @param displayString long display string
+ * @param maxLength maximum length of the string, 0 for empty string, negative or undefined for no limit
```
Benjin: IMO, 0 returning an empty string (with no ellipsis) is a behavior pretty specific to the status bar truncation (effectively making it just the icon, right?) and would be a surprise in a general utility function.  Maybe that case should move out of this util function so that this behaviors more predictable in any other place it gets used?  Not married to it, just struck me as a bit odd.

Benjin: Can you double-check that this specifc behavior is captured in the option description for `configStatusBarConnectionInfoMaxLength`?

PR #20178, opened 2025-09-16
Fixing context values for connections and running queries. 

aasimkhan30:
## Description
This fixes the context variable to match the vscode's expected values.  This fixes the issues where we are not updating toolbar icons properly when the files are connected or the query is running. 

For example,  the query has started executing (as seen in the status bar) but it doesn't reflect in toolbar. 

image

1. In mssql.connection, we store the resource using uri.toString(true) whereas vscode expects uri.toString().
https://github.com/microsoft/vscode/blob/bb5a3c607b14787009f8e9fadb720beee596133c/src/vs/workbench/…
1. In mssql.runningQueries, we store the resourcePath using uri.path where as vscode expects uri.fsPath. 
https://github.com/microsoft/vscode/blob/bb5a3c607b14787009f8e9fadb720beee596133c/src/vs/workbench/…

## Code Changes Checklist

### DIFF
--- a/src/controllers/connectionManager.ts
+++ b/src/controllers/connectionManager.ts
@@ -654,8 +654,9 @@ export default class ConnectionManager {
                 }
                 await self.handleConnectionErrors(fileUri, connection, result);
             }
-
-            await self.tryAddMruConnection(connection);
+            if (!result.errorMessage || result.errorNumber !== 0) {
+                await self.tryAddMruConnection(connection);
+            }
         };
     }
 
@@ -784,6 +785,12 @@ export default class ConnectionManager {
                 result.errorMessage ? result.errorMessage : result.messages,
             ),
         );
+        /**
+         * Making sure to clean up the connection if it exists since the connection failed
+         * and firing the onConnectionsChanged event so UI elements like codelens can update.
+         */
+        delete this._connections[fileUri];
+        this._onConnectionsChangedEmitter.fire();
         sendErrorEvent(
             TelemetryViews.ConnectionPrompt,
             TelemetryActions.CreateConnectionResult,
@@ -1065,7 +1072,7 @@ export default class ConnectionManager {
             }
 
             this.removeActiveConnection(fileUri);
-            vscode.commands.executeCommand("setContext", "mssql.connections", this._connections);
+            this.updateConnectionsContext();
             return result;
         } else if (this.isConnecting(fileUri)) {
             // Prompt the user to cancel connecting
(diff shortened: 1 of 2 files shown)


src/controllers/queryRunner.ts
```diff
@@ -1681,6 +1671,31 @@ export default class QueryRunner {
...
+    /**
+     * Vscode core expects uri.fsPath for resourcePath context value.
+     * https://github.com/microsoft/vscode/blob/bb5a3c607b14787009f8e9fadb720beee596133c/src/vs/workbench/common/contextkeys.ts#L275
+     */
+
+    /**
+     * Add query to running queries list
+     * @param ownerUri The owner URI of the query
+     */
+    private addRunningQuery(ownerUri: string): void {
```
Benjin: These two methods (and `updateRunningQueries` below) feel like they ought to be static, or part of a QueryRunnerManager singleton class.  Those methods existing on (each individual) query runner instance feels like it gives the wrong idea of scope-of-impact/responsibility.

src/controllers/connectionManager.ts
```diff
@@ -1076,6 +1077,22 @@ export default class ConnectionManager {
...
+        /**
+         * Making sure we keep encodings in the context. We need to convert the keys
+         * to Uri and back to string because the keys in _connections might have skipped encoding.
+         * This is done to match the behavior of how vscode core handles resource URIs in
+         * contexts.
+         * https://github.com/microsoft/vscode/blob/bb5a3c607b14787009f8e9fadb720beee596133c/src/vs/workbench/common/contextkeys.ts#L261C1-L262C1
+         */
+        const connectionContext = [];
+        Object.keys(this._connections).forEach((key) => {
+            const Uri = vscode.Uri.parse(key);
```
Benjin: lowercasee `uri` for var name

src/controllers/connectionManager.ts
```diff
@@ -1076,6 +1077,22 @@ export default class ConnectionManager {
         }
     }
 
+    private updateConnectionsContext() {
+        /**
+         * Making sure we keep encodings in the context. We need to convert the keys
+         * to Uri and back to string because the keys in _connections might have skipped encoding.
```
Benjin, line 1083: why might the keys in `_connections_ have skipped encoding?  Seems like that's an/the underlying issue that ought to be addressed, while this is a symptom - still useful to address, but more of a patch.

src/controllers/connectionManager.ts
```diff
@@ -1076,6 +1077,22 @@ export default class ConnectionManager {
...
+         * This is done to match the behavior of how vscode core handles resource URIs in
+         * contexts.
+         * https://github.com/microsoft/vscode/blob/bb5a3c607b14787009f8e9fadb720beee596133c/src/vs/workbench/common/contextkeys.ts#L261C1-L262C1
+         */
+        const connectionContext = [];
+        Object.keys(this._connections).forEach((key) => {
```
Benjin: Why not `.map()`?  That seems like a more natural use than forEach() and push()

src/controllers/queryRunner.ts
```diff
@@ -1681,6 +1671,31 @@ export default class QueryRunner {
...
+        const key = vscode.Uri.parse(ownerUri).fsPath;
+        QueryRunner._runningQueries.push(key);
+        this.updateRunningQueries();
+    }
+
+    /**
+     * Remove current query from running queries list
+     */
+    private removeRunningQuery(): void {
+        QueryRunner._runningQueries = QueryRunner._runningQueries.filter(
```
Benjin, line 1693: It'd odd to me that we're storing _runningQueries as an array (rather than Set) and _connections as a record (rather than a Map).  Not saying we have to swap those now, but probably something for the future so that we can get the unqiueness guarantees that we seem to be expecting for those.

Then we can just serialize those into more primitive forms when we update the context values.

Benjin:
Can you update the PR description to describe the types of issues this has been causing?

src/controllers/connectionManager.ts
aasimkhan30: Added a TODO comment.

Benjin (approved):

lewis-sanchez (approved):

PR #20179, opened 2025-09-16
VsCode SQL Projects publish dialog Initial user interface setup 

ssreerama:

## Description
This pull request introduces a new "Publish Project" feature that enables users to publish database projects through a dedicated UI dialog in the extension. The changes span backend command registration, webview controller implementation, form component generation, and frontend React views for the publish dialog. This feature is designed to be extensible and integrates with existing project and schema comparison workflows.

**New "Publish Project" feature:**

_Backend integration and command registration:_
* Registered a new command `mssql.publishDatabaseProject` and added its handler in `MainController`, which launches the publish dialog webview for a given project file path. [[1]](diffhunk://#diff-6551749c251883d4354dde8db43c093919871c9f70d05488ad28d1d99926f6e3R53) [[2]](diffhunk://#diff-226c7e3c205fb1d29e63f899a152924e487bf890bb934ecf2542a739504d2a08R1559-R1567) [[3]](diffhunk://#diff-226c7e3c205fb1d29e63f899a152924e487bf890bb934ecf2542a739504d2a08R2623-R2643)
* Added `PublishProjectWebViewController`, which manages dialog state, form components, and reducer logic for publishing database projects. It initializes the form with sensible defaults and supports async component generation and dynamic field visibility.
* Added localization strings for the publish dialog UI in both backend (`locConstants.ts`) and frontend (`common/locConstants.ts`). [[1]](diffhunk://#diff-2d98c0c23ef655217ba0a20b80a78ca7cfc57cdffd98def8a1476a36ab83b629R1180-R1199) [[2]](diffhunk://#diff-80cf0c3f1d17e77ecb6dfa3fe3fc586e24b81468f3ee1a86eac06876139fb73dR881-R892)

_Form component and UI implementation:_
* Implemented `formComponentHelpers.ts` to generate publish form components, supporting future extensibility and async population of options.
* Added React view for the publish dialog (`PublishProject/index.tsx`, `publishProject.tsx`, and supporting components), including a custom field for publish profiles and main form actions (generate script, publish, cancel). [[1]](diffhunk://#diff-c0aeaef369728542163a7b7bcf82b1f6c3a8ef7619519ba0594103b881289391R1-R15) [[2]](diffhunk://#diff-e90853adabd0675d8b8647b1cb332407269bb59a8b71c1275349d9aa8a63237aR1-R122) [[3]](diffhunk://#diff-c23abc44e3c0d10a00911ff0f66fa2029a3724eef859043b49fb3a7174c9d567R1-R102)
* Updated the bundle configuration to include the new publish dialog React view.

These changes lay the groundwork for a user-facing publish dialog, enabling database project publishing workflows directly from the extension.

## Code Changes Checklist

Final:
image

### DIFF
--- a/src/publishProject/publishProjectWebViewController.ts
+++ b/src/publishProject/publishProjectWebViewController.ts
@@ -37,8 +37,8 @@ export class PublishProjectWebViewController extends FormWebviewController<
         super(
             context,
             _vscodeWrapper,
-            "publishDialog",
-            "publishDialog",
+            "publishProject",
+            "publishProject",
             {
                 formState: {
                     publishProfilePath: "",
@@ -47,7 +47,7 @@ export class PublishProjectWebViewController extends FormWebviewController<
                     publishTarget: "existingServer",
                     sqlCmdVariables: {},
                 },
-                formComponents: {},
+                formComponents: generatePublishFormComponents(),
                 projectFilePath,
                 inProgress: false,
                 lastPublishResult: undefined,
@@ -73,22 +73,8 @@ export class PublishProjectWebViewController extends FormWebviewController<
         // Register reducers after initialization
         this.registerRpcHandlers();
 
-        // Initialize dialog state
-        this.initializeDialog(projectFilePath);
-    }
-
-    private initializeDialog(projectFilePath: string) {
-        // Load publish form components
-        this.state.formComponents = generatePublishFormComponents();
-
-        // keep initial project path and computed database name
-        if (projectFilePath) {
-            this.state.projectFilePath = projectFilePath;
-        }
-
-        // Update visibility and initial state
+        // Update visibility based on initial publish target
         void this.updateItemVisibility();
-        this.updateState();
     }
 
     private registerRpcHandlers(): void {
--- a/src/reactviews/pages/PublishProject/components/PublishProfileSection.tsx
+++ b/src/reactviews/pages/PublishProject/components/PublishProfileSection.tsx
@@ -64,7 +64,7 @@ export const PublishProfileField: React.FC = () => {
                     size="small"
                     appearance="secondary"
                     onClick={() => context.selectPublishProfile?.()}>
-                    {loc.SelectProfile}
+                    {loc.SelectPublishProfile}
                 </Button>
                 <Button
                     size="small"
--- a/localization/l10n/bundle.l10n.json
+++ b/localization/l10n/bundle.l10n.json
@@ -1429,6 +1429,7 @@
     "Publish Target": "Publish Target",
     "Existing SQL server": "Existing SQL server",
     "Local development container": "Local development container",
+    "Save As": "Save As",
     "Schema Compare": "Schema Compare",
     "Options have changed. Recompare to see the comparison?": "Options have changed. Recompare to see the comparison?",
     "Failed to generate script: '{0}'/{0} is the error message returned from the generate script operation": {
@@ -1445,7 +1446,6 @@
     },
     "Loading Schema Designer Model...": "Loading Schema Designer Model...",
     "Schema Designer Model is ready. Changes can now be published.": "Schema Designer Model is ready. Changes can now be published.",
-    "Save As": "Save As",
     "Schema Designer": "Schema Designer",
     "Opening Publish Script. This may take a while...": "Opening Publish Script. This may take a while...",
     "Generating Report. This may take a while...": "Generating Report. This may take a while...",
--- a/src/constants/locConstants.ts
+++ b/src/constants/locConstants.ts
@@ -1291,6 +1291,8 @@ export class PublishProject {
     public static PublishTargetLabel = l10n.t("Publish Target");
     public static PublishTargetExisting = l10n.t("Existing SQL server");
     public static PublishTargetContainer = l10n.t("Local development container");
+    public static SelectPublishProfile = l10n.t("Select Profile");
+    public static SaveAs = l10n.t("Save As");
     public static GenerateScript = l10n.t("Generate Script");
     public static Publish = l10n.t("Publish");
 }
--- a/src/reactviews/common/locConstants.ts
+++ b/src/reactviews/common/locConstants.ts
@@ -894,7 +894,7 @@ export class LocConstants {
 
     public get publishProject() {
         return {
-            SelectProfile: l10n.t("Select Profile"),
+            SelectPublishProfile: l10n.t("Select Profile"),
             SaveAs: l10n.t("Save As..."),
             generateScript: l10n.t("Generate Script"),
             publish: l10n.t("Publish"),


src/reactviews/pages/PublishProject/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import ReactDOM from "react-dom/client";
+import "../../index.css";
+import { VscodeWebviewProvider } from "../../common/vscodeWebviewProvider";
+import PublishProjectPage from "./publishProject";
+
+ReactDOM.createRoot(document.getElementById("root")!).render(
+    <VscodeWebviewProvider>
```
aasimkhan30: use VscodeWebviewProvider2 and the selector mechanism to access state. This helps in keeping the renders local to the component and improve perf. 

To see examples you can see the execution plan (simpler) and query results (a little more complex). 

Here is the index file https://github.com/microsoft/vscode-mssql/blob/main/src/reactviews/pages/ExecutionPlan/index.tsx

Here is the state provider: https://github.com/microsoft/vscode-mssql/blob/main/src/reactviews/pages/ExecutionPlan/executionPlanStateProvider.tsx

and here is how to access a state prop: https://github.com/microsoft/vscode-mssql/blob/425fb9a77db99eee37e2cb65ed2156b8b5e9a7df/src/reactviews/pages/QueryResult/queryResultPane.tsx#L174-L175

ssreerama: oh! didn't notice there is a latest one. will definitely look into the latest webView provider and make the required updates. Thanks

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -0,0 +1,158 @@
...
+                state: PublishDialogState,
+                payload: Partial<IPublishForm> & { projectFilePath?: string },
+            ) => {
+                if (payload) {
+                    state.formState = { ...state.formState, ...payload };
+                    if (payload.projectFilePath) {
+                        state.projectFilePath = payload.projectFilePath;
+                    }
+                }
+                this.updateState(state);
```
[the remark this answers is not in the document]
ssreerama: Removing the state update from the reducers and just an fyi these are just a placeholder code which are not yet implemnted fully.

src/publishProject/formComponentHelpers.ts
```diff
@@ -0,0 +1,63 @@
...
+import { PublishProject as Loc } from "../constants/locConstants";
+
+/**
+ * Generate publish form components. Kept async for future extensibility
+ * (e.g. reading project metadata, fetching remote targets, etc.)
+ */
+export async function generatePublishFormComponents(): Promise<
+    Record<keyof IPublishForm | string, PublishDialogFormItemSpec>
+> {
+    const components: Record<keyof IPublishForm | string, PublishDialogFormItemSpec> = {
```
Benjin: why `| string`?  Everything here should be a key for `IPublishForm`.

src/publishProject/formComponentHelpers.ts
```diff
@@ -0,0 +1,63 @@
...
+        profileName: {
+            propertyName: "profileName",
+            label: Loc.ProfileLabel,
+            required: false,
+            type: FormItemType.Input,
+        },
+        serverName: {
+            propertyName: "serverName",
+            label: Loc.ServerLabel,
+            required: false,
```
Benjin: Out of curiosity, when is the serverName not required?

src/publishProject/formComponentHelpers.ts
```diff
@@ -0,0 +1,63 @@
...
+                    displayName: Loc.PublishTargetExisting,
+                    value: "existingServer",
+                },
+                {
+                    displayName: Loc.PublishTargetContainer,
+                    value: "localContainer",
+                },
+            ],
+        },
+    } as Record<keyof IPublishForm | string, PublishDialogFormItemSpec>;
```
Benjin: Is this cast necessary?  The type's already specified as part of the declaration, and you're populating at definition time.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -0,0 +1,165 @@
...
+        // Load publish form components asynchronously for future extensibility
+        this.state.formComponents = await generatePublishFormComponents();
+
+        // keep initial project path and computed database name
+        if (projectFilePath) {
+            this.state.projectFilePath = projectFilePath;
+        }
+
+        await this.updateItemVisibility();
+        this.updateState();
```
Benjin: This doesn't need to be called explicitly if there's nothing after it because the first thing the following `then` calls is `updateState()`, and we don't want to trigger a double-render.

src/reactviews/pages/PublishProject/publishProjectStateProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { createContext, useMemo } from "react";
+import { useVscodeWebview2 } from "../../common/vscodeWebviewProvider2";
+import { WebviewRpc } from "../../common/rpc";
+import {
+    PublishDialogReducers,
+    PublishDialogState,
+    IPublishForm,
+} from "../../../sharedInterfaces/publishDialog";
+
+export interface PublishProjectContextValue {
```
Benjin: This should be called `PublishProjectContextProps` to be consistent with the other dialogs.

src/reactviews/pages/PublishProject/publishProjectStateProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+    // Use inner state directly for form system generics
+    state?: PublishDialogState; // snapshot accessor
+    formAction: (event: PublishFormActionEvent) => void;
```
Benjin: Why are state and formAction included here?  Shouldn't these be coming from something this extends?  @aasimkhan30, Maybe this is because the form stuff hasn't been converted over to the new webview2 stuff?  I can hop on that if that's the issue here.

test/unit/publishProjectWebViewController.test.ts
```diff
@@ -0,0 +1,69 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as TypeMoq from "typemoq";
```
Benjin: All new test files should use Sinon instead of TypeMoq.  We're trying to phase out TypeMoq.

src/reactviews/pages/PublishProject/publishDialogSelector.ts
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { PublishDialogReducers, PublishDialogState } from "../../../sharedInterfaces/publishDialog";
+import { useVscodeSelector } from "../../common/useVscodeSelector";
+
+export function usePublishDialogSelector<T>(
```
Benjin, line 9: @aasimkhan30 can you review to make sure this PR is using the webview2 stuff correctly?

src/reactviews/pages/PublishProject/publishProject.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const loc = LocConstants.getInstance().publishProject;
+    const context = useContext(PublishProjectContext);
+
+    // Select pieces of state needed for this component
+    const formComponents = usePublishDialogSelector((s) => s.formComponents, Object.is);
+    const formState = usePublishDialogSelector((s) => s.formState, Object.is);
+    const inProgress = usePublishDialogSelector((s) => s.inProgress, Object.is);
+
+    // Check if component is properly initialized and ready for user interaction
+    const isComponentReady = isPublishFormContext(context) && !!formComponents && !!formState;
```
Benjin: This typeguard shouldn't be necessary... why isn't context's type already known?

src/reactviews/pages/PublishProject/publishProject.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+    // Check if all required fields are provided based on publish target
+    const isFormValid =
```
Benjin: Is the form framework not handling this for you already?  That should handle validation and required fields.

If not, then maybe the form framework isn't right for this dialog, or not hooked up correctly.

src/reactviews/pages/PublishProject/publishProject.tsx
```diff
@@ -0,0 +1,149 @@
...
+            if (formState.publishTarget === "existingServer") {
+                return !!formState.serverName;
+            }
+
+            // For local container, server name is not required
+            if (formState.publishTarget === "localContainer") {
+                return true; // Could add container-specific validations here if needed
+            }
+
+            return false;
```
Benjin: what case results in falling through to here?  What's left when publishTarget is neither `existingServer` or `localContainer`?

src/reactviews/pages/PublishProject/publishProject.tsx
```diff
@@ -0,0 +1,149 @@
...
+        })();
+
+    // Buttons should be disabled when:
+    // - Component is not ready (missing context, form components, or form state)
+    // - Operation is in progress
+    // - Form validation fails
+    const buttonsDisabled = !isComponentReady || inProgress || !isFormValid;
```
Benjin: this would be more useful to be called something like "readyToPublish".

src/reactviews/pages/PublishProject/publishProjectStateProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+    selectPublishProfile: () => void;
+    savePublishProfile: (profileName: string) => void;
+    setPublishValues: (
+        values: Partial<PublishDialogState["formState"]> & { projectFilePath?: string },
+    ) => void;
+    extensionRpc: WebviewRpc<PublishDialogReducers>;
+}
+
+// Event payload coming from shared FormField components
+export interface PublishFormActionEvent {
```
Benjin: Why not use the original FormEvent interface?

src/reactviews/pages/PublishProject/publishProjectStateProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+        () => ({
+            get state() {
+                const inner = getSnapshot(); // inner PublishDialogState
+                if (!inner || Object.keys(inner).length === 0) {
+                    return undefined;
+                }
+                return inner;
+            },
+            formAction: (event: PublishFormActionEvent) =>
+                extensionRpc.action("formAction", { event }),
```
Benjin: I think `state` and `formAction` should be available by default if you extend the correct classes/interfaces; why are these necessary?

src/sharedInterfaces/publishDialog.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/**
+ * Form item specification for Publish dialog fields.
+ */
+export interface PublishDialogFormItemSpec
+    extends FormItemSpec<IPublishForm, PublishDialogState, PublishDialogFormItemSpec> {
+    isAdvancedOption?: boolean;
+    optionCategory?: string;
+    optionCategoryLabel?: string;
```
Benjin: Are these used anywhere?

typings/vscode-mssql.d.ts
```diff
@@ -475,6 +475,15 @@ declare module "vscode-mssql" {
         cancel(operationId: string): Thenable<ResultStatus>;
     }
 
+    export interface IPublishDatabaseProjectService {
+        publishProject(
```
Benjin: I don't see this getting used anywhere.  Is there a reason this is a new `publishDatabaseProjectService` instead of using `DacFxService.deployDacpac()` (like ADS), or a new method on that service?

Where/when does the sqlproj get built into a dacpac?

src/publishProject/formComponentHelpers.ts
ssreerama: It should be true, since is a basic UI and server/all other fields have their own tasks to work on, I left it untouched.

src/publishProject/formComponentHelpers.ts
ssreerama: Initially I added a sample string message to display, that left it here, cleaning now!

src/publishProject/formComponentHelpers.ts
ssreerama: Not required, since using only few properties from IPublishForm the sqlcmd variable is missing here, so used the casting to bypass the type checking.

src/publishProject/publishProjectWebViewController.ts
ssreerama: Right! thanks for the catch, missed cleaning it!

src/reactviews/pages/PublishProject/publishProject.tsx
ssreerama: added the type guard check just by overthinking which is not necessary here. Removed the unnecessary type guard - the context was already properly typed as PublishProjectContextValue | undefined. Simplified to a basic null check and let TypeScript handle the type narrowing properly.

src/reactviews/pages/PublishProject/publishProject.tsx
ssreerama: same here, should have used the existing validation, didn't notice the framework already provide the validations, that I'm overlooking at it. thanks.

typings/vscode-mssql.d.ts
ssreerama: We’ll be using IDacFxService’s deployDacpac and generateDeployScript methods to publish the project along with its associated task. This code had become outdated with my initial changes and was unintentionally left in place

src/sharedInterfaces/publishDialog.ts
ssreerama: These are created for advance publish options dialog placeholder, currently they are not used anywhere, commenting them to not confuse others :)

src/reactviews/pages/PublishProject/publishProject.tsx
```diff
@@ -76,42 +40,28 @@ function PublishProjectDialog() {
...
+              )
+            : false;
 
     // Buttons should be disabled when:
     // - Component is not ready (missing context, form components, or form state)
     // - Operation is in progress
-    // - Form validation fails
-    const buttonsDisabled = !isComponentReady || inProgress || !isFormValid;
+    // - Form has validation errors
+    const readyToPublish = !isComponentReady || inProgress || hasValidationErrors;
```
Benjin, line 58: The name and value here doesn't match up. The name is "ready to publish", but you're actually calculating if it's _not_ ready to publish.

src/reactviews/pages/PublishProject/publishProjectStateProvider.tsx
```diff
@@ -5,45 +5,34 @@
...
-    savePublishProfile: (profileName: string) => void;
-    setPublishValues: (
-        values: Partial<PublishDialogState["formState"]> & { projectFilePath?: string },
-    ) => void;
+export interface PublishProjectContextProps extends PublishProjectProvider {
+    readonly state?: PublishDialogState;
+    log(message: string, level?: LoggerLevel): void;
+    sendActionEvent(event: WebviewTelemetryActionEvent): void;
+    sendErrorEvent(event: WebviewTelemetryErrorEvent): void;
+    /** Advanced escape hatch; prefer using typed provider methods */
```
Benjin: You shouldn't have to redefine these, but maybe there's some cleanup we need to do on the webview2 stuff here first.  Can you sync with @aasimkhan30 on that?

src/reactviews/pages/PublishProject/components/ConnectionSection.tsx
```diff
@@ -0,0 +1,69 @@
...
+    savePublishProfile: (profileName: string) => void;
+}
+
+export const ConnectionSection: React.FC<{ startIdx: number }> = ({ startIdx }) => {
+    const context = useContext(PublishProjectContext) as PublishFormContext | undefined;
+
+    const serverComponent = usePublishDialogSelector((s) => s.formComponents.serverName, Object.is);
+    const databaseComponent = usePublishDialogSelector(
+        (s) => s.formComponents.databaseName,
+        Object.is,
```
aasimkhan30: You don't have to provide comparison function here since you are using the default one.

src/reactviews/pages/PublishProject/publishProjectStateProvider.tsx
```diff
@@ -0,0 +1,74 @@
...
+    children,
+}) => {
+    const { extensionRpc, getSnapshot } = useVscodeWebview2<
+        PublishDialogState,
+        PublishDialogReducers
+    >();
+
+    const value = useMemo<PublishProjectContextProps>(
+        () => ({
+            get state() {
```
aasimkhan30: Please remove state from here and just use selector. Having a centralized state defeats the purpose of having selector.

src/reactviews/pages/PublishProject/publishProject.tsx
```diff
@@ -0,0 +1,99 @@
...
+});
+
+function PublishProjectDialog() {
+    const classes = useStyles();
+    const formStyles = useFormStyles();
+    const loc = LocConstants.getInstance().publishProject;
+    const context = useContext(PublishProjectContext);
+
+    // Select pieces of state needed for this component
+    const formComponents = usePublishDialogSelector((s) => s.formComponents, Object.is);
```
aasimkhan30, line 38: Please remove default "equals"

src/reactviews/pages/PublishProject/components/PublishProfileSection.tsx
```diff
@@ -0,0 +1,103 @@
...
+export const PublishProfileField = (props: { idx: number }) => {
+    const { idx } = props;
+    const classes = useStyles();
+    const formStyles = useFormStyles();
+    const loc = LocConstants.getInstance().publishProject;
+    const context = useContext(PublishProjectContext) as PublishFormContext | undefined;
+    if (!context || !context.state) {
+        return undefined;
+    }
+    const component = context.state.formComponents.profileName as PublishDialogFormItemSpec;
```
aasimkhan30: Please use selectors here.

src/sharedInterfaces/publishDialog.ts
```diff
@@ -33,9 +33,8 @@ export interface PublishDialogState
  */
 export interface PublishDialogFormItemSpec
     extends FormItemSpec<IPublishForm, PublishDialogState, PublishDialogFormItemSpec> {
-    isAdvancedOption?: boolean;
-    optionCategory?: string;
-    optionCategoryLabel?: string;
+    // (Removed advanced option metadata: was isAdvancedOption/optionCategory/optionCategoryLabel)
+    // Reintroduce when the Publish dialog gains an "Advanced" section with grouped fields.
```
Benjin: Can you create an issue with the type "Task" to track this work?  Add the link here with `// TODO so we can find it

src/sharedInterfaces/publishDialog.ts
```diff
@@ -64,6 +63,38 @@ export interface PublishDialogReducers extends FormReducers<IPublishForm> {
...
+        publishProfilePath?: string;
+    }): void;
+    /** Generate (but do not execute) a publish script */
+    generatePublishScript(): void;
+    /** Choose a publish profile file and apply (may partially override form state) */
+    selectPublishProfile(): void;
+    /** Persist current form state as a named profile */
+    savePublishProfile(profileName: string): void;
+    /** Bulk set form values (e.g., after loading a profile) */
+    setPublishValues(values: {
```
Benjin: This should probably be happening in the controller, then we wouldn't need a bulk set reducer.

test/unit/publishProjectWebViewController.test.ts
```diff
@@ -32,11 +27,59 @@ suite("PublishProjectWebViewController Tests", () => {
             keys: () => [] as readonly string[],
             setKeysForSync: (_keys: readonly string[]) => undefined,
         } as unknown as vscode.Memento & { setKeysForSync(keys: readonly string[]): void };
-        mockContext.setup((c) => c.globalState).returns(() => globalState);
 
-        mockVscodeWrapper = TypeMoq.Mock.ofType(VscodeWrapper);
-        const outputChannel = TypeMoq.Mock.ofType<vscode.OutputChannel>();
-        mockVscodeWrapper.setup((v) => v.outputChannel).returns(() => outputChannel.object);
+        const rawContext = {
```
Benjin: You don't need to mock out the entire context object.  Check for some code or a helper you can reuse in other tests.

src/sharedInterfaces/publishDialog.ts
ssreerama: created a task, actually I need to create tasks for all the work items.

src/reactviews/pages/PublishProject/publishProjectStateProvider.tsx
ssreerama: yeah I shouldn't redefine them here, as getCoreRPCs2 already provides them that can be used directly for telemetry events, removed for now.

aasimkhan30 (approved):

lewis-sanchez (dismissed):

scripts/bundle-reactviews.js
```diff
@@ -25,6 +25,7 @@ const config = {
         userSurvey: "src/reactviews/pages/UserSurvey/index.tsx",
         schemaDesigner: "src/reactviews/pages/SchemaDesigner/index.tsx",
         schemaCompare: "src/reactviews/pages/SchemaCompare/index.tsx",
+        publishDialog: "src/reactviews/pages/PublishProject/index.tsx",
```
Benjin: this should probably "publishProject" to align with the rest of the entries here

src/controllers/mainController.ts
```diff
@@ -1594,6 +1595,15 @@ export default class MainController implements vscode.Disposable {
                 ),
             );
 
+            this._context.subscriptions.push(
```
Benjin: This method is where all of the commands that are activated (only) from the Object Explorer are registered.  For organization's sake, this should be up near ~L300.

src/controllers/mainController.ts
```diff
@@ -2587,6 +2597,22 @@ export default class MainController implements vscode.Disposable {
         schemaCompareWebView.revealToForeground();
     }
 
+    /**
+     * Handler for the Publish Database Project command.
+     * Accepts the project file path as an argument.
+     * This method launches the Publish Project UI for the specified database project.
+     * @param projectFilePath The file path of the database project to publish.
+     */
+    public async onPublishDatabaseProject(projectFilePath: string): Promise<void> {
```
Benjin: No `await`s here. Can be made non-async since that's not doing anything here.

src/publishProject/formComponentHelpers.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Generate publish form components. Kept async for future extensibility
+ * (e.g. reading project metadata, fetching remote targets, etc.)
+ */
+export async function generatePublishFormComponents(): Promise<
+    Record<keyof IPublishForm, PublishDialogFormItemSpec>
+> {
+    const components: Record<keyof IPublishForm, PublishDialogFormItemSpec> = {
+        profileName: {
```
Benjin: Is this the selected publishProfile, or a connection profile name?  If connection profile, we should instead use the connection ID.  if it's a publishProfile, then this entry should probably be renamed to `publishProfilePath`

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -0,0 +1,145 @@
...
+            sqlCmdVariables: {},
+        };
+
+        const innerState: PublishDialogState = {
+            formState: initialFormState,
+            formComponents: {},
+            projectFilePath,
+            inProgress: false,
+            lastPublishResult: undefined,
+        } as PublishDialogState;
```
Benjin: No need to save this as both `innerState` and `initialState`.  Okay to just pass it directly to `super()` below, IMO.

Note: the executionPlan code has `innerState` just because it's a funky state that's used in multiple webviews and is sometimes nested in other state.  That's not the typical model, and isn't necessary for the forms :)

src/publishProject/formComponentHelpers.ts
```diff
@@ -0,0 +1,70 @@
...
+    PublishDialogFormItemSpec,
+    PublishDialogState,
+} from "../sharedInterfaces/publishDialog";
+import { PublishProject as Loc } from "../constants/locConstants";
+
+/**
+ * Generate publish form components. Kept async for future extensibility
+ * (e.g. reading project metadata, fetching remote targets, etc.)
+ */
+export async function generatePublishFormComponents(): Promise<
```
Benjin: Does this need to be async?  Connection Dialog is async here because it needs to load the connection options from STS and the Azure accounts for auth from the cache, but the publis project dialog doesn't have either of those reqs.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -0,0 +1,145 @@
...
+            .then(() => {
+                this.updateState();
+                this.initialized.resolve();
+            })
+            .catch((err) => {
+                this.initialized.reject(err);
+            });
+    }
+
+    private async initializeDialog(projectFilePath: string) {
```
Benjin: With the current code, I don't think any of this actually is asynchronous, so you should just be able to just call this synchronously and get rid of the `initialized` flag.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -0,0 +1,145 @@
...
+        this.registerReducer("selectPublishProfile", async (state: PublishDialogState) => {
+            // TODO: implement profile selection logic
+            return state;
+        });
+
+        this.registerReducer(
+            "savePublishProfile",
+            async (state: PublishDialogState, payload: { profileName?: string }) => {
+                if (payload?.profileName) {
+                    state.formState.profileName = payload.profileName;
```
Benjin: Is this a full path or a file name?

src/reactviews/pages/PublishProject/components/ConnectionSection.tsx
```diff
@@ -0,0 +1,88 @@
...
+    useEffect(() => setLocalServer(serverValue || ""), [serverValue]);
+    useEffect(() => setLocalDatabase(databaseValue || ""), [databaseValue]);
+
+    if (!publishCtx) {
+        return undefined;
+    }
+
+    const renderInput = (
+        component:
+            | {
```
Benjin: what does `|` here do?  I haven't seen that before.

src/reactviews/pages/PublishProject/components/ConnectionSection.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const [localDatabase, setLocalDatabase] = useState(databaseValue || "");
+
+    useEffect(() => setLocalServer(serverValue || ""), [serverValue]);
+    useEffect(() => setLocalDatabase(databaseValue || ""), [databaseValue]);
+
+    if (!publishCtx) {
+        return undefined;
+    }
+
+    const renderInput = (
```
Benjin: Looks like you couldn't use `<FormField>` here because it hasn't been upgraded to the new webview framework yet?  Guess I should prioritize that...

src/reactviews/pages/PublishProject/components/PublishProfileSection.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { LocConstants } from "../../../common/locConstants";
+import { PublishProjectContext } from "../publishProjectStateProvider";
+import { usePublishDialogSelector } from "../publishDialogSelector";
+import type { IPublishForm } from "../../../../sharedInterfaces/publishDialog";
+import { FormItemType } from "../../../../sharedInterfaces/form";
+
+/**
+ * Extended context type including the extra publish profile actions we expose.
+ */
+type PublishFormActions = {
```
Benjin: Why is this necessary?  Should this be inherited from `PublishProjectProvider`?

src/reactviews/pages/PublishProject/components/PublishProfileSection.tsx
```diff
@@ -0,0 +1,121 @@
...
+        return undefined;
+    }
+    if (component.type !== FormItemType.Input) {
+        return undefined;
+    }
+
+    return (
+        <div className={`${formStyles.formComponentDiv} ${classes.root}`}>
+            <div className={classes.fieldContainer}>
+                <Field
```
Benjin: This shouldn't be duplicated if we can avoid it...  Can this use the `renderInput()` helper from the other file, at least until I upgrade the shared form stuff?

Can you made a workitem in ADO for centralizing this generation logic?

scripts/bundle-reactviews.js
```diff
@@ -26,6 +26,7 @@ const config = {
         schemaDesigner: "src/reactviews/pages/SchemaDesigner/index.tsx",
         schemaCompare: "src/reactviews/pages/SchemaCompare/index.tsx",
         changePassword: "src/reactviews/pages/ChangePassword/index.tsx",
+        publishDialog: "src/reactviews/pages/PublishProject/index.tsx",
```
Benjin: this should be `publishProfile` instead of `publishDialog` to keep consistent.

src/sharedInterfaces/publishDialog.ts
```diff
@@ -0,0 +1,86 @@
...
+    lastPublishResult?: { success: boolean; details?: string };
+}
+
+/**
+ * Form item specification for Publish dialog fields.
+ */
+export interface PublishDialogFormItemSpec
+    extends FormItemSpec<IPublishForm, PublishDialogState, PublishDialogFormItemSpec> {
+    // (Removed advanced option metadata: was isAdvancedOption/optionCategory/optionCategoryLabel)
+    // Reintroduce when the Publish dialog gains an "Advanced" section with grouped fields.
```
Benjin: These are advanced options for publishing, not advanced options for the connection, right?

test/unit/publishProjectWebViewController.test.ts
```diff
@@ -0,0 +1,75 @@
...
+            appendLine: () => undefined,
+            clear: () => undefined,
+            replace: (_value: string) => undefined,
+            show: () => undefined,
+            hide: () => undefined,
+            dispose: () => undefined,
+        };
+
+        // Subclass VscodeWrapper to override the outputChannel getter cleanly.
+        class TestVscodeWrapper extends VscodeWrapper {
```
Benjin: there's a helper that centralizes this already.  Search for `stubVscodeWrapper()`

src/publishProject/publishProjectWebViewController.ts
ssreerama: as per the current ADS, it displays the fullpath, updating the name to path

src/sharedInterfaces/publishDialog.ts
ssreerama: yes! for publish options, updated the todo comment!

src/reactviews/pages/PublishProject/components/PublishProfileSection.tsx
ssreerama: I created a helper class for Field components to reuse in entire dialog in the publishTarget related PR here:https://github.com/microsoft/vscode-mssql/pull/20217/files#diff-ab478588221e41e538d74b331474578c38fbcaac74f5bb4a7b02b49b7fb0c051. But yes it will be nice if we can centralize the logic at parent level and reuse at all dialogs when possible.

test/unit/publishProjectWebViewController.test.ts
ssreerama: Done!

scripts/bundle-reactviews.js
ssreerama: Updated to PublishProject!

src/reactviews/pages/PublishProject/components/PublishProfileSection.tsx
ssreerama: right, it was left uncleaned, as I have upcoming tasks to work on these connection/profile/target/advanced fields, these files will get a lot of reusable code.

src/controllers/mainController.ts
ssreerama: Thanks, I have moved this to commands register section.

src/controllers/mainController.ts
ssreerama: used to have await methods in one of my iterations and left unchanged. Might be required in future when loading the profile/options etc, but not now. I have removed the async for now.

src/publishProject/formComponentHelpers.ts
ssreerama: Not an async anymore for now at least, thanks!

src/publishProject/formComponentHelpers.ts
ssreerama: modified to publishProfilePath as it is for publish profile and holds the fiull path.

src/publishProject/publishProjectWebViewController.ts
ssreerama: Ah, I understand now. Thank you! I assumed those variables were used to improve code readability

src/reactviews/pages/PublishProject/components/ConnectionSection.tsx
ssreerama: just a format thing, since the type is "list of variable" or undefined, format file adding that operand .

src/reactviews/pages/PublishProject/components/PublishProfileSection.tsx
ssreerama: created one here: https://msdata.visualstudio.com/SQLToolsAndLibraries/_workitems/edit/4730560

Benjin (approved):

PR #20181, opened 2025-09-17
Fix webview panel creation. 

aasimkhan30:
## Description

1. Create the webview panel before starting the query execution. This makes sure that all  state updates are properly sent to the webview.
1. Awaiting properly before query panel creation  is created. This ensures that webview RPCs are set properly and all the react resources are loaded properly.
1. Adding a different splash screen for query result documents.It is a simple spinner with a “Loading results” message to indicate progress while results are being prepared.
1. Adding splash screen in case query fails to start.

image

## Code Changes Checklist

### DIFF
--- a/src/models/sqlOutputContentProvider.ts
+++ b/src/models/sqlOutputContentProvider.ts
@@ -337,6 +337,10 @@ export class SqlOutputContentProvider {
                 executionPlanOptions?.includeActualExecutionPlanXml,
             this._actualPlanStatuses.includes(uri),
         );
+        if (isOpenQueryResultsInTabByDefaultEnabled()) {
+            await this._queryResultWebviewController.createPanelController(queryRunner.uri);
+            await new Promise<void>((resolve) => setTimeout(resolve, 300));
+        }
         if (queryRunner) {
             void queryCallback(queryRunner);
         }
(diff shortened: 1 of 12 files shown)


src/models/sqlOutputContentProvider.ts
```diff
@@ -337,6 +337,10 @@ export class SqlOutputContentProvider {
                 executionPlanOptions?.includeActualExecutionPlanXml,
             this._actualPlanStatuses.includes(uri),
         );
+        if (isOpenQueryResultsInTabByDefaultEnabled()) {
+            await this._queryResultWebviewController.createPanelController(queryRunner.uri);
+            await new Promise<void>((resolve) => setTimeout(resolve, 300));
```
ssreerama: Are we having any timing issue here? just wanted to understand why we need to wait here for those milliseconds?

aasimkhan30: The query results logic wasn’t waiting for the webview to fully initialize before sending data. Because of this, some initial events were missed, which caused the screen to appear blank for a few seconds until the next update event arrived and refreshed the data.
I’ve switched the flow to a promise-based method so the webview is guaranteed to be ready before any data is sent, making the behavior more reliable.

caohai (approved):

PR #20194, opened 2025-09-18
Cleaning up previous query run and  unnecessary callbacks in query execution code. 

aasimkhan30:
### Fix: Query Execution State Mismatch
There was an issue where the frontend stopped tracking a query run while the query was still executing in STS. In this state, attempting to run the query again from the frontend caused STS to throw the following error:

https://github.com/microsoft/sqltoolsservice/blob/c9283a1458269886a829c43af5a32fa0353b05e6/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs#L955-L959

To resolve this, the frontend now explicitly sends a **cancel request** before starting a new query run, even when the frontend believes no query is executing. This ensures STS does not have a stale running query.  
- The cancel request is safe: it will not error out if no query is running, as shown here: [QueryExecutionService.cs#L493-L501](https://github.com/microsoft/sqltoolsservice/blob/c9283a1458269886a829c43af5a32fa0353b05e6/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/QueryExecutionService.cs#L493-L501).

### Cleanup
1. Removed unused code in `queryRunner`.
2. Removed unnecessary callbacks when starting query runs, to make it easier to debug and step through.

## Code Changes Checklist

### DIFF
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -561,7 +561,7 @@ export default class QueryRunner {
         } catch (error) {
             // TODO: Localize
             this._vscodeWrapper.showErrorMessage(
-                "Something went wrong getting more rows: " + error.message,
+                LocalizedConstants.QueryResult.getRowsError(getErrorMessage(error)),
             );
             void Promise.reject(error);
         }
--- a/src/controllers/queryRunner.ts
+++ b/src/controllers/queryRunner.ts
@@ -231,6 +231,23 @@ export default class QueryRunner {
         return queryCancelResult;
     }
 
+    public async resetQueryRunner(): Promise<void> {
+        try {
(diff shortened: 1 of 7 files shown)


src/controllers/queryRunner.ts
```diff
@@ -603,6 +535,38 @@ export default class QueryRunner {
...
+        try {
+            const queryExecuteSubsetResult = await this._client.sendRequest(
+                QueryExecuteSubsetRequest.type,
+                queryDetails,
+            );
+            if (queryExecuteSubsetResult) {
+                return queryExecuteSubsetResult;
+            }
+        } catch (error) {
+            // TODO: Localize
```
caohai, line 562: Might as well do the loc with this PR since you are already touching this piece.😀

caohai (approved):

Issue #1480, opened 2019-11-08
Switch corresponding results pane when switching query tabs


Tyler-Oreskey:
- MSSQL Extension Version: 1.7.0
- VSCode Version: 1.40.0
- OS Version: Catalina 10.15.1

Description:
As a user I would like for SQL query panes and their corresponding results to stay paired when changing panes.

Steps to reproduce:

1. Create and run at least two SQL queries in different query panes.
2. Switch back and forth between the query panes.
3. Notice the results do not switch with their corresponding pane.

Kenobi-the-2nd:
Agreed, this is definitely needed to improve workflow. Another issue [here ](https://github.com/microsoft/vscode-mssql/issues/1587)mentions the same problem.

NicholasTuttle:
I would certainly like to see this as well. It's one of the biggest reasons I use SSMS over VS Code.

kburtram:
The results pane experience has been completely redesigned since this suggestion was filed, and within the tool windows pane the results grid does track the active editor.  Please log a new issue relate against the current experience if there are further suggestions.  Thanks for logging this issue!

PR #20196, opened 2025-09-18
Fixing new panel location logic and auto panel switching. 

aasimkhan30:
## Description
Fixes: #1480
1. Revealing the corresponding result panel when the active query editor is changed. This can be turned off in settings.
2. Refine where new results panels open based on user preference. It leads to more predictable result panel opening based on config's “Split Results To” is set to current/next/end

1. Current: Always open in current editor.
2. Next: Always open in next column. Max column is 9
3. End: Always open in the last visible column + 1. Max column is 9. 

## Code Changes Checklist

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -226,5 +226,6 @@
     "mssql.connectionSharing.clearAllConnectionSharingPermissions": "Clear All Connection Sharing Permissions",
     "mssql.statusBar.connectionInfoMaxLength.description": "The maximum number of characters to display for the connection info in the status bar. Set to -1 for no limit.",
     "mssql.schemaDesigner.enableExpandCollapseButtons": "Enable expand/collapse buttons in Schema Designer table nodes when tables have more than 10 columns",
-    "mssql.connectionManagement.rememberPasswordsUntilRestart": "Temporarily store passwords for connections with 'Saved Passwords' disabled, until the extension is restarted. This prevents repeated password prompts when reusing connections within the same session."
+    "mssql.connectionManagement.rememberPasswordsUntilRestart": "Temporarily store passwords for connections with 'Saved Passwords' disabled, until the extension is restarted. This prevents repeated password prompts when reusing connections within the same session.",
+    "mssql.autoRevealResultsPanel": "Automatically reveal the results panel when switching to an editor with query results."
 }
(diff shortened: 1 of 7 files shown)


package.nls.json
```diff
@@ -226,5 +226,6 @@
     "mssql.connectionSharing.clearAllConnectionSharingPermissions": "Clear All Connection Sharing Permissions",
     "mssql.statusBar.connectionInfoMaxLength.description": "The maximum number of characters to display for the connection info in the status bar. Set to -1 for no limit.",
     "mssql.schemaDesigner.enableExpandCollapseButtons": "Enable expand/collapse buttons in Schema Designer table nodes when tables have more than 10 columns",
-    "mssql.connectionManagement.rememberPasswordsUntilRestart": "Temporarily store passwords for connections with 'Saved Passwords' disabled, until the extension is restarted. This prevents repeated password prompts when reusing connections within the same session."
+    "mssql.connectionManagement.rememberPasswordsUntilRestart": "Temporarily store passwords for connections with 'Saved Passwords' disabled, until the extension is restarted. This prevents repeated password prompts when reusing connections within the same session.",
+    "mssql.autoRevealResultsPanel": "Automatically reveal the results panel when switching to an editor with query results."
```
caohai: Should we add something like "only effective when the query result is opened in a new tab instead of in the query pane"?

src/queryResult/queryResultWebViewController.ts
```diff
@@ -62,13 +62,31 @@ export class QueryResultWebviewController extends ReactWebviewViewController<
...
+                        /**
+                         * If the results are shown in webview view, then update the state to show the
+                         * results for the active editor
+                         */
+                        this.state = this.getQueryResultState(uri);
+                    } else {
+                        /**
+                         * If there is no results for the active editor, then show the splash screen
+                         */
+                        this.showSplashScreen();
```
caohai: If a user is not running any query and has the bottom area hidden, just editing a `.sql` file, will this line cause the splash screen to show up if they switch tabs?

aasimkhan30: Yeah, it will cause the splash screen to show up.

caohai (approved):

PR #20217, opened 2025-09-23
Publish Dialog - Adding Target field functionality

ssreerama:

## Description
This pull request adds the target field appearance and validations on target option selection.
- When the local container option is selected, four new fields appears with validations
    - Port, Admin password, confirm admin password and docker image
-  Added tests more maintainable and focused.
-  Additionally, new form fields, validation messages, and labels have been added for container and Azure SQL server scenarios.
- toggle enable and disable Publish and script generation buttons based on target field selection and with successful validation 

## Code Changes Checklist

image

1 TargetSection

### DIFF
--- a/localization/xliff/vscode-mssql.xlf
+++ b/localization/xliff/vscode-mssql.xlf
@@ -2596,11 +2596,11 @@
       <source xml:lang="en">Remove {0}</source>
       <note>{0} is the object type</note>
     </trans-unit>
-    <trans-unit id="++CODE++4850b174b713d88cfc63de107830d5388929020e78abc91fc19bba7a6821625f">
-      <source xml:lang="en">Required</source>
-    </trans-unit>
     <trans-unit id="++CODE++0b73ca14b5d89fd5c289af312a66842146e766baf6c5018848b36f086944cb0b">
       <source xml:lang="en">Replication</source>
+    </trans-unit>
+    <trans-unit id="++CODE++4850b174b713d88cfc63de107830d5388929020e78abc91fc19bba7a6821625f">
+      <source xml:lang="en">Required</source>
     </trans-unit>
     <trans-unit id="++CODE++daee7606b339f3c339076fe2c9f372a3ff40c8ee896005d829c7481b64ca5303">
       <source xml:lang="en">Reset</source>


src/publishProject/dockerUtils.ts
```diff
@@ -0,0 +1,213 @@
...
+    }
+    const n = Number(str);
+    return n >= 1 && n <= constants.MAX_PORT_NUMBER;
+}
+
+/**
+ * Returns true if password meets SQL complexity (length 8-128, does not contain login name,
+ * and contains at least 3 of 4 categories: upper, lower, digit, symbol).
+ */
+export function isValidSqlAdminPassword(password: string, userName = "sa"): boolean {
```
aasimkhan30: Maybe these should be in a more general utility function. Keeping it in dockerUtils prevents usage in other features that have nothing to do with docker. Same with other general SQL based utilities function.

aasimkhan30 (dismissed):

src/publishProject/dockerUtils.ts
ssreerama: I can move them to the ProjectsUtils file, I placed them here because those pwd fields can be only visible when the local container is selected.

lewis-sanchez (dismissed):

TEST_UPDATES_SUMMARY.md
```diff
@@ -0,0 +1,198 @@
+# Test Updates Summary
```
Benjin: This shoudn't be getting checked in

test/unit/publishProjectDialog.test.ts
```diff
@@ -0,0 +1,593 @@
+/*---------------------------------------------------------------------------------------------
```
Benjin: What's the difference between this test file and the other one in this PR? Aren't both test files for publishProjectWebViewController?

test/unit/publishProjectDialog.test.ts
```diff
@@ -0,0 +1,593 @@
...
+    let mockOutputChannel: vscode.OutputChannel;
+    let workspaceConfigStub: sinon.SinonStub;
+
+    const projectPath = "c:/work/ContainerProject.sqlproj";
+
+    setup(() => {
+        sandbox = sinon.createSandbox();
+
+        // Create mock output channel
+        mockOutputChannel = {
```
Benjin: We have premade stubs for this, vscodeWrapper, and extension context.

test/unit/publishProjectDialog.test.ts
```diff
@@ -0,0 +1,593 @@
...
+    validateSqlServerPortNumber,
+    isValidSqlAdminPassword,
+} from "../../src/publishProject/projectUtils";
+
+/**
+ * UI and Form interaction tests for Publish Project Dialog
+ * Tests form field behavior, visibility, user interactions, and validators
+ * Controller/state logic tests are in publishProjectWebViewController.test.ts
+ */
+suite("PublishProjectWebViewController", () => {
```
Benjin: Please go through all the generated tests and confirm that they're actually testing useful and relevant things.  Looks like this file has a ton of tests that test side-effects more than actual functionality, and has a bunch of duplicated code, too.

Benjin:
Please make sure that the commit descriptions - whether human-written or generated - are concise, relevant, and accurate.  For example:
* unless the purpose of the PR is specifically to localize strings, no need to bring that up
* no files were migrated from typemoq to sinon according to this diff.  The test file that was edited never included any typemoq files to begin with.

----

### It's important to review your code, docs, and PRs yourself before asking other people to spend their time reviewing them.

This is a good practice for manually-written code/docs/PRs, but _especially_ important for generated stuff.

test/unit/publishProjectDialog.test.ts
ssreerama: I have created this test file to specifically test publish dialog rendering fields validations and changes. The other file is for controller logic. They seem to be similar as we have very few changes yet.

My intension was to add more tests that covers 100% of UI operations that can be performed in the dialog. ex: toggle fields on target options selection, enable/disable buttons on successful validations etc.

ssreerama: I'm moving all tests to single file and concise the tests that feels the similar testing pattern.

src/constants/constants.ts
```diff
@@ -296,3 +296,45 @@ export const x64 = "x64";
...
+    ContainerImageTag: "containerImageTag",
+    AcceptContainerLicense: "acceptContainerLicense",
+} as const;
+export const PublishFormContainerFields = [
+    PublishFormFields.ContainerPort,
+    PublishFormFields.ContainerAdminPassword,
+    PublishFormFields.ContainerAdminPasswordConfirm,
+    PublishFormFields.ContainerImageTag,
+    PublishFormFields.AcceptContainerLicense,
+] as const;
```
Benjin: Looks like types belong in `sharedInterfaces/publishDialog.ts` instead of here.  They aren't really constants, but more defining data shapes.

src/constants/constants.ts
```diff
@@ -296,3 +296,45 @@ export const x64 = "x64";
...
+export const dockerImageDefaultTag = "latest";
+export const AzureSqlDbFullDockerImageName = "Azure SQL Database (Edge emulation)";
+export const SqlServerDockerImageName = "SQL Server";
+export const MAX_PORT_NUMBER = 65535;
+export const SqlServerName = "SQL server";
+export const AzureSqlServerName = "Azure SQL server";
+export const DefaultSqlPortNumber = "1433";
+export const RequiredFieldMessage = "Required";
+export const DefaultAdminUsername = "sa";
+export const LicenseAcceptanceMessage = "You must accept the license";
```
Benjin: Localize

src/constants/constants.ts
```diff
@@ -296,3 +296,45 @@ export const x64 = "x64";
...
+export const licenseAgreementUrl =
+    "https://github.com/microsoft/containerregistry/blob/main/legal/Container-Images-Legal-Notice.md";
+export const dockerImageDefaultTag = "latest";
+export const AzureSqlDbFullDockerImageName = "Azure SQL Database (Edge emulation)";
+export const SqlServerDockerImageName = "SQL Server";
+export const MAX_PORT_NUMBER = 65535;
+export const SqlServerName = "SQL server";
+export const AzureSqlServerName = "Azure SQL server";
+export const DefaultSqlPortNumber = "1433";
+export const RequiredFieldMessage = "Required";
```
Benjin: Localize

src/constants/constants.ts
```diff
@@ -296,3 +296,45 @@ export const x64 = "x64";
...
+// SQL Database Projects Constants
+export const DSP_PREFIX = "Microsoft.Data.Tools.Schema.Sql.Sql";
+export const DSP_SUFFIX = "DatabaseSchemaProvider";
+export const sqlServerDockerRegistry = "mcr.microsoft.com";
+export const sqlServerDockerRepository = "mssql/server";
+export const sqlServerEulaLink = "https://aka.ms/mssql-container-license";
+export const licenseAgreementUrl =
+    "https://github.com/microsoft/containerregistry/blob/main/legal/Container-Images-Legal-Notice.md";
+export const dockerImageDefaultTag = "latest";
+export const AzureSqlDbFullDockerImageName = "Azure SQL Database (Edge emulation)";
```
Benjin, line 303: Should this (and the other image names) be localized?  Might need confirmation with PM because it's branding

src/publishProject/dockerUtils.ts
```diff
@@ -0,0 +1,232 @@
...
+        const finds = regex.exec(imageName);
+        if (finds) {
+            // 0 is the full match and 1 is the number with pattern inside the first ()
+            return +finds[1];
+        }
+    }
+    return undefined;
+}
+
+// Extract a version year from a target platform string
```
Benjin: This should be an actual docstring, like you did for the other functions

src/publishProject/dockerUtils.ts
```diff
@@ -0,0 +1,232 @@
...
+    };
+}
+
+/**
+ * Loads Docker tags for a given target version and updates form component options
+ * @param targetVersion project target version
+ * @param tagComponent form component to update with tags
+ * @param formState current form state to set default tag if needed
+ */
+export async function loadDockerTags(
```
Benjin: I'm wary of this entire file.  Why are we using a different method for populating the list of available docker images from the docker deployment UI?  Seems odd that Publish Project fetches the list of available tags from the server, but the deployment UI has it hard-coded.

Virtually all of this logic should be shared between the two features, IMO.

We should be consistent so that 1) users know what to expect, and 2) we can reuse code.  If this is a PM decision, can you have your PM reach out?

src/publishProject/formComponentHelpers.ts
```diff
@@ -3,36 +3,89 @@
...
+
+/**
+ * Checks if preview features are enabled in VS Code settings for SQL Database Projects.
+ * @returns true if preview features are enabled, false otherwise
+ */
+function isPreviewFeaturesEnabled(): boolean {
+    return (
+        vscode.workspace
+            .getConfiguration(constants.DBProjectConfigurationKey)
+            .get<boolean>(constants.enablePreviewFeaturesKey) ?? false
```
Benjin: Looks like this feature flag is just for the SqlProj features.  Can you rename the variable name to something like `enableSqlProjPreviewFeaturesKey`?

src/publishProject/formComponentHelpers.ts
```diff
@@ -3,36 +3,89 @@
...
 } from "../sharedInterfaces/publishDialog";
-import { PublishProject as Loc } from "../constants/locConstants";
+import { getPublishServerName, validateSqlServerPortNumber } from "./projectUtils";
+import { validateSqlServerPassword } from "../deployment/dockerUtils";
+
+/**
+ * Checks if preview features are enabled in VS Code settings for SQL Database Projects.
+ * @returns true if preview features are enabled, false otherwise
+ */
+function isPreviewFeaturesEnabled(): boolean {
```
Benjin: I know this is only getting used in this file, but I think it would still make more sense for this to be in another file, maybe projectUtils.ts

src/publishProject/formComponentHelpers.ts
Benjin: `?? false` shouldn't be necessary, since the SqlProj extension should register a default value if it's unset.  or if that doesn't work, you can do `.get<boolean>(keyName, defaultValue)` instead to follow the VS Code pattern more.

src/publishProject/projectUtils.ts
```diff
@@ -0,0 +1,172 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as mssql from "vscode-mssql";
+import * as constants from "../constants/constants";
+import { SqlProjectsService } from "../services/sqlProjectsService";
+
+// Shape returned by sqlProjectsService.getProjectProperties (partial, only fields we use)
```
Benjin: We shouldn't maintain a separate structure if we can avoid it.  Does `Partial<GetProjectPropertiesResult>` work here instead?

src/publishProject/projectUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+    defaultCollation: string;
+    databaseSchemaProvider: string;
+    projectStyle: unknown;
+    targetVersion?: string;
+}
+
+/**
+ * Target platforms for a sql project
+ */
+export const enum SqlTargetPlatform {
```
Benjin, line 27: Can you add the same note here about them matching the values in Microsoft.Data.Tools.Schema.SchemaModel.SqlPlatformNames?

src/publishProject/projectUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+    sqlServer2012 = "SQL Server 2012",
+    sqlServer2014 = "SQL Server 2014",
+    sqlServer2016 = "SQL Server 2016",
+    sqlServer2017 = "SQL Server 2017",
+    sqlServer2019 = "SQL Server 2019",
+    sqlServer2022 = "SQL Server 2022",
+    sqlAzure = "Azure SQL Database",
+    sqlDW = "Azure Synapse SQL Pool",
+    sqlEdge = "Azure SQL Edge",
```
Benjin: Can you double-check that we still should have Edge here?  I thought it was no longer supported.

src/publishProject/projectUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+    [SqlTargetPlatform.sqlDW, "Dw"],
+    [SqlTargetPlatform.sqlDwServerless, "Serverless"],
+    [SqlTargetPlatform.sqlDwUnified, "DwUnified"],
+    [SqlTargetPlatform.sqlDbFabric, "DbFabric"],
+]);
+
+/**
+ * Get project properties from the tools service and extract the target platform version portion
+ * of the DatabaseSchemaProvider (e.g. 130, 150, 160, AzureV12, AzureDw, etc.).
+ * Returns undefined if it cannot be determined or the service call fails.
```
Benjin: This line should be `@returns ...` so that jsdoc picks it up properly

src/publishProject/projectUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+            constants.DSP_PREFIX.length,
+            dsp.length - constants.DSP_SUFFIX.length,
+        );
+        // Basic sanity check: version should be non-empty and alphanumeric-ish
+        if (!version || /[^A-Za-z0-9]/.test(version)) {
+            return undefined;
+        }
+        return version;
+    } catch {
+        return undefined; // swallow errors to keep publish dialog resilient
```
Benjin, line 92: How does the dialog respond if any of these return undefined?

src/publishProject/projectUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+    try {
+        if (!projectFilePath) {
+            return undefined;
+        }
+        const result = await sqlProjectsService.getProjectProperties(projectFilePath as string);
+        if (!result?.success) {
+            return undefined;
+        }
+        const version = await getProjectTargetVersion(sqlProjectsService, projectFilePath);
+        const props: ProjectProperties = {
```
Benjin: You should do this with a spread operator instead so that you don't have to list each property (and risk missing future changes).

src/publishProject/projectUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+        };
+        return props;
+    } catch {
+        return undefined;
+    }
+}
+
+export function getPublishServerName(target: string) {
+    return target ===
+        targetPlatformToVersion.get("Azure SQL Database" /* SqlTargetPlatform.sqlAzure */)
```
Benjin: This should pass the enum instead of hardcoding the string, even if you have to do sqlAzure.toString()

src/publishProject/projectUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+        ? constants.AzureSqlServerName
+        : constants.SqlServerName;
+}
+
+/*
+ * Validates the SQL Server port number.
+ */
+export function validateSqlServerPortNumber(port: string | number | undefined): boolean {
```
Benjin: Why not just take a number and validate that it's a whole number between 1 and MAX_PORT_NUMBER?  let the caller worry about undefined and string conversion.

src/publishProject/projectUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+    const n = Number(str);
+    return n >= 1 && n <= constants.MAX_PORT_NUMBER;
+}
+
+/**
+ * Returns true if password meets SQL complexity (length 8-128, does not contain login name,
+ * and contains at least 3 of 4 categories: upper, lower, digit, symbol).
+ */
+export function isValidSqlAdminPassword(password: string, userName = "sa"): boolean {
```
Benjin: Is this logic that can be reused/shared from the container deployment code?  I know there's a password complexity check over there as well.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -70,18 +73,74 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                        ...props,
+                    } as {
+                        [key: string]: unknown;
+                        targetVersion?: string;
+                    };
+                    projectTargetVersion = props.targetVersion;
+                }
+            }
+        } catch {
+            // swallow errors; keep dialog resilient
```
Benjin: You shouldn't swallow these entirely.  At least log them and emit telemetry.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -70,18 +73,74 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                    this._sqlProjectsService,
+                    projectFilePath,
+                );
+                if (props) {
+                    this.state.projectProperties = {
+                        ...props,
+                    } as {
+                        [key: string]: unknown;
+                        targetVersion?: string;
+                    };
```
Benjin: What's going on here?  Why doesn't `readProjectProperties()` return a type that already has `targetVersion` defined?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -70,18 +73,74 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
         void this.updateItemVisibility();
     }
 
+    /** Registers all reducers in pure (immutable) style */
     private registerRpcHandlers(): void {
         this.registerReducer("publishNow", async (state: PublishDialogState) => {
             // TODO: implement actual publish logic (currently just clears inProgress)
-            state.inProgress = false;
-            return state;
+            return { ...state, inProgress: false };
```
Benjin: Why this change?  It was better the first way because you're not recreating the state object

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -96,29 +155,47 @@ export class PublishProjectWebViewController extends FormWebviewController<
 
         this.registerReducer(
             "savePublishProfile",
-            async (state: PublishDialogState, _payload: { publishProfileName?: string }) => {
+            async (state: PublishDialogState, _payload: { publishProfileName: string }) => {
```
Benjin: Why is the type getting defined here?  That should happen automatically by defining the reducers in sharedInterfaces.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -96,29 +155,47 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
 
-    protected getActiveFormComponents(_state: PublishDialogState) {
-        return [...PublishProjectWebViewController.mainOptions];
+    protected getActiveFormComponents(state: PublishDialogState): (keyof IPublishForm)[] {
+        const activeComponents: (keyof IPublishForm)[] = [
+            constants.PublishFormFields.PublishTarget,
+            constants.PublishFormFields.PublishProfilePath,
+            constants.PublishFormFields.ServerName,
+            constants.PublishFormFields.DatabaseName,
+        ] as (keyof IPublishForm)[];
```
Benjin: This second cast shouldn't be necessary

src/reactviews/common/utils.ts
```diff
@@ -118,3 +118,48 @@ export function getEOL(): string {
...
+ * Converts <a href="url" ...>text</a> tags into React-compatible elements.
+ *
+ * @param html - HTML string potentially containing anchor tags
+ * @returns Object with parts array (text/link segments) or undefined if no HTML
+ *
+ * @example
+ * const result = parseHtmlLabel('I accept the <a href="https://example.com">Terms</a>');
+ * // Returns: { parts: ['I accept the ', { href: 'https://example.com', text: 'Terms' }] }
+ */
+export function parseHtmlLabel(
```
Benjin: I don't love this way to do inline links.  It feels fragile.  Can this be done the same way we did in `localContainerHelpers.ts` with no additional parsing funkiness and get rid of this method entirely?

src/reactviews/pages/PublishProject/components/FormFieldComponents.tsx
```diff
@@ -3,122 +3,147 @@
...
-    PublishProjectProvider,
-} from "../../../../sharedInterfaces/publishDialog";
+import type { PublishDialogFormItemSpec } from "../../../../sharedInterfaces/publishDialog";
 
-/**
- * Renders a form input field for the Publish Project dialog.
- * This is a shared helper to avoid duplication until the form infrastructure is upgraded.
- */
+// Helper to get validation state from component
+const getValidationState = (
```
Benjin: use `function` syntax instead when possible; cleaner to read.

src/reactviews/pages/PublishProject/components/FormFieldComponents.tsx
```diff
@@ -3,122 +3,147 @@
...
+                onBlur={() => options?.onBlur?.(value)}
+                contentAfter={
+                    isPasswordField && options?.onTogglePassword ? (
+                        <Button
+                            onClick={options.onTogglePassword}
+                            icon={options.showPassword ? <EyeRegular /> : <EyeOffRegular />}
+                            appearance="transparent"
+                            size="small"
+                            aria-label={options.showPassword ? "Hide password" : "Show password"}
+                            title={options.showPassword ? "Hide password" : "Show password"}
```
Benjin: Aria labels and titles should be localized

src/reactviews/pages/PublishProject/components/PublishTargetSection.tsx
```diff
@@ -3,24 +3,269 @@
...
+        display: "flex",
+        flexDirection: "column",
+        gap: "16px",
+        paddingLeft: "16px",
+        borderLeft: "2px solid var(--vscode-editorWidget-border, #8883)",
+    },
+});
 
 export const PublishTargetSection: React.FC = () => {
+    const classes = useStyles();
```
Benjin: name these "styles" to keep consistent

src/reactviews/pages/PublishProject/components/PublishTargetSection.tsx
```diff
@@ -3,24 +3,269 @@
...
-        return undefined;
+    // Auto-populate defaults and revalidate passwords
+    useEffect(() => {
+        if (!publishCtx || !isContainer) {
+            return;
+        }
+
+        // Default container port if not set
+        if (!portValue) {
+            publishCtx.formAction({
```
Benjin, line 98: Please take a look at how the other form components are structured.  We shouldn't be repeating onChange and onBlur for every input.

src/reactviews/pages/PublishProject/components/PublishTargetSection.tsx
```diff
@@ -3,24 +3,269 @@
...
+                propertyName: PublishFormFields.ContainerAdminPasswordConfirm,
+                isAction: false,
+                value: confirmPasswordValue as string,
+                updateValidation: true,
+            });
+        }
+    }, [isContainer, passwordValue, publishCtx]);
+
+    if (!publishCtx || !targetComponent) {
+        return <></>;
```
Benjin: why not `undefined`?

src/reactviews/pages/PublishProject/publishProject.tsx
```diff
@@ -35,33 +36,56 @@ function PublishProjectDialog() {
...
               )
             : false;
 
-    // Buttons should be disabled when:
-    // - Component is not ready (missing context, form components, or form state)
-    // - Operation is in progress
-    // - Form has validation errors
-    const readyToPublish = !isComponentReady || inProgress || hasValidationErrors;
+    // Identify missing required values for visible components (treat empty string / whitespace as missing)
+    const hasMissingRequiredValues =
```
Benjin: All this should be handled automatically in the controller.  If you're handling it here, then there's no point in using the form framework.

test/unit/publishProjectWebViewController.test.ts
```diff
@@ -53,4 +58,268 @@ suite("PublishProjectWebViewController Tests", () => {
...
+
+        await controller.initialized.promise;
+
+        // Initial state
+        expect(controller.state.inProgress).to.be.false;
+        expect(controller.state.lastPublishResult).to.be.undefined;
+
+        // Can be updated
+        controller.state.inProgress = true;
+        expect(controller.state.inProgress).to.be.true;
```
Benjin: Is there a reason it wouldn't be able to be updated?  I'm not sure why this is useful.

test/unit/publishProjectWebViewController.test.ts
```diff
@@ -53,4 +58,268 @@ suite("PublishProjectWebViewController Tests", () => {
...
+    });
+
+    test("container target values are properly saved to formState", async () => {
```
Benjin: I know this duplication of formAction logic is temporary until we can convert the old stuff to the new one, but surely we just need to test one case here, since they're all the same (except when `isAction` changes)

test/unit/publishProjectWebViewController.test.ts
```diff
@@ -53,4 +58,268 @@ suite("PublishProjectWebViewController Tests", () => {
...
+                databaseSchemaProvider:
+                    "Microsoft.Data.Tools.Schema.Sql.SqlAzureV12DatabaseSchemaProvider",
+            }),
+        };
+
+        const controller = new PublishProjectWebViewController(
+            contextStub,
+            vscodeWrapperStub,
+            "c:/work/AzureProject.sqlproj",
+            mockSqlProjectsService as any, // eslint-disable-line @typescript-eslint/no-explicit-any
```
Benjin: better to cast the `mockSqlProjectService` properly above than here.  Less error prone if changes are made.

src/constants/constants.ts
ssreerama: I'll check with Iqra!

src/publishProject/projectUtils.ts
ssreerama: yeah, we shouldn't use the duplication, I can use the existing interface here!

src/publishProject/projectUtils.ts
ssreerama: Updating the note accordingly, these names are how we display the available target platform names in sqlproj UI, not coming from dacfx.

src/publishProject/projectUtils.ts
ssreerama: when returned undefined, dialog still opens, but the docker images dropdown will be empty. In such scenario, we could show a warning message to the user or provide a default docker tags

src/publishProject/projectUtils.ts
ssreerama: I think i have missed cleaning this one, I'm already using the deployment validation from the docker utils and left this uncleaned here.

src/reactviews/pages/PublishProject/components/PublishTargetSection.tsx
ssreerama: yea it should undefined or null!

src/publishProject/dockerUtils.ts
ssreerama: I'll check what the current deployment UI has.

Initially I moved the method from sql-database-projects and refactored it to get the docker images of selected target version.

src/publishProject/projectUtils.ts
ssreerama: true.

src/deployment/dockerUtils.ts
```diff
@@ -979,13 +979,93 @@ export async function findAvailablePort(startPort: number): Promise<number> {
...
+}
+
+/**
+ * Retrieves and filters SQL Server container tags based on project target version.
+ * Used by Publish Project dialog to provide granular tag selection.
+ *
+ * @param targetVersion - The SQL Server version (e.g., "160" for SQL Server 2022)
+ * @returns Sorted array of tags filtered by version and organized by year
+ */
+export async function getSqlServerContainerTagsForTargetVersion(
```
ssreerama, line 1003: Updated the publish dialog to use the existing dockerUtils file from the deployment UI instead of the new one.

Currently, the raw image tags are displayed to align with the ADS publish dialog. I’m working with my PM to determine if these should match the deployment UI image tags. If changes are needed, I’ll address them in a separate PR.

This update does not impact the existing implementation.

.vscode/launch.json
```diff
@@ -7,9 +7,15 @@
             "type": "extensionHost",
             "request": "launch",
             "runtimeExecutable": "${execPath}",
-            "args": ["--extensionDevelopmentPath=${workspaceRoot}"],
+            "args": [
```
Benjin: Revert this file for the PR

ssreerama: this should be for my local validation, got merged accidentally!

src/deployment/dockerUtils.ts
Benjin, line 1003: This should also be confirmed with Iqra and (just as importantly) @croblesm.  In the container deployment UX, we are intentionally _only_ allowing users to select the latest/GAed tags for each version because we're told we should not be providing a way for users to install version of SQL that Microsoft doesn't provide support for.  By that understanding, we shouldn't be pulling all the tags, and should just provide options for GA versions, which means this can likely just be hardcoded.

src/publishProject/projectUtils.ts
```diff
@@ -107,66 +112,31 @@ export async function readProjectProperties(
...
 
+/**
+ * Gets the appropriate server name display string based on the target version.
+ * @param target - The target version string (e.g. "AzureV12", "150")
+ * @returns "Azure SQL server" for Azure SQL Database, "SQL server" for all other targets
+ */
 export function getPublishServerName(target: string) {
-    return target ===
-        targetPlatformToVersion.get("Azure SQL Database" /* SqlTargetPlatform.sqlAzure */)
+    return target === targetPlatformToVersion.get(SqlTargetPlatform.sqlAzure)
```
Benjin, line 130: Out of curiosity, what is the target platform value for Fabric?  Is that still `sqlAzure`?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -125,11 +130,23 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
         // Fetch Docker tags for the container image dropdown
-        if (projectTargetVersion) {
-            const tagComponent =
-                this.state.formComponents[constants.PublishFormFields.ContainerImageTag];
-            if (tagComponent) {
-                await loadDockerTags(projectTargetVersion, tagComponent, this.state.formState);
+        // Use the deployment UI function with target version filtering
+        const tagComponent = this.state.formComponents[PublishFormFields.ContainerImageTag];
+        if (tagComponent) {
+            try {
```
Benjin, line 136: This should end up being removed to just use `-latest` for all, but waiting PM confirmation.

Benjin (approved):

aasimkhan30 (approved):

Issue #20223, opened 2025-09-23
[Bug]: Update container image label to show CTP2.0 for macOS M-based users


croblesm:
### Description

When running the Local SQL Server database container feature on macOS with M-based (ARM) processors, selecting **SQL Server 2025 – latest** does not actually pull the latest (RC1) image. Instead, it falls back to the older **CTP2.0** image. Because of the SQL Server 2025 container, known issues with macOS:
https://github.com/microsoft/mssql-docker/issues/940
https://github.com/microsoft/mssql-docker/issues/942
https://github.com/microsoft/mssql-docker/issues/943

This behavior is inconsistent with what the UI shows and causes confusion for users. Intel-based macOS users, as well as Windows and Linux users, are not affected.  

This is the dialog in question:
<img width="1340" height="914" alt="Image" src="https://github.com/user-attachments/assets/4752ae29-9453-41cf-98dc-995178382553" />

**Expected Behavior**
For M-based macOS users, the UI should explicitly display **CTP2.0** instead of **latest**, so that the image being pulled is clear and accurate.  

### Steps to Reproduce

1. On a macOS machine with an Apple M-based (ARM) processor, open VS Code with the MSSQL extension.  
2. Go to **New Deployment → Local SQL Server database container**.  
3. From the **Select image** dropdown, choose **SQL Server 2025 – latest**.  
4. Create the container.  
5. Observe that the container pulls **CTP2.0** instead of **RC1**, despite the UI indicating **latest**.  

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [x] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

- VS Code version: 1.104.1
- MSSQL extension version: 1.36
- macOS 15.6.1

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [x] I want to work on this issue

laurenastrid1:
Closing because this is not the right fix anymore

PR #20227, opened 2025-09-24
Match display name to version tag for broken mac image

laurenastrid1:

## Description
Screenshot 2025-09-24 161913

Fixes https://github.com/microsoft/vscode-mssql/issues/20223

## Code Changes Checklist

### DIFF
--- a/src/deployment/dockerUtils.ts
+++ b/src/deployment/dockerUtils.ts
@@ -59,6 +59,7 @@ export const dockerLogger = Logger.create(
 );
 
 const dockerInstallErrorLink = "https://www.docker.com/products/docker-desktop/";
+const sqlServerImageListLink = "https://mcr.microsoft.com/v2/mssql/server/tags/list";
 // Exported for testing purposes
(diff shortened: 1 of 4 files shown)


src/deployment/dockerUtils.ts
```diff
@@ -587,14 +588,15 @@ export async function getDockerPath(executable: string): Promise<string> {
 /**
  * Temp fix for the SQL Server 2025 version issue on Mac.
  * Returns the last working version of SQL Server 2025 for Mac.
+ * If we are getting the display name, add some formatting
  */
-export function constructVersionTag(version: string): string {
+export function constructVersionTag(version: string, isVersionDisplayName?: boolean): string {
```
Benjin, line 593: @croblesm and I were chatting about some additional improvements.

One idea is to include an info ℹ️ icon + tooltip to explain why it's the not `latest` when the other options are, another is to add some simple/clean way for a user to choose a tag other than `latest` after having selected the major version.  But I don't think that needs to happen in this PR.

PR #20231, opened 2025-09-25
Adding Tool Tip to aid into Differentiating Connections

allancascante:
## Description

Adding tooltips to OE including the attributes having not default values for each connection.

image

## Code Changes Checklist

### DIFF
--- a/localization/l10n/bundle.l10n.json
+++ b/localization/l10n/bundle.l10n.json
@@ -728,6 +728,8 @@
     "Object Explorer Filter": "Object Explorer Filter",
     "Azure MFA": "Azure MFA",
     "Windows Authentication": "Windows Authentication",
+    "Enabled": "Enabled",
+    "Disabled": "Disabled",
     "User": "User",
     "Port": "Port",
     "SQL Container Name": "SQL Container Name",
--- a/localization/xliff/vscode-mssql.xlf
+++ b/localization/xliff/vscode-mssql.xlf
@@ -1038,6 +1038,9 @@
     <trans-unit id="++CODE++6fbed214e0ea21c683db0ba2d648ef09b11edf4b4bb23835d4fdda84b5d097f3">
       <source xml:lang="en">Disable intellisense and syntax error checking on current document</source>
     </trans-unit>
+    <trans-unit id="++CODE++75081b593d15cf6e631971bc6768723f593b88b172477e40ae7d363e4829816d">
+      <source xml:lang="en">Disabled</source>
+    </trans-unit>
     <trans-unit id="++CODE++acfc5be785a9bb3da11c9f05adb69bd5633096f09002aa44813a1b3207e912ad">
       <source xml:lang="en">Disconnect</source>
     </trans-unit>
@@ -1124,6 +1127,9 @@
     </trans-unit>
     <trans-unit id="++CODE++45e33ceba31737f1cb793c231fafc9dd99074ca1f26476a70b0c830ef155c9d3">
       <source xml:lang="en">Enable Trust Server Certificate</source>
+    </trans-unit>
+    <trans-unit id="++CODE++92c1cdfdf4cb9cf6fcca962f206de36fd5d60db1178bc9461052f8de703a0e06">
+      <source xml:lang="en">Enabled</source>
     </trans-unit>
     <trans-unit id="++CODE++76ad7b53af548367c78d7216972ce81ee53ecf7ee1d6fe5f3d16a73633bd4e49">
       <source xml:lang="en">Encountering a problem?  Share the details with us by opening a GitHub issue so we can improve!</source>
@@ -3735,6 +3741,9 @@
     <trans-unit id="mssql.objectExplorer.disableGroupBySchema">
       <source xml:lang="en">Disable Group By Schema</source>
     </trans-unit>
+    <trans-unit id="mssql.objectExplorer.Tooltip.disabledLabel">
+      <source xml:lang="en">Disabled</source>
+    </trans-unit>
     <trans-unit id="mssql.disconnect">
       <source xml:lang="en">Disconnect</source>
     </trans-unit>
@@ -3845,6 +3854,9 @@
     </trans-unit>
     <trans-unit id="mssql.walkthroughs.getStarted.enableModernFeatures.description">
       <source xml:lang="en">Enable the new set of data development features that provide a modern way to work with your SQL database in VS Code.&#10;[Enable New Experiences](command:mssql.enableRichExperiences)</source>
+    </trans-unit>
+    <trans-unit id="mssql.objectExplorer.Tooltip.enabledLabel">
+      <source xml:lang="en">Enabled</source>
     </trans-unit>
     <trans-unit id="mssql.enableRichExperiences.description">
       <source xml:lang="en">Enables UI-based features in the MSSQL extension for richer and more powerful features. Restart Visual Studio Code after changing this setting.</source>


src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,48 @@ export class ConnectionNode extends TreeNodeInfo {
...
+            trustServerCertificate: false,
+            persistSecurityInfo: false,
+            azureAuthType: AzureAuthType.AuthCodeGrant,
+            multipleActiveResultSets: false,
+            connectTimeout: Constants.defaultConnectionTimeout,
+            commandTimeout: Constants.defaultCommandTimeout,
+            applicationName: Constants.connectionApplicationName,
+            savePassword: false,
+            emptyPasswordInput: false,
+            profileSource: 0,
```
Benjin: Let's exclude any that aren't part of the actual connection info, so:
`profileSource`, `savePassword`, `emptyPasswordInput`, and `azureAuthType` as well.

src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,48 @@ export class ConnectionNode extends TreeNodeInfo {
...
+            }
+
+            if (key in defaultValues && value === defaultValues[key]) {
+                return;
+            }
+
+            props.push(`${key}: ${value}`);
+        });
+
+        return props.length > 0 ? props.join("\n") : "";
```
Benjin: if props is empty, will this show an empty tooltip, or the default tooltip (just the display text of the node)?

src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,48 @@ export class ConnectionNode extends TreeNodeInfo {
...
             this.iconPath = ObjectExplorerUtils.iconPath(ICON_SERVER_DISCONNECTED);
         }
+
+        // Tooltip logic: show all non-default properties except those in the label (database, user, server)
+        this.tooltip = this.getConnectionTooltip(connectionProfile);
+    }
+
+    private getConnectionTooltip(connectionProfile: IConnectionProfile): string {
+        // Properties to exclude
+        const exclude = ["database", "user", "server", "profileName", "id", "groupId"];
```
aasimkhan30: Let's not exclude server, database, user. These will be helpful when the connection has a profile name.

aasimkhan30: Also make sure we are handling different auth types properly. For example, windows auth shouldn't have user, Entra auth should have Entra login id.

src/objectExplorer/nodes/connectionNode.ts
allancascante: tooltip for "" works as not having a tooltip at all, I'll update the code to not do anything if the results is ""

src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,82 @@ export class ConnectionNode extends TreeNodeInfo {
...
+        //handle auth types properly, if auth type is integrated or azureMfa don't show user in tooltip
+        if (
+            connectionProfile.authenticationType === Constants.integratedauth ||
+            connectionProfile.authenticationType === Constants.azureMfa
+        ) {
+            exclude.push("user");
+        }
+
+        // if connection has a profile name, don't exclude server, user and database from tooltip
+        if (connectionProfile.profileName) {
```
aasimkhan30: Lets include these details even if profile name isn't present. A lot of times database and user gets truncated from the autogenerated name. The only way to see everything will be to resize OE (which isn't the best experience)

aasimkhan30 (changes requested):

src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,82 @@ export class ConnectionNode extends TreeNodeInfo {
...
+            trustServerCertificate: (connectionNodeDefaults as any).trustServerCertificate,
+            persistSecurityInfo: (connectionNodeDefaults as any).persistSecurityInfo,
+            azureAuthType: (connectionNodeDefaults as any).azureAuthType,
+            multipleActiveResultSets: (connectionNodeDefaults as any).multipleActiveResultSets,
+            connectTimeout: (connectionNodeDefaults as any).connectTimeout,
+            commandTimeout: (connectionNodeDefaults as any).commandTimeout,
+            applicationName: (connectionNodeDefaults as any).applicationName,
+            savePassword: (connectionNodeDefaults as any).savePassword,
+            emptyPasswordInput: (connectionNodeDefaults as any).emptyPasswordInput,
+            profileSource: (connectionNodeDefaults as any).profileSource,
```
Benjin: `savePassword`, `emptyPasswordInput`, and `profileSource` are not connection properties (they're metadata to help us save the conenction), so we should exclude all of those.

src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,82 @@ export class ConnectionNode extends TreeNodeInfo {
...
+            "profileSource",
+            "savePassword",
+            "emptyPasswordInput",
+            "azureAuthType",
+            "password",
+        ];
+        // Default values for comparison
+        const connectionNodeDefaults: Partial<IConnectionProfile> = getDefaultConnection();
+
+        const defaultValues = {
```
Benjin: How did you choose which properties to include vs. exclude in this list?  There are a ton (and new get added occasionally), so ideally we're not trying to maintain a separate (and complete) list.

allancascante: I was hoping to have a discussion about what to exclude, in this PR

allancascante: ✔ Always include:

profileName (for accessibility)
server, database
Auth-specific credential (user for SQL Auth, accountId for Entra Auth)

✔ Also include:

Other non-secret, connection-related properties that differ from defaults (because they indicate intentional configuration changes).

❌ Exclude:

Metadata (id, groupId, profileSource, etc.)
Secrets (password, azureAccountToken, etc.)
Entra auth metadata (tenantId, expiresOn)
connectionString

src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,161 @@ export class ConnectionNode extends TreeNodeInfo {
...
+     * - extra non-default properties
+     * It should exclude:
+     * - password or any sensitive info
+     * - properties that are default values or empty
+     * @param connectionProfile The connection profile to generate the tooltip for
+     * @returns the tooltip string or undefined if no properties to show
+     */
+    private getConnectionTooltip(connectionProfile: IConnectionProfile): string | undefined {
+        // Properties to exclude
+        const exclude = [
```
aasimkhan30: Given how big this list is, IMO it will be hard to keep it updated. We might accidentally miss updating it and leak new properties that are not meant to be shown in a tooltip. 

It will be safer to have an allowlist for properties. 

For now, we can just display these props.
```
- profileName if present (without the profileName key)
- server
- database
- creds based on auth type:
- user for sql auth
- email for entra auth.
- Container Name (if container based connection)
- container version (if container based connection).
```

We can add more props in the future if needed. 

@Benjin  and @croblesm what do you think?

Benjin: Conclusion from talking about this offline:

```
[<profileName>] # note: no tag
Server: <serverName>
Database: <db name>
Auth: <username or type>
[Port: <port>]
[SQL Container Name: <container name>]
[SQL Container Version: <version>]
[Application intent: <intent>]
[Connection timeout: <timeout>]
[Command timeout: <timeout>]
[Always encrypted: <enabled/disabled>]
[Replication: <enabled/disabled>]
```

`[optional]` items (square brackets) are only if set/differs from the default value

README.md
aasimkhan30, line 1: Merge from main to get rid of formatting changes.

src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,170 @@ export class ConnectionNode extends TreeNodeInfo {
...
+                defaultValue: (connectionNodeDefaults as any).database,
+            },
+            {
+                key: "authenticationType",
+                value: "",
+                label: "Authentication Type",
+                defaultValue: (connectionNodeDefaults as any).authenticationType,
+            },
+            {
+                key: "user",
```
aasimkhan30, line 118: How about account name for Entra connections.

AGENTS.md
aasimkhan30, line 1: remove this change as well. Just merge from main.

test/unit/AGENTS.md
aasimkhan30, line 1: same here.

src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,170 @@ export class ConnectionNode extends TreeNodeInfo {
...
+            }
+        });
+
+        let lines = props.map((p) => {
+            if (p.value) {
+                if (excludedLabelKeys.find((k) => k === p.key)) {
+                    return `${p.value}`;
+                } else {
+                    if (p.label) {
+                        const localizedLabel = vscode.l10n.t(p.label);
```
aasimkhan30: Loc only for works for static strings. This might not work.

src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,170 @@ export class ConnectionNode extends TreeNodeInfo {
...
+            connectionProfile.authenticationType === Constants.integratedauth
+        ) {
+            // If auth type is not SQL Login, remove user property
+            const userIndex = props.findIndex((p) => p.key === "user");
+            if (userIndex !== -1) {
+                props.splice(userIndex, 1);
+            }
+        }
+
+        Object.keys(connectionProfile).forEach((key) => {
```
aasimkhan30: Why not reduce everything to a singular loop over props. Maybe you can use the reduce function.

src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -59,6 +62,170 @@ export class ConnectionNode extends TreeNodeInfo {
...
+                    } else {
+                        return `${p.label}: ${p.value}`;
+                    }
+                }
+            } else {
+                return undefined;
+            }
+        });
+        lines = lines.filter((line): line is string => line !== undefined);
+        return lines.length > 0 ? lines.join("\n") : undefined;
```
aasimkhan30: use os.eol to join strings.

src/objectExplorer/nodes/connectionNode.ts
allancascante: this is what we capture when we talk about what to include in the tooltip:

Server name
Container name / SQL container name
Database name
Username
Port number (non-default)
SQL container version / Container version
Additional settings
Application intent
Connection timeout
Command timeout
Always encrypted
Replication

So you saying something else is missing? @aasimkhan30

PR #20270, opened 2025-10-02
[CodeCoverage] Enabling CodeCov github actions on vscode-mssql repo

manujoseph85:
## Description

As part of improving our release process and focusing on quality, we need to closely track test code coverage in our repo. Currently, we generate code coverage data but we do no enforce any checks on it. The html code coverage report generated is uploaded as an artifact and is not easily accessible. We only post the overall code coverage data as a comment to the PR.

To improve code coverage and overall quality we need to

- Enforce code coverage checks on each PR
- Present the patch coverage data to the developer on each PR
- Easily view which changes in the PR are missing coverage

CodeCov is the tool that some of the other Microsoft OSS repos are using to achieve these goals. To use the service we need to upload our Cobertura coverage report to CodeCov during CI. We also need to supply a codecov.yml configuration file, with the settings we choose for CodeCov, in the repo root.

This PR adds the step to upload coverage reports to CodeCov using approved github action for CodeCov. It also adds the CodeCov configuration file at the root.

In the first version of the CodeCov configuration we have configured 50% overall coverage target and 70% patch target. We can slowly increase overall target as we achieve more coverage by adding tests.

## Code Changes Checklist


.github/workflows/build-and-test.yml
```diff
@@ -289,6 +289,13 @@ jobs:
         with:
           name: CoverageReport # Artifact name
           path: coveragereport # Directory containing files to upload
+      
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
Benjin, line 296: Can you also send a PR for the internal docs repo (I'll send the link) with instructions for how to regenerate this?

.github/workflows/build-and-test.yml
```diff
@@ -289,6 +289,13 @@ jobs:
         with:
           name: CoverageReport # Artifact name
           path: coveragereport # Directory containing files to upload
+      
+      - name: Upload coverage to Codecov
```
Benjin, line 293: Does this replace or complement the existing step that comments on PRs with code coverage?

Benjin (approved):

.github/workflows/build-and-test.yml
manujoseph85: Yes, I was going to ask for the link to internal docs. Will add a page for CodeCov setup

.github/workflows/build-and-test.yml
manujoseph85: Current step only adds one-line to a comment with overall code coverage. CodeCov will offer much more like patch coverage and line-wise breakdown of hits/misses.

I think we can remove that line for code coverage from the comment after we get comfortable with CodeCov. There are other useful info in that comment like VSIX size which we need to keep.

aasimkhan30:
@manujoseph85 
Does this warning mean that codecov hasn't run on main yet?

image

manujoseph85:
> @manujoseph85 Does this warning mean that codecov hasn't run on main yet?
> 
> image

Yes, that is correct. Once this change goes through, all subsequent PRs should have a base code coverage run to compare against.

manujoseph85:
Linking ADO workitem AB#4724522

caohai (approved):

manujoseph85:
AB#4724522 closed

PR #20274, opened 2025-10-03
Converting TypeMoq tests to sinon

Benjin:
First batch of TypeMoq -> Sinon test conversions, using AI with a good bit of coaching.

I've rolled most of the prompt/rules I've been using into `test/unit/AGENTS.md`.

## Code Changes Checklist


test/unit/AGENTS.md
```diff
@@ -2,27 +2,27 @@
...
-  * Default to chai.expect; when checking Sinon interactions, use sinon-chai.
-  * Avoid Object.defineProperty hacks and (if possible) fake/partial plain objects; use sandbox.createStubInstance(type) and sandbox.stub(obj, 'prop').value(...).
-  * Add shared Sinon helpers to test/unit/utils.ts when they’ll be reused.
-  * If updating preexisting tests, preserve relevant inline comments from the original tests.
-  * When introducing a Sinon helper to replace a TypeMoq helper (e.g., capabilities mock), follow the utils.ts pattern: accept an optional
-  sandbox, create stub instances, and return them.
-  * Avoid unnecessary casts, like `myVar as unknown as MyType` when myVar is already a sinon-stubbed instance of MyType.
-  * Maintain existing formatting conventions, line endings, and text encoding.
-  * Nest test suits as necessary to group tests in a logical manner.
+-   Do not edit application/source files unless the refactor demands it. Confirm before editing files outside of /test/unit, and justify why you need to make those changes.
```
Benjin, line 5: all .md formatting is a side-effect of the new prettier rules.  https://github.com/microsoft/vscode-mssql/pull/20275 addresses those separately, and will be pulled into this PR once merged.

test/unit/databaseObjectSearchService.test.ts
```diff
@@ -14,12 +14,18 @@ import { DatabaseObjectSearchService } from "../../src/services/databaseObjectSe
...
+    let sandbox: sinon.SinonSandbox;
+    let client: sinon.SinonStubbedInstance<SqlToolsServiceClient>;
     let searchService: DatabaseObjectSearchService;
 
     setup(() => {
-        client = TypeMoq.Mock.ofType(SqlToolsServiceClient, TypeMoq.MockBehavior.Loose);
-        searchService = new DatabaseObjectSearchService(client.object);
+        sandbox = sinon.createSandbox();
+        client = sandbox.createStubInstance(SqlToolsServiceClient);
+        searchService = new DatabaseObjectSearchService(client as unknown as SqlToolsServiceClient);
```
lewis-sanchez: Is this the kind of casting you were trying to avoid or is this fine?

test/unit/utils.ts
```diff
@@ -36,21 +39,33 @@ export function stubVscodeWrapper(
...
-    stubber.stub(vscodeWrapper, "outputChannel").get(() => {
-        return outputChannel;
-    });
+    const outputChannel: vscode.OutputChannel = {
+        name: "",
+        append: stubber.stub(),
+        appendLine: stubber.stub(),
+        clear: stubber.stub(),
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        show: stubber.stub() as any,
```
lewis-sanchez, line 50: Why did you need to cast this to any, but not for the others?

test/unit/utils.ts
```diff
@@ -65,3 +80,15 @@ export function stubUserSurvey(
 
     return userSurvey;
 }
+
+export function getMockContext(): vscode.ExtensionContext {
+    return {
+        extensionUri: vscode.Uri.parse("file://test"),
+        extensionPath: "path",
+    } as unknown as vscode.ExtensionContext;
```
lewis-sanchez, line 141: Why did you need to cast to unknown before casting to a specific type?

test/unit/mssqlProtocolHandler.test.ts
```diff
@@ -308,29 +245,14 @@ suite("MssqlProtocolHandler Tests", () => {
     });
 
     suite("readProfileFromArgs", () => {
-        test("Should include connection properties that aren't sourced from SQL Tools Service", async () => {
```
lewis-sanchez, line 311: Were these removed tests moved somewhere else or do we not want them anymore?

test/unit/mssqlProtocolHandler.test.ts
```diff
@@ -154,58 +143,6 @@ suite("MssqlProtocolHandler Tests", () => {
             expect(openConnectionDialogStub).to.not.have.been.called;
         });
 
-        test("Should ignore partially-matching profile when additional specifiers do not match", async () => {
```
lewis-sanchez, line 157: Did we intentionally remove this test or something Copilot did while making changes?

test/unit/queryNotificationHandler.test.ts
```diff
@@ -28,165 +27,114 @@ suite("QueryNotificationHandler tests", () => {
...
-                runnerMock.object.setHasCompleted();
-            });
-
-        // Get handlers
+        runnerMock = sandbox.createStubInstance(QueryRunner);
+        runnerMock.handleQueryComplete.callsFake(() => {
+            runnerMock.setHasCompleted();
+        });
+
+        runner = runnerMock as unknown as QueryRunner;
```
lewis-sanchez: Same question as before about casting to unknown as an intermediary cast. Just commenting in case this needs to change.

test/unit/databaseObjectSearchService.test.ts
Benjin: Exactly the kind I was hoping to avoid with the prompts.  Not all of them can be eliminated (without some type-system sorcery), but thanks for calling out the ones that have snuck through so I can double-check them!

test/unit/utils.ts
Benjin: I wasn't able to get rid of this `any` because `show()` has several overloads that are confusing sinon's extraction of the valid types through the type system.  If `OutputChannel` was a class, I could just stub that directly, but that isn't an option for interfaces.

test/unit/utils.ts
Benjin: These are because:
1. `vscode.Webview` and `vscode.WebviewPanel` are interfaces (so I can't use `createStubInstance()`), and
2. I'm not mocking out everything in those interfaces - just the relevant properties - so the type-checker doesn't allow casting directly (insufficient overlap error).

test/unit/mssqlProtocolHandler.test.ts
Benjin: Good catch; looks like the AI really messed up with this file.

test/unit/mssqlProtocolHandler.test.ts
Benjin: Good catch; AI messed up this file a lot.

lewis-sanchez (approved):

Issue #20187, opened 2025-09-17
[Bug]: Average, Count, and Sum Not Calculating in Query Results when Clicking Row or Column Headers


obviousSailfish:
### Description

Is is expected behavior coming from Azure Data Studio that if you click on a column header or a row header in the query results panel that the average, count, and sum of the selected cells is calculated and shown in the query results footer. This does not appear to work when using the microsoft/vscode-mssql extension in Visual Studio Code.

It appears that the only way to get average, count, and sum calculations to show in the footer is to select a range of cells with the mouse. This is very limiting because there appears to be no way to select anything beyond what is visible on screen when selection by mouse begins.

### Steps to Reproduce

1. Execute query.
2. Click on a column or row header in the query results grid.
3. Observe that average, count, and sum are not calculated.

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [x] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

* VS Code Version: 1.104.0 (Universal)
* MSSQL Extension Version: 1.36.0
* Operating System: macOS Tahoe 26.0
* Connection Type: Azure SQL

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #20277, opened 2025-10-03
Cleaning up Summary

aasimkhan30:

## Description

For large selections:

1. Initial summary
    image
2. Summary while loading 
    image
3. If canceled.
    image
4. on completion
    image

Making sure tooltip shows summary items in new line 

image

## Code Changes Checklist

### DIFF
--- a/localization/l10n/bundle.l10n.json
+++ b/localization/l10n/bundle.l10n.json
@@ -1224,11 +1224,33 @@
         "message": "Average: {0}  Count: {1}  Sum: {2}",
         "comment": ["{0} is the average, {1} is the count, {2} is the sum"]
     },
-    "Average: {0}  Count: {1}  Distinct Count: {2}  Max: {3}  Min: {4}  Null Count: {5}  Sum: {6}/{0} is the average, {1} is the count, {2} is the distinct count, {3} is the max, {4} is the min, {5} is the null count, {6} is the sum": {
-        "message": "Average: {0}  Count: {1}  Distinct Count: {2}  Max: {3}  Min: {4}  Null Count: {5}  Sum: {6}",
-        "comment": [
-            "{0} is the average, {1} is the count, {2} is the distinct count, {3} is the max, {4} is the min, {5} is the null count, {6} is the sum"
-        ]
+    "Average: {0}/{0} is the average": {
+        "message": "Average: {0}",
+        "comment": ["{0} is the average"]
+    },
+    "Count: {0}/{0} is the count": {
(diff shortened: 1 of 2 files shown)


src/queryResult/utils.ts
```diff
@@ -596,3 +596,8 @@ export function calculateSelectionSummaryFromData(
 
     return summary;
 }
+
+export const inMemoryGridDataProcessingThreshold =
```
Benjin: Since the main query results pane is initialized just the once, we're stuck with this value once it's loaded.  It should either get updated in the main controller/state when we detect an update, or be added to the list of settings that requires a reload when it's changed.

src/controllers/queryRunner.ts
```diff
@@ -1069,90 +1072,93 @@ export default class QueryRunner {
...
             this.fireSummaryChangedEvent(requestId, {
-                text: LocalizedConstants.QueryResult.summaryLoadingCanceled,
-                tooltip: LocalizedConstants.QueryResult.summaryLoadingCanceledTooltip,
+                command: {
+                    title: Constants.cmdHandleSummaryOperation,
+                    command: Constants.cmdHandleSummaryOperation,
+                    arguments: [this.uri],
+                },
+                continue: proceed,
+                text: `$(play-circle) ${LocalizedConstants.QueryResult.summaryFetchConfirmation(totalRows)}`,
```
Benjin, line 1090: A couple random questions about the summary (for my own learning):
* Is the summary calculated in MSSQL for just the rows that have been loaded, or for the entire resultset (in STS, I guess)?
* If the first one, is the summary updated as results are streamed in or do we wait until we've loaded everything we will

Benjin (approved):

src/controllers/queryRunner.ts
aasimkhan30: We calculate summaries for selected items using STS. The selection can include only the currently loaded rows, or the entire table (e.g., "Select All" or selecting a column by clicking on column header). Previously, we only performed calculations on loaded rows, which was incorrect.  Bug filed for it: **https://github.com/microsoft/vscode-mssql/issues/20187**

Now, we load the selected rows directly from STS before running calculations. While it would be possible to skip this step when only visible rows are selected, I chose not to optimize for that case. Since users typically select only a handful of rows manually, the extra roundtrip is inexpensive and not worth maintaining separate flows.

Benjin (approved):

lewis-sanchez (approved):

PR #20298, opened 2025-10-13
[CodeCoverage]Adding CodeCov badge to README.md to display CodeCoverage

manujoseph85:
## Description

Now that we have onboarded to CodeCov, we can display a badge with CodeCoverage percentage in README. Lets try to get to the next level there!

## Code Changes Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 [![Build and Test (Unit + E2E)](https://github.com/microsoft/vscode-mssql/actions/workflows/build-and-test.yml/badge.svg?branch=main)](https://github.com/microsoft/vscode-mssql/actions/workflows/build-and-test.yml)
 ![GitHub Discussions](https://img.shields.io/github/discussions/microsoft/vscode-mssql)
+[![codecov](https://codecov.io/github/microsoft/vscode-mssql/graph/badge.svg?token=NXLNAwJgRB)](https://codecov.io/github/microsoft/vscode-mssql)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,6 @@
 [![Build and Test (Unit + E2E)](https://github.com/microsoft/vscode-mssql/actions/workflows/build-and-test.yml/badge.svg?branch=main)](https://github.com/microsoft/vscode-mssql/actions/workflows/build-and-test.yml)
 ![GitHub Discussions](https://img.shields.io/github/discussions/microsoft/vscode-mssql)
+[![codecov](https://codecov.io/github/microsoft/vscode-mssql/graph/badge.svg?token=NXLNAwJgRB)](https://codecov.io/github/microsoft/vscode-mssql)
```
Benjin, line 3: This README also gets packaged with the VSIX when we publish to the VS Code marketplace.  Do we want to advertise our code coverage there (even if we were above our target threshold)?

manujoseph85: Good point. I think Code Coverage is very similar to 'Build and Test' badge above. Will check how that is displayed in the marketplace.

manujoseph85: Looks to me like a lot of OSS projects that have onboarded to codecov display this badge. I think its ok to show this in the marketplace as 50%+ is decent and it shows that we care about quality. Will also be motivation for us to improve this no to 70+.

So, I will go ahead and add this now. We can revisit this later if needed.

Issue #20261, opened 2025-10-01
[Feature Request]: Provide docs about connection-sharing (API)


m4ss1m0g:
### Feature Description

Provide documentation about connection-sharing (API), now all infos are inside issues and PR:
https://github.com/microsoft/vscode-mssql/pull/19542
https://github.com/microsoft/vscode-mssql/issues/19456
https://github.com/microsoft/vscode-mssql/issues/19544

And my discussion about this
https://github.com/microsoft/vscode-mssql/discussions/19519

An example code could be very helpful.

Also, add a notice of this possibility on the main README.

### Problem and Motivation

I would port the SqlPackage extension on ADS, to VScode.

### Related Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

### If you selected "Other", please describe the area

Connection-Sharing, API

### Confirmation

- [x] I have searched existing feature requests and couldn't find a match
- [ ] I want to help implement this feature

kburtram:
@aasimkhan30 let's discuss where we'd like to document our extensibility APIs.  Probably in the wiki?

aasimkhan30:
@m4ss1m0g while I get the docs ready. Why don't you look at the example extension here: 

Please let me know if you have more questions.
https://github.com/microsoft/vscode-mssql/tree/main/samples/connection-sharing-sample

m4ss1m0g:
Thanks!

m4ss1m0g:
I created a brand new extension, and I  tried to copy and paste the `extension.ts` code, but I encountered many problems:

## Missing imports

These imports are missing

``` typescript
import { IConnectionProfile } from 'azdata';
import * as mssql from 'vscode-mssql';
```
From `mssql`, the extension uses _only_ `IConnectionSharingService` from `/typings/vscode-mssql.d.ts`, but importing this file is quite impossible because there are many dependencies.

So I tried to copy the `IConnectionSharingService` and `IConnectionProfile` and add the missing dependencies, below is 
my step to create a working file called `external.ts`

## external.ts

### IConnectionSharingService

[...]

PR #20300, opened 2025-10-14
Adding ability to share connection strings through connection sharing API.

aasimkhan30:

## Description
This in response to https://github.com/microsoft/vscode-mssql/issues/20261#issuecomment-3385362393

The user needs a connection string in order to run some sqlpackage command through their extension. 

This API checks if the user has granted permissions to share connection with the external extension and then shares the connection string for the connection id. 

## Code Changes Checklist

### DIFF
--- a/samples/connection-sharing-sample/src/extension.ts
+++ b/samples/connection-sharing-sample/src/extension.ts
@@ -20,7 +20,7 @@ export async function getConnectionSharingServiceApi(): Promise<
 
     await mssqlExtension.activate();
 
-    const mssqlExtensionApi = mssqlExtension.exports as any;
+    const mssqlExtensionApi = mssqlExtension.exports as mssql.IExtension;
     if (!mssqlExtensionApi) {
         vscode.window.showErrorMessage("Unable to access MSSQL extension API");
         return;


samples/connection-sharing-sample/src/extension.ts
```diff
@@ -5,6 +5,36 @@ import * as mssql from "vscode-mssql";
...
+    if (!mssqlExtension) {
+        vscode.window.showErrorMessage(
+            "MSSQL extension is not installed. Please install it first.",
+        );
+        return;
+    }
+
+    await mssqlExtension.activate();
+
+    const mssqlExtensionApi = mssqlExtension.exports as any;
```
Benjin: Can this be more specific than `as any` (for the sake of being a good example)?

src/extension.ts
```diff
@@ -166,7 +166,13 @@ export async function activate(context: vscode.ExtensionContext): Promise<IExten
                     scriptingObject,
                 );
             },
-        },
+            getConnectionString: (extensionId: string, connectionId: string): Promise<string> => {
```
Benjin, line 169: should we also include a flag for including/censoring the password?  We have the functionality for it already, so it's just a simple matter of piping the param, and - if we default it to censoring the password - it requires users to be intentional about fetching the secret.

aasimkhan30: I wasn't able to find a use case for excluding credentials, so I excluded it.

Benjin (approved):

Issue #19733, opened 2025-07-11
[Feature Request]: Auto-prompt to clear Entra token cache


croblesm:
### Feature Description

Add an **automatic or guided flow to clear the Microsoft Entra account token cache** when the `multiple_matching_tokens_detected` error occurs during Active Directory Interactive authentication in the extension.

Currently, users must manually run `MS SQL: Clear Microsoft Entra account token cache` from the command palette to resolve this issue. This feature would:

- Detect the specific error (`multiple_matching_tokens_detected`).
- Prompt the user with a clear notification:
  - Explain that clearing the token cache will likely resolve the issue.
  - Provide a **one-click button** to clear the cache automatically.
- Optionally allow enabling **auto-clear on next connection attempt** if the error persists.

<img width="1446" height="262" alt="Image" src="https://github.com/user-attachments/assets/aec88718-1af2-4b46-a939-ff92c9df39e4" />

### Problem and Motivation

Right now, the resolution requires users to:

1. Open the VS Code command palette.
2. Search for `mssql clear`.
3. Run `MS SQL: Clear Microsoft Entra account token cache`.

This workflow is **not obvious or discoverable**, especially under time pressure or for less experienced users.

By **automating or guiding the user to clear the token cache**, we will:

- Improve the extension’s user experience.
- Reduce connection troubleshooting frustration.
- Align with modern VS Code extension practices by providing **self-healing or assisted resolution** for known error states.
- Prevent repeated issues that result in user confusion and support tickets.

### Related Area

- [x] Connection dialog
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local container
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Confirmation

- [x] I have searched existing feature requests and couldn't find a match
- [x] I want to help implement this feature

PR #20305, opened 2025-10-15
Adding actionable message banners and clear token cache button

Benjin:

## Description

* Adds a button to clear the token cache if a `multiple_matching_tokens` error is detected.
* Adds the ability to add action buttons and specify intent (error/warning/success/info) for messages.

Fixes https://github.com/microsoft/vscode-mssql/issues/19733

image

## Code Changes Checklist

### DIFF
--- a/test/unit/connectionDialogWebviewController.test.ts
+++ b/test/unit/connectionDialogWebviewController.test.ts
@@ -19,6 +19,7 @@ import MainController from "../../src/controllers/mainController";
 import VscodeWrapper from "../../src/controllers/vscodeWrapper";
 import { ObjectExplorerProvider } from "../../src/objectExplorer/objectExplorerProvider";
 import {
+    AddFirewallRuleDialogProps,
     AuthenticationType,
     AzureSqlServerInfo,
     ConnectionInputMode,
@@ -53,6 +54,9 @@ import { multiple_matching_tokens_error } from "../../src/azure/constants";
 import { Logger } from "../../src/models/logger";
 import { MsalAzureController } from "../../src/azure/msal/msalAzureController";
 import { errorPasswordExpired } from "../../src/constants/constants";
+import { FirewallRuleSpec } from "../../src/sharedInterfaces/firewallRule";
+import { FirewallService } from "../../src/firewall/firewallService";
+import { AddFirewallRuleState } from "../../src/sharedInterfaces/addFirewallRule";
 
 chai.use(sinonChai);
 
@@ -443,7 +447,7 @@ suite("ConnectionDialogWebviewController Tests", () => {
             expect(
                 controller.state.formMessage,
                 "Error should be cleared after loading the connection",
-            ).to.equal(undefined);
+            ).to.be.undefined;
 
             expect(
                 controller.state.readyToConnect,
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -88,7 +88,7 @@
         "@dagrejs/dagre": "^1.1.4",
         "@eslint/compat": "^1.1.0",
         "@fluentui-contrib/react-data-grid-react-window": "^1.2.0",
-        "@fluentui/react-components": "^9.64.0",
+        "@fluentui/react-components": "^9.72.2",
```
aasimkhan30, line 91: Any reason for the update. Not saying we shouldn't just curious what this brings in.

src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -744,6 +748,17 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
             return state;
         });
 
+        this.registerReducer("messageButtonClicked", async (state, payload) => {
+            if (payload.buttonId === CLEAR_TOKEN_CACHE) {
+                this._mainController.onClearAzureTokenCache();
```
aasimkhan30: Should we call the method directly on connectionManager (or possibly azureController)?
I'm trying to move these calls out of mainController.

Over time, we tend to add extra logic or guardrails in mainController, but not in the corresponding core service methods. This causes the behavior of these methods to diverge, which can lead to inconsistencies. 

Let's just get rid of mainController version and just use the inner ones directly.

src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -994,9 +1009,17 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
                     return await this.handleConnectionErrorCodes(connectionInfo, state);
                 }
             } catch (error) {
-                this.state.formError = getErrorMessage(error);
+                this.state.formMessage = { message: getErrorMessage(error) };
                 this.state.connectionStatus = ApiStatus.Error;
 
+                if (
```
aasimkhan30, line 1017: should this go into handleConnectionErrrorCodes method? Also do you want to handle this error on the connectionManager level? 
https://github.com/microsoft/vscode-mssql/blob/097b79d2d077f7615bf3f5071c872924ddeff7bd/src/controllers/connectionManager.ts#L2001

aasimkhan30: Maybe we can create a new error type and handle the error type in both connection manager and connection dialog webview controller.

package.json
Benjin: I was seeing some odd behavior with the <Message> component formatting, thought maybe bumping the version might bring in a bugfix, then realized it was my own mistake. 🙃

And then I figured that there's no reason to _revert_ a minor vBump and may as well keep it.

src/connectionconfig/connectionDialogWebviewController.ts
Benjin: Good call.

src/connectionconfig/connectionDialogWebviewController.ts
Benjin: I looked at that, but since this is an error occurring before SqlClient comes into play (during connection preparation), this doesn't fit there.  It already has the same two methods for handling as the SqlClient errors: one default (in connectionManager) and one specialized for Connection Dialog.

PR #20307, opened 2025-10-15
Adding fixes for get rows out of bounds. 

aasimkhan30:

## Description

While I was not able to repo this issue, I am adding more guardrails around the getRows code so it doesn't lead to negative row fetches. 

This also adds debouncing when scrolling through tables so that we don't spam sts with too many requests and also adding checks so that cell values don't keep fluctuating as we try to update from state results.

Fixes: #20283 

Before:

Code_mWBpYpLpnj

Fixed:

Code_94vWT6pDYj

## Code Changes Checklist

### DIFF
--- a/src/reactviews/pages/QueryResult/table/asyncDataView.ts
+++ b/src/reactviews/pages/QueryResult/table/asyncDataView.ts
@@ -3,8 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-// import * as vscode from 'vscode';
 import { IDisposableDataProvider } from "./dataProvider";
+import { v4 as uuid } from "uuid";
 
 export interface IObservableCollection<T> {
     getLength(): number;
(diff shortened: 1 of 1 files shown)


src/reactviews/pages/QueryResult/table/asyncDataView.ts
```diff
@@ -22,6 +22,11 @@ class DataWindow<T> {
     private _data: T[] | undefined;
     private _length: number = 0;
     private _offsetFromDataSource: number = -1;
+    private _currentRequestId: number = 0;
+    private _debounceTimeout: NodeJS.Timeout | undefined;
+    private readonly _debounceDelay: number = 50; // ms to wait before sending request
```
Benjin: Good to include the unit in the var name: `_debounceDelayMs`

src/reactviews/pages/QueryResult/table/asyncDataView.ts
```diff
@@ -55,28 +64,71 @@ class DataWindow<T> {
...
+        }
+
+        // Clear any pending debounced requests
+        if (this._debounceTimeout) {
+            clearTimeout(this._debounceTimeout);
+        }
+
+        const executeLoad = () => {
+            // Double-check that this request is still current
+            if (currentRequestId !== this._currentRequestId) {
```
Benjin, line 105: Thinking about ways this might trip over itself...
something feels odd about about the requestId being incremented (rather than randomly generated), but then only checking for inequality rather than greater/less than.  But I'm not immediately spotting a way that `this.currentRequestId` might be _behind_ `currentRequestId` (rather than ahead, as it would be for a rapid scroll), so it's probably fine.  Just wanted to call it out.

Benjin (approved):

Benjin (approved):

PR #20309, opened 2025-10-15
Publish Dialog - Connection section

ssreerama:

## Description
- This PR Introduces a new workflow for selecting SQL Server connections in the Publish Project dialog.
- Enhances user experience by allowing:
    - Connecting to a server.
    - Selecting a database via a dropdown.
- Reduces manual input errors and streamlines the publishing process.
- UI Improvements:
    - Added Server and Database fields in the dialog.
- Clicking the New Connection icon:
    - Opens the connection dialog.
    - Allows users to connect to their target server.

Once connected:

The Database field auto-populates with available databases.
## Code Changes Checklist

3 ConnectionSection

### DIFF
--- a/src/publishProject/publishProjectWebViewController.ts
+++ b/src/publishProject/publishProjectWebViewController.ts
@@ -36,6 +36,8 @@ export class PublishProjectWebViewController extends FormWebviewController<
     PublishDialogFormItemSpec,
     PublishDialogReducers
 > {
+    private _cachedDatabaseList?: { displayName: string; value: string }[];
+    private _cachedSelectedDatabase?: string;
     public readonly initialized: Deferred<void> = new Deferred<void>();
     private readonly _sqlProjectsService?: SqlProjectsService;
     private readonly _dacFxService?: mssql.IDacFxService;
@@ -89,7 +91,6 @@ export class PublishProjectWebViewController extends FormWebviewController<
             },
         );
 
-        // Store the SQL Projects Service and Connection Manager
         this._sqlProjectsService = sqlProjectsService;
         this._dacFxService = dacFxService;
         this._connectionManager = connectionManager;
@@ -102,7 +103,6 @@ export class PublishProjectWebViewController extends FormWebviewController<
             this.state.deploymentOptions.excludeObjectTypes.value = [];
         }
 
-        // Register reducers after initialization
         this.registerRpcHandlers();
 
         // Listen for successful connections
@@ -116,7 +116,6 @@ export class PublishProjectWebViewController extends FormWebviewController<
             }),
         );
 
-        // Initialize async to allow for future extensibility and proper error handling
         void this.initializeDialog(projectFilePath)
             .then(() => {
                 this.updateState();
@@ -128,7 +127,6 @@ export class PublishProjectWebViewController extends FormWebviewController<
     }
 
     private async initializeDialog(projectFilePath: string) {
-        // keep initial project path and computed database name
         if (projectFilePath) {
             this.state.projectFilePath = projectFilePath;
         }
@@ -153,7 +151,7 @@ export class PublishProjectWebViewController extends FormWebviewController<
                 TelemetryViews.SqlProjects,
                 TelemetryActions.PublishProjectChanges,
                 error instanceof Error ? error : new Error(String(error)),
-                false, // don't include error message in telemetry for privacy
+                false,
             );
         }
 
@@ -163,11 +161,9 @@ export class PublishProjectWebViewController extends FormWebviewController<
             this.state.formState.databaseName,
         );
 
-        // Update state to notify UI of the project properties and form components
         this.updateState();
 
         // Fetch Docker tags for the container image dropdown
-        // Use the deployment UI function with target version filtering
         const tagComponent = this.state.formComponents[PublishFormFields.ContainerImageTag];
         if (tagComponent) {
             try {
@@ -183,7 +179,6 @@ export class PublishProjectWebViewController extends FormWebviewController<
                 }
             } catch (error) {
                 console.error("Failed to fetch Docker container tags:", error);
-                // Keep dialog resilient - don't block if Docker tags fail to load
             }
         }
 
@@ -193,14 +188,13 @@ export class PublishProjectWebViewController extends FormWebviewController<
         await this.validateForm(this.state.formState, undefined, true);
     }
 
-    /** Registers all reducers in pure (immutable) style */
     private registerRpcHandlers(): void {
         this.registerReducer("openConnectionDialog", async (state: PublishDialogState) => {
             // Set waiting state to detect new connections
             state.waitingForNewConnection = true;
             this.updateState(state);
 
-            // Execute the command to open the connection dialog (same as "+" button in servers panel)
+            // Execute the command to open the connection dialog
             void vscode.commands.executeCommand(constants.cmdAddObjectExplorer);
 
             return state;
@@ -266,13 +260,21 @@ export class PublishProjectWebViewController extends FormWebviewController<
                             parsedProfile.deploymentOptions || state.deploymentOptions,
                     };
                 } catch (error) {
-                    void vscode.window.showErrorMessage(
-                        `${Loc.PublishProfileLoadFailed}: ${error}`,
-                    );
+                    return {
+                        ...state,
+                        formMessage: {
+                            message: `${Loc.PublishProfileLoadFailed}: ${error}`,
+                            intent: "error",
+                        },
+                    };
                 }
             }
 
             return state;
+        });
+
+        this.registerReducer("closeMessage", async (state: PublishDialogState) => {
+            return { ...state, formMessage: undefined };
         });
 
         this.registerReducer(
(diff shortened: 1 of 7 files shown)


src/publishProject/publishProjectWebViewController.ts
```diff
@@ -95,6 +102,29 @@ export class PublishProjectWebViewController extends FormWebviewController<
         // Register reducers after initialization
         this.registerRpcHandlers();
 
+        // Listen for new connections
+        this.registerDisposable(
+            this._connectionManager.onConnectionsChanged(async () => {
```
aasimkhan30: Please listen to onSuccessfulConnection, it gives you the connection info with the URI.

aasimkhan30 (approved):

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -319,6 +354,67 @@ export class PublishProjectWebViewController extends FormWebviewController<
         );
     }
 
+    /** Auto-select a new connection and populate server/database fields */
+    private async autoSelectNewConnection(connectionUri: string): Promise<void> {
+        try {
+            // IMPORTANT: Get the connection profile FIRST, before any async calls
+            // The connection URI may be removed from activeConnections during async operations
+            const connection = this._connectionManager.activeConnections[connectionUri];
```
aasimkhan30: Why don't you modify the event and add connectionInfo to it. That should have the connectionId and it will make this process more deterministic.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -319,6 +354,67 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+            // Get connection string first
+            const connectionString = await this._connectionManager.getConnectionString(
+                connectionUri,
+                true, // includePassword
+                true, // includeApplicationName
+            );
+            this.state.connectionString = connectionString;
+
+            // Get databases
+            const databases = await this._connectionManager.listDatabases(connectionUri);
```
aasimkhan30: Maybe a simple follow up task. If the list Database API fails, maybe show an error message to the user.

src/publishProject/publishProjectWebViewController.ts
aasimkhan30: @Benjin, I have seen this pattern a few times. Do you think we should add create and return "ConnectionInfo" api to connection manager. Just so that we don't add multiple different implementations of this.

Benjin: > Why don't you modify the event and add connectionInfo to it. That should have the connectionId and it will make this process more deterministic.

@lewis-sanchez can you remind me why you did it this way?

Maybe we can create a Deferred<IConnectionProfile> that's then resolved when the (modified) event is emitted.  Seems much cleaner

Benjin: > Do you think we should add create and return "ConnectionInfo" api to connection manager?

So we'd have `cm.CreateConnection() : Promise<void>` and `cmd.CreateAndReturnConnection(): Promise<IConnectionProfile>`?  Yeah, that's pretty reasonable.

...however, while I think that API would be nice to have, IMO the event is ultimately the way to go. I can see users having multiple connection dialogs going and easily getting them mixed up because they're not modal.

src/reactviews/pages/PublishProject/components/ConnectionSection.tsx
```diff
@@ -4,12 +4,27 @@
...
+        gap: "16px",
+        maxWidth: "640px",
+        width: "100%",
+    },
+});
 
 export const ConnectionSection: React.FC = () => {
     const publishCtx = useContext(PublishProjectContext);
+    const formStyles = useFormStyles();
+    const classes = useStyles();
```
Benjin: just call this `styles` to keep the semantics consistent

src/reactviews/index.css
```diff
@@ -23,10 +23,16 @@
...
+/* Dropdown and Combobox styling */
 .fui-Dropdown {
     border: 1px solid var(--vscode-input-border, transparent);
 }
+.fui-Combobox {
+    background-color: var(--vscode-settings-dropdownBackground, var(--vscode-dropdown-background));
+    color: var(--vscode-settings-dropdownForeground, var(--vscode-dropdown-foreground));
+    border: 1px solid var(--vscode-settings-dropdownBorder, var(--vscode-input-border, transparent));
+    border-radius: 2px;
+}
```
Benjin, line 35: What was missing with the styling that this needed to be added?  I hadn't noticed issues with the other comboboxes.

ssreerama: I noticed that we only use comboboxes in a couple of places in the code. Without these styles, the combobox displays a white border, and its inner background color doesn’t align with the input or dropdown fields. Perhaps the existing ones didn’t require these styles earlier. I'll have to check how the existing ones displays now with these styles.

image
image

src/publishProject/publishProjectWebViewController.ts
ssreerama: right! onSuccessfulEvent already contains the connection details, which I overlooked and instead retrieved again using fileUrl from the active connections list. Removed the redundant logic.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -84,9 +89,10 @@ export class PublishProjectWebViewController extends FormWebviewController<
             },
         );
 
-        // Store the SQL Projects Service
+        // Store the SQL Projects Service and Connection Manager
```
Benjin: These sorts of comments are just noise; we can remove them.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -175,7 +195,18 @@ export class PublishProjectWebViewController extends FormWebviewController<
 
     /** Registers all reducers in pure (immutable) style */
```
Benjin: What does this comment even mean?  If it's not a useful comment, we should remove it.

Benjin: Same for many other comments in this PR.

We only want to keep comments that are useful (provide information that isn't immediately apparent from the code) and descriptive.  Ideally, the code itself is "self-documenting" through good variable and function names, so comments should mostly be used when that isn't as possible.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -286,8 +317,12 @@ export class PublishProjectWebViewController extends FormWebviewController<
 
                 try {
                     const databaseName = state.formState.databaseName || projectName;
-                    // TODO: Build connection string from connection details when server/database selection is implemented
-                    const connectionString = "";
+                    // Connection string depends on publish target:
+                    // - For container targets: empty string (no server connection)
```
Benjin: Why do container targets have no connection string?  Is that because we're provisioning a new container (and _then_ would use a connection string for that new container)?  Can you add that to this comment?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -320,6 +355,81 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                        value: db,
+                    }));
+                }
+
+                // Optionally select the first database if available
+                if (databases.length > 0 && !this.state.formState.databaseName) {
+                    this.state.formState.databaseName = databases[0];
+                }
+            } catch (dbError) {
+                // Show error message to user when database listing fails
```
Benjin, line 416: I think this would be better with a proper react message component.  See `dialogMessage.tsx` and how it's used.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -320,6 +355,81 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                    TelemetryActions.PublishProjectConnectionError,
+                    { error: dbError instanceof Error ? dbError.message : String(dbError) },
+                );
+            }
+
+            // Validate form to update button state after connection
+            await this.validateForm(this.state.formState, undefined, false);
+
+            // Update UI immediately to reflect the new connection
+            this.updateState();
```
Benjin: Should this happen after `waitingForNewConnection = false` in the `finally` block to ensure that state change is also reflected in the UI?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -335,6 +445,58 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                    if (
+                        this.state.previousDatabaseList &&
+                        this.state.previousDatabaseList.length > 0
+                    ) {
+                        databaseComponent.options = [...this.state.previousDatabaseList];
+
+                        // Restore previously selected database
+                        if (this.state.previousSelectedDatabase) {
+                            this.state.formState.databaseName = this.state.previousSelectedDatabase;
+                        }
```
Benjin: Why clear and restore?  Seems like you could just show/hide certain components instead.

If you do want to go with this design (saving values when the publish target changes), why are the previous values stored in the state instead of in the controller?

src/reactviews/index.css
Benjin: Hmm, but I haven't seen this issue with comboboxes in other UIs?  And the Publish Target combobox in the first screenshot looks fine.  That tells me that the issue is somewhere else, and we shouldn't just patch it with some one-off CSS.

test/unit/publishProjectWebViewController.test.ts
```diff
@@ -23,6 +23,8 @@ suite("PublishProjectWebViewController Tests", () => {
     let mockSqlProjectsService: any;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     let mockDacFxService: any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let mockConnectionManager: any;
```
Benjin: Any time we work around the type-checker by casting something as `any` or `unknown`, we expose ourselves to breakage later down the line so it needs to be for a good reason.

Why couldn't `sinon.SinonStubbedInstance<DacFxService>` be used for these three?

src/publishProject/publishProjectWebViewController.ts
ssreerama: clearing and restoring required for database list when switching target options from exisitng server to container, where for existing server if the databsase has the list populated, switching to container options should have an empty list. But I have cached the list now to a local variable instead of state. Thank you

PR #20313, opened 2025-10-16
Adding extra padding in result grid

allancascante:
## Description

Adding extra padding around the cells in the result grid to improve readability.

Before

image

After

image

## Code Changes Checklist

### DIFF
--- a/src/reactviews/pages/QueryResult/table/table.ts
+++ b/src/reactviews/pages/QueryResult/table/table.ts
@@ -637,6 +637,10 @@ export class Table<T extends Slick.SlickData> implements IThemable {
             );
         }
 
+        content.push(
+            `.monaco-table .${this.idPrefix} .grid-cell-value-container { padding: 2px; }`,
+        );
+
         if (styles.tableHeaderForeground) {
             content.push(
                 `.monaco-table .${this.idPrefix} .slick-header .slick-header-column { color: ${styles.tableHeaderForeground}; }`,


aasimkhan30:
Please add some screenshots.

allancascante:
> Please add some screenshots.

done

aasimkhan30:
It seems like the padding is applied only to the left. Let's limit the css styling to just that as it doesn't seem to affect other directions.

src/reactviews/pages/QueryResult/table/table.ts
```diff
@@ -637,6 +637,10 @@ export class Table<T extends Slick.SlickData> implements IThemable {
             );
         }
 
+        content.push(
+            `.monaco-table .${this.idPrefix} .grid-cell-value-container { padding: 2px; }`,
```
Benjin, line 641: I think what we want here is vertical-centering, not just adding some predefined padding (that might scale poorly if the user changes the result set font size, too.

Try setting `"mssql.resultsFontSize": 24` and you'll see that it still looks a bit funky:

Image

allancascante: @Benjin  this is what I see with 24px on a default font family with the extra 1 px padding.

image

and this is without the extra 1px on the padding

image

The funkiness you describe if you change the font is because the cells don't scale to fit the font size and that should be addressed when the setting to allow changing cell height is added.

slick-cell predefined padding is 1px 2px 2px 1px  so the difference is 1px on the left and right of the cell, and it is not altering the top and bottom paddings

allancascante:
dropping as not needed

PR #20317, opened 2025-10-17
Switching to using keycode instead of key in web view keyboard events.

aasimkhan30:
## Description

Switching keyboard events to use `code` instead of `key` as it is not case sensitive and a better prop to use for keyboard combinations.  

## Code Changes Checklist

### DIFF
--- a/src/reactviews/pages/QueryResult/table/plugins/copyKeybind.plugin.ts
+++ b/src/reactviews/pages/QueryResult/table/plugins/copyKeybind.plugin.ts
@@ -15,7 +15,7 @@ import {
 } from "../utils";
 import { KeyCode } from "../../../../common/keys";
 import { QueryResultReactProvider } from "../../queryResultStateProvider";
-import { GetPlatformRequest } from "../../../../../sharedInterfaces/webview";
+import { isMetaKeyPressed } from "../../../../common/utils";
 
 /**
  * Implements the various additional navigation keybindings we want out of slickgrid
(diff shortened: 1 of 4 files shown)


src/reactviews/pages/QueryResult/table/plugins/copyKeybind.plugin.ts
```diff
@@ -48,18 +48,9 @@ export class CopyKeybind<T extends Slick.SlickData> implements Slick.Plugin<T> {
 
     private async handleKeyDown(e: KeyboardEvent): Promise<void> {
         let handled = false;
-        let platform = await this._qrContext.extensionRpc.sendRequest(GetPlatformRequest.type);
```
aasimkhan30, line 51: Removed this as it is async and it doesn't work when we want to stop default behavior or stop propagation for the event. 
while we are not doing that here,  I don't want to give a wrong idea to someone copying from this code.

src/reactviews/common/keys.ts
```diff
@@ -3,14 +3,14 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-export enum Keys {
+export enum KeyCode {
```
Benjin, line 6: Can you check to see if we're already consuming https://www.npmjs.com/package/@fluentui/keyboard-keys indirectly via the fluent components?  Maybe we can use that instead of maintaining a partial list of our own?

aasimkhan30: Unfortunately, this won't work as I want to move from `Keys` to  `Code` and that package only has options `Keys` and `KeyCode`. 
Keys  is case sensitive and 
key code is deprecated.

PR #20336, opened 2025-10-22
Align publish dialog Docker image tags with Deployment UI

ssreerama:

## Description

This PR removes the old logic of providing all raw tags in the docker image tag dropdown when user selected the container publish option, and matches with the current deployment UI image tag results. But for Publish dialog we cannot provide all available images as user cannot publish 2025 project on 2022 server or earlies, we are filtering the tags based on the project version.

## Code Changes Checklist


aasimkhan30 (approved):

src/publishProject/projectUtils.ts
```diff
@@ -140,3 +159,41 @@ export function getPublishServerName(target: string) {
...
+): Promise<FormItemOptions[]> {
+    try {
+        // Get the deployment UI versions first
+        const deploymentVersions = await getSqlServerContainerVersions();
+        if (!deploymentVersions || deploymentVersions.length === 0) {
+            return [];
+        }
+
+        // Determine minimum year based on target version
+        let minYear = 2017; // Default to SQL Server 2017
```
aasimkhan30: NIT: Make this a constant.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -125,28 +125,20 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
-                    // Set default to first option (most recent -latest) if not already set
-                    if (!this.state.formState.containerImageTag && tagOptions[0]) {
-                        this.state.formState.containerImageTag = tagOptions[0].value;
-                    }
+                tagComponent.options = tagOptions;
+                if (!this.state.formState.containerImageTag && tagOptions.length > 0) {
+                    this.state.formState.containerImageTag = tagOptions[0].value;
                 }
             } catch (error) {
                 console.error("Failed to fetch Docker container tags:", error);
```
aasimkhan30: NIT: use Logger to log these errors. Switch all the console logs to Logger.

src/publishProject/projectUtils.ts
Benjin: Why are we defaulting to such an old version instead of the latest GA?  FYI: there's a bug with 2025 on ARM that won't be fixed at GA.  Please sync with PMs and @laurenastrid1 about the solution here.

src/publishProject/projectUtils.ts
```diff
@@ -140,3 +159,41 @@ export function getPublishServerName(target: string) {
...
+        if (targetVersion) {
+            const versionNum = parseInt(targetVersion, 10);
+            // Find the corresponding year, or default to 2017 if not found
+            minYear = DSP_VERSION_TO_YEAR.get(versionNum) ?? 2017;
```
Benjin: This should refer to a `const DEFAULT_YEAR` instead of hardcoded value.  It'll be easier to change without missing a spot.

src/deployment/dockerUtils.ts
```diff
@@ -993,72 +993,6 @@ export async function getAllSqlServerContainerTags(): Promise<string[]> {
     }
 }
 
-/**
```
Benjin, line 998: Yay, much cleaner ❤️

src/publishProject/projectUtils.ts
ssreerama: In ADS we get max available year image tag incase no verison number is found from the project. Since we have current available images from 2017-2025 in deploymentUI, I used the min number 2017, but it seems what ADS is doing is right, we should provide the very latest image or all available images(checking with PM)

src/publishProject/projectUtils.ts
Benjin, line 32: Can you list the full path in the DacFx codebase?  Also, if there isn't a comment over there that says to update this list, can you send a PR to add one?

src/publishProject/projectUtils.ts
```diff
@@ -143,6 +162,63 @@ export function validateSqlServerPortNumber(port: number): boolean {
...
+ * Used by Publish Project dialog to provide granular tag selection.
+ *
+ * This follows the same filtering logic as ADS (Azure Data Studio):
+ * - If target version is known: filter to show versions >= target version year
+ * - If target version is unknown: fallback to max available version year (like ADS)
+ *
+ * @param targetVersion - The SQL Server version (e.g., "160" for SQL Server 2022)
+ * @returns Sorted array of tags filtered by version from deployment UI versions
+ */
+export async function getSqlServerContainerTagsForTargetVersion(
```
Benjin, line 193: Be sure to add unit tests!

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -142,28 +142,20 @@ export class PublishProjectWebViewController extends FormWebviewController<
 
         // Load publish form components
         this.state.formComponents = generatePublishFormComponents(projectTargetVersion);
-
-        // Update state to notify UI of the project properties and form components
         this.updateState();
 
-        // Fetch Docker tags for the container image dropdown
-        // Use the deployment UI function with target version filtering
+        // Use deployment UI method to get filtered image tags
```
Benjin: Is this comment still relevant?  I only see you calling the publish projects utils.

src/publishProject/projectUtils.ts
ssreerama: done! added a direct test that validates the expected image tags

src/publishProject/projectUtils.ts
ssreerama: Updated the comment, I dont see it was mentioned in DacFx, will create a PR there with a note.

src/publishProject/publishProjectWebViewController.ts
Benjin: Why was this changed back?

ssreerama: seems rebasing to main branch reset the change. I missed noticing the diff while resolving them, reverting   it.

src/publishProject/projectUtils.ts
```diff
@@ -143,6 +163,63 @@ export function validateSqlServerPortNumber(port: number): boolean {
...
+): Promise<FormItemOptions[]> {
+    try {
+        // Get the deployment UI versions first
+        const deploymentVersions = await getSqlServerContainerVersions();
+        if (!deploymentVersions || deploymentVersions.length === 0) {
+            return [];
+        }
+
+        // Extract years from all available versions to find the maximum
+        const availableYears = deploymentVersions
```
Benjin: How about building a Map<int, string> where the key is the year and the value is the full tag?  I think that would clean this up.

src/publishProject/projectUtils.ts
```diff
@@ -143,6 +163,63 @@ export function validateSqlServerPortNumber(port: number): boolean {
...
+
+        // Filter deployment versions based on minimum year
+        const filteredVersions = deploymentVersions.filter((option) => {
+            const year = parseInt(option.displayName.match(/\d{4}/)?.[0] || "0", 10);
+            return year >= minYear;
+        });
+
+        return filteredVersions;
+    } catch (e) {
+        dockerLogger.appendLine(`Error filtering SQL Server container tags: ${getErrorMessage(e)}`);
```
Benjin: dockerLogger.error() instead

localization/xliff/vscode-mssql.xlf
```diff
@@ -1,4684 +1,4687 @@
...
-      <source xml:lang="en">New Query</source>
-    </trans-unit>
-    <trans-unit id="mssql.newTable">
-      <source xml:lang="en">New Table</source>
-    </trans-unit>
-    <trans-unit id="mssql.walkthroughs.nextSteps.title">
-      <source xml:lang="en">Next Steps with MSSQL for Visual Studio Code</source>
-    </trans-unit>
-    <trans-unit id="mssql.logRetentionMinutes">
-
```
Benjin, line 1: revert this file; if we need to fix the line endings, that should be done in a separate PR

src/publishProject/projectUtils.ts
```diff
@@ -143,6 +163,63 @@ export function validateSqlServerPortNumber(port: number): boolean {
...
+        if (availableYears.length === 0) {
+            return deploymentVersions; // No years found, return all
+        }
+
+        const maxYear = Math.max(...availableYears);
+
+        // Determine minimum year based on target version
+        let minYear: number;
+        if (targetVersion) {
+            const versionNum = parseInt(targetVersion, 10);
```
Benjin: the `,10` isn't necessary because it defaults to base-10

localization/xliff/vscode-mssql.xlf
ssreerama: reverted to the previous commit that this file has changed from. I'm still not seeing a valid changes it shows, this file got changed after I merged to latest main.

aasimkhan30 (approved):

Benjin (approved):

Benjin (approved):

Issue #20042, opened 2025-08-27
[Bug]: "Container does not exist" when trying to connect


dzsquared:
### Description

I have a saved connection that I added for a container running on localhost, but when I click to connect in OE I get an error asking to remove the connection

<img width="294" height="250" alt="Image" src="https://github.com/user-attachments/assets/0f395ed1-d8bf-4131-a793-272435ccc984" />

when I hit cancel it connects anyways, because... yeah the container does exist.

<img width="605" height="694" alt="Image" src="https://github.com/user-attachments/assets/64661a98-8e95-4c2f-ae73-dd1de4205567" />

Identifier
ms-mssql.mssql
Version
1.35.0
Last Updated
2025-08-20, 14:07:00

Version: 1.103.2
Commit: 6f17636121051a53c88d3e605c491d22af2ba755
Date: 2025-08-20T16:45:34.255Z
Electron: 37.2.3
ElectronBuildId: 12035395
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0

### Steps to Reproduce

1. have a connection to a local container that had been working for months
2. try to connect to the container after starting it in docker desktop

### Affected Area

- [ ] Connection dialog
- [ ] Query results panel
- [ ] Query editor
- [x] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local container
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

Version: 1.103.2
Commit: 6f17636121051a53c88d3e605c491d22af2ba755
Date: 2025-08-20T16:45:34.255Z
Electron: 37.2.3
ElectronBuildId: 12035395
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

faresbouzayen:
Hey, I’d love to work on this issue! Can I take it?

dzsquared:
Behavior has changed, now it doesn't prompt to remove the connection - but the OE spins for a long while "starting docker" (which is already running) and eventually connects while throwing an error toast

[...]

PR #20347, opened 2025-10-22
Fix container connection bug

laurenastrid1:

## Description
Fixes https://github.com/microsoft/vscode-mssql/issues/20042

Previously, the container would not be recognized if the port was left unspecified.

## Code Changes Checklist


aasimkhan30 (approved):

src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -888,6 +888,9 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
             cleanedConnection.enclaveAttestationUrl = undefined;
         }
 
+        // If there's no port specified, set it to the default port number
+        cleanedConnection.port = cleanedConnection.port ?? defaultPortNumber;
```
Benjin, line 892: We typically don't want to (or need to) explicitly set default values in the connection itself because if we have to set it for all the connections, then it's not really a default, just a mandatory field.  I suspect that the fix for this bug probably lies elsewhere (e.g. at connection time)

Issue #18846, opened 2025-02-25
XML and Execution plans


tune007:
When running a script like e.g. Sp_BlitzCache I am always able in ADS or SSMS to just click on the result column with the XML for the query plan, this will automatically open an Execution plan. With this extension I have to go and save the XML as a .sqlpan before it opens as an execution plan

----

Setup:
1. Run this to grab the handle for a query (doesn't matter which):

```
SELECT TOP 1 
    cp.plan_handle, 
    st.text AS sql_text
FROM 
    sys.dm_exec_cached_plans AS cp
CROSS APPLY 
    sys.dm_exec_sql_text(cp.plan_handle) AS st
```

2. Replace `0x6000...` with the `plan_handle` from above, and run this:

```
SELECT 
    qp.query_plan
FROM 
    sys.dm_exec_query_plan(0x06000...) AS qp; -- Replace with actual plan_handle
```

3. Click `<ShowPlan>` XML

Expectation:
* XML starting with `<ShowPlan>` is detected, so it's opened as a query plan

Reality:
* Unformatted XML is displayed; user must save as `.sqlplan` file then reopen to view.

Benjin:
When you use the "Enable Actual Plan" and "Estimated Plan" buttons, there should be a Query Plan tab alongside the Results and Messages:

![Image](https://github.com/user-attachments/assets/7fa22f3a-e0d0-477b-805b-7e5ba83ccd65)

![Image](https://github.com/user-attachments/assets/92df007a-6ecd-4f21-9408-883bc491a6b3)

If you want to view the XML from that plan without saving it, you can click this button:

![Image](https://github.com/user-attachments/assets/34595206-192c-41ca-850f-811bdd77e48e)

Are these features not working for you?  Note that you'll need to have [preview features enabled](https://learn.microsoft.com/en-us/sql/tools/visual-studio-code-extensions/mssql/mssql-extension-visual-studio-code?view=sql-server-ver16#enable-the-new-ui-features) for these to appear.

amatlock930:
@Benjin  - I think the request is to be able to click on the plan xml returned in queries against the table-valued function like this: sys.dm_exec_query_plan(a.plan_handle). In ADS and SSMS, the results return the "ShowPlanXML" in the grid and it is clickable and opens the XML in the plan viewer.

Benjin:
Ah, understood.  I'll update the bug.

croblesm:
Hi @tune007,

[...]

PR #20361, opened 2025-10-27
Show query  plan tab when query plan xml result is clicked

laurenastrid1:

## Description
Fixes https://github.com/microsoft/vscode-mssql/issues/18846

Previously, when a Query Plan XML was clicked in the result grid, it would open up a new XML document with the Query Plan XML. Now, it pulls up the Query Plan View in the Query Results Pane

showPlanTabWhenQueryPlanXmlClicked

## Code Changes Checklist


aasimkhan30:
I don't think this will work if the query plan is not displayed in its tab. 


Sometimes, query plans are stored in tables and people can run select on those tables returning just the xml in results grid. 

The right fix will be to try getting the execution plan json for the xml and open it in a new execution plan document, if it fails, just open it as a regular xml document.

Issue #19516, opened 2025-05-30
[Bug]: Unable to list subscription from several tenants


webartoli:
### Description

My account is enabled in 2 tenants. Each tenant has its own subscription.

From mssql extension I'm unable to select the subscription from other tenants and no tenant dropdown is avaiable.

![Image](https://github.com/user-attachments/assets/445893b2-35ef-403b-8af6-29db01a54194)

### Steps to Reproduce

Azure Data Studio is showing as follow
![Image](https://github.com/user-attachments/assets/16dd5b63-03a0-4360-a0e8-e873a5af2a4c)

AZ Cli is showing as follow:

```
az account list 
[
  {
    "cloudName": "AzureCloud",
    "homeTenantId": "cemasked-forx-secu-rity-reasonxxxx68",
    "id": "b9masked-forx-secu-rity-reasonxxxx52",
    "isDefault": true,
    "managedByTenants": [],
    "name": "Personal Subscription - Visual Studio Enterprise - MPN",
    "state": "Enabled",
    "user": {
      "name": "claudio@masked.com",
      "type": "user"
    }
  },
  {
    "cloudName": "AzureCloud",
    "homeTenantId": "9bmasked-forx-secu-rity-reasonxxxx85",
    "id": "38masked-forx-secu-rity-reasonxxxx61",
    "isDefault": false,
    "managedByTenants": [],
    "name": "FARM ON CLOUD MASKED",
    "state": "Enabled",
    "tenantId": "9bmasked-forx-secu-rity-reasonxxxx85",
    "user": {
      "name": "claudio@masked.com",
      "type": "user"
    }
  }
]
```

### Affected Area

- [x] Connection dialog
- [ ] Query editor
- [ ] IntelliSense/auto-completion
- [ ] Query results panel
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

- VSCode Version: 1.100.2 (latest)
- ms-mssql.mssql Version: 1.32.0 (Latest)
- OS Version: Windows 11 24H2

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

TWolversonReply:
Please fix this

allancascante:
[...]

PR #20367, opened 2025-10-28
Fix for azure subscriptions on multiple tenants not showing

allancascante:
## Description

When a user has access to the same Azure subscription through multiple accounts (e.g., personal and work accounts), the subscription filter dialog exhibited several issues, reported [here](https://github.com/microsoft/vscode-mssql/issues/19516):

Duplicate Selection Bug: Selecting one instance of a subscription would automatically select all instances across different accounts
Missing Subscriptions: Not all subscriptions were visible in the filter dropdown, even when no filtering was applied
Missing Tenants: Not all tenants for an account were displayed in the connection dialog
Error on Load: "Error loading Azure subscriptions" appeared when the same subscription existed under multiple accounts

### Root Cause Analysis

**Issue 1**: Tenant Discovery
Problem: auth.getTenants(account) with an account parameter wasn't reliably returning all tenants for an account, particularly for newly added tenants where the user hadn't explicitly signed in through VS Code.

**Fix**: Changed from auth.getTenants(account) to auth.getTenants() (without account parameter) to fetch ALL tenants across all accounts, then filter by account ID client-side. This ensures all tenants are discovered regardless of authentication state.

**Issue 2**: Token Refresh
Problem: VS Code's Azure authentication system only returns subscriptions from tenants where it has active authentication tokens. Newly added tenants/subscriptions weren't discovered until explicit authentication.

**Fix**: Added refreshAzureTokensForAllAccounts() method that proactively signs into all tenants for all accounts when the connection dialog opens, ensuring all subscriptions are discoverable.

**Issue 3**: Subscription Storage Collisions
Problem: The subscription filter used a composite key format tenantId/subscriptionId, which is not unique when the same subscription is accessible from multiple accounts. This caused:

- Subscriptions stored in a Map<subscriptionId, AzureSubscription> to overwrite each other
- Filter configuration to store duplicate entries
- UI to show both instances as selected when only one was chosen

**Fix**: Changed composite key format from tenantId/subscriptionId to account/tenantId/subscriptionId to ensure uniqueness. Updated all related code:

- Subscription Map storage
- Filter configuration format
- getTenantFilters() and getSubscriptionFilters() parsing logic
- Tenant ID population in subscription info

Changes Made

Core Logic Changes

1. azureHelpers.ts

- Modified getTenantsForAccount() to fetch all tenants then filter by account
- Updated composite key generation: ${sub.account.label}/${sub.tenantId}/${sub.subscriptionId}
- Updated promptForAzureSubscriptionFilter() to use new composite key format
- Enhanced getSubscriptionQuickPickItems() to include account label in description for disambiguation

2. connectionDialogWebviewController.ts

- Added refreshAzureTokensForAllAccounts() method to proactively ensure tokens exist for all tenants
- Modified loadAzureSubscriptions() to call token refresh before fetching subscriptions
- Updated subscription Map key from subscriptionId to ${tenantId}/${subscriptionId}
- Updated loadAzureServersForSubscription() to accept both tenantId and subscriptionId parameters
- Modified reducer to extract tenantId from state when loading servers
- Added tenant count to telemetry for tracking scenario where users have multiple tenants

3. MssqlVSCodeAzureSubscriptionProvider.ts

- Updated getTenantFilters() to parse index 1 from 3-part composite key
- Updated getSubscriptionFilters() to parse index 2 from 3-part composite key

4. connectionDialog.ts

- Added tenantId: string field to AzureSubscriptionInfo interface

5. azureBrowsePage.tsx

- Changed server selection mode from SelectFirstIfAny to AlwaysSelectNone (prevents auto-selection)
- Added proper cleanup of database selection when server selection becomes invalid

### Testing

New Test File: MssqlVSCodeAzureSubscriptionProvider.test.ts:
- getTenantFilters extracts tenant IDs from composite keys
- getSubscriptionFilters extracts subscription IDs from composite keys
- getTenantFilters handles empty configuration
- getSubscriptionFilters handles empty configuration
- filters correctly handle duplicate subscriptions from different accounts
- filters correctly parse complex email addresses

Enhanced Test File: azureHelpers.test.ts:
- creates pick items with account-based composite keys
- respects previous selection with account-based composite keys
- handles complex email addresses in composite keys

UI Updates

Before:

- Subscription filter showed duplicate selections for shared subscriptions:
image
image

- Not all subscriptions visible even without filters
- Auto-selected servers inappropriately
- Error messages when loading subscriptions

After:

- Each subscription+account combination is independently selectable 
image
- All subscriptions from all tenants are discoverable on dialog open
image
- Subscription filter accurately reflects user selection
- No auto-selection of servers/databases unless explicitly chosen
- Clear visual distinction in UI showing account email in subscription description

### Breaking Changes

Configuration Format Change: Existing subscription filters stored as tenantId/subscriptionId will need to be reset. Users with active filters will see all subscriptions unfiltered on first use after update and will need to re-select their desired subscriptions. This is intentional to avoid migration complexity and ensure clean state.

The change maintains backward compatibility with the VS Code Azure SDK and doesn't affect users who:

- Only have one Azure account
- Don't have duplicate subscriptions across accounts
- Haven't set up subscription filters
- Related: This fix also resolves the server auto-selection issue mentioned in the original report where servers were being selected even when filters weren't specified.

## Code Changes Checklist

### DIFF
--- a/test/unit/MssqlVSCodeAzureSubscriptionProvider.test.ts
+++ b/test/unit/MssqlVSCodeAzureSubscriptionProvider.test.ts
@@ -8,12 +8,6 @@ import * as sinon from "sinon";
 import * as vscode from "vscode";
 import { MssqlVSCodeAzureSubscriptionProvider } from "../../src/azure/MssqlVSCodeAzureSubscriptionProvider";
 
-// Interface to access protected methods for testing
-interface ITestableProvider {
-    getTenantFilters(): Promise<string[]>;
-    getSubscriptionFilters(): Promise<string[]>;
-}
-
 suite("MssqlVSCodeAzureSubscriptionProvider Tests", () => {
     let sandbox: sinon.SinonSandbox;
     let configStub: {
@@ -51,9 +45,7 @@ suite("MssqlVSCodeAzureSubscriptionProvider Tests", () => {
                 .returns(mockFilterConfig);
 
             const provider = MssqlVSCodeAzureSubscriptionProvider.getInstance();
-            const tenantFilters = await (
-                provider as unknown as ITestableProvider
-            ).getTenantFilters();
+            const tenantFilters = await provider["getTenantFilters"]();
 
             expect(tenantFilters).to.have.lengthOf(3);
             expect(tenantFilters).to.include("tenant1");
@@ -72,9 +64,7 @@ suite("MssqlVSCodeAzureSubscriptionProvider Tests", () => {
                 .returns(mockFilterConfig);
 
             const provider = MssqlVSCodeAzureSubscriptionProvider.getInstance();
-            const subscriptionFilters = await (
-                provider as unknown as ITestableProvider
-            ).getSubscriptionFilters();
+            const subscriptionFilters = await provider["getSubscriptionFilters"]();
 
             expect(subscriptionFilters).to.have.lengthOf(3);
             expect(subscriptionFilters).to.include("subscription1");
@@ -88,9 +78,7 @@ suite("MssqlVSCodeAzureSubscriptionProvider Tests", () => {
                 .returns([]);
 
             const provider = MssqlVSCodeAzureSubscriptionProvider.getInstance();
-            const tenantFilters = await (
-                provider as unknown as ITestableProvider
-            ).getTenantFilters();
+            const tenantFilters = await provider["getTenantFilters"]();
 
             expect(tenantFilters).to.be.an("array").that.is.empty;
         });
@@ -101,9 +89,7 @@ suite("MssqlVSCodeAzureSubscriptionProvider Tests", () => {
                 .returns([]);
 
             const provider = MssqlVSCodeAzureSubscriptionProvider.getInstance();
-            const subscriptionFilters = await (
-                provider as unknown as ITestableProvider
-            ).getSubscriptionFilters();
+            const subscriptionFilters = await provider["getSubscriptionFilters"]();
 
             expect(subscriptionFilters).to.be.an("array").that.is.empty;
         });
@@ -120,9 +106,7 @@ suite("MssqlVSCodeAzureSubscriptionProvider Tests", () => {
                 .returns(mockFilterConfig);
 
             const provider = MssqlVSCodeAzureSubscriptionProvider.getInstance();
-            const subscriptionFilters = await (
-                provider as unknown as ITestableProvider
-            ).getSubscriptionFilters();
+            const subscriptionFilters = await provider["getSubscriptionFilters"]();
 
             // Should return both entries (even though subscription ID is the same)
             // because they're from different accounts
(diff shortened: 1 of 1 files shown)


src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -1486,6 +1496,11 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
             state.loadingAzureSubscriptionsStatus = ApiStatus.Loading;
             this.updateState();
 
+            // Step 3: Ensure we have valid tokens for all tenants in all accounts
+            // This is critical for discovering new subscriptions that were added after initial sign-in
```
Benjin: Did you end up being able to reproduce this issue manually before these changes?  IIRC, you were having trouble getting the bug to repro.

allancascante: yes, I was able to repro; just had to add an account to two different subscriptions in two different tenants.

src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -1540,6 +1562,63 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
...
+
+                // Get all tenants for this account
+                const tenants = await auth.getTenants(accountInfo);
+                this.logger.verbose(
+                    `Found ${tenants.length} tenant(s) for account ${account.name}`,
+                );
+
+                // Attempt to get a token for each tenant
+                // This ensures VS Code has authentication sessions for all tenants
+                for (const tenant of tenants) {
```
Benjin: Will this end up prompting users multiple times (a prompt for each tenant)?  We don't want to barrage people with sign-ins just because we're trying to list a bunch of databases.

test/unit/MssqlVSCodeAzureSubscriptionProvider.test.ts
```diff
@@ -0,0 +1,207 @@
...
+import { expect } from "chai";
+import * as sinon from "sinon";
+import * as vscode from "vscode";
+import { MssqlVSCodeAzureSubscriptionProvider } from "../../src/azure/MssqlVSCodeAzureSubscriptionProvider";
+
+// Interface to access protected methods for testing
+interface ITestableProvider {
+    getTenantFilters(): Promise<string[]>;
+    getSubscriptionFilters(): Promise<string[]>;
+}
```
Benjin: You can get rid of this and use `MssqlVSCodeAzureSubscriptionProvider["somePrivateMethod"]` instead, and we'll preserve the type constraints better.

src/connectionconfig/connectionDialogWebviewController.ts
allancascante: removed this logic, in my testing it was asking by account not by each tenant; nonetheless it might be annoying if you have many accounts, if a subscription has been recently added, user will have to manually trigger the refresh for the account.

Benjin (changes requested):
Let's chat about this PR and the proposed fixes after Ignite.

Benjin:
This was completed in another PR

Issue #20188, opened 2025-09-17
[Bug]: JSON column data type detection is incorrectly making links


kburtram:
### Description

The logic to detect JSON data type detection logic is making values like `[WideWorldImporters].[Sales].[Customers]` into links.  This logic should be more robust to ensure only actual JSON columns are links.

<img width="1005" height="501" alt="Image" src="https://github.com/user-attachments/assets/3ba18eb1-7705-4296-aa4f-2938d43a4b61" />

### Steps to Reproduce

1. Run a query like 
```SQL

CREATE SYNONYM dbo.CustomersSyn
FOR WideWorldImporters.Sales.Customers;


SELECT 
    s.name AS SynonymName,
    s.base_object_name,
    s.create_date
FROM sys.synonyms s
WHERE s.name = 'CustomersSyn';
```

Expected: No linked columns

Actual: The synonym object name is detected as JSON 

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [x] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

Latest

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #20375, opened 2025-10-29
Enable full Json validation before setting column type as Json

manujoseph85:
## Description

Bug ref: [[Bug]: JSON column data type detection is incorrectly making links](https://github.com/microsoft/vscode-mssql/issues/20188)

We were doing a simple regex match for 'starts with [ or { and ends with ] or }' to determine if a column is Json. This will lead to many invalid values being picked as Json such as the one in mentioned in the bug.

This PR adds a multi-level check to quickly filter with regex and then parse with JSON.parse to confirm if a string is Json. It also fixes the Regex to match brackets correctly ({ to } and [ to ]).

Moved Json util function to a separate file under sharedInterfaces for unit testing. PR also adds unit tests for isJson function and IsJsonRegex regex.

## Code Changes Checklist

### DIFF
--- a/src/reactviews/common/jsonUtils.ts
+++ b/src/reactviews/common/jsonUtils.ts
@@ -3,11 +3,19 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-/**
- * Regular expression to match JSON objects and arrays
- * Matches strings that start with { or [ and end with } or ] respectively
+/*
+ * Determines if a string value has the shape of a JSON object or array
+ * @param value - string value to check
+ * @returns true if the value has the shape of a JSON object or array, false otherwise
  */
-export const IsJsonRegex = /^\s*(\{[\S\s]*\}|\[[\S\s]*\])\s*$/;
+function isJsonShape(value: string): boolean {
+    const trimmedValue = value?.trim();
+    return (
+        !!trimmedValue &&
+        ((trimmedValue.startsWith("{") && trimmedValue.endsWith("}")) ||
+            (trimmedValue.startsWith("[") && trimmedValue.endsWith("]")))
+    );
+}
 
 /**
  * Determines if a string value contains valid JSON data
@@ -16,7 +24,7 @@ export const IsJsonRegex = /^\s*(\{[\S\s]*\}|\[[\S\s]*\])\s*$/;
  * @returns true if the value contains valid JSON, false otherwise
  */
 export function isJson(value: string): boolean {
-    if (value?.match(IsJsonRegex)) {
+    if (isJsonShape(value)) {
         try {
             JSON.parse(value);
             return true;
--- a/src/reactviews/pages/QueryResult/resultGrid.tsx
+++ b/src/reactviews/pages/QueryResult/resultGrid.tsx
@@ -360,10 +360,6 @@ function getColumnFormatter(columnInfo: qr.IDbColumn): (
     };
 }
 
-function isJsonCell(value: qr.DbCellValue): boolean {
-    return !!(value && !value.isNull && value.displayValue?.match(IsJsonRegex));
-}
-
 function isXmlCell(value: DBCellValue, log?: LogCallback): boolean {
     let isXML = false;
     try {
--- a/test/unit/jsonUtils.test.ts
+++ b/test/unit/jsonUtils.test.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { isJson, IsJsonRegex } from "../../src/reactviews/common/jsonUtils";
+import { isJson } from "../../src/reactviews/common/jsonUtils";
 import { expect } from "@playwright/test";
 
 suite("isJsonCell Tests", () => {
@@ -138,56 +138,6 @@ suite("isJsonCell Tests", () => {
         test("Should return false for empty string", () => {
             const value = "";
             expect(isJson(value), "Empty string should return false").toBe(false);
-        });
-    });
-
-    suite("Regex validation", () => {
-        test("IsJsonRegex should match empty object", () => {
-            expect("{}".match(IsJsonRegex), "Empty object should match regex").toBeTruthy();
-        });
-
-        test("IsJsonRegex should match empty array", () => {
-            expect("[]".match(IsJsonRegex), "Empty array should match regex").toBeTruthy();
-        });
-
-        test("IsJsonRegex should match object with whitespace", () => {
-            expect(
-                "  { }  ".match(IsJsonRegex),
-                "Object with whitespace should match regex",
-            ).toBeTruthy();
-        });
-
-        test("IsJsonRegex should match array with whitespace", () => {
-            expect(
-                "  [ ]  ".match(IsJsonRegex),
-                "Array with whitespace should match regex",
-            ).toBeTruthy();
-        });
-
-        test("IsJsonRegex should not match plain string", () => {
-            expect(
-                !"hello world".match(IsJsonRegex),
-                "Plain string should not match regex",
-            ).toBeTruthy();
-        });
-
-        test("IsJsonRegex should not match quoted string", () => {
-            expect(
-                !'"hello world"'.match(IsJsonRegex),
-                "Quoted string should not match regex",
-            ).toBeTruthy();
-        });
-
-        test("IsJsonRegex should not match number", () => {
-            expect(!"123".match(IsJsonRegex), "Number should not match regex").toBeTruthy();
-        });
-
-        test("IsJsonRegex should match complex nested structure", () => {
-            const complexJson = '{"a": [1, {"b": []}]}';
-            expect(
-                complexJson.match(IsJsonRegex),
-                "Complex nested structure should match regex",
-            ).toBeTruthy();
         });
     });
 });


src/sharedInterfaces/jsonUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+export const IsJsonRegex = /^\s*(\{[\S\s]*\}|\[[\S\s]*\])\s*$/;
+
+/**
+ * Determines if a string value contains valid JSON data
+ * Uses a regular expression to quickly check the structure before attempting to parse
+ * @param value - string value to check
+ * @returns true if the value contains valid JSON, false otherwise
+ */
+export function isJson(value: string): boolean {
+    if (value && value.match(IsJsonRegex)) {
```
aasimkhan30: NIT: value?.match(isJsonRegex)

src/reactviews/common/jsonUtils.ts
aasimkhan30, line 1: I’d avoid putting this method in sharedInterfaces. That folder is intentionally types-only, and adding runtime code there means we’d need to guarantee both Node and browser compatibility (we don’t have lint rules or runtime tests for that yet). It’s also not an intuitive place to look for helpers.
Since this function is only used in one spot, could you move it back under Webviews and import it locally? We can revisit a proper shared runtime module later if we have more instances of shared code.

manujoseph85: Moved it back to the reactviews/common. Initially, an eslint rule was not letting me do that but updated the eslint rule to allow import of reactviews files in test folder

src/sharedInterfaces/jsonUtils.ts
manujoseph85: Changed

test/unit/jsonUtils.test.ts
```diff
@@ -0,0 +1,185 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from "assert";
```
Benjin: I'm in the process of swapping over all of our tests to use `chat`'s `expect(x).to.be(y)` so that it reads more fluently and we get better/more accurate failure messages by default.  Could you use that instead?

Benjin: Also, out of curiosity, how much (if at all) were these test cases generated - and if so, by which tool?  `test/unit/AGENTS.md` should instruct agents to write assertions in that way, but I might have to improve that if it isn't getting followed.

Last I looked, all the main tools except Claude know to look for that file (though maybe I should add a CLAUDE.md file that just redirects, or see if Git knows how to deal with a simlink between files)

src/reactviews/common/jsonUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Determines if a string value contains valid JSON data
+ * Uses a regular expression to quickly check the structure before attempting to parse
+ * @param value - string value to check
+ * @returns true if the value contains valid JSON, false otherwise
+ */
+export function isJson(value: string): boolean {
+    if (value?.match(IsJsonRegex)) {
+        try {
+            JSON.parse(value);
```
Benjin, line 29: Does SQL allow users to store invalid JSON (or XML for that matter) in a a JSON/XML column?  If they do, I don't think we should gate the ability to open a cell on whether the user's data is messed up, given that we don't actually _need_ to do anything with any successfully-parsed JSON.  I think it'd be confusing to a user who expects to be able to open a cell if some cells can be clicked and others can't, especially if the best way to see what's wrong with the data is to try to open it in the larger viewer.

If not, I guess this `parse()` call is just ceremonial for SELECTs.

test/unit/jsonUtils.test.ts
```diff
@@ -0,0 +1,185 @@
...
+    suite("Invalid JSON strings", () => {
+        test("Should return false for invalid object syntax", () => {
+            const value = '{name: "John"}'; // Missing quotes around key
+            assert.strictEqual(isJson(value), false, "Invalid object syntax should return false");
+        });
+
+        test("Should return false for invalid array syntax", () => {
+            const value = "[1, 2, 3,]"; // Trailing comma
+            assert.strictEqual(isJson(value), false, "Invalid array syntax should return false");
+        });
```
Benjin, line 89: ...huh.  Interesting how these are very common in JS for building objects and arrays, but are technically invalid in _JavaScript_ Object Notation...

src/reactviews/common/jsonUtils.ts
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Regular expression to match JSON objects and arrays
+ * Matches strings that start with { or [ and end with } or ] respectively
+ */
+export const IsJsonRegex = /^\s*(\{[\S\s]*\}|\[[\S\s]*\])\s*$/;
```
Benjin: If we're just checking for strings that are either beginning and ending with `{}` or `[]`, a regex seems heavy for that.  Why not a `trim()` followed by checking first and last characters to see if they're matching brackets?

src/reactviews/common/jsonUtils.ts
aasimkhan30: Does this regex have a perf penalty over the previous regex. 

Looking at the code in ADS. there does seem to be some perf issue on doing a more extensive regex.

https://github.com/microsoft/azuredatastudio/blob/e80988413d24e2107d45570e4b33fd86f776c7be/src/sql/workbench/contrib/query/browser/gridPanel.ts#L77-L85.

aasimkhan30: Here is the PR that fixes the issue. So a heavier regex is indeed pretty bad. 

https://github.com/microsoft/azuredatastudio/pull/21421

src/reactviews/pages/QueryResult/resultGrid.tsx
```diff
@@ -196,7 +197,11 @@ const ResultGrid = forwardRef<ResultGridHandle, ResultGridProps>((props: ResultG
                                           columnDef,
                                           dataContext,
                                       );
-                                  } else if (isJsonCell(value) && props.resultSetSummary) {
+                                  } else if (
```
aasimkhan30: How about we render all cells as plain text by default, and allow users to double-click (or press Enter/Space) to open a Fluent dialog with a Monaco editor?

1. On double-click or keyboard activation (Enter/Space), open a dialog containing a Monaco editor.
2. Parse the cell’s value to detect its data type.
3. Set the Monaco editor’s language mode based on the detected type.

This approach keeps the grid lightweight and delays the parsing to until user expand the cell. 


Image

However, it is a bigger feature work than this fix.

@kburtram and @croblesm, what do you think?

aasimkhan30 (changes requested):

test/unit/jsonUtils.test.ts
manujoseph85: I used GHC with Claude Sonnet to generate these tests. So GHC does not seem to be respecting that Agents.md file. We should figure out why. I can do some digging too.

Used another prompt to explicitly convert all the assertions to the chai format now.

test/unit/jsonUtils.test.ts
manujoseph85: Yep, I think JSON parsers are stricter than JS interpreters. Interesting learning for me too.

src/reactviews/common/jsonUtils.ts
manujoseph85: Looked at the regex that was fixed in the ADS PR (/({(.|\s)*?})|(\[(.|\s)*?\])/g). This is much heavier because it uses a global search and has multiple capture groups.

The modified regex that I am adding (/^\s*(\{[\S\s]*\}|\[[\S\s]*\])\s*$/) does not do global search for multiple capture groups. Its anchored at both ends with one simple alternation. So it should be very similar to the current regex that we have in terms of performance.

Our current regex (/^\s*[\{|\[][\S\s]*[\}\]]\s*$/g) is incorrect because it does not match brackets ({ to } or [ to ]). So I added a simple alternation to fix that. Theoretically, I believe that should not cause any perf impact but I will also run some tests to verify that.

manujoseph85: Ran the below perf benchmark on all the 3

```
// 2 MB of 'a', then a closing brace at the end
const bigObj = "{" + "a".repeat(2_000_000) + "}";
// same but with NO closing brace (pathological for lazy scans)
const badObj = "{" + "a".repeat(2_000_000);

const newRegex = /^\s*(?:\{[\s\S]*\}|\[[\s\S]*\])\s*$/;
const adsRegex = /({(.|\s)*?})|(\[(.|\s)*?\])/g;
const currentRegex = /^\s*[\{|\[][\S\s]*[\}\]]\s*$/g;

var b1_1 = bench("newRegex bigObj", () => bigObj.match(newRegex));
var b2_1 = bench("adsRegex bigObj", () => bigObj.match(adsRegex));
var b3_1 = bench("currentRegex bigObj", () => bigObj.match(currentRegex));

var b1_2 = bench("newRegex badObj", () => badObj.match(newRegex)); // fast fail
var b2_2 = bench("adsRegex badObj", () => badObj.match(adsRegex)); // likely much slower
var b3_2 = bench("currentRegex badObj", () => badObj.match(currentRegex));

function bench(label: string, fn: () => unknown): number {
    const t0 = performance.now();
    let r: unknown;
    for (let i = 0; i < 200; i++) r = fn(); // repeat to smooth noise
    const t1 = performance.now();
    console.log(label, (t1 - t0).toFixed(2), "ms", r && "");
    return t1 - t0;
}
```

manujoseph85: Results came out like below

newRegex bigObj 123.56 ms
**adsRegex bigObj 1193.37 ms**
currentRegex bigObj 125.80 ms

newRegex badObj 472.59 ms null
**adsRegex badObj 2137.31 ms null**
currentRegex badObj 474.46 ms null

ADS regex is really bad (heavy). Our current regex and new one proposed in the PR are similar (with new one performing slightly better)

manujoseph85: @Benjin to answer your original question, we are not using SQL columntype to determine if we need to show Json or Xml formatting. Perhaps because most people just store those in nvarchar. We parse the content to make the determination.

src/reactviews/common/jsonUtils.ts
manujoseph85: Thanks for the suggestion, this sounds like a great idea to me. I suspect that we used a Regex based validation originally because its extensible and results in clean code. This Json is efficient (does not cause any backtracking) and should be O(n). Trimming would be technically O(n) as well but in practice when I ran a perf test, the trim + check came out much superior in total time.

Since grid is very perf sensitive, I will switch to trim + check method. We are giving up some elegance/extensibility with Regex but in this case that's Ok, I think.

src/reactviews/common/jsonUtils.ts
manujoseph85: To conclude this thread, I am removing the Json and using a manual isJsonShape method (trim + check start and end) which is even better perf than the efficient Json.

src/reactviews/pages/QueryResult/resultGrid.tsx
manujoseph85: I think "double click to open any cell in monaco" is a nice idea. We give up the hyperlink styling for xml/json content so it might be less intuitive for users probably.

I can file this as a Feature Request and then get it triaged by Karl/Carlos.

PR #20387, opened 2025-10-31
Throttling query result set updates to prevent spamming webview rpc. 

aasimkhan30:
## Description

Doing this prevents spamming webview rpc when a large grid is streamed and we get a bunch of resultSet updates. This is the same logic we used in messages to prevent message spamming. 

## Code Changes Checklist


src/queryResult/queryResultWebViewController.ts
aasimkhan30, line 1: just cleanup

Issue #20349, opened 2025-10-23
[Bug]: “MSSQL: Select Top 1000” command fails with Cannot read properties of undefined (reading 'nodeType')


croblesm:
### Description

When running the MSSQL: Select Top 1000 command on a table from Object Explorer, the operation fails with the following error: `Cannot read properties of undefined (reading 'nodeType')`

Expected Behavior:
The command should open a new query editor with a SELECT TOP (1000) * FROM [schema].[table] statement for the selected object.

Actual Behavior:
The command throws a runtime error and does not generate the query.

### Steps to Reproduce

1. Set the `MSSSQL: Select Top 1000` key binding
2. Open Object Explorer and connect to a database.
3. Right-click on a table.
4. Choose Select Top 1000.
5. Observe the error message.

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

### If you selected "Other", please describe the affected area

Core

### Environment Information

- VS Code Version: Latest
- MSSQL extension version: Latest
- OS: macOS 15.7.1

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

Issue #20259, opened 2025-10-01
Scripting operations aren't cancelable


kburtram:
> Hi [@nycgavin](https://github.com/nycgavin) & [@match-gabeflores](https://github.com/match-gabeflores) just to clarify, I have two questions:
> 
> 1. Are you saying the cancel button (seen in screenshot below) is hidden while the query is still running?
> 
> I cannot see this button
> 
> 2. Do you see the cancel button right after starting the query execution, and it disappears later on, or is it never there in the first place?
> 
> ![Image](https://github.com/user-attachments/assets/3d5b054a-5ce1-46cd-bbff-0ce72a1b8676)
> 
> Cannot see the button 
> 
> All I see is this loading floating box:
> 
> <img width="470" height="107" alt="Image" src="https://github.com/user-attachments/assets/ba4b589f-a681-47ad-ab02-c3657cedbf27" />
> 
> It stays there for like 5 minutes straight. There is literally no way to remove it, unless I disable the extension's notifications. 
> 
> I am also on VPN
>  

 _Originally posted by @dbersan in [#18837](https://github.com/microsoft/vscode-mssql/issues/18837#issuecomment-3347722570)_

PR #20409, opened 2025-11-04
Scripting service refactor + keybinding support.

aasimkhan30:

## Description

1. Consolidating scripiting code 
2. Making script operation async, We immediately get an operation id from STS that is used to listen to progress/completion events.
STS PR: https://github.com/microsoft/sqltoolsservice/pull/2534
3. It also lets user cancel the scripting operation through the notification 

Fixes: #20349 and Fixes: #20259

## Code Changes Checklist

### DIFF
--- a/src/scripting/scriptingService.ts
+++ b/src/scripting/scriptingService.ts
@@ -292,7 +292,7 @@ export class ScriptingService {
             scriptTelemetryActivity.endFailed(error, false /* do not include error message */);
         }
 
-        UserSurvey.getInstance().promptUserForNPSFeedback("scriptAs");
+        UserSurvey.getInstance()?.promptUserForNPSFeedback("scriptAs");
     }
 
     public static getScriptCompatibility(serverMajorVersion: number, serverMinorVersion: number) {
--- a/test/unit/scriptingService.test.ts
+++ b/test/unit/scriptingService.test.ts
@@ -30,7 +30,13 @@ import * as Constants from "../../src/constants/constants";
 import * as LocalizedConstants from "../../src/constants/locConstants";
 import { ActivityObject, ActivityStatus } from "../../src/sharedInterfaces/telemetry";
 import { Logger } from "../../src/models/logger";
-import { getMockContext, initializeIconUtils, stubVscodeWrapper } from "./utils";
+import {
+    getMockContext,
+    initializeIconUtils,
+    stubUserSurvey,
+    stubVscodeWrapper,
+    stubWithProgress,
+} from "./utils";
 
 chai.use(sinonChai);
 
@@ -154,23 +160,23 @@ suite("Scripting Service", () => {
         telemetryActivities = [];
         registeredCommands = {};
 
-        withProgressStub = sandbox
-            .stub(vscode.window, "withProgress")
-            .callsFake(async (_options, task) => {
-                const progress = {
-                    report: sandbox.stub(),
-                } as vscode.Progress<{ message?: string; increment?: number }>;
+        stubUserSurvey(sandbox);
 
-                const token: vscode.CancellationToken = {
-                    isCancellationRequested: false,
-                    onCancellationRequested: (callback: (e: unknown) => void) => {
-                        cancellationCallback = callback;
-                        return { dispose: () => undefined } as vscode.Disposable;
-                    },
-                } as vscode.CancellationToken;
+        withProgressStub = stubWithProgress(sandbox, async (_options, task) => {
+            const progress = {
+                report: sandbox.stub(),
+            } as vscode.Progress<{ message?: string; increment?: number }>;
 
-                return await task(progress, token);
-            });
+            const token: vscode.CancellationToken = {
+                isCancellationRequested: false,
+                onCancellationRequested: (callback: (e: unknown) => void) => {
+                    cancellationCallback = callback;
+                    return { dispose: () => undefined } as vscode.Disposable;
+                },
+            } as vscode.CancellationToken;
+
+            return await task(progress, token);
+        });
 
         showErrorMessageStub = sandbox.stub(vscode.window, "showErrorMessage").resolves(undefined);
         showInformationMessageStub = sandbox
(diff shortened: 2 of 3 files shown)


src/scripting/scriptingService.ts
```diff
@@ -148,31 +409,156 @@ export class ScriptingService {
...
+                    return undefined;
+                }
+
+                this._onGoingScriptingOperations.set(operationId, scriptPromise);
+
+                const scriptResult = await scriptPromise.promise;
+
+                if (scriptResult.errorMessage) {
+                    scriptTelemetryActivity.endFailed(new Error(scriptResult.errorMessage), false);
+                    if (scriptResult.errorMessage === SCRIPT_OPERATION_CANCELED_ERROR) {
```
Benjin, line 556: `SCRIPT_OPERATION_CANCELED_ERROR` doesn't need localization because it's never actually displayed to the user, right?

src/scripting/scriptingService.ts
```diff
@@ -148,31 +409,156 @@ export class ScriptingService {
...
+                    ScriptingRequest.type,
+                    scriptingParams,
+                );
+
+                operationId = result?.operationId;
+
+                if (!operationId) {
+                    const error = new Error("Missing operation id from scripting response");
+                    scriptTelemetryActivity.endFailed(error, false);
+                    scriptTelemetryActivity.endFailed(error, true);
```
Benjin: Duplicate `endFailed()` calls.

Also, please add something like `/* includeErrorMessage */` next to the magic constants (I really wish TS would support named params...)

src/scripting/scriptingService.ts
```diff
@@ -11,26 +11,281 @@ import {
...
+        );
+
+        // Handle scripting progress notifications
         this._client.onNotification(ScriptingProgressNotification.type, (params) => {
-            this._client.logger.verbose(JSON.stringify(params));
-            if (params.errorMessage) {
-                const errorText = `Scripting progress error: ${params.errorMessage} - ${params.errorDetails}`;
-                this._client.logger.error(errorText);
-                vscode.window.showErrorMessage(errorText);
+            if (!params.errorMessage) {
```
Benjin, line 123: ...scripting operations report errors as a ProgressNotification, not CompleteNotification?  That's weird (unless it's able to continue through errors, I guess).

src/scripting/scriptingService.ts
```diff
@@ -11,26 +11,281 @@ import {
...
+        private _sqlDocumentService: SqlDocumentService,
+        private _sqlOutputContentProvider: SqlOutputContentProvider,
+        private _statusview: StatusView,
+        private _objectExplorerTree: vscode.TreeView<TreeNodeInfo>,
+    ) {
         this._client = this._connectionManager.client;
+
+        this._logger = Logger.create(this._vscodeWrapper.outputChannel, "ObjectExplorerService");
+
+        const pushDisposable = (disposable: vscode.Disposable): void => {
```
Benjin, line 65: Can you shift these registrations and onNotifications into an `initialize()` method to make this more readable?

src/scripting/scriptingService.ts
```diff
@@ -11,26 +11,281 @@ import {
...
+import { Logger } from "../models/logger";
+import VscodeWrapper from "../controllers/vscodeWrapper";
+import { UserSurvey } from "../nps/userSurvey";
+
+export const SCRIPT_OPERATION_CANCELED_ERROR = "Scripting operation cancelled by user.";
 
 export class ScriptingService {
     private _client: SqlToolsServiceClient;
+    private _onGoingScriptingOperations: Map<
+        string,
```
Benjin: Can you add a comment for what the key is supposed to be?  (I assume either a node ID or an editor URI, but would be nice to have that confirmed in code)

src/scripting/scriptingService.ts
```diff
@@ -11,26 +11,281 @@ import {
...
+                );
+                await queryPromise;
+                await this._connectionManager.connectionStore.removeRecentlyUsed(
+                    <IConnectionProfile>connectionCreds,
+                );
+            }
+
+            scriptTelemetryActivity.end(ActivityStatus.Succeeded);
+        } catch (error) {
+            this._logger.error("Scripting failed: ", error);
```
Benjin: `getErrorMessage()`

test/unit/scriptingService.test.ts
```diff
@@ -113,96 +140,581 @@ suite("Scripting Service Tests", () => {
...
+            server: "test-server",
+            database,
+        } as any);
+    }
 
-        expect(scriptingObject).to.include({
-            type: "Table",
-            schema: "dbo",
-            name: "test_table",
+    suiteSetup(() => {
```
Benjin, line 157: Why `suiteSetup()` instead of just `setup()`?  It'd run more often (but who cares because it's unit tests), and you wouldn't need to manually reset everything (which risks people updating this file forgetting to reset a new stub).

test/unit/scriptingService.test.ts
```diff
@@ -113,96 +140,581 @@ suite("Scripting Service Tests", () => {
...
+        sandbox = sinon.createSandbox();
+        initializeIconUtils();
+        telemetryActivities = [];
+        registeredCommands = {};
+
+        withProgressStub = sandbox
```
Benjin: Check `/test/unit/utils.ts` to see if there are any stubbers over there that you can reuse.

Also, add new ones that have generic behavior (vscode.window/window/workspace/command, telemetry, and logger seem like good candidates).

test/unit/scriptingService.test.ts
```diff
@@ -113,96 +140,581 @@ suite("Scripting Service Tests", () => {
...
 
-        const scriptingParams = scriptingService.createScriptingParamsFromNode(
-            testNode,
-            "test_uri",
+    suiteTeardown(() => {
+        sandbox.restore();
+    });
+
+    test("registers scripting commands", () => {
+        expect(registerCommandStub.callCount).to.equal(5);
```
Benjin: `(stub).to.have.callCount(5)` should yield a more useful failure message

Benjin: Also, pull the call count directly from the array of expected calls

test/unit/scriptingService.test.ts
```diff
@@ -113,96 +140,581 @@ suite("Scripting Service Tests", () => {
...
+            expect(sqlOutputContentProvider.runQuery).to.not.have.been.called;
+            expect(loggerStub.error).to.have.been.calledOnce;
+            expect(telemetryActivities[0].endFailed).to.have.been.calledOnce;
+        } finally {
+            scriptTreeStub.restore();
+        }
+    });
+
+    test("scriptNode logs error when scripting object is missing", async () => {
+        const node = new TreeNodeInfo(
```
Benjin, line 394: I think treeNodeInfo.test.ts's `createTreeNode()` can eliminate some redundancy.

src/scripting/scriptingService.ts
aasimkhan30: Maybe we should switch to an options params with props as they are easy to read.

src/scripting/scriptingService.ts
aasimkhan30: Some times it reports partials errors in progress and then returns an empty/incomplete script in completeEvent without any error. For our extension's use, I am treating any error as fatal as we don't want to give incorrect incomplete script to users.

test/unit/scriptingService.test.ts
aasimkhan30: It can but I think it is easier to just set it up locally for the test as I am adding object metadata to some of the treeNodes here.

Benjin (approved):

PR #20417, opened 2025-11-04
Publish Dialog - SQLCMD Variables section

ssreerama:

## Description

This PR adds the ability to view and edit the SQLCMD variables in the project, below are the operations that user can perform within the dialog
- Opening a publish dialog will read the sqlcmd variables from the sqlproj and display them in a table 
- values can be editable and at the same time can revert the edits using the revert icon preset in the table header.
- table header is sticky and can scroll the entire variables within the table.
- loading profile will append the list of sqlcmd varialbes to the table (variables from project + variables from profile)
- reset any variable will update the value to its default, variables coming from project will be default to the project values and variables coming from the profile will be reset to the profile values.
- save profile will add all variables present in the table to the publish.xml file

## Code Changes Checklist

5 SQLCMDVariables

### DIFF
--- a/localization/l10n/bundle.l10n.json
+++ b/localization/l10n/bundle.l10n.json
@@ -1512,7 +1512,7 @@
     "Database name is required": "Database name is required",
     "Publish Target": "Publish Target",
     "Existing SQL server": "Existing SQL server",
-    "Existing Azure SQL logical server": "Existing Azure SQL logical server",
+    "Existing Azure SQL server": "Existing Azure SQL server",
     "Local development container": "Local development container",
     "New SQL Server local development container": "New SQL Server local development container",
     "New Azure SQL logical server (Preview)": "New Azure SQL logical server (Preview)",
--- a/localization/xliff/vscode-mssql.xlf
+++ b/localization/xliff/vscode-mssql.xlf
@@ -1392,8 +1392,8 @@
     <trans-unit id="++CODE++bd705ebabe1919a636a0e09b2ee27d43a5aba05d645cc0339451e54b37f42bd6">
       <source xml:lang="en">Execution Plan</source>
     </trans-unit>
-    <trans-unit id="++CODE++b5f54c19df1649916f59331f9decf53dc337a2c241a3afb20d58810309ebe30f">
-      <source xml:lang="en">Existing Azure SQL logical server</source>
+    <trans-unit id="++CODE++6e70f2efe82744e86406531b7767f19ad25970b8075f1ef14fd2b91a94ab1b93">
+      <source xml:lang="en">Existing Azure SQL server</source>
     </trans-unit>
     <trans-unit id="++CODE++1b2da57e60f89f7e5fa102ef76d7ef2ca1d7d212380c0200d648e2215a5f555d">
       <source xml:lang="en">Existing SQL server</source>
--- a/src/constants/locConstants.ts
+++ b/src/constants/locConstants.ts
@@ -1346,7 +1346,7 @@ export class PublishProject {
     public static SqlCmdVariablesLabel = l10n.t("SQLCMD Variables");
     public static PublishTargetLabel = l10n.t("Publish Target");
     public static PublishTargetExisting = l10n.t("Existing SQL server");
-    public static PublishTargetExistingLogical = l10n.t("Existing Azure SQL logical server");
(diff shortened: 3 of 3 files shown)


src/reactviews/pages/PublishProject/components/sqlCmdVariablesSection.tsx
```diff
@@ -0,0 +1,231 @@
...
+            }));
+
+            const updatedVariables = {
+                ...sqlCmdVariables,
+                [varName]: newValue,
+            };
+            publishCtx.formAction({
+                propertyName: sqlCmdComponent.propertyName,
+                isAction: false,
+                value: updatedVariables as unknown as string | boolean,
```
Benjin: This type doesn't match up at all with the definition above, right?  You're casting a record or object to a `string | boolean`.  What's happening here?

ssreerama: The FormEvent.value is typed as string | boolean to handle simple form fields (text inputs, checkboxes). However, sqlCmdVariables is a complex property - an object with dynamic keys { [key: string]: string }.

At runtime, the form controller handles this correctly: It uses as any to accept any type, not just string | boolean.

The double-cast as unknown as string | boolean is the TypeScript workaround to satisfy the type checker while passing objects through a system designed for simple values.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -806,9 +851,33 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+     */
+    private allSqlCmdVariablesFilled(): boolean {
+        const sqlCmdVariables = this.state.formState.sqlCmdVariables;
+        if (!sqlCmdVariables) {
+            return true; // No variables to validate
+        }
+
+        for (const varName in sqlCmdVariables) {
+            const value = sqlCmdVariables[varName];
+            if (value === "" || value === undefined) {
```
llali: what about empty string? and if there's already a validation somewhere, why not using a shared code?

ssreerama: the comment is misleading, it was added to tell that the logic is similar to ADS sql projects validation, which can be removed.  also it is checking for the empty sting already

src/publishProject/projectUtils.ts
```diff
@@ -291,3 +291,26 @@ export async function parsePublishProfileXml(
...
+
+    // Check all variable values are non-empty
+    for (const varName in sqlCmdVariables) {
+        if (Object.prototype.hasOwnProperty.call(sqlCmdVariables, varName)) {
+            const value = sqlCmdVariables[varName];
+            if (value === "" || value === undefined) {
+                return false;
+            }
+        }
+    }
```
Benjin: I think this can be simplified to `return Object.values(sqlCmdVariables).every(v => v !== "" && v !== undefined);`

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -471,14 +472,15 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
 
                     // Update immutable default values: project defaults + profile overrides
-                    this._defaultSqlCmdVariables = {
-                        ...this._defaultSqlCmdVariables,
+                    const updatedDefaults = {
+                        ...state.defaultSqlCmdVariables,
                         ...parsedProfile.sqlCmdVariables,
                     };
 
                     // Update state with loaded profile data immediately (non-blocking)
```
Benjin: ...isn't this blocking?  Not that this is a problem because this is just a bunch of assignments, but the comment isn't accurate

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -541,6 +545,35 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                state: PublishDialogState,
+                payload: { variables: { [key: string]: string } },
+            ) => {
+                return {
+                    ...state,
+                    formState: {
+                        ...state.formState,
+                        sqlCmdVariables: payload.variables,
+                    },
+                };
```
Benjin: Does `state.formState.sqlCmdVariables = payload.variables` not work for some reason?  I'm seeing this class create new state objects instead of modifying the existing one a bunch.

Benjin (dismissed):

Benjin (dismissed):

Issue #20331, opened 2025-10-21
[Bug]: Results Grid keyboard shortcuts not working


Andrew0002:
### Description

I'm trying to switch from ADS to the VS Code MS SQL Extension. For the VS Code extension, I tried configuring keyboard shortcuts for copying grid results with headers, without headers, and only headers by following the readme [here](https://github.com/microsoft/vscode-mssql/tree/e95c11ad5081d4b3ef5916bb13ce0f92c02e4de3?tab=readme-ov-file#options) and customizing the shortcuts (delivered list [here](https://github.com/microsoft/vscode-mssql/blob/b8d588bf3416d6ecc9e46f24cb11889396071f92/package.json#L1542)). Tried this on both a macOS system and a Windows VM, couldn't get it working on either one. Maybe it's user error on my end - not sure.

Not sure if this is related, but I noticed that a little over a year ago there was a port of results grid code from Azure Data Studio to this plugin [here](https://github.com/microsoft/vscode-mssql/pull/18041). In Azure Data Studio, these are configured in the Keyboard Shortcuts page rather than settings.json. I looked at the Keyboard Shortcuts for the VS Code Plugin and don't see any commands for the copy functions.

(Azure Data Studio Screenshot)
<img width="856" height="200" alt="Image" src="https://github.com/user-attachments/assets/9a69f233-b916-4bbd-9b8e-98218f678308" />

### Steps to Reproduce

1. Go to VS Code's global Keyboard Shortcuts window and make sure the key combinations you want aren't already being used
2. Edit settings.json, add shortcut keys for event.copySelection, event.copyWithHeaders, and event.copyAllHeaders

    "mssql.shortcuts": {
        "_comment": "Short cuts must follow the format (ctrl)+(shift)+(alt)+[key]",
        "event.toggleResultPane": "ctrl+alt+R",
        "event.focusResultsGrid": "ctrl+alt+G",
        "event.toggleMessagePane": "ctrl+alt+Y",
        "event.prevGrid": "ctrl+up",
        "event.nextGrid": "ctrl+down",
        "event.copySelection": "ctrl+C",
        "event.copyWithHeaders": "ctrl+shift+C",
        "event.copyAllHeaders": "ctrl+shift+H",
        "event.maximizeGrid": "",
        "event.selectAll": "ctrl+A",
        "event.saveAsJSON": "",
        "event.saveAsCSV": "",
        "event.saveAsExcel": "",
[...]

PR #20418, opened 2025-11-04
Adding keyboard shortcuts back to result grid. 

aasimkhan30:
# Description

Adds a new framework to add shortcuts to items in results grid. 
1. Shortcuts and actions  are defined in the existing "mssql.shortcuts" configuration from old Webviews.
2. `vsCodeWebviewProvider` parses this config information and gives you the key combination for the keyboard event and a readable shortcut label. 
3. Added utility to match the keyboard event to the webviewProvider's key combinations.
4. Labels can be use to give information about those actions to users.
Example:
image

 Fixes: #20331 

## Code Changes Checklist

### DIFF
--- a/test/unit/keyboardUtils.test.ts
+++ b/test/unit/keyboardUtils.test.ts
@@ -1,0 +1,490 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { expect } from "chai";
+import * as sinon from "sinon";
+import {
+    getShortcutInfo,
+    parseWebviewKeyboardShortcutConfig,
+    eventMatchesShortcut,
+} from "../../src/reactviews/common/keyboardUtils";
+import {
+    WebviewAction,
+    WebviewKeyCombination,
+    WebviewKeyBindingConfiguration,
+} from "../../src/sharedInterfaces/webview";
+import * as utils from "../../src/reactviews/common/utils";
+
+suite("keyboardUtils Tests", () => {
+    let isMacStub: sinon.SinonStub;
+
+    setup(() => {
+        // Mock navigator.platform for isMac() function
+        // Default to non-Mac platform
+        isMacStub = sinon.stub(utils, "isMac").returns(false);
+    });
+
+    teardown(() => {
+        isMacStub.restore();
+    });
+    suite("getShortcutInfo", () => {
+        test("should return empty label and combination for undefined", () => {
+            const result = getShortcutInfo(undefined);
+            expect(result.label).to.equal("");
+            expect(result.keyCombination).to.deep.equal({});
+        });
+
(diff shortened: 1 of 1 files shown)


src/deployment/localContainersHelpers.ts
aasimkhan30, line 1: Unrelated test fixes for Mac.

src/reactviews/common/keyboardUtils.ts
```diff
@@ -0,0 +1,284 @@
...
+    event: KeyboardEvent | ReactKeyBoardEvent<any>,
+    combo: WebviewKeyCombination,
+): boolean {
+    if (!combo) return false;
+
+    // Treat undefined modifier as "don't care".
+    if (combo.ctrlKey !== undefined && combo.ctrlKey !== event.ctrlKey) return false;
+    if (combo.metaKey !== undefined && combo.metaKey !== event.metaKey) return false;
+    if (combo.altKey !== undefined && combo.altKey !== event.altKey) return false;
+    if (combo.shiftKey !== undefined && combo.shiftKey !== event.shiftKey) return false;
```
Benjin: For modifier keys, shouldn't we treat undefined modifiers as "must not be pressed"?

src/reactviews/pages/QueryResult/table/plugins/cellSelectionModel.plugin.ts
```diff
@@ -604,92 +607,123 @@ export class CellSelectionModel<T extends Slick.SlickData>
...
-            keyCode === KeyCode.ArrowLeft ||
-            keyCode === KeyCode.ArrowRight ||
-            keyCode === KeyCode.ArrowUp ||
-            keyCode === KeyCode.ArrowDown;
-
-        if (isArrow && e.shiftKey && !e.altKey && !metaOrCtrlPressed) {
-            this.expandSelection(keyCode);
+        if (
+            !isHandled &&
+            eventMatchesShortcut(
```
Benjin, line 615: Can `eventMatchesShortcut()` pull the keyCombination property automatically to make this even more concise?

Benjin: Actually, given that the reactwebviewbasee has the keybindings prop set, could you stick an `eventMatchesShortcut()` helper call somewhere in the context provider to get this consolidated down to:

`matchesShortcut(e, WebviewAction.ResultGridExpandSelectionLeft)`

Not sure if possible, just an idea

Benjin (dismissed):

lewis-sanchez (dismissed):

src/reactviews/pages/QueryResult/table/plugins/cellSelectionModel.plugin.ts
aasimkhan30: It’s possible, but it becomes much simpler if we keep the keybindings as props. That way they’re easy to use in non-React code, and the React parent can update them through a useEffect. Otherwise, we’d have to pass around a method and ensure it stays in sync, which is less intuitive and harder to maintain.

Benjin (approved):

Issue #20428, opened 2025-11-05
[Bug]: “Save Changes” button remains disabled after deleting a row


yoleichen:
### Description

https://github.com/user-attachments/assets/afe65ad2-74d4-4e46-a585-eb81db87ffb9

After deleting a row, the “Save Changes” button stays disabled even though there are unsaved changes. This makes it impossible to persist the deletion.

**Expected behavior:** 
 The “Save Changes” button should be enabled after a deletion, allowing the user to commit the change.

### Steps to Reproduce

1. Open a table with at least one row.
2. Delete a row.
3. Observe the “Row deleted” notification.
4. Notice that the “Save Changes” button is disabled.
5. Close the table window.
6. You’ll get a warning that there are unsaved changes.
7. Reopen the table — the deleted row is still present.

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

### If you selected "Other", please describe the affected area

View/Edit Data

### Environment Information

VS Code Version: Latest
MSSQL extension Version: Latest
OS: Windows

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #20435, opened 2025-11-05
Enable save button when a row is deleted

lewis-sanchez:

## Description
This PR fixes https://github.com/microsoft/vscode-mssql/issues/20428

Here's a demo of the save button enabling after a row is deleted.
Delete Row Enables Save Button

## Code Changes Checklist

### DIFF
--- a/src/reactviews/pages/TableExplorer/TableDataGrid.tsx
+++ b/src/reactviews/pages/TableExplorer/TableDataGrid.tsx
@@ -38,6 +38,7 @@ interface TableDataGridProps {
     onRevertRow?: (rowId: number) => void;
     onLoadSubset?: (rowCount: number) => void;
     onCellChangeCountChanged?: (count: number) => void;
+    onDeletionCountChanged?: (count: number) => void;
 }
 
 export interface TableDataGridRef {
@@ -61,6 +62,7 @@ export const TableDataGrid = forwardRef<TableDataGridRef, TableDataGridProps>(
             onRevertRow,
             onLoadSubset,
             onCellChangeCountChanged,
+            onDeletionCountChanged,
         },
         ref,
     ) => {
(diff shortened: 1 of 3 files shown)


src/reactviews/pages/TableExplorer/TableDataGrid.tsx
```diff
@@ -108,6 +112,9 @@ export const TableDataGrid = forwardRef<TableDataGridRef, TableDataGridProps>(
             if (onCellChangeCountChanged) {
                 onCellChangeCountChanged(0);
             }
+            if (onDeletionCountChanged) {
```
Benjin, line 115: Is there risk that this might not be set in the component props?  Could also simplify to `onDeletionCountChanged?(0)`, I _think_, since the check is more of a formality.

src/reactviews/pages/TableExplorer/TableExplorerToolbar.tsx
```diff
@@ -36,9 +38,8 @@ export const TableExplorerToolbar: React.FC<TableExplorerToolbarProps> = ({
         context.createRow();
     };
 
-    // Use cell-level change count directly
-    // This provides accurate granularity: each cell edit counts as one change
-    const changeCount = cellChangeCount;
+    // Total changes includes both cell edits and row deletions
+    const changeCount = cellChangeCount + deletionCount;
```
Benjin, line 42: the correctness of this sum doesn't matter, just whether it's 0, right?  otherwise, I'd wonder how it handles editing a cell then deleting the row it's in, but I don't think that matters.

lewis-sanchez: The only part that matters is that the count isn't 0. The current implementation should just count it as two changes.

PR #20436, opened 2025-11-05
Publish Dialog - Generate Script and Publish project to the database functionality

ssreerama:

## Description

This PR adds the ability to generate script and publish the project to a database. We are using vscode task execution to do the dotnet build same as in ADS, for this I have created a project Controller specifically for the project related methods, in future if we are moving the build and other functionalities to vscode repo, we can simply utilize and update it.

- Generate Script: Builds the project and calls DacFx API to get the script opened in the new window, disposes the publish dialog
- Publish: Button click builds the project and calls DacFx API to publish the project and disposes the dialog.
- Also, added bunch of telemetry points for generate script, build projects and publish project along with some failure telemetry.
- Added couple of tests to validate the generate Script and publish functionality and also for sdk and legacy project builds.

## Code Changes Checklist

6 1 GenerateScript

### DIFF
--- a/localization/LCL/de/vscode-mssql.xlf.lcl
+++ b/localization/LCL/de/vscode-mssql.xlf.lcl
@@ -757,6 +757,15 @@
         </Str>
         <Disp Icon="Str" />
       </Item>
+      <Item ItemId=";bundle.++CODE++11b16df18b896654fb041aa90fd64079d60b741da33d4b5f931d75f8dcc7d1a4" ItemType="0" PsrId="308" Leaf="true">
+        <Str Cat="Text">
+          <Val><![CDATA[Failed to list databases]]></Val>
+          <Tgt Cat="Text" Stat="Loc" Orig="New">
+            <Val><![CDATA[Fehler beim Auflisten von Datenbanken]]></Val>
+          </Tgt>
+        </Str>
+        <Disp Icon="Str" />
+      </Item>
       <Item ItemId=";bundle.++CODE++11fc4c8107afd167c617e0099298c647d423931e67c1257d1f4eb7811b2512fd" ItemType="0" PsrId="308" Leaf="true">
         <Str Cat="Text">
           <Val><![CDATA[Database Name is required]]></Val>
@@ -1666,6 +1675,15 @@
         </Str>
         <Disp Icon="Str" />
       </Item>
+      <Item ItemId=";bundle.++CODE++28b9471fbcdd3a3ce3799551c1cee3825afe4ea4ce67688c764bd4a364028bd9" ItemType="0" PsrId="308" Leaf="true">
+        <Str Cat="Text">
+          <Val><![CDATA[Scaffold backend components (e.g., data-access layers) based on your current database context]]></Val>
+          <Tgt Cat="Text" Stat="Loc" Orig="New">
+            <Val><![CDATA[Erstellen eines Gerüsts für Back-End-Komponenten (z. B. Datenzugriffsebenen) basierend auf Ihrem aktuellen Datenbankkontext]]></Val>
+          </Tgt>
+        </Str>
+        <Disp Icon="Str" />
+      </Item>
       <Item ItemId=";bundle.++CODE++28c3a8199a6dcd1373eb25c32820524f38fbba49d1e1fb378bd30dfb6d61ec2d" ItemType="0" PsrId="308" Leaf="true">
         <Str Cat="Text">
           <Val><![CDATA[MSSQL - Azure Auth Logs]]></Val>
@@ -1815,6 +1833,15 @@
           <Val><![CDATA[{0} rows selected, click to load summary]]></Val>
           <Tgt Cat="Text" Stat="Loc" Orig="New">
             <Val><![CDATA[{0} Zeilen ausgewählt, klicken Sie, um die Zusammenfassung zu laden]]></Val>
+          </Tgt>
+        </Str>
+        <Disp Icon="Str" />
+      </Item>
+      <Item ItemId=";bundle.++CODE++2c1ff609e9acedfcd56ba1c75b3cc704ce0c59afafb9f280e75d37376028e9cd" ItemType="0" PsrId="308" Leaf="true">
+        <Str Cat="Text">
+          <Val><![CDATA[Open SQL editor and connect]]></Val>
+          <Tgt Cat="Text" Stat="Loc" Orig="New">
+            <Val><![CDATA[Öffnen des SQL-Editor, und herstellen einer Verbindung]]></Val>
           </Tgt>
         </Str>
         <Disp Icon="Str" />
@@ -2616,6 +2643,15 @@
           <Val><![CDATA[Publish Project]]></Val>
           <Tgt Cat="Text" Stat="Loc" Orig="New">
             <Val><![CDATA[Projekt veröffentlichen]]></Val>
+          </Tgt>
+        </Str>
+        <Disp Icon="Str" />
+      </Item>
+      <Item ItemId=";bundle.++CODE++42e746a45011321b661459a4a5bff6312360244050814e0813284fb729dfc90d" ItemType="0" PsrId="308" Leaf="true">
+        <Str Cat="Text">
+          <Val><![CDATA[How many tables are in this database?]]></Val>
+          <Tgt Cat="Text" Stat="Loc" Orig="New">
+            <Val><![CDATA[Wie viele Tabellen befinden sich in dieser Datenbank?]]></Val>
           </Tgt>
         </Str>
         <Disp Icon="Str" />
@@ -4339,6 +4375,15 @@
         </Str>
         <Disp Icon="Str" />
       </Item>
+      <Item ItemId=";bundle.++CODE++6aafab6eae72092eb7c25ffb6f34dab1a3eb70f1ef79e6e159f0fe540ca6b675" ItemType="0" PsrId="308" Leaf="true">
+        <Str Cat="Text">
+          <Val><![CDATA[Receive natural language explanations to help developers unfamiliar with T-SQL understand code]]></Val>
+          <Tgt Cat="Text" Stat="Loc" Orig="New">
+            <Val><![CDATA[Empfangen von Erklärungen in natürlicher Sprache, um Entwicklern, die mit T-SQL nicht vertraut sind, zu helfen, Code zu verstehen]]></Val>
+          </Tgt>
+        </Str>
+        <Disp Icon="Str" />
+      </Item>
       <Item ItemId=";bundle.++CODE++6ab58ce9ca23e017bf594366b66b620458a5580640ddaf3aebf96f3c8fdf0977" ItemType="0" PsrId="308" Leaf="true">
         <Str Cat="Text">
           <Val><![CDATA[A firewall rule is required to access this server.]]></Val>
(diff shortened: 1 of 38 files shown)


src/controllers/projectController.ts
```diff
@@ -0,0 +1,144 @@
...
+                projectProperties.projectStyle,
+            );
+
+            // Create task arguments
+            const args: string[] = [constants.build, projectFilePath, ...buildArgs];
+
+            // Create task definition
+            const taskDefinition: vscode.TaskDefinition = {
+                type: constants.sqlProjBuildTaskType,
+                label: `Build ${projectName}`,
```
Benjin: Should this be localized?

src/controllers/projectController.ts
```diff
@@ -0,0 +1,144 @@
...
+            await this.executeBuildTask(buildTask, projectName);
+
+            // Return the DACPAC output path
+            return projectProperties.dacpacOutputPath;
+        } catch (error) {
+            // Send error telemetry
+            sendErrorEvent(
+                TelemetryViews.SqlProjects,
+                TelemetryActions.BuildProject,
+                error instanceof Error ? error : new Error(String(error)),
```
Benjin: getErrorMessage()

src/controllers/projectController.ts
```diff
@@ -0,0 +1,144 @@
...
+        buildDirPath: string,
+        projectStyle: mssql.ProjectType,
+    ): string[] {
+        const args: string[] = ["/p:NetCoreBuild=true", `/p:SystemDacpacsLocation=${buildDirPath}`];
+
+        // Adding NETCoreTargetsPath only for non-SDK style projects
+        // ProjectType.SdkStyle = 0, ProjectType.LegacyStyle = 1
+        const isSdkStyle = projectStyle === 0;
+
+        if (!isSdkStyle) {
```
Benjin: Why not just do `if(projectStyle === mssql.ProjectType.LegacyStyle) {}`?

src/controllers/projectController.ts
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Executes a build task and waits for it to complete
+     * @param buildTask The VS Code task to execute
+     * @param projectName Name of the project being built
+     */
+    private async executeBuildTask(buildTask: vscode.Task, projectName: string): Promise<void> {
+        return vscode.window.withProgress(
+            {
+                location: vscode.ProgressLocation.Notification,
+                title: `Building ${projectName}...`,
```
Benjin: Loc

src/controllers/projectController.ts
```diff
@@ -0,0 +1,144 @@
...
+                            disposable.dispose();
+                            if (e.exitCode === 0) {
+                                resolve();
+                            } else {
+                                const errorMsg = `Build failed with exit code ${e.exitCode}`;
+                                reject(new Error(errorMsg));
+                            }
+                        }
+                    });
+                });
```
Benjin, line 147: Does this actually need to be in a Promise with resolve/reject since it's in an async block?  Would writing normal async code with a "throw" work fine here?

src/publishProject/projectUtils.ts
```diff
@@ -104,7 +105,14 @@ export async function getProjectTargetVersion(
 export async function readProjectProperties(
     sqlProjectsService: SqlProjectsService | mssql.ISqlProjectsService,
     projectFilePath: string,
-): Promise<(mssql.GetProjectPropertiesResult & { targetVersion?: string }) | undefined> {
+): Promise<
+    | (mssql.GetProjectPropertiesResult & {
+          targetVersion?: string;
+          projectFilePath: string;
+          dacpacOutputPath: string;
+      })
```
Benjin: This is getting to be a little ugly... can you either update GetProjectPropertiesResult or define a new type that has this type union and use that here instead?

src/publishProject/projectUtils.ts
```diff
@@ -114,9 +122,20 @@ export async function readProjectProperties(
...
         }
         const version = await getProjectTargetVersion(sqlProjectsService, projectFilePath);
+
+        // Calculate DACPAC output path
+        const projectDir = path.dirname(projectFilePath);
+        const projectName = path.basename(projectFilePath, path.extname(projectFilePath));
+        const outputPath = path.isAbsolute(result.outputPath)
+            ? result.outputPath
+            : path.join(projectDir, result.outputPath);
+        const dacpacOutputPath = path.join(outputPath, `${projectName}.dacpac`);
```
Benjin: there should already be a constant with the dacpac file extension.  If you can't find one, can you add it?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -126,6 +134,117 @@ export class PublishProjectWebViewController extends FormWebviewController<
             });
     }
 
+    /**
+     * Executes publish and generate script operations
+     * @param state Current dialog state
+     * @param isPublish If true, publishes to database; if false, generates script
+     */
+    private async executePublishAndGenerateScript(
```
Benjin: I think this method would benefit from being broken up into build, publish, and script helpers

src/publishProject/publishProjectWebViewController.ts
Benjin, line 69: Can you extract these out to something like `const SQLPROJ_PUBLISH_VIEW_ID = "publishProject"`?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -126,6 +134,117 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                    databaseName,
+                    connectionUri,
+                    TaskExecutionMode.script,
+                    sqlCmdVariables,
+                    state.deploymentOptions,
+                );
+            }
+
+            // Check result and send telemetry
+            if (result.success) {
```
Benjin, line 194: can you also add this here:

```
UserSurvey.getInstance().promptUserForNPSFeedback(SQLPROJ_PUBLISH_VIEW_ID);
```

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -126,6 +134,117 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                isPublish
+                    ? TelemetryActions.PublishProjectChanges
+                    : TelemetryActions.GenerateScript,
+                error instanceof Error ? error : new Error(String(error)),
+                false,
+            );
+
+            // Show error notification
+            void vscode.window.showErrorMessage(
+                `${isPublish ? "Publish" : "Generate script"} failed: ${getErrorMessage(error)}`,
```
Benjin: Localize

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -280,6 +403,32 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                    if (parsedProfile.connectionString) {
+                        void this.connectAndPopulateDatabases(parsedProfile.connectionString).then(
+                            (connectionResult) => {
+                                // Update connection fields after background connection completes
+                                this._connectionUri =
+                                    connectionResult.connectionUri || this._connectionUri;
+                                if (connectionResult.errorMessage) {
+                                    this.state.formMessage = {
+                                        message: Loc.ProfileLoadedConnectionFailed,
+                                        intent: "error" as const,
```
Benjin: Does this need "as const"?

test/unit/projectController.test.ts
```diff
@@ -0,0 +1,176 @@
...
+import * as constants from "../../src/constants/constants";
+
+suite("ProjectController Tests", () => {
+    let sandbox: sinon.SinonSandbox;
+    let projectController: ProjectController;
+
+    // Common test constants
+    const projectFilePath = "c:/work/TestProject.sqlproj";
+    const projectName = "TestProject";
+    const dacpacOutputPath = "c:/work/bin/Debug/TestProject.dacpac";
```
Benjin, line 26: Can you also add tests that ensure this works for mac/linux-style paths, too?  Might need to stub `os.platform` for testing

test/unit/projectController.test.ts
```diff
@@ -0,0 +1,176 @@
...
+        const shellExec = taskArg.execution as vscode.ShellExecution;
+        const args = shellExec.args as string[];
+        expect(args[0]).to.equal(constants.build);
+        expect(args[1]).to.equal(projectFilePath);
+        expect(args).to.include("/p:NetCoreBuild=true");
+        expect(args.some((arg) => arg.includes("SystemDacpacsLocation"))).to.be.true;
+        expect(
+            args.some((arg) => arg.includes("NETCoreTargetsPath")),
+            "SDK-style should NOT include NETCoreTargetsPath",
+        ).to.be.false;
```
Benjin: Any time you're asserting against true or false, you should check to see if there's a better (more natural-language) assertion option that'll give you better failures messages for free.  In this case:

`expect(args).to.not.include(NETCoreTargetsPath)`

test/unit/projectController.test.ts
```diff
@@ -0,0 +1,176 @@
...
+        // Act
+        const buildPromise = projectController.buildProject(projectProperties);
+        triggerTaskCompletion(0); // Simulate successful task completion
+        await buildPromise;
+
+        // Assert - Only verify the difference: Legacy-style SHOULD include NETCoreTargetsPath
+        const taskArg = executeTaskStub.firstCall.args[0] as vscode.Task;
+        const shellExec = taskArg.execution as vscode.ShellExecution;
+        const args = shellExec.args as string[];
+        expect(
```
Benjin: Here as well

test/unit/publishProjectWebViewController.test.ts
```diff
@@ -659,4 +661,107 @@ suite("PublishProjectWebViewController Tests", () => {
...
+        const reducerHandlers = controller["_reducerHandlers"] as Map<string, Function>;
+        const generatePublishScript = reducerHandlers.get("generatePublishScript");
+        expect(generatePublishScript, "generatePublishScript reducer should be registered").to
+            .exist;
+
+        // Invoke the reducer
+        await generatePublishScript(controller.state, {});
+
+        // Verify dialog was closed
+        expect(
```
Benjin: In this case, you can use sinon-chat (see adapter.test.ts) to do something like:

`expect(panelDisposeSpy).to.have.been.calledOnce()`

Benjin: Can you also fix up the other assertions in this file?

src/controllers/projectController.ts
ssreerama: The Promise pattern is necessary, can't use simple async/await here because executeTask() doesn't wait for completion. This ensures the build completes successfully before the publish proceeds.

src/controllers/projectController.ts
ssreerama: getting circular dependency issue when using the mssql.ProjectType enum while runtime. 

SQL Database Projects extension declares a dependency on vscode-mssql extension
If projectController.ts uses regular import (import * as mssql), it tries to activate the vscode-mssql extension at load time
But vscode-mssql extension might not be fully activated yet
. This causes the activation failure: "dependency failed to activate" The solution is to keep import type to avoid triggering activation and use the numeric literal 1 for the enum value.

src/publishProject/projectUtils.ts
ssreerama: True, I felt it when i add new property to it. In future it will be ended up with more properties and having an extended property result would be easy to maintain.

Benjin (approved):

Issue #20426, opened 2025-11-05
[Feature Request]: Improve Discoverability of Newly Added Rows


yoleichen:
### Feature Description

https://github.com/user-attachments/assets/7153eb9e-4283-4e8d-85e1-683d93eebc30

When clicking "Add Row", the new row is added to the last page of the paginated table. This reduces discoverability, as users may not realize a new row has been added if they are on an earlier page.

**Expected Behavior:**
After adding a new row, the user should be able to see the new row immediately.

Possible solutions:
- Automatically navigate to the last page containing the new row.
- Insert the new row at the top of the current page.
- Insert the new rows at the bottom of the row where the cursor is at / the cell that's selected

(@jdesai-24  Any suggestion on which possible solution is better?) 

### Problem and Motivation

This behavior causes confusion, interrupts workflow, and requires extra steps to confirm that the row was actually created (e.g., manually navigating to the last page or refreshing the table).

### Related Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

### If you selected "Other", please describe the area

View/Edit Data

### Confirmation

- [x] I have searched existing feature requests and couldn't find a match
- [ ] I want to help implement this feature

yoleichen:
@lewis-sanchez Per discussion today, the new rows will be added at the last page, and we automatically focus / take user to the last page.

Issue #20419, opened 2025-11-04
[Bug]: Unable to create row in empty table


yoleichen:
### Description

https://github.com/user-attachments/assets/2c452043-fbc6-4695-9940-7c721e32b205

Click "Add Row" in an empty table with no data. The "Row created" notification shows up, and the SQL Script also add a new INSERT statement. However, the new row does not appear in the table UI.

New observation: If I change the "Rows per page" #, the new row appears in the UI. This bug might be caused by new rows not rendering correctly but it's actually already created in the backend. 

**Expected Behavior**
A new row should appear immediately in the table view after creation.

### Steps to Reproduce

1. Open the Edit Data grid for an empty table with no data
2. Click "Add Row"

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

### If you selected "Other", please describe the affected area

View/Edit Data

### Environment Information

VS Code Version: Latest
MSSQL extension Version: Latest
OS: Windows

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #20440, opened 2025-11-05
Fix bug around adding a row to an empty grid

lewis-sanchez:

## Description

This PR fixes https://github.com/microsoft/vscode-mssql/issues/20419
This PR fixes https://github.com/microsoft/vscode-mssql/issues/20426

Here's a demo of adding a new row to an empty data grid
Add Row To Empty Grid

I also made an adjustment to the grid, so that adding new rows takes users directly to where that row is, so that they don't have to look for it.
Adding row makes new row visible

## Code Changes Checklist

### DIFF
--- a/src/tableExplorer/tableExplorerWebViewController.ts
+++ b/src/tableExplorer/tableExplorerWebViewController.ts
@@ -413,6 +413,8 @@ export class TableExplorerWebViewController extends ReactWebviewPanelController<
                     );
 
                     this.updateState();
+                } else {
+                    this.logger.warn("Cannot add row: result set is undefined");
                 }
 
                 await this.regenerateScriptIfVisible(state);


src/tableExplorer/tableExplorerWebViewController.ts
```diff
@@ -401,17 +401,18 @@ export class TableExplorerWebViewController extends ReactWebviewPanelController<
...
                         rowCount: state.resultSet.rowCount + 1,
                     };
 
-                    this.updateState();
-
                     this.logger.info(
                         `Added new row to result set, now has ${state.resultSet.rowCount} rows (${state.newRows.length} new)`,
                     );
+
+                    this.updateState();
```
ssreerama, line 415: RegisterReducer already updates the state, so there’s no need to handle it explicitly unless you have a specific fix in mind. You can simply return the state.

lewis-sanchez: I ran into an issue with this while developing this controller with one of the reducers. I can't remember which one, but after I ran into that issue, I started calling it explicitly.

I'll make this change in a separate PR since it looks like that's no longer the case while tryiing things out, but in case something does happen then I want to be able to revert just that change without removing this change of adding a new row to an empty grid.

ssreerama (approved):

PR #20508, opened 2025-11-09
Publish Dialog - Deploy project to local container backend functionality

ssreerama:

Issue - https://github.com/microsoft/vscode-mssql/issues/20520

## Description

This PR:
1. Adds the back-end functionality of deploying project to a new local development container, which purely reuse the existing deployment UI helper methods.
2. The process will validate three prerequisites and then create and start a new container, builds the project and deploy the latest built dacpac to the created container.
3. Also, we display live updates to the user using vscode progress window notifications.
4. Along with the container changes, have fixed some sqlcmd table styles(better UI) and publishProject.tsx dialogMessage ele that are causing the formmessage text to flow out of the container and overlaps with publish fields.(attahed the images below)
5. Added end-to-end tests that validates the full project deployment to container flow and helper methods too.

## Code Changes Checklist

Deploy to container full flow:
8 deployToContainer

FormMessage element UI fix 
Before:
image
After:
8 FormMsgFix_SqlCmdTableFix

### DIFF
--- a/src/publishProject/publishProjectWebViewController.ts
+++ b/src/publishProject/publishProjectWebViewController.ts
@@ -372,7 +372,7 @@ export class PublishProjectWebViewController extends FormWebviewController<
         const containerName = await dockerUtils.validateContainerName("");
 
         // Parse port (already validated by form)
-        const port = parseInt(state.formState.containerPort || constants.DefaultSqlPortNumber);
+        const port = parseInt(state.formState.containerPort);
 
         return { containerName, port };
     }
@@ -623,7 +623,7 @@ export class PublishProjectWebViewController extends FormWebviewController<
                     const containerResult = await vscode.window.withProgress(
                         {
                             location: vscode.ProgressLocation.Notification,
-                            title: "Creating SQL Server container...",
+                            title: Loc.CreatingSqlServerContainer,
                             cancellable: false,
                         },
                         async () => {
--- a/localization/l10n/bundle.l10n.json
+++ b/localization/l10n/bundle.l10n.json
@@ -1622,6 +1622,7 @@
     "Microsoft SQL Server License Agreement": "Microsoft SQL Server License Agreement",
     "Select Connection": "Select Connection",
     "Checking Docker prerequisites...": "Checking Docker prerequisites...",
+    "Creating SQL Server container...": "Creating SQL Server container...",
     "Port must be a number between 1 and 65535": "Port must be a number between 1 and 65535",
     "Invalid SQL Server password for {0}. Password must be 8–128 characters long and meet the complexity requirements.  For more information see https://docs.microsoft.com/sql/relational-databases/security/password-policy": "Invalid SQL Server password for {0}. Password must be 8–128 characters long and meet the complexity requirements.  For more information see https://docs.microsoft.com/sql/relational-databases/security/password-policy",
     "{0} password doesn't match the confirmation password": "{0} password doesn't match the confirmation password",
--- a/localization/xliff/vscode-mssql.xlf
+++ b/localization/xliff/vscode-mssql.xlf
@@ -1013,6 +1013,9 @@
       <source xml:lang="en">Creating SQL Database for workspace {0}</source>
       <note>{0} is the workspace ID</note>
     </trans-unit>
+    <trans-unit id="++CODE++13329d1a1f2527d959fddce415af48d6141533dbec36dd8d8e33fae7bce865df">
+      <source xml:lang="en">Creating SQL Server container...</source>
+    </trans-unit>
     <trans-unit id="++CODE++9c37c64c878515cc2b5cd8ce05b7513d90d2134a1a2c36469aaa548d6fdfa6e9">
       <source xml:lang="en">Creating and starting your SQL Server container</source>
     </trans-unit>
--- a/src/constants/locConstants.ts
+++ b/src/constants/locConstants.ts
@@ -1366,6 +1366,7 @@ export class PublishProject {
     public static SqlServerLicenseAgreement = l10n.t("Microsoft SQL Server License Agreement");
     public static ServerConnectionPlaceholder = l10n.t("Select Connection");
     public static CheckingDockerPrerequisites = l10n.t("Checking Docker prerequisites...");
+    public static CreatingSqlServerContainer = l10n.t("Creating SQL Server container...");
     // Validation messages
     public static InvalidPortMessage = l10n.t("Port must be a number between 1 and 65535");
     public static InvalidSQLPasswordMessage(name: string) {


src/reactviews/pages/PublishProject/publishProject.tsx
```diff
@@ -51,11 +51,13 @@ function PublishProjectDialog() {
         <form className={formStyles.formRoot} onSubmit={(e) => e.preventDefault()}>
             <div className={formStyles.formDiv} style={{ overflow: "auto" }}>
                 {formMessage && (
-                    <DialogMessage
-                        message={formMessage}
-                        onMessageButtonClicked={() => {}}
-                        onCloseMessage={context.closeMessage}
-                    />
+                    <div style={{ marginBottom: "15px" }}>
```
Benjin: Can you just add `style={{ marginBottom: "15px" }}> to DialogMessage instead of wrapping it in another div?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -309,6 +323,149 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+
+        // Run prerequisite steps up to and including checkDockerEngine
+        for (let stepIndex = 0; stepIndex <= DockerStepOrder.checkDockerEngine; stepIndex++) {
+            const currentStep = dockerSteps[stepIndex];
+            const args = currentStep.argNames.map(
+                (argName) => (dummyProfile as unknown as Record<string, unknown>)[argName],
+            );
+            const result = await currentStep.stepAction(...args);
+
+            if (!result.success) {
```
laurenastrid1, line 346: are you planning to show the user some kind of error message if any of the prereq steps error?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -309,6 +323,149 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+            stepIndex <= DockerStepOrder.checkContainer;
+            stepIndex++
+        ) {
+            const currentStep = dockerSteps[stepIndex];
+            const args = currentStep.argNames.map(
+                (argName) => (dockerProfile as unknown as Record<string, unknown>)[argName],
+            );
+            const result = await currentStep.stepAction(...args);
+
+            if (!result.success) {
```
laurenastrid1, line 425: same as before- do we need to show the user an error if it fails?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -309,6 +323,149 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+     */
+    private async prepareContainerConfiguration(state: PublishDialogState): Promise<{
+        containerName: string;
+        port: number;
+    }> {
+        // Auto-generate unique container name
+        const containerName = await dockerUtils.validateContainerName("");
+
+        // Parse port (already validated by form)
+        const port = parseInt(state.formState.containerPort || constants.DefaultSqlPortNumber);
```
laurenastrid1: Is this port also checked to make sure it's not already in use?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -309,6 +323,149 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+
+            if (!result.success) {
+                return {
+                    success: false,
+                    error: result.error,
+                };
+            }
+
+            // Show success message matching deployment UI format
+            void vscode.window.showInformationMessage(`✓ ${currentStep.headerText}`);
```
Benjin, line 354: I can live with these for now (maybe others will disagree), but we should really make these updates part of the UI rather than toasts for GA.

src/publishProject/publishProjectWebViewController.ts
ssreerama: We are already showing the error when prerequisite fails, but not here when the method returns the success and error, there at line 580, we are displaying the form message.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -309,6 +323,149 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                encrypt: false,
+                trustServerCertificate: true,
+            },
+        };
+
+        try {
+            const connectionInfo = ConnectionCredentials.createConnectionInfo(connectionDetails);
+            await this._mainController.connectionManager.connect(fileUri, connectionInfo, {
+                shouldHandleErrors: false,
+            });
```
Benjin, line 469: Can you add a comment for why we don't want errors to be handled automatically?

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -401,10 +558,102 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                state.inProgress = true;
+                this.updateState(state);
 
-            return state;
+                try {
+                    // STEP 1: Run Docker prerequisite checks (Docker install, start, engine)
+                    const prereqResult = await vscode.window.withProgress(
+                        {
+                            location: vscode.ProgressLocation.Notification,
+                            title: "Checking Docker prerequisites...",
```
Benjin: Localization

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -309,6 +323,149 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                    fullErrorText: result.fullErrorText,
+                };
+            }
+
+            // Show success message matching deployment UI format
+            void vscode.window.showInformationMessage(`✓ ${currentStep.headerText}`);
+        }
+
+        // Register connection for DacFx
+        const fileUri = `mssql://publish-container-${validatedContainerName}`;
```
Benjin: Why are we using a custom file URI here?  We should be following the same pattern as the local container dialog:

1. create the container
2. save the connection to config, establish connection, and open it in Object Explorer
3. deploy dacpac to new connection

If we're doing #2 properly, then we should have a real connection URI to use for dacFx.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -401,10 +558,102 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                        state,
+                        dacpacPath,
+                        state.formState.databaseName,
+                        false,
+                    );
+
+                    state.inProgress = false;
+                    this.panel?.dispose();
+                } catch (error) {
+                    this.logger.error("Failed during container publish:", error);
```
Benjin, line 677: send telemetry

test/unit/publishProjectWebViewController.test.ts
```diff
@@ -870,4 +880,233 @@ suite("PublishProjectWebViewController Tests", () => {
...
+        expect(controller["publishToDatabase"]).to.have.been.calledOnce;
+
+        // Verify panel was closed at the end
+        expect(panelDisposeSpy).to.have.been.calledOnce;
+
+        // Verify final state
+        expect(newState.inProgress).to.be.false;
+    });
+
+    test("publishNow with LocalContainer handles error scenarios", async () => {
```
Benjin: This seems like it'll be annoying to debug if something goes wrong.  Can you break it up into a different test for each scenario, build the stubs manually for each one, then use a shared helper for doing the error validation?

src/publishProject/publishProjectWebViewController.ts
Benjin: This flow is critical/core to this "deploy to new container" feature, and IMO is incomplete without it.

src/publishProject/publishProjectWebViewController.ts
ssreerama: No, we have the port validation that validates just 1<=port<=maxPort number range. But if the port is already taken then we are directly showing them error on publish button click.

test/unit/publishProjectWebViewController.test.ts
ssreerama: sure, first I wrote individual tests and since most of the code is repeatingly used, I merged all into single one, sure, will segregate them.

src/publishProject/publishProjectWebViewController.ts
ssreerama: I'll check with the PM if we can have any better approach to show the UI for all these steps (prereqs, containter creation, building, publishing) like we show in deployment UI and can work on it.

src/reactviews/pages/PublishProject/publishProject.tsx
ssreerama: style isn't necessary but the Div element is required otherwise im getting the form message shown as below.

image

src/publishProject/publishProjectWebViewController.ts
Benjin: Eventually, the MSSQL extension will need a general way to handle showing steps, but this is probably out of scope for November since it would need to be designed.

src/publishProject/publishProjectWebViewController.ts
Benjin: I think the local container deployment UX does that; hopefully there's a helper function you can call to check.

This can be done in a follow-up.

src/publishProject/publishProjectWebViewController.ts
```diff
@@ -401,10 +580,121 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+
+                    // STEP 2: Prepare container configuration (generate name, parse port)
+                    // Form validation already ensured password, port, and EULA are valid
+                    const config = await this.prepareContainerConfiguration(state);
+
+                    // STEP 3: Create Docker container (pull, start, check, connect)
+                    const containerResult = await vscode.window.withProgress(
+                        {
+                            location: vscode.ProgressLocation.Notification,
+                            title: "Creating SQL Server container...",
```
Benjin: Localization

Benjin (approved):

Benjin (approved):

Issue #20568, opened 2025-11-13
[Bug]: DACPAC import to new Fabric SQL database fails


amthomas46:
### Description

Exported a dacpac using the new experience
Tried to import to a new database
(note theres not a strong concept of server for fabric sql, so that may be the issue)
fails with message:

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/b1b4dfb0-92de-4c9f-b616-dbefd91f6d32" />

### Steps to Reproduce

see description

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [x] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

MacOS
1.37

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

tauseefsiddique10:
This bug has been documented in all preview release documentation, blog posts and within the UI itself. At this time, it's a limitation for preview using Fabric targets with this feature. We will be improving this experience.

amthomas46:
This seems to be working for me now for a fabric sql database

<img width="771" height="765" alt="Image" src="https://github.com/user-attachments/assets/8d9ab693-6223-4115-b527-a21947205c75" />

<img width="472" height="146" alt="Image" src="https://github.com/user-attachments/assets/4ebc94fc-1f85-450c-a002-c4b27b5cc4d5" />

Issue #20567, opened 2025-11-13
[Bug]: Dacpac Dialog Bottom actions get cut off when zoomed for data tier application experience


amthomas46:
### Description

Not a big issue, very minor, but action buttons get slightly cut off.

### Steps to Reproduce

1. go into data tier application experience
2. use cmd + to increase size 
3. see scrolling works but cutoff at bottom

<img width="1512" height="946" alt="Image" src="https://github.com/user-attachments/assets/e7bd3d8c-1b58-496e-8722-16ced9f79059" />

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [x] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

1.37

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

allancascante:
Te fix for this bug was released in version 1.37

PR #20599, opened 2025-11-15
Adding fabric warning for dacpac dialog and small fix to styles

allancascante:
## Description

Dacpac deploy to new database and bacpac import to new database is not supported, adding message and UI changes to allow some partial support. Also fixed the layout to show scrollbars when the window is too small to fit the dialog.

On Deploy dacpac:
image

On import bacpac
image

This was reported as part of #20568 and #20567 

## Code Changes Checklist

### DIFF
--- a/mssql/localization/xliff/vscode-mssql.xlf
+++ b/mssql/localization/xliff/vscode-mssql.xlf
@@ -1570,6 +1570,9 @@
       <note>{0} is the error code
 {1} is the error message</note>
     </trans-unit>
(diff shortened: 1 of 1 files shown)


mssql/src/controllers/dacpacDialogWebviewController.ts
```diff
@@ -715,6 +715,19 @@ export class DacpacDialogWebviewController extends ReactWebviewPanelController<
         }
     }
 
+    /**
+     * Checks if a connection profile is connected to a Fabric server
+     */
+    private isFabricConnection(profile: IConnectionDialogProfile): boolean {
```
aasimkhan30, line 721: NIT. Maybe move this to general utils so we don't duplicate this.

PR #20644, opened 2025-11-20
Updating pipelines to cache vsix size and fix codecov

aasimkhan30:

## Description

1. Getting rid of building base branch every time we create a PR.
2. Now we run a separate workflow that just caches the size of main/release vsix sizes and reuses that cache to compare against the PR branch.
3. Removes code coverage comparison from the pipeline as we now have codecov.
4. Fixes codecov exclusions.
5. Remove web view size comparison as web views have stabilized now and we mostly care about vsix size. 

## Code Changes Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -211,7 +211,7 @@ jobs:
             echo "sqlproj_vsix_change_icon=$no_change_icon" >> $GITHUB_ENV
           fi
 
-      - name: Run unit tests
+      - name: MSSQL - Run unit tests
         continue-on-error: true
         run: |
           set +e  # Don't exit on errors
@@ -220,7 +220,7 @@ jobs:
           UNIT_EXIT_CODE=$?
           echo "unit_tests_pr_exit_code=$UNIT_EXIT_CODE" >> $GITHUB_ENV
 
-      - name: Unit Test Report
+      - name: MSSQL - Unit Test Report
         uses: dorny/test-reporter@v2
         if: success() || failure()
         with:


.github/workflows/publish-baseline.yml
```diff
@@ -0,0 +1,118 @@
...
+      - name: Upload baseline artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: baseline-artifacts
+          path: |
+            extensions/mssql/dist/views
+            extensions/mssql/*.vsix
+            extensions/mssql/coverage/cobertura-coverage.xml
+            extensions/sql-database-projects/*.vsix
+            baseline-sizes.json
```
Benjin: Can you add some comments somewhere (maybe a README in .github/workflows) about how specifically the baseline-sizes.json file works?

Looks like every time a commit is merged into main, it kicks off `publish-baseline.yml`, which saves the file... somewhere? and then the PR pipeline pulls the latest from there for comparison?

lewis-sanchez (dismissed):

.github/workflows/publish-baseline.yml
```diff
@@ -0,0 +1,118 @@
...
+      - name: SqlProj - Install dependencies and build
+        uses: ./.github/actions/build-sqlproj
+        with:
+          source-dir: "./extensions/sql-database-projects"
+
+      - name: MSSQL - Run tests for baseline coverage
+        continue-on-error: true
+        run: |
+          cd extensions/mssql
+          DISPLAY=:10 yarn test || true
```
Benjin: @copilot what does `|| true` do here?  I've seen in a few spots in this PR

.github/workflows/build-and-test.yml
```diff
@@ -262,110 +282,61 @@ jobs:
...
-          export PROFILE_NAME=test-server
-          cd target/extensions/mssql
-          export BUILT_VSIX_PATH=$(find ./ -name "*.vsix" | head -n 1)
-          DISPLAY=:10 yarn smoketest
-
       - name: Merge Smoke and Unit Test Coverage Reports
         run: |
-          if [ -f target/extensions/mssql/test/resources/mergeReports.js ]; then
-            node target/extensions/mssql/test/resources/mergeReports.js target/extensions/mssql/coverage/coverage-e2e/cobertura-coverage.xml target/extensions/mssql/coverage/cobertura-coverage.xml
+          if [ -f extensions/mssql/test/resources/mergeReports.js ]; then
```
Benjin, line 301: We should probably (eventually) move `mergeReports.js` to some shared folder in the repo root instead of within mssql.  Doesn't have to happen now, though.

.github/workflows/build-and-test.yml
```diff
@@ -68,154 +72,170 @@ jobs:
...
           elif [ "$percentage_change" -lt 0 ]; then
-            echo "vsix_change_icon=$better_change_icon" >> $GITHUB_ENV
+            echo "sqlproj_vsix_change_icon=$better_change_icon" >> $GITHUB_ENV
           else
-            echo "vsix_change_icon=$no_change_icon" >> $GITHUB_ENV
+            echo "sqlproj_vsix_change_icon=$no_change_icon" >> $GITHUB_ENV
           fi
 
-      - name: Run unit tests for PR branch
+      - name: Run unit tests
```
Benjin: While you're here, can you update the names of these extension-specific tests to say which extension they're for?  e.g.

`MSSQL - Run unit tests`
`MSSQL - Unit Test Report`

Benjin (dismissed):

.github/workflows/build-and-test.yml
aasimkhan30: I will tackle tests after this.

manujoseph85 (approved):

.github/workflows/publish-baseline.yml
aasimkhan30: It is to make sure that step never fails.

Issue #20606, opened 2025-11-17
[Bug]: Data-tier Application dialog has duplicate notifications


kburtram:
### Description

The Data-tier Application dialog has multiple prompts when it completes that basically say similar content - the operation completed successfully.  These notifications should be combined into one.

<img width="935" height="467" alt="Image" src="https://github.com/user-attachments/assets/50235f4a-ec44-4e9d-8c58-4953daf3fe62" />

### Steps to Reproduce

1. Run Export Bacpac

Expected: one notification when operation completes

Actual: two notifications

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [x] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

Latest 1.37

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

allancascante:
@kburtram after reviewing this bug, I found the problem is related to how SQL tools service reports success on dacFx operations.

**Problem: Duplicate Completion Notifications for DacFx Export/Extract Operations**
Current Behavior
When users export a BACPAC or extract a DACPAC, they see two notifications:

Generic notification (from sqlTasksService.ts):

"Export bacpac: Succeeded. Completed"
Plain information toast, no action buttons
Custom notification (from dacpacDialogWebviewController.ts):

"BACPAC exported successfully to '<filename>'"
Has "Reveal in File Explorer" button for better UX

**Potential Solutions**

Option 1 (Simplest - Client-side only):
Modify sqlTasksService.ts to check task name before showing completion notification, pull the information on target from the completed operation and show the action button in just those cases.

Option 2 (Requires backend changes):
Modify SQL Tools Service (C#) to add a flag in TaskInfo notification to suppress the notification 

Option 3 (Requires backend and Client-side changes):
Modify SQL Tools Service (C#) to include information about the file path or database resulting from the operation and use that information in the Client-side to display the action in the notification. (checked the code and the info seems to be passed back)

[...]

PR #20651, opened 2025-11-20
Changes to allow reporting success with an action on sqlTasksService

allancascante:
## Description

This PR refactors the task completion notification system to eliminate duplicate notifications for DacFx operations (Export BACPAC, Extract DACPAC, Import BACPAC, Deploy DACPAC) and provides a generic, configurable solution for custom task completion handlers, as reported in this [bug](https://github.com/microsoft/vscode-mssql/issues/20606).

Problem
Previously, DacFx operations showed two notifications:

A custom notification with action button from dacpacDialogWebviewController
A generic completion notification from sqlTasksService

Solution

Centralized all task completion notification logic in sqlTasksService with a configurable handler registration system.

## Code Changes Checklist


extensions/mssql/src/controllers/mainController.ts
```diff
@@ -628,11 +629,76 @@ export default class MainController implements vscode.Disposable {
...
 
             this.registerLanguageModelTools();
+            this.registerTaskCompletionHandlers();
 
             return true;
         }
     }
 
+    /**
+     * Helper method to register task completion handlers for DacPac operations
```
kburtram: nit: the P isn't usually capitalized in `DacPac` - `dacpac` or `DACPAC` is what I normally see

extensions/mssql/src/controllers/mainController.ts
```diff
@@ -628,11 +629,76 @@ export default class MainController implements vscode.Disposable {
...
+     */
+    private registerTaskCompletionHandlers(): void {
```
kburtram: If this method is only used for DACPAC-related completion handlers, could we rename it to something more specific like registerDacpacTaskCompletionHandlers?

Also, I’m wondering if we should start moving some of this logic out of mainController into feature-specific modules (e.g. a DACPAC-specific file) and just import/register them here. This file is starting to get quite large, and splitting by feature area might make it easier to maintain.

extensions/mssql/src/services/sqlTasksService.ts
```diff
@@ -156,21 +221,51 @@ export class SqlTasksService {
...
+                                void vscode.commands.executeCommand(command, ...args);
+                            }
+                        });
+                } else {
+                    // Show notification without action button
+                    void vscode.window.showInformationMessage(successMessage);
+                }
+            } else {
+                // Show generic completion message for tasks without custom handlers
+                const lastMessage =
```
kburtram, line 274: In the case that taskProgressInfo.message is truthy AND  taskProgressInfo.message.toLowerCase() !== taskStatusString.toLowerCase() is true, this statement evaluates to the boolean `true` instead of the actual message content. This means lastMessage would be `true` rather than the message string, causing "true" to display in the UI. 

Should this use a ternary operator instead?

extensions/mssql/src/controllers/mainController.ts
Benjin: I'm thinking this registration should be done by the dacpacDialogWebviewController when it launches the first time.  It wouldn't need to unregister upon webview disposal because that would revert to the less helpful task completion.

Benjin: Or maybe the dacfxService constructor would be even better.

extensions/mssql/src/services/sqlTasksService.ts
```diff
@@ -65,12 +66,63 @@ type ActiveTaskInfo = {
 };
 type ProgressCallback = (value: { message?: string; increment?: number }) => void;
 
+/**
+ * Configuration for a custom task completion handler that shows a notification with an action button
+ */
+export interface TaskCompletionHandler {
+    /**
+     * The name of the task to handle (must match taskInfo.name from SQL Tools Service)
```
Benjin: are these taskNames localized at all when coming to SQL Tools Service?  I don't love matching on a human-friendly string (rather than a task type ID that's less likely to change and won't get localized).

extensions/mssql/src/services/sqlTasksService.ts
```diff
@@ -65,12 +66,63 @@ type ActiveTaskInfo = {
...
+    getActionCommand?: () => string;
+
+    /**
+     * Gets the command arguments to pass when executing the action
+     * Optional - required only if getActionButtonText is provided
+     * @param taskInfo The task information
+     * @param targetLocation The resolved target location
+     * @returns The command arguments
+     */
+    getActionCommandArgs?: (taskInfo: TaskInfo, targetLocation: string) => unknown[];
```
Benjin: feels like the sorta thing that should be an array of `{buttonText, actionCommand, args}[]`

extensions/mssql/src/services/sqlTasksService.ts
```diff
@@ -156,21 +221,51 @@ export class SqlTasksService {
...
-                      taskInfo.taskInfo.name,
-                      taskStatusString,
-                      lastMessage.toString(),
-                  )
-                : localizedConstants.taskStatusWithName(taskInfo.taskInfo.name, taskStatusString);
-            showCompletionMessage(taskProgressInfo.status, taskMessage);
+            const targetLocation = handler
+                ? handler.getTargetLocation(taskInfo.taskInfo)
+                : undefined;
+            if (taskProgressInfo.status === TaskStatus.Succeeded && handler && targetLocation) {
```
Benjin, line 249: Is there a failure state that we'd also want to handle with custom?

extensions/mssql/test/unit/sqlTasksService.test.ts
```diff
@@ -0,0 +1,513 @@
...
+            executeCommandStub = vscode.commands.executeCommand as sinon.SinonStub;
+            executeCommandStub.reset();
+        }
+
+        if (!(vscode.window.withProgress as unknown as sinon.SinonStub).restore) {
+            withProgressStub = sandbox.stub(vscode.window, "withProgress");
+        } else {
+            withProgressStub = vscode.window.withProgress as sinon.SinonStub;
+            withProgressStub.reset();
+        }
```
Benjin: This can be significantly cleaned up.  The teardown() method should handle all those restore calls, and you can probably use vscodeWrapper (with the prebuild stubs from `stubVscodeWrapper` for most of these.

extensions/mssql/src/services/sqlTasksService.ts
```diff
@@ -65,12 +66,63 @@ type ActiveTaskInfo = {
 };
 type ProgressCallback = (value: { message?: string; increment?: number }) => void;
 
+/**
+ * Configuration for a custom task completion handler that shows a notification with an action button
+ */
+export interface TaskCompletionHandler {
```
Benjin, line 82: Overall, I like the idea of what you're doing here, but I think we can/should think more about the API to find the right balance between flexibility, power, and simplicity here.

extensions/mssql/src/services/sqlTasksService.ts
allancascante: @kburtram This is the original code, just moved to the else just to function as before in cases where there is no extra handler: [            // Get the message to display, if the last status doesn't have a valid message then get the last valid one
            const lastMessage =
                (taskProgressInfo.message &&
                    taskProgressInfo.message.toLowerCase() !== taskStatusString.toLowerCase()) ??
                taskInfo.lastMessage;
            // Only include the message if it isn't the same as the task status string we already have - some (but not all) task status
            // notifications include this string as the message
            const taskMessage = lastMessage
                ? localizedConstants.taskStatusWithNameAndMessage(
                      taskInfo.taskInfo.name,
                      taskStatusString,
                      lastMessage.toString(),
                  )
                : localizedConstants.taskStatusWithName(taskInfo.taskInfo.name, taskStatusString);
            showCompletionMessage(taskProgressInfo.status, taskMessage);](https://github.com/microsoft/vscode-mssql/blob/4a78f14f50fdaff36476f502e3b4f23cc1903b4c/extensions/mssql/src/services/sqlTasksService.ts#L159) I don't know for sure why the author did that back in the original code, and I didn't want to alter that behavior if it was already working fine and tested

Benjin: Looks like that might've been a bug, based on the comment in the code you linked:
```
// Get the message to display, if the last status doesn't have a valid message then get the last valid one
```

Think it should be:

```
const lastMessage =(taskProgressInfo.message.toLowerCase() !== taskStatus.toLowerCase()) ? taskProgressInfo.message : taskInfo.lastMessage;
```

Good example of why code comments are important :)

extensions/mssql/src/services/sqlTasksService.ts
allancascante: currently there is no need for this support, it will be nice to have but not required.

extensions/mssql/src/services/sqlTasksService.ts
allancascante: currently this is not needed, sounds like a nice extension for the future when the time comes

extensions/mssql/src/services/sqlTasksService.ts
allancascante: this are just the minimal things needed for the fix of the bug, eventually if there is such need we can come back and add more support

extensions/mssql/src/constants/constants.ts
```diff
@@ -336,3 +336,15 @@ export const build = "build";
 export const sqlProjBuildTaskType = "sqlproj-build";
 export const msBuildProblemMatcher = "$msCompile";
 export const buildDirectory = "BuildDirectory";
+
+// DacFx task names (as reported by SQL Tools Service)
+export const taskNameExportBacpac = "Export bacpac";
+export const taskNameExtractDacpac = "Extract dacpac";
+export const taskNameImportBacpac = "Import bacpac";
+export const taskNameDeployDacpac = "Deploy dacpac";
```
Benjin: Are these still needed since you addedthe operation IDs below to match more confidently on?

extensions/mssql/src/services/dacFxService.ts
```diff
@@ -7,9 +7,84 @@ import SqlToolsServiceClient from "../languageservice/serviceclient";
...
+        const platformInfo = new PlatformInformation(process.platform, process.arch, undefined);
+
+        // Determine the OS-specific reveal button text
+        let revealButtonText: string;
+        if (platformInfo.isMacOS) {
+            revealButtonText = LocalizedConstants.DacpacDialog.RevealInFinder;
+        } else if (platformInfo.isLinux) {
+            revealButtonText = LocalizedConstants.DacpacDialog.OpenContainingFolder;
+        } else {
+            // Windows or any other platform
```
Benjin: Exceedingly minor nit: since "Explorer" and "Finder" are brand-names specific to those platform, the "any other platform" should be Linux's case.  "Open containing folder" is completely brand/platform-agnostic.

extensions/mssql/src/services/sqlTasksService.ts
Benjin: We should either define this type now, or remove the call for until we need it.  Adding an `unknown[]` or `any[]` now invites a mess to need cleaning up later.

extensions/mssql/src/services/sqlTasksService.ts
```diff
@@ -65,16 +68,68 @@ type ActiveTaskInfo = {
...
+     * @returns The localized button text
+     */
+    getActionButtonText?: () => string;
+
+    /**
+     * Gets the VS Code command to execute when the action button is clicked
+     * Optional - required only if getActionButtonText is provided
+     * @returns The VS Code command ID
+     */
+    getActionCommand?: () => string;
```
Benjin: Is there a benefit to these being functions/callbacks instead of just regular strings if there's no argument getting passed in?  If there's an arg (like the others), then the function could at least react to some sort of input, but that's not happening here.

I think we talked about this, but I can't remember what we ended up deciding...

extensions/mssql/src/services/sqlTasksService.ts
```diff
@@ -84,6 +139,16 @@ export class SqlTasksService {
...
     }
 
+    /**
+     * Registers a custom completion handler for a specific task type.
+     * When a task with the specified operation ID completes successfully, the handler will be invoked
+     * to show a custom notification with an action button.
+     * @param handler The task completion handler configuration
+     */
+    public registerCompletionHandler(handler: TaskCompletionHandler): void {
+        this._completionHandlers.set(handler.operationName, handler);
```
Benjin: If a completion handler is set twice, the first one would just get overwritten (with no logging or error).

What should happen if there's already a completion handler set for this operation?  I think it's okay to only support one (for now?), but then we need to decide whether the second registration should fail or overwrite.  If overwrite, it needs to log (and/or emit telemetry) that this is happening because that could be a bug.

extensions/mssql/src/services/sqlTasksService.ts
```diff
@@ -156,21 +226,51 @@ export class SqlTasksService {
...
+                          taskInfo.taskInfo.name,
+                          taskStatusString,
+                          lastMessage.toString(),
+                      )
+                    : localizedConstants.taskStatusWithName(
+                          taskInfo.taskInfo.name,
+                          taskStatusString,
+                      );
+                this.showCompletionMessage(taskProgressInfo.status, taskMessage);
+            }
```
Benjin, line 290: nit: leave extra line between logic blocks for readability

extensions/mssql/src/services/sqlTasksService.ts
Benjin: The registration call needs a docstring that tells devs that it's only for successful task completions.

Maybe the registration event should also be renamed to `registerCompletionSuccessHandler()`?

Benjin (approved):

PR #20655, opened 2025-11-21
GitHub Copilot: Add display names with connection IDs to Copilot tool messages

caohai:
## Description
1. Improves Copilot tool UX by displaying human-readable connection names alongside connection IDs in all tool(that require a connection ID) confirmation and invocation messages.
2. Add tests covering related tools.

image

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/test/unit/changeDatabaseTool.test.ts
+++ b/extensions/mssql/test/unit/changeDatabaseTool.test.ts
@@ -3,8 +3,6 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-/* eslint-disable @typescript-eslint/no-explicit-any */
-
 import * as vscode from "vscode";
 import { expect } from "chai";
 import * as sinon from "sinon";
@@ -56,9 +54,7 @@ suite("ChangeDatabaseTool Tests", () => {
         mockToken = {} as vscode.CancellationToken;
 
         // Create the tool instance
-        changeDatabaseTool = new ChangeDatabaseTool(
-            mockConnectionManager as unknown as ConnectionManager,
-        );
+        changeDatabaseTool = new ChangeDatabaseTool(mockConnectionManager);
     });
 
     teardown(() => {
@@ -83,11 +79,10 @@ suite("ChangeDatabaseTool Tests", () => {
 
         test("should return error when connection is not found", async () => {
             const noConnectionMock = sandbox.createStubInstance(ConnectionManager);
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
             noConnectionMock.getConnectionInfo.returns(undefined as any);
 
-            const toolWithNoConnection = new ChangeDatabaseTool(
-                noConnectionMock as unknown as ConnectionManager,
-            );
+            const toolWithNoConnection = new ChangeDatabaseTool(noConnectionMock);
 
             const options = {
                 input: {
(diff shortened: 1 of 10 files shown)


extensions/mssql/src/copilot/tools/toolsUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export function getDisplayNameForTool(connInfo: ConnectionInfo | undefined): string {
+    if (connInfo) {
+        // We have connection info - return the formatted display name
+        return getConnectionDisplayName(connInfo.credentials);
+    } else {
+        // Connection info not available - use a placeholder that makes it clear
+        // This will show as "Unknown Connection (ID: connection-123)" which is better than
+        // "connection-123 (ID: connection-123)"
+        return loc.unknownConnection;
```
Benjin, line 25: What would result in this being unknown?  Is there value in informing the user that this is "Unknown"?

extensions/mssql/test/unit/changeDatabaseTool.test.ts
```diff
@@ -0,0 +1,201 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Benjin: Why are we disabling this lint rule for the entire file?  Because disabling the linter rule introduces a bit of risk, I think we should do this line-by-line (or small, scoped block by small, scoped block) so that it's an intentional decision each time.

extensions/mssql/test/unit/changeDatabaseTool.test.ts
```diff
@@ -0,0 +1,201 @@
...
+        mockConnectionManager.isConnected.returns(true);
+        mockConnectionManager.disconnect.resolves();
+        mockConnectionManager.connect.resolves(true);
+
+        // Mock CancellationToken
+        mockToken = {} as vscode.CancellationToken;
+
+        // Create the tool instance
+        changeDatabaseTool = new ChangeDatabaseTool(
+            mockConnectionManager as unknown as ConnectionManager,
```
Benjin: this recast shouldn't be necessary; `mockConnectionManager` already should be a ConnectionManager-like to the type-checker.

I've noticed that AI _really_ likes to do this for some reason, even with explicit insturctions that it shouldn't.  Can you do a scan and clean it up?

extensions/mssql/src/constants/locConstants.ts
```diff
@@ -1741,33 +1741,39 @@ export class MssqlChatAgent {
...
-    public static disconnectToolConfirmationMessage = (connectionId: string) => {
+    public static disconnectToolConfirmationMessage = (
+        displayName: string,
+        connectionId: string,
+    ) => {
         return l10n.t({
-            message: "Disconnect from connection '{0}'?",
-            args: [connectionId],
-            comment: ["{0} is the connection ID"],
+            message: "Disconnect from connection '{0}' (ID: {1})?",
```
Benjin, line 1749: Out of curiosity, what's the value of including the connection ID (especially when it's a session ID, not the ID that shows up in connection config files) in a user-facing message?  Is that useful for debugging/logging?

Benjin (dismissed):

extensions/mssql/test/unit/changeDatabaseTool.test.ts
caohai: fixed

extensions/mssql/test/unit/changeDatabaseTool.test.ts
caohai: did a scan and removed unnessary ones.

extensions/mssql/src/copilot/tools/toolsUtils.ts
caohai: If for whatever reason the underlying connection associated with the connection ID is cleaned up, copilot would not know that (unless we do a connection state check before every tool call) and the invocation message would show "Unknown Connection".
One example: Tool call with stale chat history (note there's a dotted line indicating a difference chat session.)
image

extensions/mssql/src/constants/locConstants.ts
caohai: I kept this info mainly for debugging purposes, so a user can see what connectionID they get from the connect tool call and verify the same ID is being used by following tool calls.

Issue #20601, opened 2025-11-17
[Bug]: Column shown as NULL in a non-nullable column


yoleichen:
### Description

https://github.com/user-attachments/assets/735a9b18-a60b-4407-a173-c5fe5b816c84

If I clear a cell in a non-nullable columns, the cell is shown as valid (yellow highlight) but the text label says NULL.

**Expected behavior:**
Same as ADS, shown as empty but if user type NULL, show as invalid

### Steps to Reproduce

1. clear a cell in a non-nullable columns
2. observe it says NULL but still valid

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### If you selected "Other", please describe the affected area

Edit Data

### Environment Information

- VS Code version: Latest:
- MSSQL extension version: Latest
- OS: Windows

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #20684, opened 2025-11-24
Non-nullable columns don't show NULL when they're empty.

lewis-sanchez:

## Description

This PR fixes https://github.com/microsoft/vscode-mssql/issues/20601

Here is a demo of this working. In the table shown in the screenshot, the "Middle Name' column is defined as nullable while the last name column is not. When a cell is cleared in a column that is nullable "NULL" appears, and when a cell is cleared in a column that is not nullable, then it is empty ("").
image

This PR bumps the STS version and the changes between the previous version and the updated version are these:
image

## Code Changes Checklist


lewis-sanchez:
Need to bump STS version once the STS build completes.

PR #20687, opened 2025-11-25
Fixing query disposal, connection cleanup and telemetry apis. 

aasimkhan30:

## Description
1. Adds call stack option to telemetry to send a call stack. The call stack option gets rid of unnecessary file path info 
and make it readable.
    ```
    QueryResultWebviewController.onNotification < registerCommonRequestHandlers <      QueryResultWebviewController.registerRpcHandlers < QueryResultWebviewController.initialize < new QueryResultWebviewController < new SqlOutputContentProvider < MainController.initialize < async MainController.activate 
    ```
1. Fixes telemetry api to send correlation id for all activity lifecycle events.
2. Wrap webview controller requests and reducers in activity telemetry to get perf issues.
4. Adds activity telemetry with timeouts to these operations:
   - Connection start (with 5s timeout for "longRunningInitialization" marker)
   - Connection cancel
   - Query start (statement and selection/document)
   - Query cancel  
   - Get rows (with bucketized row counts)
1. Removed noisy webview telemetry and logs.
2. Improves connection error handling with proper cleanup of connection state on failures.
6. Changes query document close behavior cancels query but keeps results when panel is open.

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/telemetry/telemetry.ts
+++ b/extensions/mssql/src/telemetry/telemetry.ts
@@ -46,17 +46,14 @@ const SKIP_FUNCTIONS = new Set([
 ]);
 
 /**
- * Captures a call stack with minimal overhead.
- * Uses V8's captureStackTrace for fastest possible capture.
- * Filters out telemetry internal functions.
+ * Filters a stack trace string to remove internal telemetry functions
+ * and limit the number of frames.
+ * @param stack The stack trace string to filter
  */
-function captureCallStack(): string {
-    const err = { stack: "" };
-    Error.captureStackTrace(err, captureCallStack);
-
+export function filterStack(stack: string): string {
     const frames: string[] = [];
-    for (const line of err.stack.split("\n")) {
-        if (frames.length >= 8) break;
+    for (const line of stack.split("\n")) {
+        if (frames.length >= 20) break;
 
         const match = line.match(/at (\S+)/);
         if (!match) continue;
@@ -65,12 +62,28 @@ function captureCallStack(): string {
 
         // Extract the last part of the name for filtering (e.g., "Foo.bar" -> "bar")
         const funcName = name.split(".").pop() || name;
-        if (SKIP_FUNCTIONS.has(funcName)) continue;
+        if (SKIP_FUNCTIONS.has(funcName)) {
+            // Only skip if the function is global
+            // This prevents skipping user methods that happen to share names with internal functions (e.g. 'update')
+            if (name === funcName) {
+                continue;
+            }
+        }
 
         frames.push(name);
     }
 
     return frames.join(" < ");
+}
+
+/**
+ * Captures a call stack and filters out internal telemetry functions
+ * and user file paths.
+ */
+export function captureCallStack(): string {
+    const err = { stack: "" };
+    Error.captureStackTrace(err, captureCallStack);
+    return filterStack(err.stack || "");
 }
 
 /**
--- a/extensions/mssql/test/unit/telemetry.test.ts
+++ b/extensions/mssql/test/unit/telemetry.test.ts
@@ -1,0 +1,93 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { expect } from "chai";
+import * as Telemetry from "../../src/telemetry/telemetry";
+
+suite("Telemetry Tests", () => {
+    test("captureCallStack should return a string", () => {
+        const stack = Telemetry.captureCallStack();
+        expect(stack).to.be.a("string");
+    });
+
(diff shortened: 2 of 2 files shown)


extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -1187,31 +1207,36 @@ export default class ConnectionManager {
...
+                    });
+                }
+            }, Constants.stsImmediateActivityTimeout);
+            initResponse = await this.client.sendRequest(
                 ConnectionContracts.ConnectionRequest.type,
                 connectParams,
             );
+            isIntialized = true;
         } catch (error) {
+            isIntialized = true;
```
Benjin: Why would this be set to true if the connection initialization threw an error?

extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -1258,15 +1276,28 @@ export default class ConnectionManager {
              */
 
             await this.handleConnectionSuccess(fileUri, connectionInfo, result);
+            connectionActivity.end(
```
Benjin, line 1279: Refresh my memory: `activity.end()` includes all the same properties that it was started with, right?  Need to make sure that we don't lose the server/cloud type info.

extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -1258,15 +1276,28 @@ export default class ConnectionManager {
...
                     result,
                     connectionInfo.credentials,
                 );
 
+                errorType = errorHandlingResult?.errorHandled;
+                connectionActivity.update({
+                    retryConnection: errorHandlingResult?.isHandled ? "true" : "false",
+                });
                 if (errorHandlingResult.isHandled) {
+                    connectionActivity.end(ActivityStatus.Pending);
```
Benjin: Why does this end as "pending"?

extensions/mssql/src/sharedInterfaces/telemetry.ts
```diff
@@ -203,6 +211,8 @@ export type FinishActivity = (
     activityStatus: Exclude<ActivityStatus, ActivityStatus.Failed>,
     additionalProperties?: Record<string, string>,
     additionalMeasurements?: Record<string, number>,
+    connectionProfile?: any, //TODO fix any with IConnectionProfile
```
Benjin, line 217: Planning to do these TODOs?

extensions/mssql/src/telemetry/telemetry.ts
```diff
@@ -34,14 +34,80 @@ const telemetryReporter = new AdsTelemetryReporter<
...
+
+        // Find opening paren (if exists) to know where function name ends
+        let nameEnd = raw.indexOf(" (", atPos + 3);
+        if (nameEnd === -1 || nameEnd > lineEnd) {
+            nameEnd = lineEnd;
+        }
+
+        const name = raw.slice(atPos + 3, nameEnd).trim();
+
+        // Skip empty, anonymous, Object. prefixed, and telemetry internal functions
```
Benjin: I don't have a better solution for it, but JS's proclivity for anonymous functions is probably going to result in a bunch of lost context here.

extensions/mssql/src/telemetry/telemetry.ts
```diff
@@ -34,14 +34,80 @@ const telemetryReporter = new AdsTelemetryReporter<
...
+    "end",
+    "endFailed",
+]);
+
+/**
+ * Captures a call stack with minimal overhead.
+ * Uses V8's captureStackTrace for fastest possible capture.
+ * Filters out telemetry internal functions.
+ */
+function captureCallStack(): string {
```
Benjin: Would love to understand this logic a bit better, maybe a unit test would be a good way to do that?  Can you extract everything after `Error.captureStackTrace()` into a helper function for deterministic test execution and give (just two or three) test examples of what this should look like?

extensions/mssql/src/controllers/connectionManager.ts
Benjin: Looks like it does; nm

extensions/mssql/src/controllers/connectionManager.ts
aasimkhan30: Let me fix the name

extensions/mssql/src/sharedInterfaces/telemetry.ts
aasimkhan30: That will require a big change of moving a bunch of interfaces into shared. Will do this as a follow up task.

extensions/mssql/src/controllers/connectionManager.ts
aasimkhan30: Switched to "retrying"

extensions/mssql/src/sharedInterfaces/telemetry.ts
Benjin: Swap with `IConnectionDialogProfile`

Benjin (dismissed):

PR #20714, opened 2025-12-02
Add Column Reordering, Column Hiding, and Filtering to Edit Data Grid

lewis-sanchez:

## Description

This PR adds support for column sorting, column reordering, column hiding, and filtering to the Edit Data grid.

### Column Reordering
Column headers can be dragged to reorder columns
Column Reordering

### Column Hiding
Columns can be hidden by opening the grid menu and selecting which columns should be hidden.
Column Hiding

### Column filtering:
Filter mechanism is added to the top of each column to filter data rfecords
Grid Filtering

The filter row that appears at the top of the grid can also be hidden if the user chooses to hide them.

This is how the grid appears when the filters row is visible:
image

This is how the grid appears when the filters row is hidden:
image

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/reactviews/pages/TableExplorer/FluentCompoundFilterComponent.tsx
+++ b/extensions/mssql/src/reactviews/pages/TableExplorer/FluentCompoundFilterComponent.tsx
@@ -4,7 +4,14 @@
  *--------------------------------------------------------------------------------------------*/
 
 import React, { useState, useEffect, useRef, useCallback } from "react";
-import { Dropdown, Option, Input, makeStyles, DropdownProps, InputProps } from "@fluentui/react-components";
+import {
+    Dropdown,
+    Option,
+    Input,
+    makeStyles,
+    DropdownProps,
+    InputProps,
+} from "@fluentui/react-components";
 
 export interface FilterOperator {
     operator: string;


extensions/mssql/src/reactviews/pages/TableExplorer/TableDataGrid.tsx
```diff
@@ -132,12 +137,33 @@ export const TableDataGrid = forwardRef<TableDataGridRef, TableDataGridProps>(
                     void reactGridRef.current.paginationService.goToPageNumber(1);
                 }
             },
+            getSelectedRowIds: () => {
+                if (reactGridRef.current?.slickGrid) {
+                    const selectedRows = reactGridRef.current.slickGrid.getSelectedRows();
```
ghiscoding: you could use the Grid State Service for this, it would be simpler to use and that kind of logic should be handled already. 

See [`gridService.getCurrentRowSelections()`](https://github.com/ghiscoding/slickgrid-universal/blob/625e688be1bcd823612d67b1ed0190d312bfc7b2/packages/common/src/services/gridState.service.ts#L301-L314)

or use [`dataView.dataView.getAllSelectedIds()`](https://github.com/ghiscoding/slickgrid-universal/blob/625e688be1bcd823612d67b1ed0190d312bfc7b2/packages/common/src/core/slickDataview.ts#L1533-L1539)

What I'm trying to say is that the code logic you added here should already be available in some of the services  or functions that I added over the years. I was tired of copying the same logic over and over and I created a bunch of Services to remove these duplications like the `GridService` and the `GridStateService`. If docs are missing, let me know and I can add some more (I actually haven't looked at the docs, I went straight to the service)

extensions/mssql/src/reactviews/pages/TableExplorer/TableDataGrid.tsx
```diff
@@ -329,22 +364,122 @@ export const TableDataGrid = forwardRef<TableDataGridRef, TableDataGridProps>(
                             minHeight: 250, // Minimum height to prevent unnecessary scrollbar
                         },
                         forceFitColumns: false, // Allow horizontal scrolling for many columns
-                        enableColumnReorder: false,
-                        enableHeaderMenu: false,
+
+                        // Localization for grid UI
+                        locales: {
+                            TEXT_CONTAINS: loc.tableExplorer.filterContains,
```
ghiscoding: I know you only have 1 Slickgrid-React grid so far, but if you're ever considering adding multiple grids, I would suggest to move these common options (like `locales` and others) into a separate common grid option file. So far it's only 1 grid, but who knows if you're thinking of adding/replacing more grids 😉

extensions/mssql/src/reactviews/pages/TableExplorer/TableDataGrid.tsx
lewis-sanchez: Thank you, @ghiscoding! I'll make these changes.

extensions/mssql/src/reactviews/pages/TableExplorer/TableDataGrid.tsx
```diff
@@ -582,6 +732,20 @@ export const TableDataGrid = forwardRef<TableDataGridRef, TableDataGridProps>(
             }
         }
 
+        // Handle row selection changes
+        function handleSelectedRowsChanged(_e: any, args: any) {
+            if (onSelectedRowsChanged && reactGridRef.current?.dataView) {
+                const selectedRowIndices = args.rows || [];
```
ghiscoding: You could maybe use [`dataView.dataView.getAllSelectedIds()`](https://github.com/ghiscoding/slickgrid-universal/blob/625e688be1bcd823612d67b1ed0190d312bfc7b2/packages/common/src/core/slickDataview.ts#L1533-L1539) and/or [`dataView.dataView.getAllSelectedItems()`](https://github.com/ghiscoding/slickgrid-universal/blob/625e688be1bcd823612d67b1ed0190d312bfc7b2/packages/common/src/core/slickDataview.ts#L1597-L1608) the latter being the data context (object) items. You should give it a try first though because as the function comment says...

> Note: when using Pagination it will also include hidden selections assuming `preserveHiddenOnSelectionChange` is set to true.

and this is the default:

```ts
globalGridOptions = {
  dataView: {
    // when enabled, this will preserve the row selection even after filtering/sorting/grouping
    syncGridSelection: {
      preserveHidden: false,
      preserveHiddenOnSelectionChange: true,
    },
    syncGridSelectionWithBackendService: false, // but disable it when using backend services
  },
}
```
so you might need to toggle these flags depending if you want to have the functions return the selections on the other pages or not. So anyway, all your code logic is roughly a duplication of something that already exists so...

Basically this is what allows me to remember what the Row Selection is/was on all other pages as demoed in 
[Example 10](https://[ghiscoding.github.io/slickgrid-react-demos/#/example10).

extensions/mssql/src/reactviews/pages/TableExplorer/TableDataGrid.tsx
ghiscoding: @lewis-sanchez out of topic, but also don't forget to also remove the unnecessary `editCommandHandler` that we discussed in my previous PR 😉

lewis-sanchez: The `editCommandHandler`  is gone too. 😄

extensions/mssql/src/reactviews/pages/TableExplorer/TableDataGrid.tsx
lewis-sanchez: Thanks, @ghiscoding! I appreciate you taking the time to explain this.

ghiscoding:
@lewis-sanchez since you're now using the Grid Menu again (hamburger menu), you might want to re-enable the `alwaysShowVerticalScroll` grid option, that you disabled in PR #20650 

One of the reason to keep the scroll is to avoid the grid size change when the scroll is showing/hiding, in other words if the scroll is always shown then the grid won't flicker and it looks like the data is being pushed to the left by the size of the scrollbar (we can see a small flickering because of this in your video above on "Column Filtering"). It's really up to you, but that's my personal preference to always keep it. Cheers

lewis-sanchez:
> @lewis-sanchez since you're now using the Grid Menu again (hamburger menu), you might want to re-enable the `alwaysShowVerticalScroll` grid option, that you disabled in PR #20650
> 
> One of the reason to keep the scroll is to avoid the grid size change when the scroll is showing/hiding, in other words if the scroll is always shown then the grid won't flicker and it looks like the data is being pushed to the left by the size of the scrollbar (we can see a small flickering because of this in your video above on "Column Filtering"). It's really up to you, but that's my personal preference to always keep it. Cheers

Good catch, @ghiscoding! I will re-enable the option.

extensions/mssql/src/reactviews/pages/TableExplorer/TableDataGrid.css
```diff
@@ -235,3 +235,188 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    padding: 6px 12px;
+    background-color: var(--vscode-editor-background);
+    border-bottom: 1px solid var(--vscode-panel-border);
+    gap: 8px;
+}
+
+.search-input-container {
```
aasimkhan30, line 256: Can we scope these css rules to grid? I worry that generic names like these might accidentally change styling in other places

aasimkhan30 (dismissed):

extensions/mssql/src/reactviews/common/locConstants.ts
```diff
@@ -1312,6 +1312,60 @@ export class LocConstants {
             noDataAvailable: l10n.t("No data available"),
             noPendingChanges: l10n.t("No pending changes. Make edits to generate a script."),
             closeScriptPane: l10n.t("Close Script Pane"),
+            filterContains: l10n.t("Contains"),
```
Benjin, line 1320: Lots of new strings here!  If there are all for slickgrid's localization, let's move them to their own loc grouping (maybe "slickGrid" instead of "tableExplorer" since they're specific to its implementation)

extensions/mssql/src/reactviews/common/locConstants.ts
```diff
@@ -1312,6 +1312,60 @@ export class LocConstants {
...
+            filterNotContains: l10n.t("Not contains"),
+            filterEquals: l10n.t("Equals"),
+            filterNotEqualTo: l10n.t("Not equal to"),
+            filterStartsWith: l10n.t("Starts With"),
```
Benjin: I think all of these strings should be label-cased:

* "Starts With" -> "Starts with"
* "Clear all Filters" -> "Clear all filters"

etc.

Benjin (dismissed):

croblesm:
Hi @lewis-sanchez, 

I noticed a small thing that only occurs after applying a filter. The grid has a solid line around the border

image

This solid line disappears, you can see that the last row in the grid has no border

image

ghiscoding:
> Hi @lewis-sanchez,
> 
> I noticed a small thing that only occurs after applying a filter. The grid has a solid line around the border

the reason is because your 2nd print screen has less data than the minimum grid height and it looks like there is a border on the entire grid. I don't fully remember but I think that the cell border is only on the top row, so if you combine all of that, it is probably why it might be missing the border on the bottom row in your 2nd print screen. I never add both top/bottom borders because that becomes ugly since they append on top of each other and it shows as 2px instead of 1px

lewis-sanchez:
@croblesm, I added the missing bottom border that you called out.
image

Benjin (approved):

ghiscoding:
@lewis-sanchez is the word "Contains" in each column filter using my select dropdown for filter Operators? If so, it wasn't exactly in your animated gif, I mean in my version I had a 1px border to tell the user to provide a bit more highlight to that dropdown... So if it is the same select dropdown, then you might need to provide a bit more highlight to that fact? I'm not sure how you do it in Fluent but it's something to maybe consider?

image

lewis-sanchez:
Hi @ghiscoding, I need to update the screenshots in the PR description, but the filter controls now look like this:
image

image

Is the blue highlighting the highlighting you were referring to?

ghiscoding:
> Is the blue highlighting the highlighting you were referring to?

no in mine, I really meant just a 1px border around the select since my original select is much thinner (I mean all filters with a compound operator select and I've added a print screen above) it didn't require a focus or anything, it was just a quick way of visualizing that it was different... but regardless, in your new print screen it now shows the caret representing a select dropdown, so it's much better and addresses the issue I was trying to emphasizes, so good job 👍🏻 At least now it shows that it's clickable as opposed to your original print screen (below) which wasn't clear that the text "Contains" was clickable


image

Issue #19238, opened 2025-04-23
[Bug:] Unable to connect to a database using ActiveDirectoryServicePrincipal


emil-eklund:
- MSSQL Extension Version: 1.30.0
- VSCode Version: 1.99.3
- OS Version: Microsoft Windows 11 Business 10.0.26100 Build 26100

Steps to Reproduce:

1. Click on `Add Connection (preview)`
2. Select input type `Connection String`
3. Enter a service principal-based connection string
4. Press `Connect`

### Error shown in messagebar:
Error: Cannot open server "`<guid>`" requested by the login. The login failed.

### Error in toaster:
mssql: Error 40532: Cannot open server "`<guid>`" requested by the login.  The login failed.

### Error in output:
Error connecting to server "`<database-name>`.database.fabric.microsoft.com". Details: Cannot open server "`<guid>`" requested by the login.  The login failed.

![Image](https://github.com/user-attachments/assets/6eed5196-7ac2-4982-aeaa-272d6ea24f75)

The exact same connection string works perfectly using a normal SqlConnection through .NET.
Interestingly, the server-name is not a guid, but does contain one. It is not the same one as given in the error message. The guid shown in the error message is the **Tenant ID** of the service principal.

Benjin:
We currently only support SQL auth, Integrated auth, and Universal Entra auth.  We'll keep this issue open for users to vote on so that we can gauge feedback regarding additional auth types.

emil-eklund:
Hi @Benjin , would it be possible to display an error message at least? Some users (me) will be confused by the option "Connection String" if they are not allowed to connection strings. Having wasted several hours debugging my credentials, a simple "Not supported" message would have helped immensely.

kburtram:
@Benjin I think improving the error message in this scenario makes sense, if possible.

PR #20722, opened 2025-12-03
Showing error when trying to parse a connection string with not valid auth

allancascante:
## Description

Fixes https://github.com/microsoft/vscode-mssql/issues/19238

Extension only supports Integrated, SqlLogin or AzureMFA so when parsing connection strings it should show an error if it uses any different that those.

Includes an STS bump to take https://github.com/microsoft/sqltoolsservice/pull/2549:
image

## Code Changes Checklist


extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -499,6 +499,16 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
                 TelemetryActions.LoadFromConnectionString,
             );
 
+            // Helper function to set error message in the appropriate place
+            const setConnectionStringError = (errorMessage: string) => {
```
Benjin: nit: prefer `function setConnectionStringError(errorMessage) { }` syntax for readability

extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -510,6 +520,27 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
                     state.connectionProfile,
                 );
 
+                // trying to parse an unsupported authentication type will result in a undefined authenticationType
+                if (!state.connectionProfile.authenticationType) {
+                    // Show error for unsupported authentication type
+                    const errorMessage = l10n.t(
```
Benjin: Localized strings should be in the locConstants file with the rest of them

extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -499,6 +499,16 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
...
+                    (state.dialog as ConnectionStringDialogProps).connectionStringError =
+                        errorMessage;
+                } else {
+                    state.formMessage = { message: errorMessage };
+                }
+            };
+
             try {
                 const connDetails =
                     await this._mainController.connectionManager.parseConnectionString(
```
Benjin, line 509: Does this actually throw an error if the authentication type isn't one of "AzureMFA", "SqlLogin", or "Integrated"?  I thought the STS PR removes the code that forces it into one of these valid types, so we'll need to do that up here instead.

extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -510,6 +520,27 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
                     state.connectionProfile,
                 );
 
+                // trying to parse an unsupported authentication type will result in a undefined authenticationType
+                if (!state.connectionProfile.authenticationType) {
+                    // Show error for unsupported authentication type
+                    const errorMessage = l10n.t(
+                        "Unsupported authentication type in connection string. Only SQL Login, Integrated, and Azure MFA authentication types are supported.",
```
Benjin: In general, if an error message says that an input in incorrect, it's good to have the error repeat the invalid input back to the user (in this case, the "bad" authentication type).

extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
Benjin: We'll need to add a test that confirms this, too

Benjin (dismissed):

extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -510,6 +520,27 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
...
+                    );
+
+                    setConnectionStringError(errorMessage);
+
+                    sendErrorEvent(
```
Benjin: If the user supplies an unsupported auth type, that's a user error, not an error in our code.  We can capture that as telemetry, but it should be `sendActionEvent()`, not `sendErrorEvent()`

Issue #20579, opened 2025-11-13
[Bug]: Object Explorer incorrectly things local/xcopy server is a docker container, and throws error about it not existing


Benjin:
### Description

Docker container detection logic sees a local xcopy server and thinks it's a docker container based on the port.  Because of that, it tries (and fails) to run docker-specific logic and prompts me to delete it based on the fact that it can't find a matching container running in docker.

The repro feels a bit convoluted, but ultimately the issue is that we're not correctly detecting when a connection is a docker container.  We should probably _only_ label a connection as a docker container if

1) it's created through the container deployment UI
2) if it's created through the publish project UI
or 3) after a successful connection, we find a docker image that's both running and on the same port

### Steps to Reproduce

1. Create xcopy server on default port and connect/save.
2. Disconnect xcopy and stop process.
3. Create docker server on default port and connect/save.
4. Disconnect docker and stop container (might also need to delete?)
5. Restart xcopy process
6. Reload VS Code
7. Attempt to connect to xcopy server

Expected:
* server icon does not change to docker container image
* extension doesn't prompt me to delete my docker container

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

Windows, 1.37.0

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

Issue #20536, opened 2025-11-11
[Bug]: Password not passed correctly when creating a container


dlevy-msft-sql:
### Description

Expected the extension to create a container with the sa password set to `P@$$WORD2`.
Container was created with sa password set to `P@7698WORD2`.

### Steps to Reproduce

1. Create a container using the extension, use the password above
2. Observe the login failure
3. Use `docker inspect <containerName> | grep -v grep | grep -i password` to look at what password was actually used

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [x] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

VS Code Version: 1.105
Extension Version: 1.36.0
Operating system: Linux/codespaces

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [x] I want to work on this issue

croblesm:
Hi @dlevy-msft-sql,

I was able to reproduce this issue. This is not a problem with the extension per se, but a synchronization issue. As you can see, I have two "Local Dev" connections, but the `VS Live` profile is failing:

<img width="1472" height="903" alt="Image" src="https://github.com/user-attachments/assets/bd15d047-5268-4538-b94c-61d5820c9821" />

After reporting that error message, I see the prompt to enter the password:

<img width="1472" height="903" alt="Image" src="https://github.com/user-attachments/assets/13f1e5c2-c47a-4b4f-9cdf-773a49e16bf1" />

This creates the false impression that the container was created with the wrong password. However, that is not the case. The container in question does not exist on my local machine:

```bash
[croblesm] ~ > docker ps -a --format "table {{.Names}}\t{{.Status}}"

NAMES         STATUS
sql_copilot   Exited (137) 10 minutes ago
fabcon-dab    Exited (255) 2 months ago
sqldb_seed    Exited (255) 7 days ago
```

This is the connection metadata. Notice that the container given for the VS Live profile is sql_server_container, but again, this container does not exist on my local environment.
```JSON
        {
            "version": "2025-RC1-ubuntu-24.04",
            "user": "SA",
            "password": "",
            "savePassword": true,
[...]

Issue #19606, opened 2025-06-13
[Bug]: Local SQL Server Incorrectly Identified as Docker Container


caohai:
### Description

The Object Explorer incorrectly identifies locally installed SQL Server instances as Docker containers when both exist on the system.

![Image](https://github.com/user-attachments/assets/f6a5fafa-8935-4e4e-b733-a3472586ab86)

### Steps to Reproduce

1. Install SQL Server Developer Edition locally (using default Shared Memory protocol)
2. Create a connection to this local instance using 'localhost' with Windows Authentication
3. Deploy a SQL Server Docker container on port 1433
4. Connect to the local SQL Server instance

### Observed Behavior
1. The locally installed SQL Server instance displays a Docker container icon
2. Connecting to the local instance attempts to start the Docker container first
3. The Docker container can be stopped from the container menu of the local instance connection
 
### Expected Behavior
The application should properly distinguish between locally installed SQL Server instances and Docker containers, with appropriate icons and management options for each.

### Related doc:
https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/default-sql-server-network-protocol-configuration?view=sql-server-ver16#default-configuration
https://learn.microsoft.com/en-us/sql/tools/configuration-manager/aliases-sql-server-configuration-manager?view=sql-server-ver16#connect-to-the-local-server-using-shared-memory

### Affected Area

- [ ] Connection dialog
- [ ] Query editor
- [ ] IntelliSense/auto-completion
- [ ] Query results panel
- [x] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

Version: 1.100.3 (user setup)
Commit: 258e40fedc6cb8edf399a463ce3a9d32e7e1f6f3
Date: 2025-06-02T13:30:54.273Z
Electron: 34.5.1
ElectronBuildId: 11369351
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

MSSQL Version: 1.33.0-RC

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #20728, opened 2025-12-03
Identify whether a connection is a docker container in OE

laurenastrid1:

## Description

Fixes https://github.com/microsoft/vscode-mssql/issues/20579
Fixes https://github.com/microsoft/vscode-mssql/issues/19606

Check whether a connection is a docker container by matching a connections' server info's machine name to the existing docker container ids.

Also fixes https://github.com/microsoft/vscode-mssql/issues/20536 by properly escaping container inputs on container creation

serverInfoCheckContainer
serverInfoCheckNonContainer

Code Changes Checklist

-   [X] New or updated **unit tests** added


extensions/mssql/src/objectExplorer/objectExplorerService.ts
```diff
@@ -779,6 +764,20 @@ export class ObjectExplorerService {
             return;
         }
 
+        if (!connectionProfile.containerName) {
+            const serverInfo = this._connectionManager.getServerInfo(connectionProfile);
+            let machineName = "";
+            if (serverInfo) {
+                machineName = (serverInfo as any)["machineName"];
```
caohai: when is this filed populated (I assume in user config json file)?

extensions/mssql/src/deployment/dockerUtils.ts
```diff
@@ -915,43 +917,13 @@ async function getUsedPortsFromContainers(containerIds: string[]): Promise<Set<n
...
-        } catch {
-            // skip container if inspection fails
-        }
-    }
-
-    return undefined;
-}
-
 /**
  * Checks if a connection is a Docker container by inspecting the server name.
```
caohai: Stale doc, can you update it to the latest parameter and add some decription to it?

Benjin:
What's the overall logic for determining whether a connection is a local docker image or a different local instance (e.g. an xcopy or LOCALDB instance)?  Can you add that as a comment somewhere?

extensions/mssql/src/objectExplorer/objectExplorerService.ts
```diff
@@ -779,6 +764,20 @@ export class ObjectExplorerService {
             return;
         }
 
+        if (!connectionProfile.containerName) {
```
Benjin: I think this logic belongs in `connectionManager.connect()` instead.  We can upgrade that call's `options` arg to include a flag like "hydrateProfileMetadata" (default off, but `true` for here and connection dialog), and if that's true, then check these _after_ the connection is triggered.

That way:
* the logic stays within connection code (rather than OE code)
* doesn't block initial connection attempt
* we eliminate false positives for local servers that aren't docker

extensions/mssql/src/objectExplorer/objectExplorerService.ts
laurenastrid1: populated when a connection is made; which is guaranteed at this point in the code

laurenastrid1:
> What's the overall logic for determining whether a connection is a local docker image or a different local instance (e.g. an xcopy or LOCALDB instance)? Can you add that as a comment somewhere?

once a connection is made, figure out the machine name the connection is hosted on documented in the connection's server info; on docker connections, the machine name is the UUID corresponding to the container's id.

extensions/mssql/src/objectExplorer/nodes/connectionNode.ts
```diff
@@ -40,6 +40,19 @@ const createDisconnectedNodeContextValue = (
     };
 };
 
+const createConnectedNodeContextValue = (
+    connectionProfile: ConnectionProfile,
+): vscodeMssql.TreeNodeContextValue => {
+    let nodeSubType = connectionProfile.database ? DATABASE_SUBTYPE : undefined;
+    if (connectionProfile.containerName) nodeSubType = dockerContainer;
```
lewis-sanchez: Maybe wrap the conditional body in curly braces. It looks odd the way it's currently formatted.

lewis-sanchez (approved):

PR #20751, opened 2025-12-09
Add configuration option to show batch start/completion messages (default enabled)

adamhartford:
This PR adds a new configuration option `mssql.showBatchMessages` that allows users to control the display of batch start and completion messages in the SQL output panel. 

**Problem**: When executing large SQL files (thousands of lines), the output panel becomes cluttered with "Started executing query at..." and "Commands completed successfully" messages, making it difficult to spot actual error messages. This differs from SQL Server Management Studio behavior, which doesn't show these verbose batch messages by default.

**Solution**: Added a configurable option that:
- Defaults to `true` (preserves current behavior for backward compatibility)
- When set to `false`, suppresses batch start/completion messages while preserving all error messages
- Allows users to achieve SSMS-like output behavior for cleaner error visibility

### DIFF
--- a/extensions/mssql/src/models/sqlOutputContentProvider.ts
+++ b/extensions/mssql/src/models/sqlOutputContentProvider.ts
@@ -481,31 +481,41 @@ export class SqlOutputContentProvider {
             );
 
             const batchStartListener = queryRunner.onBatchStart(async (batch) => {
-                let time = new Date().toLocaleTimeString();
-                if (batch.executionElapsed && batch.executionEnd) {
-                    time = new Date(batch.executionStart).toLocaleTimeString();
-                }
-
-                // Build a message for the selection and send the message
-                // from the webview
-                let message: IMessage = {
-                    message: LocalizedConstants.runQueryBatchStartMessage,
-                    selection: batch.selection,
-                    isError: false,
-                    time: time,
-                    link: {
-                        text: LocalizedConstants.runQueryBatchStartLine(
-                            batch.selection.startLine + 1,
-                        ),
-                        uri: queryRunner.uri,
-                    },
-                };
-
-                const resultWebviewState = this._queryResultWebviewController.getQueryResultState(
-                    queryRunner.uri,
+                // Check configuration to see if batch messages should be shown
+                let queryUri = queryRunner.uri;
+                let extConfig = this._vscodeWrapper.getConfiguration(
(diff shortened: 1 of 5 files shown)


adamhartford:

extensions/mssql/src/models/sqlOutputContentProvider.ts
```diff
@@ -481,31 +481,41 @@ export class SqlOutputContentProvider {
...
-                    queryRunner.uri,
+                // Check configuration to see if batch messages should be shown
+                let queryUri = queryRunner.uri;
+                let extConfig = this._vscodeWrapper.getConfiguration(
+                    Constants.extensionConfigSectionName,
+                    queryUri,
                 );
-                resultWebviewState.messages.push(message);
-                this.scheduleThrottledUpdate(queryRunner.uri);
+                let showBatchMessages: boolean = extConfig.get(Constants.configShowBatchMessages) ?? true;
```
Benjin: default value in the config definition should handle the `?? true` case already

extensions/mssql/src/models/sqlOutputContentProvider.ts
```diff
@@ -481,31 +481,41 @@ export class SqlOutputContentProvider {
...
+                
+                if (showBatchMessages) {
```
Benjin: since checking this flag would be an early bail-out, it'd be a bit cleaner to do

```
if (showBatchMesssages === false) {
  return;
}
```

and leave the rest untouched

Benjin:
@croblesm @kburtram Thoughts on taking this?

Benjin:
@adamhartford looks like the test build is failing due to linter errors; can you clean those up?  Should be able to just run the vscode formatter.

adamhartford:
@Benjin Thanks for the feedback. I've made the changes you suggested and fixed the linter errors.

Benjin:
Hey @adamhartford, this fell through the cracks on our end; sorry about that.  We're happy to take it, but will need to rebase a lot of code at this point.  If you're still interested in contributing this, can you resubmit a new PR?  Otherwise, I'll pull it in (and credit you, of course!)

Benjin:
Hey @adamhartford, I've ported this functionality to an up-to-date branch, and sent it out for PR with https://github.com/microsoft/vscode-mssql/pull/22762, so I'm closing this PR.  Thanks for adding this feature!

Issue #20432, opened 2025-11-05
[Bug]: Open .scmp file using database connections does nothing


justinburnscorpcu:
### Description

I am expecting the ability to load my "Options" selections as well as my "Include" toggles between my Source and Target database connections from a .scmp file. 
What is actually happening is that after clicking "Open .scmp file" and choosing my file that I have saved previously via the VS Code extension, nothing happens. I don't see any error messages. The Source and Target fields don't populate, thus the comparison and "Include" selections don't populate. The Options remain unchanged.

### Steps to Reproduce

1. Add connection to two Azure SQL databases, authenticating via "Microsoft Entra ID - Universal with MFA support"
2. Open Schema Compare tab
3. Open .scmp file
4. After opening file, nothing happens. No error message, and no fields or settings are changed. I have tried repeating this after running a comparison first, and still no luck. At some point in the process, I am wanting the ability to load all my "Options" settings as well as my "Include" selections between the Source and Target when working with direct database connections authenticated via Microsoft Entra ID

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [x] Schema Compare
- [ ] Schema Designer
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

VS Code: 1.105.1
Extension SQL Server (mssql): 1.36.0
OS: Windows_NT x64 10.0.22621
Connection Type: Microsoft SQL Azure (Authenticated with Microsoft Entra ID)

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #20762, opened 2025-12-10
Fixing the open scmp with azure mfa src and trg endpoints

ssreerama:

Issue: https://github.com/microsoft/vscode-mssql/issues/20432

## Description
Save scmp doesn't save accountID info because the accountID is not a valid SQL Server connection property.  for the azure server and loading that scmp file fails to make the server connection due to the missing accountID info. This fix makes sures to get the matching profile from the available connections and get the accountID from there to connect to the server.

1. Added the same logic to SC and Publish profile.xml load scenario for azure mfa. 
2. Place the `helper method in shared utils.
3. Added telemetry event in connection manager
4. Added error message to the SC UI

## Code Changes Checklist

Added error message to the UI on SC failed scenario: 
image
image
image

### DIFF
--- a/extensions/mssql/src/constants/locConstants.ts
+++ b/extensions/mssql/src/constants/locConstants.ts
@@ -1507,13 +1507,6 @@ export class SchemaCompare {
             args: [errorMessage],
             comment: ["{0} is the error message from the connection attempt"],
         });
-    public static profileLoadedConnectionFailedForAzureMfa = (serverName: string) =>
-        l10n.t({
-            message:
-                "Profile loaded, but the connection could not be automatically established. Please create a connection to {0} then try again.",
-            args: [serverName],
-            comment: ["{0} is the server name"],
-        });
 }
 
 export class SchemaDesigner {
--- a/extensions/mssql/src/schemaCompare/schemaCompareWebViewController.ts
+++ b/extensions/mssql/src/schemaCompare/schemaCompareWebViewController.ts
@@ -2599,7 +2599,7 @@ export class SchemaCompareWebViewController extends ReactWebviewPanelController<
 
                         // Show the error message as no matched profile is found
                         vscode.window.showErrorMessage(
-                            locConstants.SchemaCompare.profileLoadedConnectionFailedForAzureMfa(
+                            locConstants.PublishProject.ProfileLoadedConnectionFailed(
                                 connInfo.server,
                             ),
                         );


extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -1019,6 +1019,13 @@ export default class ConnectionManager {
             account = await this.accountStore.getAccount(connectionInfo.accountId);
             profile = new ConnectionProfile(connectionInfo);
         } else {
+            // Send telemetry to identify code paths where accountId is missing
+            sendErrorEvent(
+                TelemetryViews.ConnectionManager,
+                TelemetryActions.Connect,
+                new Error("Azure MFA connection missing accountId in confirmEntraTokenValidity"),
+                false, // includeErrorMessage
```
Benjin: this can be "true" here since we know the erorr message won't contain PII

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -970,6 +970,14 @@ export class PublishProjectWebViewController extends FormWebviewController<
                 await this._connectionManager.parseConnectionString(connectionString);
             const connectionInfo = ConnectionCredentials.createConnectionInfo(connectionDetails);
 
+            // Ensure accountId is present for Azure MFA connections before connecting
+            if (connectionInfo.authenticationType === azureMfa && !connectionInfo.accountId) {
+                await Utils.ensureAccountIdForAzureMfa(
+                    connectionInfo,
+                    this._connectionManager.findMatchingProfile.bind(this._connectionManager),
```
Benjin: Why not just pass in `connectionManager` then call `findMatchingProfile` directly in the helper function?  Binding shouldn't be necessary here

extensions/mssql/src/models/utils.ts
```diff
@@ -540,6 +541,27 @@ export function isSameScmpConnection(
...
+ * @param findMatchingProfileFn Function to find matching profile from saved profiles
+ */
+export async function ensureAccountIdForAzureMfa(
+    connectionInfo: IConnectionInfo,
+    findMatchingProfileFn: (
+        profile: IConnectionProfile,
+    ) => Promise<{ profile: IConnectionProfile; score: MatchScore }>,
+): Promise<void> {
+    // Try to find accountId from saved connection profiles
+    const matchResult = await findMatchingProfileFn(new ConnectionProfile(connectionInfo));
```
Benjin: Should this also check the MatchScore to ensure it's a sufficiently-close connection?  What if it's not?

Can you add comments here for that logic (or why it doesn't need to check MatchScore if not)

extensions/mssql/src/models/utils.ts
```diff
@@ -540,6 +541,27 @@ export function isSameScmpConnection(
...
 }
 
+/**
+ * For Azure MFA connections, ensures the accountId is present in the connection info.
+ * If accountId is missing, attempts to find it from saved connection profiles.
+ * This is needed when opening connections from stored profiles like .scmp and publish.xml files.
+ * @param connectionInfo The connection info to populate with accountId if missing (modified in place)
+ * @param findMatchingProfileFn Function to find matching profile from saved profiles
+ */
+export async function ensureAccountIdForAzureMfa(
```
Benjin: see comment about just passing in ConnectionManager.  I don't see the benefit of passing a func

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: the reason for not passing the connectionManager as parameter is because, connection manager already import utils, if i do that I have to import connectionmanaegr in utils which causes circular dependency. I can place the ensureAccountId method in differnt place, but I noticed Utils file already is being used for connection related helper methods, i chose this place.

ssreerama: Also, to avoid binding, i can call the findMatchingProfile method exists in utils directly, but for that i have to get the savedConnectionsList from the connection store and then call the helper method.

extensions/mssql/src/models/utils.ts
ssreerama: Done, have moved the helper method to conn mgr class.

extensions/mssql/src/models/utils.ts
ssreerama: added comment with min of server level match we are going with

extensions/mssql/src/schemaCompare/schemaCompareWebViewController.ts
```diff
@@ -2588,9 +2588,26 @@ export class SchemaCompareWebViewController extends ReactWebviewPanelController<
                 );
                 ownerUri = utils.generateQueryUri().toString();
 
-                isConnected = await this.connectionMgr.connect(ownerUri, connInfo, {
-                    connectionSource: "schemaCompare",
-                });
+                // Ensure accountId is present for Azure MFA connections before connecting
+                if (connInfo.authenticationType === azureMfa && !connInfo.accountId) {
+                    await this.connectionMgr.ensureAccountIdForAzureMfa(connInfo);
```
Benjin: change this to return true/false depending on if it was successful

then if `false`, provide a more helpful error message (something about needing to create the connection)

ssreerama: Added more informational message

image

extensions/mssql/src/constants/locConstants.ts
```diff
@@ -1408,7 +1408,7 @@ export class PublishProject {
         return l10n.t("Failed to fetch Docker container tags: {0}", errorMessage);
     };
     public static ProfileLoadedConnectionFailed = l10n.t(
```
Benjin: We shouldn't have two identical error messages as separate entries here.  It's okay for Schema Compare to call a PublishProject string or vice-versa.

extensions/mssql/src/constants/locConstants.ts
```diff
@@ -1408,7 +1408,7 @@ export class PublishProject {
...
-        "Profile loaded but connection failed. Please connect to the server manually.",
+        "Profile loaded, but the server connection could not be established. Manual intervention is required to connect to the server.",
```
Benjin: This doesn't tell the user what "manual intervention" they should do.  How about:

> Profile loaded, but the connection could not be automatically established.  Please create a connection to {serverName} with {Entra account} to continue.

PR #20783, opened 2025-12-11
Bumping minor versions for mssql and sqlrpoj

Benjin:


extensions/sql-database-projects/package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "sql-database-projects-vscode",
   "displayName": "SQL Database Projects",
   "description": "Enables users to develop and publish database schemas for MSSQL Databases",
-  "version": "1.5.6",
```
caohai, line 5: any reason to bump the version for sqlproj given we are not releasing it this time?

Benjin: I sent this out before I'd confirmed that we weren't releasing.  Reverted.

PR #20754, opened 2025-12-09
Add tree filter to Schema Designer Filter button

laurenastrid1:
# Pull Request Template – vscode-mssql

## Description

schemaDesignerFilterButtonTree

Fixes https://github.com/microsoft/vscode-mssql/issues/19120
## Code Changes Checklist

-   [X] New or updated **unit tests** added
-   [X] All existing tests pass (`npm run test`)
-   [X] Code follows [contributing guidelines](https://github.com/microsoft/vscode-mssql/blob/main/CONTRIBUTING.md)
-   [X] Telemetry/logging updated if relevant
-   [X] No regressions or UX breakage

Issue #19120, opened 2025-04-11
[Bug]: Add tree-based table filtering to Schema Designer


yoleichen:
**Current**

![Image](https://github.com/user-attachments/assets/bbb47b33-fe36-4798-b0bb-67838edd149d)

**Expected (see design)**

![Image](https://github.com/user-attachments/assets/089c7a55-a582-4442-9fbf-71a6c6690a38)

PR #20789, opened 2025-12-11
Porting "Add tree filter to Schema Designer Filter button (#20754)"

Benjin:
Original PR: #20754 
Bug: https://github.com/microsoft/vscode-mssql/issues/19120


Benjin:
Leaving this for Jan; still some cleanup to do with this enhancement.

PR #20812, opened 2025-12-15
Add Copying and Exporting Support to Edit Data

lewis-sanchez:

## Description

This PR adds support for the following copy and export operations:
* Copy - Copies the selected grid data
* Copy with Headers - Copies selected grid data with column headers
* Copy Headers - Doesn't copy grid data, but only the column headers of selected columns.
* Export to CSV
* Export to Excel
* Export to JSON

Here's a demo of copying with these 3 copy options:
Copying Demo

This PR also adds support for exporting to CSV, Excel, and JSON.
Here's a demo of exporting to CSV
Export to CSV

Here's a demo of exporting to Excel
Export to Excel

Here's a demo of exporting to JSON
Export to JSON

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/test/unit/tableExplorerService.test.ts
+++ b/extensions/mssql/test/unit/tableExplorerService.test.ts
@@ -727,16 +727,17 @@ suite("TableExplorerService Tests", () => {
             const result = await tableExplorerService.serializeData(filePath, "csv", headers, rows);
 
             expect(result).to.equal(mockResult);
-            expect(result.succeeded).to.be.true;
-            expect(mockClient.sendRequest.calledOnce).to.be.true;
+            expect(result.succeeded, "Serialization should succeed").to.be.true;
+            expect(mockClient.sendRequest.calledOnce, "sendRequest should be called exactly once")
+                .to.be.true;
 
             const callArgs = mockClient.sendRequest.firstCall.args;
             const params = callArgs[1] as any;
             expect(callArgs[0]).to.equal(SerializeStartRequest.type);
             expect(params.saveFormat).to.equal("csv");
             expect(params.filePath).to.equal(filePath);
-            expect(params.isLastBatch).to.be.true;
-            expect(params.includeHeaders).to.be.true;
+            expect(params.isLastBatch, "isLastBatch should be true for single batch").to.be.true;
+            expect(params.includeHeaders, "includeHeaders should be true by default").to.be.true;
         });
 
         test("should successfully serialize data to JSON format", async () => {
@@ -756,7 +757,7 @@ suite("TableExplorerService Tests", () => {
                 rows,
             );
 
-            expect(result.succeeded).to.be.true;
+            expect(result.succeeded, "JSON serialization should succeed").to.be.true;
 
             const callArgs = mockClient.sendRequest.firstCall.args;
             expect((callArgs[1] as any).saveFormat).to.equal("json");
@@ -779,7 +780,7 @@ suite("TableExplorerService Tests", () => {
                 rows,
             );
 
-            expect(result.succeeded).to.be.true;
+            expect(result.succeeded, "Excel serialization should succeed").to.be.true;
 
             const callArgs = mockClient.sendRequest.firstCall.args;
             expect((callArgs[1] as any).saveFormat).to.equal("excel");
@@ -860,7 +861,7 @@ suite("TableExplorerService Tests", () => {
 
             const result = await tableExplorerService.serializeData(filePath, "csv", headers, rows);
 
-            expect(result.succeeded).to.be.false;
+            expect(result.succeeded, "Serialization should return failure status").to.be.false;
             expect(result.messages).to.equal("Failed to write file");
         });
 
@@ -875,7 +876,7 @@ suite("TableExplorerService Tests", () => {
                 expect.fail("Should have thrown an error");
             } catch (err) {
                 expect(err).to.equal(error);
-                expect(mockLogger.error.calledOnce).to.be.true;
+                expect(mockLogger.error.calledOnce, "Error should be logged").to.be.true;
                 expect(mockLogger.error.firstCall.args[0]).to.equal("Serialization failed");
             }
         });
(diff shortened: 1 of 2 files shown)


extensions/mssql/src/sharedInterfaces/tableExplorer.ts
```diff
@@ -208,6 +208,7 @@ export interface TableExplorerContextProps {
     copyScriptToClipboard: () => void;
     toggleScriptPane: () => void;
     setCurrentPage: (pageNumber: number) => void;
+    saveResults: (format: "csv" | "json" | "excel", data: ExportData) => void;
```
Benjin: This type union can be extracted out to a reusable type:

`export type SuportedSaveTypes = "csv" | "json" | "excel";`

extensions/mssql/src/tableExplorer/tableExplorerWebViewController.ts
```diff
@@ -1154,6 +1155,154 @@ export class TableExplorerWebViewController extends ReactWebviewPanelController<
...
+            } catch (error) {
+                this.logger.error(
+                    `Error saving results: ${getErrorMessage(error)} - OperationId: ${this.operationId}`,
+                );
+
+                endActivity.endFailed(
+                    new Error(`Failed to save results: ${getErrorMessage(error)}`),
+                    true,
+                    undefined,
+                    undefined,
```
Benjin: add inline comments whenever you have mystery constants.

extensions/mssql/src/tableExplorer/tableExplorerWebViewController.ts
```diff
@@ -1154,6 +1155,154 @@ export class TableExplorerWebViewController extends ReactWebviewPanelController<
...
+                    this.logger.info("Save dialog cancelled by user");
+                }
+            } catch (error) {
+                this.logger.error(
+                    `Error saving results: ${getErrorMessage(error)} - OperationId: ${this.operationId}`,
+                );
+
+                endActivity.endFailed(
+                    new Error(`Failed to save results: ${getErrorMessage(error)}`),
+                    true,
```
Benjin: If this is going to be `true`, we need to be absolutely certain that the error message cannot have any user data (like the content being copied/saved).  Filepaths and URIs should be automatically scrubbed by the telemetry library, though.

extensions/mssql/src/tableExplorer/tableExplorerWebViewController.ts
```diff
@@ -1154,6 +1155,154 @@ export class TableExplorerWebViewController extends ReactWebviewPanelController<
...
+            }
+
+            return state;
+        });
+    }
+
+    /**
+     * Format data as CSV
+     */
+    private formatAsCsv(headers: string[], rows: string[][]): string {
```
Benjin: Do we not already have functions that do this?  Check how the results grid does it, or with @aasimkhan30

extensions/mssql/package.json
```diff
@@ -165,6 +165,7 @@
         "axios": "^1.12.0",
         "dotenv": "^16.4.5",
         "error-ex": "^1.3.0",
+        "excel-builder-vanilla": "^4.2.1",
```
Benjin: Given that we're already able to save to excel files elsewhere in the extension, I assume we have existing code that doesn't require a new package?

ghiscoding: `excel-builder-vanilla` is a package that I also maintain, a fork of `excel-builder`, and that can be used in Slickgrid-React when enabling grid Excel export (you would also need `@slickgrid-universal/excel-export`). So that might be the reason to choose it!? `excel-builder-vanilla` is also the smallest JS library available, it's ESM and 7x smaller than other libs like XLXS

extensions/mssql/src/models/contracts.ts
```diff
@@ -127,3 +127,110 @@ export namespace SaveResultsAsInsertRequest {
     >("query/saveInsert");
 }
 // --------------------------------- </ Save Results as INSERT Request > ------------------------------------------
+
+// --------------------------------- < Serialize Data Request > ------------------------------------------
+// Serialize data to CSV, JSON, or Excel format using the backend serialization service
```
Benjin, line 132: If these are missing from the STS contracts, how is the query results grid doing serialization?

Not necessarily a problem, just surprised to see new contracts needed.

extensions/mssql/test/unit/tableExplorerService.test.ts
```diff
@@ -705,6 +706,217 @@ suite("TableExplorerService Tests", () => {
...
+            expect(result.succeeded).to.be.true;
+            expect(mockClient.sendRequest.calledOnce).to.be.true;
+
+            const callArgs = mockClient.sendRequest.firstCall.args;
+            const params = callArgs[1] as any;
+            expect(callArgs[0]).to.equal(SerializeStartRequest.type);
+            expect(params.saveFormat).to.equal("csv");
+            expect(params.filePath).to.equal(filePath);
+            expect(params.isLastBatch).to.be.true;
+            expect(params.includeHeaders).to.be.true;
```
Benjin: for true/false verifications, we should include failure messages because otherwise it's pretty useless: "expected false to be true".

The others at least give an indication which check failed because they'll say "expected undefined to be csv"

extensions/mssql/test/unit/tableExplorerWebViewController.test.ts
Benjin: instead of stubbing out each function individually and casting dangerously through an `any`, we should do `mockTableExplorerService = sandbox.createStubInstance(TableExplorerService)`.

See other test files for examples.

extensions/mssql/src/models/contracts.ts
lewis-sanchez: There's a total of 3 endpoints to export data that query results is using `query/saveExcel`, `query/saveCsv`, and `query/saveJson`. The three endpoints are using the `QueryExecutionService` which stores query data in a dictionary named `ActiveQueries` and this is where the data to serialize to Excel, CSV, and JSON is coming from.

The `EditDataService` manages it's own queries, so using the save endpoints that are prefixed with `query/*` wouldn't work because the mechanism to lookup the data is different and wouldn't work with the 3 `query/save*` endpoints. Using `serialize/start` just requires that we pass the data to serialize along with a formatting type (xlsx, json, csv) so that it can serialize the data to a file.

PR #20828, opened 2025-12-17
Sql Package telemetry events are failing to capture after the decoupling

ssreerama:

## Description

Telemetry events are getting lost when properties are missing from package json file after we decouple sql proj extension from ADS. This fixes the issue to capture the correct key from the json. And also removed ADS related checks and conditions from the telemetry event.

## Code Changes Checklist

### DIFF
--- a/extensions/sql-database-projects/src/common/utils.ts
+++ b/extensions/sql-database-projects/src/common/utils.ts
@@ -401,23 +401,15 @@ function readJsonFromExtensionRoot(relativePath: string): any {
 
 export function getPackageInfo(packageJson?: any): IPackageInfo | undefined {
 	if (!packageJson) {
-		packageJson = readJsonFromExtensionRoot('package.ads.json');
+		packageJson = readJsonFromExtensionRoot('package.json');
 	}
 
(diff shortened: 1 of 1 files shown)


extensions/sql-database-projects/src/common/utils.ts
```diff
@@ -401,23 +401,15 @@ function readJsonFromExtensionRoot(relativePath: string): any {
 
 export function getPackageInfo(packageJson?: any): IPackageInfo | undefined {
 	if (!packageJson) {
-		packageJson = readJsonFromExtensionRoot('package.ads.json');
+		packageJson = readJsonFromExtensionRoot('package.json');
```
Benjin, line 404: Be sure to confirm via Fiddler capture that these events are getting emitted properly before merging this in.

Benjin (approved):

extensions/sql-database-projects/src/common/utils.ts
ssreerama: I have verified the VScode Telemetry and fiddler classic for the telemetry events and confirming telemetry is looking good.

Issue #19717, opened 2025-07-09
[Bug]: Inconsistent Target Platform Selection When Creating New Project via "+" Button


jasonhorner:
### Description

Description:

When creating a new SQL project using the "+" button in the vscode-mssql extension, the only selectable target platforms are "Azure SQL Database" and "SQL Server". However, after the project is created, the Target Platform drop-down allows you to choose from the full list of valid platforms (e.g., SQL Server 2019, Azure SQL DB for Fabric etc.).

Initial selction:

<img width="858" height="384" alt="Image" src="https://github.com/user-attachments/assets/bd81fe82-fcfb-40e7-98bf-00b4186d125d" />

Changing target after creation:

<img width="825" height="415" alt="Image" src="https://github.com/user-attachments/assets/9b2e8da6-5c6b-4670-a512-731e7559cffa" />

This inconsistency limits flexibility during initial project creation and may confuse users who expect the full list of target platforms to be available from the start.

### Steps to Reproduce

Steps to Reproduce:

Open the SQL Projects pane in VS Code.

Click the "+" button to create a new SQL project.

Observe the limited target platform options (only Azure SQL DB and SQL Server).

Create the project.

After creation, open the project settings.

Observe that the Target Platform drop-down now shows many more options.

Expected Behavior:

The full list of valid target platforms should be available when creating a new SQL project, not just after creation.

### Affected Area

- [ ] Connection dialog
- [ ] Query results panel
- [ ] Query editor
- [ ] Object Explorer
- [ ] Table Designer
- [ ] Schema Compare
- [ ] Schema Designer
- [ ] Local container
- [ ] GitHub Copilot integration
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

### If you selected "Other", please describe the affected area

Database Projects - this should be added to the list of checkboxes in affected area

### Environment Information

SQL Projects: 1.5.3

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [x] I want to work on this issue

ssreerama:
Hey @jasonhorner , could you help clarify the issue?

It looks like there might be some confusion between the project type and the target platform options. When you select "Azure SQL DB" as the project type, the target platform options are not available. However, if you choose "SQL Server Database", you’ll see the supported target platform options.

Looping in @dzsquared for additional insights.

ssreerama:
@jasonhorner , Do you have any update on my ask?

jasonhorner:
[...]

PR #20848, opened 2025-12-18
Fix - Adding Target Platform options while Azure project creation

ssreerama:

# ISSUE: https://github.com/microsoft/vscode-mssql/issues/19717

## Description
- We previously did not provide projectTypes for Azure projects.
- This caused the targetPlatform quick pick to stay disabled when creating Azure projects for years.
- To ensure consistent behavior across all project creation flows, we are adding back the Azure project projectTypes.
- This re-enables the target platform quick pick for Azure projects.
- Although the underlying fix is in the sqlprojects code, the New Project feature belongs to the Data Workspace, so the tests are added there.
- For the new test file, I imported the necessary Chai and Sinon‑Chai modules.
- As part of the Data Workspace Phase 2 work, all existing tests will be migrated to this updated testing pattern.

## Code Changes Checklist

Before Fix: ( NO Target Platform quickpick will appear)
AzureProjTargetPlatform_BeforeFix

After Fix:
AzureProjTargetPlatform_AfterFix


Benjin:
Can you add screenshots for the before/after flows?

extensions/data-workspace/src/test/dialogs/newProjectQuickpick.test.ts
```diff
@@ -0,0 +1,225 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from "vscode";
```
Benjin, line 6: Out of curiosity, did you use AI to generate these tests?  If so, which tool and which prompt?

I noticed that the tests look like they're using the same instructions from `/extensions/mssql/test/AGENT.md` (which is good!), but I didn't expect that file to be found/used since it isn't is this path.

I wonder if we should centralize the testing instructions, or add an AGENTS.md file here that just says "follow the instructions in <path to other one>"

ssreerama: I'm using the vscode copilot to generate the test file, but I had provided the prompt like to follow the mssql test suite structure, since data-workspace and sql-proj extensions have different test structure compatible with ADS repo. 

it would be more beneficial to make the agents.md file as centralized.

PR #20853, opened 2025-12-19
Adding support for reading from/writing to workspace connections.

Benjin:
## Description

This PR adds support for reading from and writing to workspace connections and connection groups.  With this change, users should be able to check workspace config files into their repos to share connections and groups with coworkers.

Note: only workspace configuration is supported, not workspace _folder_ config.  If requested from users, that support can come later.  Anticipated upgrade for that involve tracking the specific workspace folder config file for each connection/group (currently it's just whether it's global or workspace JSON, of which there can only be one each at a time)

Fixes https://github.com/microsoft/vscode-mssql/issues/19780

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/test/unit/connectionUI.test.ts
+++ b/extensions/mssql/test/unit/connectionUI.test.ts
@@ -219,9 +219,7 @@ suite("Connection UI tests", () => {
 
         await connectionUI.removeProfile();
 
-        expect(connectionStoreStub.getProfilePickListItems).to.have.been.calledOnceWithExactly(
-            false,
-        );
+        expect(connectionStoreStub.getProfilePickListItems).to.have.been.calledOnce;
         expect(promptStub).to.have.been.calledOnce;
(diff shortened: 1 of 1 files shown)


extensions/mssql/src/connectionconfig/connectionconfig.ts
```diff
@@ -53,54 +55,20 @@ export class ConnectionConfig implements IConnectionConfig {
      * Get a list of all connections in the connection config. Connections returned
      * are sorted first by whether they were found in the user/workspace settings,
      * and next alphabetically by profile/server name.
+     * @param _alsoGetFromWorkspace deprecated parameter; connections are always retrieved from both user and workspace settings.
```
aasimkhan30: NIT: Should we just remove this?

aasimkhan30: nvm saw your todo

aasimkhan30 (approved):

PR #20915, opened 2026-01-16
Added Azure Blob Logic

laurenastrid1:

## Description

Added azure blob logic. This includes the service itself as well as azure storage+blob helpers. The service and helpers are used in the backup+restore functionalities.

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/test/unit/azureHelpers.test.ts
+++ b/extensions/mssql/test/unit/azureHelpers.test.ts
@@ -323,6 +323,7 @@ suite("Azure Helpers", () => {
                 mockSubscriptions[0],
                 clientStub as unknown as armStorage.StorageManagementClient,
             );
+            expect.fail("Expected fetchStorageAccountsForSubscription to throw");
         } catch (e) {
             expect(e.message).to.equal("Test error");
         }
@@ -367,6 +368,7 @@ suite("Azure Helpers", () => {
                 mockAzureResources.storageAccount,
                 clientStub as unknown as armStorage.StorageManagementClient,
             );
+            expect.fail("Expected fetchBlobContainersForStorageAccount to throw");
         } catch (e) {
             expect(e.message).to.equal("Test error");
         }
@@ -414,6 +416,7 @@ suite("Azure Helpers", () => {
                 mockAzureResources.storageAccount,
                 clientStub as unknown as armStorage.StorageManagementClient,
             )) as armStorage.StorageAccountsListKeysResponse;
+            expect.fail("Expected getStorageAccountKeys to throw");
         } catch (e) {
             expect(e.message).to.equal("Test error");
         }


extensions/mssql/src/utils/utils.ts
```diff
@@ -204,3 +204,15 @@ export function formatXml(xml: string): string {
...
     }
 }
+
+/**
+ * This function adds one year to the current date and returns it in the UTC format.
+ * It's used to pass an expiration date argument to the create shared access signature RPC.
+ * It returns the date in the UTC format for locale time zone independence.
+ * @returns next year's UTC date
+ */
+export function nextYear(): string {
```
Benjin: Based on the docstring, this should probably be named something like "getExpirationDateForSas" or something, in case we end up wanting to change the TTL for the SAS.

Benjin: Why one year?  Is the SAS persisted anywhere or just used for this session?  Can it be shorter (to avoid potential security leakage opportunities)?

Benjin:
Patch coverage looks pretty low.  Can you add additional tests to bring that up to 60+%?

extensions/mssql/src/utils/utils.ts
laurenastrid1: This is what we did in ADS

laurenastrid1: i copied this function over from ADS but i can definetely rename it

extensions/mssql/src/connectionconfig/azureHelpers.ts
```diff
@@ -282,66 +282,93 @@ export class VsCodeAzureHelper {
...
-            storageAccountResourceGroup,
-            storageAccount.name,
-        );
+            return await storage.storageAccounts.listKeys(
+                storageAccountResourceGroup,
+                storageAccount.name,
+            );
+        } catch (error) {
+            console.error("Error fetching storage account keys:", error);
+            return {};
```
Benjin: Why do we want to return an empty object instead of something that more explicitly indicates there was an error (like throwing, or returning an explicit error object that includes the message?

extensions/mssql/src/connectionconfig/azureHelpers.ts
```diff
@@ -282,66 +282,93 @@ export class VsCodeAzureHelper {
...
-        // get resource group for storage account
-        return listAllIterator(
-            storage.blobContainers.list(storageAccountResourceGroup, storageAccount.name),
-        );
+            // get resource group for storage account
+            return listAllIterator(
+                storage.blobContainers.list(storageAccountResourceGroup, storageAccount.name),
+            );
+        } catch (error) {
+            console.error("Error fetching blob containers for storage account:", error);
```
Benjin, line 335: Same question here about error handling: we want to make sure we "fail elegantly", i.e. we can surface useful messages to users.

extensions/mssql/src/connectionconfig/azureHelpers.ts
```diff
@@ -282,66 +282,93 @@ export class VsCodeAzureHelper {
...
+            const storage =
+                storageClient ??
+                new StorageManagementClient(sub.credential, sub.subscriptionId, {
+                    endpoint: getCloudProviderSettings().settings.armResource.endpoint,
+                });
 
-        return listAllIterator(storage.storageAccounts.list());
+            return listAllIterator(storage.storageAccounts.list());
+        } catch (error) {
+            console.error("Error fetching storage accounts for subscription:", error);
```
Benjin, line 301: same comment here about error handling

extensions/mssql/test/unit/azureHelpers.test.ts
```diff
@@ -316,6 +316,17 @@ suite("Azure Helpers", () => {
             "storageAccounts.list should be called once",
         ).to.be.true;
 
+        clientStub.storageAccounts.list = sinon.stub().rejects(new Error("Test error"));
+
+        try {
```
Benjin, line 321: If this is broken and doesn't throw for some reason, then the test will (incorrectly) pass, right?

Probably should instead be using

```
           expect(() =>
                await azureHelpers.VsCodeAzureHelper.fetchStorageAccountsForSubscription()
            ).to.throw(message);
```

PR #20916, opened 2026-01-16
Change Edit Data grid default page size to 50

lewis-sanchez:

## Description

This PR adjusts the default page size for the Edit Data grid from 10 to 50.

Before:
image

After:
image

Users can still change the page size after the grid loads if they want to see less or more records. A demo of that is here:
Change Default Page Size

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/reactviews/pages/TableExplorer/TableExplorerPage.tsx
+++ b/extensions/mssql/src/reactviews/pages/TableExplorer/TableExplorerPage.tsx
@@ -136,7 +136,6 @@ export const TableExplorerPage: React.FC = () => {
                                             ref={gridRef}
                                             resultSet={resultSet}
                                             themeKind={themeKind}
-                                            pageSize={50}
                                             currentRowCount={currentRowCount}
                                             failedCells={failedCells}
                                             deletedRows={deletedRows}


extensions/mssql/src/reactviews/pages/TableExplorer/TableDataGrid.tsx
```diff
@@ -62,7 +62,7 @@ export const TableDataGrid = forwardRef<TableDataGridRef, TableDataGridProps>(
         {
             resultSet,
             themeKind,
-            pageSize = 100,
+            pageSize = 50,
```
ssreerama, line 65: this is just a default size right, and user can still change the view from the dorpdown for page size right! and what are we showing there? 

50 100 150 200 .... in the dropdown?
not sure what we show when it was 10, was it multiples of the (default size)?

lewis-sanchez: Yeah, this is just the default size and users can still change the page size using the dropdown.

With the previous default page size of 10, if your result set had 50 records you would previously see that split across 5 pages with each page having 10 rows. If you changed the page size to 50, then you would just see 1 page because all 50 records could fit on a single 50 row page.

PR #20917, opened 2026-01-16
Create databases dialog. 

aasimkhan30:

## Description

1. Create database dialog
image

2. Drop Database Dialog
image

3. Rename Database
image

Created a  base `ObjectManagementWebviewController` to port more Object management dialogs in the future. 

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/reactviews/pages/ObjectManagement/createDatabaseDialogPage.tsx
+++ b/extensions/mssql/src/reactviews/pages/ObjectManagement/createDatabaseDialogPage.tsx
@@ -15,6 +15,7 @@ import {
     ObjectManagementSubmitRequest,
 } from "../../../sharedInterfaces/objectManagement";
 import { locConstants } from "../../common/locConstants";
+import { getErrorMessage } from "../../common/utils";
 import { ObjectManagementDialog } from "../../common/objectManagementDialog";
 import { ObjectManagementContext } from "./objectManagementStateProvider";
 import { CreateDatabaseForm, CreateDatabaseFormState } from "./createDatabaseForm";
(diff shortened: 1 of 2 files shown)


extensions/mssql/src/constants/locConstants.ts
```diff
@@ -26,6 +26,50 @@ export class Common {
...
+export let databaseNameRequired = l10n.t("Database name is required");
+export let msgSelectServerNodeToCreateDatabase = l10n.t(
+    "Please select a server node in Object Explorer to create a database.",
+);
+export let msgSelectDatabaseNodeToDrop = l10n.t(
+    "Please select a database node in Object Explorer to drop.",
+);
+export let msgSelectDatabaseNodeToRename = l10n.t(
+    "Please select a database node in Object Explorer to rename.",
+);
```
Benjin: If the menu item is only showing up on valid nodes, do these messages need to be localized?

extensions/mssql/src/controllers/createDatabaseWebviewController.ts
```diff
@@ -0,0 +1,228 @@
...
+    ObjectManagementDialogType,
+} from "../sharedInterfaces/objectManagement";
+import * as Constants from "../constants/constants";
+import * as LocConstants from "../constants/locConstants";
+import { ObjectManagementService } from "../services/objectManagementService";
+import { getErrorMessage } from "../utils/utils";
+import VscodeWrapper from "./vscodeWrapper";
+import { ObjectManagementWebviewController } from "./objectManagementWebviewController";
+
+interface OptionsCollection {
```
laurenastrid1: why not add these interfaces to the object management shared interface file?

extensions/mssql/package.json
```diff
@@ -578,6 +583,16 @@
                     "when": "view == objectExplorer && viewItem =~ /\\btype=(disconnectedServer|Server|Database)\\b/",
                     "group": "2_MSSQL_serverDbActions@2"
                 },
+                {
```
laurenastrid1, line 586: if the View Item is a direct connection to a database instead of a database node under a server, we should also show these options. I think the way to check this is by also checking if the subtype is a Database

laurenastrid1:
I'm not sure if this is just my machine, but with the create database dialog, it doesn't register server connection. Also, the advanced options button doesn't have any options under it. 

createDatabaseReview

aasimkhan30:
@laurenastrid1 , what error are you seeing in console?

extensions/mssql/package.json
laurenastrid1: like this sort of connection: 
image

PR #20925, opened 2026-01-19
Profiler STS Integration adding UI foundation and multi-session support

allancascante:
## Description

Creating the UI for the profiler, adding integration with STS backend to perform all profiler operations: Create, Start, Stop and Pause profile sessions, creating the ring buffer and integrated it with multi-session support, added notifications and toolbar to allow user to perform the ui operations, added an action to launch the profiler, integrate it with slickgrid and created the template configuration for the session create sql commands and template views; also implemented notifications, status bar display and logging. 

UI 

image

image

image

Create New Session (Click on + New Session) : 

image

image

Note:
- This is missing Telemetry it will be added in a later PR.

## Code Changes Checklist


extensions/mssql/src/profiler/profilerSession.ts
```diff
@@ -0,0 +1,664 @@
...
+import { v4 as uuidv4 } from "uuid";
+import { RingBuffer } from "./ringBuffer";
+import { EventRow, Filter, SessionType, SessionState, ViewTemplate } from "./profilerTypes";
+import { ProfilerService } from "../services/profilerService";
+import { ProfilingSessionType } from "../models/contracts/profiler";
+
+/**
+ * Default event buffer capacity
+ */
+const DEFAULT_BUFFER_CAPACITY = 10000;
```
Benjin: For larger numbers, it's good to use spacers:

```suggestion
const DEFAULT_BUFFER_CAPACITY = 10_000;
```

extensions/mssql/src/profiler/profilerSession.ts
```diff
@@ -0,0 +1,664 @@
...
+
```
Benjin: How'd we land at 10K events?  Do we have an estimate for:
* about how long of a profiler session would fill this up with default options?
* about how much memory would a full (10k) buffer take up?

extensions/mssql/src/profiler/profilerDefaultConfig.ts
```diff
@@ -0,0 +1,536 @@
...
+
+import { EngineType, ProfilerConfig } from "./profilerTypes";
+
+/**
+ * Default profiler configuration with templates and views
+ * Based on Azure Data Studio profiler configuration
+ */
+export const defaultProfilerConfig: ProfilerConfig = {
+    views: {
+        "Standard View": {
```
Benjin: I'm always suspicious of Record entries because it's difficult/impossible to have type-safety on the keys, which means:
* we lose type-safety
* we have to use uglier Object.values/keys to traverse the structure

How much flexibility do we have with the definition of this structure?  Can ADS users customize/save their config that we have to be compatible with?

extensions/mssql/src/profiler/profilerDefaultConfig.ts
```diff
@@ -0,0 +1,536 @@
...
+    WHERE ([package0].[equal_boolean]([sqlserver].[is_system],(0)) AND [duration]>=(1000))),
+ADD EVENT sqlserver.sql_batch_starting(
+    ACTION(package0.event_sequence,sqlserver.client_app_name,sqlserver.client_pid,sqlserver.database_id,sqlserver.database_name,sqlserver.nt_username,sqlserver.query_hash,sqlserver.server_principal_name,sqlserver.session_id)
+    WHERE ([package0].[equal_boolean]([sqlserver].[is_system],(0))))
+ADD TARGET package0.ring_buffer(SET max_memory=(25600))
+WITH (MAX_MEMORY=4096 KB,EVENT_RETENTION_MODE=ALLOW_SINGLE_EVENT_LOSS,MAX_DISPATCH_LATENCY=5 SECONDS,MAX_EVENT_SIZE=0 KB,MEMORY_PARTITION_MODE=NONE,TRACK_CAUSALITY=ON,STARTUP_STATE=OFF)`,
+        },
+    },
+    viewToSessionMap: {
+        "Standard View": ["Standard_OnPrem", "Standard_Azure"],
```
Benjin: Are these "sessions" or "session templates"?  I thought sessions were a runtime artifact when a user starts using the profiler?

extensions/mssql/src/profiler/profilerTypes.ts
```diff
@@ -0,0 +1,218 @@
...
+ */
+export interface ProfilerConfig {
+    /** Available view templates keyed by ID */
+    views: Record<string, ViewTemplate>;
+    /** Available session templates keyed by ID */
+    templates: Record<string, ProfilerTemplate>;
+    /** Maps view IDs to compatible session template IDs */
+    viewToSessionMap: Record<string, string[]>;
+    /** Maps session template IDs to compatible view IDs */
+    sessionToViewMap: Record<string, string[]>;
```
Benjin: Should these maps be something that gets calculated from the JSON when the profile is loaded?  Storing these in the JSON itself risks drift.

extensions/mssql/src/profiler/profilerTypes.ts
```diff
@@ -0,0 +1,218 @@
...
+}
+
+/**
+ * Configuration structure for profiler templates and views
+ */
+export interface ProfilerConfig {
+    /** Available view templates keyed by ID */
+    views: Record<string, ViewTemplate>;
+    /** Available session templates keyed by ID */
+    templates: Record<string, ProfilerTemplate>;
```
Benjin: Part of me thinks that the config should just be an array of views and array of templates - with the IDs/keys stored as a property in each view/template - then at Profiler launch, these get read into a map, and the view-to-session and session-to-view maps get calculated.  Cleans up the code and config file a lot, IMO.

extensions/mssql/src/profiler/profilerConfigService.ts
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Get a field value from an EventRow, checking both direct properties and additionalData
+     */
+    private getFieldValue(event: EventRow, field: string): string | number | null {
+        // Check if it's a direct property of EventRow
+        if (field in event && field !== "additionalData") {
+            const value = (event as unknown as Record<string, unknown>)[field];
+            if (value === undefined || value === null) {
```
Benjin, line 201: Linter blocks use of `null`, and wants us to use `undefined` where possible (honestly don't know why, tbh.)  May as well make it happy, and use eslint disable lines where the value _needs_ to be `null`.

extensions/mssql/src/profiler/profilerConfigService.ts
```diff
@@ -0,0 +1,242 @@
...
+        // Check additionalData
+        if (event.additionalData && field in event.additionalData) {
+            return event.additionalData[field] ?? null;
+        }
+
+        return null;
+    }
+
+    /**
+     * Format a timestamp for display
```
Benjin: Give example of target timestamp format

extensions/mssql/src/profiler/profilerConfigService.ts
```diff
@@ -0,0 +1,242 @@
...
+     */
+    private getFieldValue(event: EventRow, field: string): string | number | null {
+        // Check if it's a direct property of EventRow
+        if (field in event && field !== "additionalData") {
+            const value = (event as unknown as Record<string, unknown>)[field];
+            if (value === undefined || value === null) {
+                return null;
+            }
+            // Format specific fields
+            if (field === "timestamp") {
```
Benjin: use a const instead (here and in default config)

extensions/mssql/src/profiler/profilerController.ts
```diff
@@ -0,0 +1,527 @@
...
+            webviewController.setCreatingSession(false);
+            vscode.window.showErrorMessage(LocProfiler.failedToCreateSession(String(e)));
+        }
+    }
+
+    /**
+     * Launches the profiler UI with a connection to manage profiling sessions.
+     * This is the main entry point for opening a profiler window.
+     */
+    public async launchProfiler(): Promise<void> {
```
Benjin: Nit: good practice to try to put public methods above private methods in the code; easier to see what the external interface for the class is

extensions/mssql/src/profiler/profilerController.ts
```diff
@@ -0,0 +1,527 @@
...
+
+        try {
+            // Prompt user to select a server and create a dedicated profiler connection
+            this._logger.verbose("Prompting user to select a server for profiling...");
+            const profilerUri = await this.promptForNewConnection();
+            if (!profilerUri) {
+                this._logger.verbose("User cancelled or connection failed");
+                return;
+            }
+            this._logger.verbose(`Profiler connection created: ${profilerUri}`);
```
Benjin: Shouldn't we already have the selected connection from when the user launched this via right-click menu?  I assume this feature is only able to be launched by right-clicking on a database (not a server, not from command palette)

extensions/mssql/src/profiler/profilerController.ts
```diff
@@ -0,0 +1,527 @@
...
+
+        this._logger.verbose("Profiler commands registered");
+    }
+
+    /**
+     * Prompts the user to select a saved connection profile and connects to it.
+     * Creates a dedicated profiler connection.
+     * @returns The connection URI if successful, undefined if cancelled or failed
+     */
+    private async promptForNewConnection(): Promise<string | undefined> {
```
Benjin: Why does the user need to select which connection?  I'm hoping we can get rid of this.

extensions/mssql/src/profiler/ringBuffer.ts
```diff
@@ -0,0 +1,225 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { IndexedRow, Filter, QueryResult } from "./profilerTypes";
+
+/**
+ * A circular buffer for storing rows with automatic sequence numbering.
+ *
+ * @template T - Row type that extends IndexedRow
+ */
+export class RingBuffer<T extends IndexedRow> {
```
Benjin, line 18: Did we get this code from somewhere (that we need to reference), or written just for this?

extensions/mssql/src/profiler/ringBuffer.ts
```diff
@@ -0,0 +1,225 @@
...
+    private _size: number = 0;
+    private _capacity: number;
+    private _paused: boolean = false;
+
+    /**
+     * Creates a new RingBuffer with the specified capacity.
+     * @param capacity - Maximum number of rows to store
+     * @param _indexedFields - Unused, kept for API compatibility
+     */
+    constructor(capacity: number, _indexedFields: string[] = []) {
```
Benjin, line 30: How does a user change the ring buffer capacity?  If they do it in the middle of a session, does the history reset?

extensions/mssql/src/profiler/profilerSessionManager.ts
```diff
@@ -0,0 +1,424 @@
...
+     * @param sessionId - The session ID
+     * @returns True if the session is now paused, false if resumed
+     * @throws Error if session not found
+     */
+    togglePauseSession(sessionId: string): boolean {
+        const session = this._sessions.get(sessionId);
+        if (!session) {
+            throw new Error(`Session '${sessionId}' not found`);
+        }
+        if (session.isPaused) {
```
Benjin, line 364: better to have the "toggle" behavior in the front-end, so all calls to the backend are explicitly "pause" or "start"

extensions/mssql/src/profiler/profilerController.ts
```diff
@@ -0,0 +1,527 @@
...
+        private _sessionManager: ProfilerSessionManager,
+    ) {
+        this._logger = Logger.create(this._vscodeWrapper.outputChannel, "Profiler");
+        this.registerCommands();
+    }
+
+    private registerCommands(): void {
+        // Launch Profiler command
+        this._context.subscriptions.push(
+            vscode.commands.registerCommand("mssql.profiler.launch", async () => {
```
Benjin: I think we can get rid of this command and force users through the context menu.  Don't forget to update the package.json to remove these from the command palette.

extensions/mssql/src/profiler/ringBuffer.ts
Benjin: Would be good to add a comment saying that this is an adaptation of the ring buffer code in STS

extensions/mssql/src/profiler/ringBuffer.ts
Benjin: If the user has to change the config setting, that means that the session is reset

Benjin: Sounds like this is by design for now.  Support for changing ring buffer size  while preserving session can be added later.

extensions/mssql/src/profiler/profilerSession.ts
allancascante: Current ADS implementation has a 1000 ring buffer, the selected number is a 10X that number; memory for 10 000 events will vary particurly as Event might have very large SQL statements; i.e. a simple SELECT 1 would be about 10 bytes of text, while a large SQL statement might be 10-100 times that. In the implementation the ring buffer will keep one copy of the entire event data, the value for text send to slick grid will be trimmed down to 10k max. But 10 000 events with each event about 2KB will be about 20MB of memory, this has to be multiplied by 2 for the data actually kept in the slick grid.

allancascante: There is no way to know when this is going to fill up as it depends on how the users use the feature, that's why there is a setting for them to adjust, and we will have telemetry for when events are being drop, until this feature is being used by users we will have data to know the answers to your questions.

extensions/mssql/src/profiler/profilerSessionManager.ts
allancascante: this is a port of the ADS behavior, this will be a change for later version

extensions/mssql/src/profiler/profilerConfigService.ts
```diff
@@ -11,19 +11,62 @@ import {
...
+        this._viewToSessionMap = new Map();
+        this._sessionToViewMap = new Map();
+        this.computeMaps();
+    }
+
+    /**
+     * Computes the view-to-session and session-to-view maps from the templates' defaultView property.
+     * This ensures the maps stay in sync with the actual template definitions.
+     */
+    private computeMaps(): void {
```
Benjin, line 46: Doesn't particularly matter for this case, but...

Generally a good practice to either have this return its own maps (and have the caller assign to the instance variables) or to have this _also_ handle the initialization (/reinitialization) of the maps.  Splitting the responsibility like this makes it more likely for bugs to be introduced.

extensions/mssql/src/profiler/profilerConfigService.ts
```diff
@@ -40,10 +83,7 @@ export class ProfilerConfigService {
      * Get all available templates
      */
     public getTemplates(): ProfilerTemplate[] {
-        return Object.entries(this._config.templates).map(([id, template]) => ({
-            ...template,
-            id,
-        }));
+        return this._config.templates;
```
Benjin, line 86: Thanks for making these changes!  This looks much cleaner.

Issue #20937, opened 2026-01-20
[Bug]: Publish Project server selection should be a dropdown of servers, not launch the connection dialog


Benjin:
Publish Project dialog currently launches the connection dialog when selecting a server, which is disorienting and confusing to users.

Instead, we should follow the model that the Dacpac wizard dialog uses, with a dropdown of servers (and another one for databases, which is populated once the server dropdown connects)

Change this:

<img width="688" height="49" alt="Image" src="https://github.com/user-attachments/assets/9e9b9152-1ea0-48bd-8613-7822decd0530" />

to this:

<img width="724" height="154" alt="Image" src="https://github.com/user-attachments/assets/5084df64-be54-4705-8d76-b1d77907da08" />

PR #20964, opened 2026-01-21
Publish Dialog - Changing the server field to combo box that holds the active connections

ssreerama:

# Issue: https://github.com/microsoft/vscode-mssql/issues/20937

## Description

This PR:
- Converts the server input field to combobox
- User can select the connections from the dropdown as well as search the sever from the list.
- Any recently added connections from connection dialog adds automatically to the server dd.
- little improvements like error handling and loading databases placeholder are placed.
- If user wants to add a new connection, they can add it from the server connection dialog, and the server dropdown will be pulling those automatically

## Code Changes Checklist

## Reviewers: 
[Please read our reviewer guidelines](https://github.com/microsoft/vscode-mssql/blob/main/.github/REVIEW_
GUIDELINES.md)

Before Fix: Please observer Server field(each new connection requires to open connection dialog) and profile loading connecting the server and Db in background.
Server_Profile_Improvement_BeforeFix

After fix: Server DD improvement
PublishProjectDialog_NewServerComboFinal

Profile load scenario - connecting to server:
profile Load

### DIFF
--- a/extensions/mssql/src/publishProject/publishProjectWebViewController.ts
+++ b/extensions/mssql/src/publishProject/publishProjectWebViewController.ts
@@ -1032,25 +1032,12 @@ export class PublishProjectWebViewController extends FormWebviewController<
                 this.updateState();
 
                 const result = await this.connectToServerByConnectionId(payload.connectionId);
-
-                state.loadConnectionStatus = result.errorMessage
-                    ? ApiStatus.Error
-                    : ApiStatus.Loaded;
-
-                // Update database dropdown options (clear on error, populate on success)
                 const databaseComponent = state.formComponents[PublishFormFields.DatabaseName];
-                if (result.errorMessage || !result.databases) {
-                    // Clear options on error or no databases
-                    databaseComponent.options = [];
-                } else {
-                    // Populate with database list
-                    databaseComponent.options = result.databases.map((db) => ({
-                        displayName: db,
-                        value: db,
-                    }));
-                }
 
+                // Handle error case: set error status, clear options, show message, and return early
                 if (result.errorMessage) {
+                    state.loadConnectionStatus = ApiStatus.Error;
+                    databaseComponent.options = [];
                     state.formMessage = {
                         message: result.errorMessage,
                         intent: "error",
@@ -1058,7 +1045,13 @@ export class PublishProjectWebViewController extends FormWebviewController<
                     return state;
                 }
 
-                // Update connection info
+                // Success case: update status, populate databases, and set connection info
+                state.loadConnectionStatus = ApiStatus.Loaded;
+                databaseComponent.options = (result.databases || []).map((db) => ({
+                    displayName: db,
+                    value: db,
+                }));
+
                 this._connectionUri = result.ownerUri;
                 state.formState.serverName = result.serverName || "";
 


extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +504,95 @@ export class PublishProjectWebViewController extends FormWebviewController<
         }
     }
 
+    /**
+     * Gets active connections formatted for dropdown display (deduplicated by server)
```
Benjin: This list show be displaying all saved connections, not just the active ones.  Can you reference how the dacpac dialog implements this?  It works well over there.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,167 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+     */
+    private updateServerDropdownOptions(): void {
+        const serverComponent = this.state.formComponents[PublishFormFields.ServerName];
+        if (serverComponent && this.state.availableConnections) {
+            serverComponent.options = this.state.availableConnections.map((profile) => ({
+                displayName: profile.profileName || "",
+                value: profile.id || "",
+            }));
+        }
+    }
```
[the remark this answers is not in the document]
ssreerama: I removed that functionality and will be updating the description.

extensions/mssql/src/reactviews/common/locConstants.ts
```diff
@@ -1077,6 +1077,12 @@ export class LocConstants {
             copySqlPackageCommandToClipboard: l10n.t("Copy command to clipboard"),
             showUnmaskedCommand: l10n.t("Show unmasked command (reveals sensitive information)"),
             showMaskedCommand: l10n.t("Show masked command (hides sensitive information)"),
+            selectServer: l10n.t("Select a server"),
+            noConnectionsAvailable: l10n.t(
+                "No connections available. Please create a connection first.",
+            ),
+            addConnection: l10n.t("Add Connection..."),
```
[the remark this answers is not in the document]
ssreerama: cleaned!

extensions/mssql/src/reactviews/pages/PublishProject/components/FormFieldComponents.tsx
aasimkhan30, line 248: It will be nice to switch this to searchable dropdown.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,186 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+     * Returns all connection profiles, not just active connections.
+     */
+    private async listSavedConnections(): Promise<IConnectionDialogProfile[]> {
+        try {
+            const savedConnections =
+                await this._connectionManager.connectionStore.readAllConnections();
+
+            return savedConnections.map((conn) => {
+                const profile = conn as IConnectionProfile;
+                const displayName = profile.profileName || getConnectionDisplayName(profile);
```
Benjin: just `getConnectionDisplayName()` is good-enough; it'll check `.profileName` first.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,186 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                    database: profile.database,
+                    user: profile.user,
+                    password: profile.password,
+                    authenticationType: profile.authenticationType,
+                    profileName: displayName,
+                    id: profile.id,
+                    groupId: profile.groupId,
+                    savePassword: profile.savePassword,
+                    azureAuthType: profile.azureAuthType,
+                } as IConnectionDialogProfile;
```
Benjin: Why are we constructing a new object here, and why manually?

If there's no display name, we could have a call to fetch that (like [here](https://github.com/microsoft/vscode-mssql/blob/a86c0d1b0ceba4f78b0930cbb58c51d7bb1c4b4b/extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts#L650)), or we could repackage:

```
{
    displayName: string;
    profile: IConnectionProfile
}
```

I don't love the idea of reconstructing an identical object unless we have to... and if we do, then we should use deepClone() or something.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,186 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+            // Find the profile in saved connections
+            const savedConnections =
+                await this._connectionManager.connectionStore.readAllConnections();
+            const profile = savedConnections.find((conn) => {
+                const connProfile = conn as IConnectionProfile;
+                return connProfile.id === profileId;
+            }) as IConnectionProfile | undefined;
+
+            if (!profile) {
+                // Profile not found in saved connections.
```
Benjin: When would a profile not be found in the saved connections?  Is that a situation we want to support, or should this instead be an error state?

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,186 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+            return [];
+        }
+    }
+
+    /**
+     * Connects to a server using the specified profile ID.
+     * If already connected, returns the existing ownerUri.
+     * Otherwise, establishes a new connection.
+     */
+    private async connectToServerByProfileId(profileId: string): Promise<{
```
Benjin: Is this the GUID saved as the connection's ID in the JSON, or something else?  If it's the JSON ID, then we should call this "connectionId" to keep consistent across the code.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,186 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+        errorMessage?: string;
+    }> {
+        try {
+            // Find the profile in saved connections
+            const savedConnections =
+                await this._connectionManager.connectionStore.readAllConnections();
+            const profile = savedConnections.find((conn) => {
+                const connProfile = conn as IConnectionProfile;
+                return connProfile.id === profileId;
+            }) as IConnectionProfile | undefined;
```
Benjin: Didn't all the connections get read from the store earlier?  Maybe those can be cached into a map, and we fetch the full profile from there instead of reading from disk again.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
Benjin: Maybe this block can be removed; looks like it's doing some random things, and I'm not sure why

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,186 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+            }
+
+            // Check if already connected
+            let ownerUri = this._connectionManager.getUriForConnection(profile);
+            if (ownerUri && this._connectionManager.isConnected(ownerUri)) {
+                // Connection is active - try to get databases
+                let databases: string[] = [];
+                try {
+                    databases = await this._connectionManager.listDatabases(ownerUri);
+                    this._connectionString = await this._connectionManager.getConnectionString(
```
Benjin: Seems odd to save the connection string now.  Why not get it when the user clicks "publish"?  Less time for it to hang around in memory (possible with a secret)

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,186 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                        ownerUri,
+                        true,
+                        true,
```
Benjin: Add comments for these "magic constants" so people know what they are.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,186 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                    );
+                } catch (dbError) {
+                    this.logger.warn(`Failed to list databases: ${dbError}`);
```
Benjin: getErrorMessage()

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,186 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                return {
+                    ownerUri,
+                    isConnected: true,
+                    serverName: profile.server,
+                    databases,
+                };
+            }
+
+            // Not connected - establish new connection
+            const result = await this._connectionManager.connect("", profile);
```
Benjin, line 573: should the fileURI instead be a GUID so that it doesn't risk overlapping/interfering with other connections?

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -493,6 +494,186 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+
+            if (result) {
+                ownerUri = this._connectionManager.getUriForConnection(profile);
+
+                // Try to get databases, but don't fail the connection if this errors
+                let databases: string[] = [];
+                try {
```
Benjin: Given that the database list needs to be updated whenever the user changes the dropdown, this could probably belongs in a helper method.

extensions/mssql/src/reactviews/pages/PublishProject/components/FormFieldComponents.tsx
ssreerama: I see we have a searchable dropdown component defined already, I will use it for the server, where the dropdown field is a freeform type which needs to take user input, so combobox is perfect apt for the scenario.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: If the connection is added in the connection profile, we will be seeing the prepopulated server connections since the dropdown is populated from the saved connections, the profile should always be found. this was a fallback logic i added for checking the connections when the connection is removed from connection dialog but server still have the saved connections. this can be removed and simply return the error when profile is not found. User can always make a new connection or add missing connections.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: saving the connection string is specifically for the sqlpackage command generation API call and for save publish profile. We can also get it on demand.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: You're right, it is the GUID from the connection's id field in the JSON settings. Let me rename profileId to connecitonID for no confusion.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: Actually we don't need these changes as we can get the display name from getConnectionDisplayName

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -808,10 +982,11 @@ export class PublishProjectWebViewController extends FormWebviewController<
                     // If profile has a connection string, connect in background (non-blocking)
                     if (parsedProfile.connectionString) {
                         void this.connectAndPopulateDatabases(parsedProfile.connectionString).then(
```
Benjin: Why are we using `then()` here instead of awaiting?

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -846,6 +1035,57 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                this.updateState();
+
+                const result = await this.connectToServerByProfileId(payload.profileId);
+
+                state.isConnecting = false;
+                state.isLoadingDatabases = false;
+
+                // Always update database dropdown options (clear on error, populate on success)
+                const databaseComponent = state.formComponents[PublishFormFields.DatabaseName];
+                if (databaseComponent) {
```
Benjin: when would databaseComponent not exist?

extensions/mssql/test/unit/publishProjectWebViewController.test.ts
```diff
@@ -624,6 +633,103 @@ suite("PublishProjectWebViewController Tests", () => {
...
+        );
+        (mockConnectionManager.connect as sinon.SinonStub).resolves(true);
+        (mockConnectionManager.getUriForConnection as sinon.SinonStub).returns(
+            "mssql://testserver",
+        );
+        (mockConnectionManager.isConnected as sinon.SinonStub).returns(false);
+        (mockConnectionManager.listDatabases as sinon.SinonStub).resolves(mockDatabases);
+        (mockConnectionManager.getConnectionString as sinon.SinonStub).resolves(
+            "Server=testserver;Database=master",
+        );
```
Benjin: Why do these all have `as sinon.SinonStub` casts?  I wouldn't think that be necessary.

extensions/mssql/test/unit/publishProjectWebViewController.test.ts
```diff
@@ -624,6 +633,103 @@ suite("PublishProjectWebViewController Tests", () => {
...
+        await controller.initialized.promise;
+
+        // Verify server dropdown is populated from saved connections
+        expect(controller.state.availableConnections).to.have.length(1);
+        const serverComponent = controller.state.formComponents.serverName;
+        expect(serverComponent.options).to.have.length(1);
+        expect(serverComponent.options[0].displayName).to.equal("Test Server");
+
+        // Get and call the connectToServer reducer
+        const reducerHandlers = controller["_reducerHandlers"] as Map<string, Function>;
```
Benjin: Is this `as` necessary?  Any time we do a cast like that, it potentially breaks out of the type-checker's safety, so that should only be done when required.

Double-check other spots, too.

extensions/mssql/test/unit/publishProjectWebViewController.test.ts
```diff
@@ -624,6 +633,103 @@ suite("PublishProjectWebViewController Tests", () => {
...
+            mockSavedConnections,
+        );
+        (mockConnectionManager.connect as sinon.SinonStub).resolves(false);
+        (mockConnectionManager.getUriForConnection as sinon.SinonStub).returns("");
+        (mockConnectionManager.isConnected as sinon.SinonStub).returns(false);
+
+        const controller = createTestController();
+        await controller.initialized.promise;
+
+        const reducerHandlers = controller["_reducerHandlers"] as Map<string, Function>;
```
Benjin: unnecessary casts?

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
Benjin: Yeah, let's just remove and error.  If we really want to handle that better, just subscribe to the connectionRemoved event and delete it from the dropdown here when that fires.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
Benjin: Let's get it on-demand since that could potentially contain a password; I don't think we gain anything by generating it preemptively.

Benjin:
Whenever we make UI changes, it's good practice to include before/after screenshots/videos in the PR description.  Make sure you don't post pics with PII (email, subscription ID, server URL, etc.)

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: The connection manager already handles generating a unique URI when you pass an empty string. If this is what you comment is referening to.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: When loading a publish profile, the dialog shows the parsed profile data immediately (database name, server name, etc.) without waiting for the connection to complete, but the connectinon is made in background. If we used await, the entire profile loading would block until the connection completes, which could make the UI feel unresponsive.

may be we can change this to await and make some loading experience on the server dropdown while making connection.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: this check is unneccesaary now, before I added "Loading databases" while making the serve connection instead of server dd loader and left the check here. cleaning it.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: added a gif for how we now load the profile and make server connection visibly.

ssreerama:
> Whenever we make UI changes, it's good practice to include before/after screenshots/videos in the PR description. Make sure you don't post pics with PII (email, subscription ID, server URL, etc.)

Done! usually I add the after changes video/ss, updates the before changes video.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -495,92 +494,54 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+     * Lists all saved connections from the connection store.
      * Returns all connection profiles, not just active connections.
      */
     private async listSavedConnections(): Promise<IConnectionDialogProfile[]> {
         try {
             const savedConnections =
                 await this._connectionManager.connectionStore.readAllConnections();
+            return savedConnections as IConnectionDialogProfile[];
+        } catch (error) {
+            this.logger.error(`Failed to list saved connections: ${getErrorMessage(error)}`);
```
Benjin, line 508: Should emit error telemetry here

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -495,92 +494,54 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
         try {
-            // Find the profile in saved connections
-            const savedConnections =
-                await this._connectionManager.connectionStore.readAllConnections();
-            const profile = savedConnections.find((conn) => {
-                const connProfile = conn as IConnectionProfile;
-                return connProfile.id === profileId;
-            }) as IConnectionProfile | undefined;
+            // Find the profile from cached available connections (already loaded from store)
+            const profile = this.state.availableConnections?.find(
```
Benjin: Why is `availableConnections` part of the state?  Looks like it's only used in the controller, and the options get sent to the frontend by setting them as serverComponent options.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -662,13 +603,35 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+        }
+
+        try {
+            return await this._connectionManager.getConnectionString(
+                this._connectionUri,
+                false, // includePassword
+                true, // includeApplicationName
+            );
+        } catch (error) {
+            this.logger.warn(`Failed to get connection string: ${getErrorMessage(error)}`);
```
Benjin: If it can't get a connection string, it's a total blocker, right?  In which case:

* this should be an error rather than warning
* we should emit error telemetry
* we should display the error to the user in the UI so they can see that something went wrong

Benjin: Double-check that we're doing all of these three things for other spots where an error would block the user from deploying.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -979,41 +942,46 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                        this.state.isLoadingDatabases = true;
+                        this.updateState();
+
+                        const connectionResult = await this.connectAndPopulateDatabases(
+                            parsedProfile.connectionString,
                         );
+
+                        // Clear loading indicators
+                        this.state.isConnecting = false;
+                        this.state.isLoadingDatabases = false;
```
Benjin: Instead of booleans, we should use `Status` or `ApiStatus` from `webview.ts`.  That'll allow us to track state better and provide a message if needed.

The model we should follow:
* use `ApiStatus` (+ a `state.errorMessage` property) if the error message will be shown in a generic area (like a banner at the top)
* use `Status` if the error message needs to show up in a specific place (like next to the component doing the loading)

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -1035,35 +1003,33 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
                 state.isLoadingDatabases = true;
 
                 // Update state to show loading indicator
                 this.updateState();
 
-                const result = await this.connectToServerByProfileId(payload.profileId);
+                const result = await this.connectToServerByConnectionId(payload.connectionId);
 
                 state.isConnecting = false;
                 state.isLoadingDatabases = false;
```
Benjin: I think you should pass `state` to this helper, let it own the connection and database-loading status updates (because it's performing those actions), have it display its own error (because it has the most context on the failure), and return undefined if it failed.  Then this call can return early if it gets a failure.

Take a look at how `loadAzureSubscriptions()` works in extensions\mssql\src\connectionconfig\connectionDialogWebviewController.ts for a cleaner model that keeps ownership compartmentalized.  The benefit of this compartmentalization is that it's easier to test the full cases for the helper because all the relevant side-effects (like errors). are self-contained there.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: correct, I found another state also unnecessary in the scope and cleaning that. a local variable is sufficient since they are being used in controller only.

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
ssreerama: If we can't get conection string, yes, 
- SqlPackage command generation - will fail with error shown to user
- Save Publish Profile - will fail with error shown to user

added telemetry and converted warning to error, also at couple of other places, I feel sending the telemetry would be beneficial, so making those updates too. Thanks

extensions/mssql/src/publishProject/publishProjectWebViewController.ts
```diff
@@ -1008,16 +1026,16 @@ export class PublishProjectWebViewController extends FormWebviewController<
...
+                state.loadConnectionStatus = ApiStatus.Loading;
 
                 // Update state to show loading indicator
                 this.updateState();
 
                 const result = await this.connectToServerByConnectionId(payload.connectionId);
 
-                state.isConnecting = false;
-                state.isLoadingDatabases = false;
+                state.loadConnectionStatus = result.errorMessage
```
Benjin: The error state logic should be consolidated instead of sprinkled around below.  That way, you cleanly have one block that does all the error stuff (then returns), and the rest is known to be a good state.  It's much easier to read the code and understand the different states this way.

```
if (result.errorMessage) {
  state.loadConnectionStatus;
  databaseComponent.options = [];
  state.formMessage = result.errorMessage;

  return state;
}

// all successful connection stuff together
// ...
```

PR #20985, opened 2026-01-22
Fix: creating SQL project inside already open workspace results in tasks definition in wrong spot

ssreerama:

# Issue: #20821 

## Description

This pull request refactors how SQL project build tasks are created and managed.
- Moving away from template-based task generation to a programmatic approach. 
- The changes improve reliability, cross-platform compatibility, and error handling when updating or creating `tasks.json` files for SQL projects.
- Scenarios it covers
  - When workspace exists, tasks.json will be created at workspace level, if already task.json present, then we warn the user with info that we are updating their task.json file.
  - When no workspace exists, we fall back to existing functionality that task.json will be created at project level
  - Any child projects creates under the parent project, we always update the workspace level task.json file

## Code Changes Checklist

image
TaskJson_updateExtJson_After

### DIFF
--- a/extensions/sql-database-projects/src/common/constants.ts
+++ b/extensions/sql-database-projects/src/common/constants.ts
@@ -729,7 +729,6 @@ export const unhandledMoveNode = l10n.t("Unhandled node type for move");
 
 //#region tasks.json
 export const updatingExistingTasksJson = l10n.t("A SQL Projects build task has been added to the existing tasks.json file.");
-export const sqlProjectBuildTaskLabel = 'sqlproj: Build';
 export const netCoreBuildArg = '/p:NetCoreBuild=true';
 export const systemDacpacsLocationArgPrefix = '/p:SystemDacpacsLocation=';
 export const netCoreTargetsPathArgPrefix = '/p:NETCoreTargetsPath=';


extensions/sql-database-projects/src/common/utils.ts
```diff
@@ -872,3 +872,42 @@ export function throwIfFailed(result: azdataType.ResultStatus | vscodeMssql.Resu
...
+export function createSqlProjectBuildTask(options: SqlProjectBuildTaskOptions): object {
+	return {
+		label: constants.getSqlProjectBuildTaskLabel(options.projectName),
+		type: constants.processTaskType,
+		command: constants.dotnet,
+		args: options.buildArgs,
+		problemMatcher: constants.problemMatcher,
+		isBackground: true,
+		group: {
+			kind: constants.build,
```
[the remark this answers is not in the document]
ssreerama: since the buildGroupKind is not used anywhere, removed it and using the build, since both constants are same and duplicated.

extensions/sql-database-projects/test/templates.test.ts
```diff
@@ -25,7 +25,7 @@ suite('Templates', function (): void {
 	test('Should load all templates from files', async function (): Promise<void> {
 		// check expected counts
 
-		const numScriptObjectTypes = 17;
+		const numScriptObjectTypes = 16;
```
Benjin, line 28: What changed here?

extensions/sql-database-projects/src/controllers/projectController.ts
```diff
@@ -276,21 +273,24 @@ export class ProjectsController {
 				const existingContent = await fs.readFile(tasksJsonPath, 'utf8');
 				const existingTasksJson = JSON.parse(existingContent);
 
-				// Ensure tasks array exists
-				if (!existingTasksJson.tasks) {
+				// Ensure tasks array exists and is actually an array
```
Benjin: If this exists but isn't an array, that means that the user (or something else) has edited the json to something unexpected/incorrect.  Maybe we should throw an error (something like "invalid format: expected 'tasks' to be an array) instead of overwriting whatever changes they had in case they're intentional, or a simple typo.

extensions/sql-database-projects/src/controllers/projectController.ts
```diff
@@ -300,16 +300,60 @@ export class ProjectsController {
...
+	/**
+	 * Creates a build task object for the given project
+	 */
+	private createBuildTaskForProject(project: ISqlProject, projectName: string, configureDefaultBuild: boolean): object {
+		// Use forward slashes for cross-platform compatibility (dotnet accepts both)
+		const projectFilePathNormalized = utils.getPlatformSafeFileEntryPath(project.projectFilePath);
+
+		// Build args array for process execution (avoids shell escaping issues)
+		const buildArgs: string[] = [constants.build, projectFilePathNormalized];
+		const isSdkStyle = project.sqlProjStyleName === 'SdkStyle';
```
Benjin: This doesn't need to be saved as a const.

Benjin: Unless `sqlProjStyleName` is a string union, then `"SdkStyle"` should instead be saved as a const somewhere to avoid typo risk.

Benjin:
> image

Where is this message defined?  I think a better way to word it would be:

> A SQL Projects build task [will be | has been] added to the existing file.

extensions/sql-database-projects/test/templates.test.ts
ssreerama: we are removing the physical tasks.json template file from the templated folder. that reduced the count.

that file was used to create the task when default build was selected but which is now programmatically created by vscode tasks.

ssreerama:
> > image
> 
> Where is this message defined? I think a better way to word it would be:
> 
> > A SQL Projects build task [will be | has been] added to the existing file.

Updated the message. It was reusing an existing message that wasn’t previously used anywhere, which is why it didn’t appear in the changes.

extensions/sql-database-projects/src/controllers/projectController.ts
ssreerama: No the string is not an union, i have moved it to constants and updated the similar references

extensions/sql-database-projects/src/controllers/projectController.ts
ssreerama: Correct. If the tasks.json file is not supported, it may be corrupted. We throw an error in that case to notify the user about the build task issue, while still allowing project creation to continue.

ssreerama:
Updated the gif and image with after addressing the proposed changes.

extensions/sql-database-projects/src/controllers/projectController.ts
```diff
@@ -276,40 +273,90 @@ export class ProjectsController {
...
+
+					// Merge the new task into existing tasks
+					existingTasksJson.tasks.push(newTask);
 
 					// Write back the merged tasks.json
 					await fs.writeFile(tasksJsonPath, JSON.stringify(existingTasksJson, null, '\t'), 'utf8');
 
 					// Show notification to user
 					void vscode.window.showInformationMessage(constants.updatingExistingTasksJson);
 				}
```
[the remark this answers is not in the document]
ssreerama: That is the correct, we are adding default build to each project, if tasks.json file has multiple default build then all those projects will be shown when user select build commands.

extensions/sql-database-projects/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to the SQL Database Projects extension will be documented in
 
 ## [1.5.7] - 2026-02-27
 
+- Added five new SQL object templates: Schema, Table-Valued Function, Trigger, Database Trigger, and Sequence.
+- Fixed an issue where the SQL project build task was being created at the project level instead of the workspace level.
 - Fixed an issue where system dacpac files were missing from the BuildDirectory, causing build failures for projects with system database references.
```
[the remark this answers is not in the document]
ssreerama: its intentional change

PR #20997, opened 2026-01-23
Creating initial diff panel for schema Designer.

aasimkhan30:

## Description

It creates a diff panel for schema designer which contains changes made actively in a session. 

Overall Design:
image

Panel:
image

Each change item has reveal and undo:
image

Search:
image

## Code Changes Checklist

### DIFF
--- a/localization/xliff/vscode-mssql.xlf
+++ b/localization/xliff/vscode-mssql.xlf
@@ -161,11 +161,11 @@
     <trans-unit id="++CODE++fe80d12c71bbc300158260094cf676c1a9f3546b792f027c84e7a91d9fc75b20">
       <source xml:lang="en">Advanced Publish Options</source>
     </trans-unit>
-    <trans-unit id="++CODE++a52ace420f2175d08b1577a1bea5445e36801229c074ef9ed6c55a73401fd9c2">
-      <source xml:lang="en">All</source>
-    </trans-unit>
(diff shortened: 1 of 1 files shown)


lewis-sanchez:
Is it possible to break this down into smaller PR's? It's a big PR.

extensions/mssql/src/reactviews/pages/SchemaDesigner/diff/diffUtils.ts
```diff
@@ -0,0 +1,489 @@
...
+ * Foreign Keys (grouped under source table):
+ * - Added: New foreign keys
+ * - Deleted: Removed foreign keys
+ * - Modified: FKs with changes to name, columns, referenced table/columns, or actions
+ *
+ * @param oldSchema - The original schema (baseline)
+ * @param newSchema - The current schema (with modifications)
+ * @returns ChangesSummary with changes grouped by table
+ */
+export function calculateSchemaDiff(
```
manujoseph85, line 201: We should do some analysis on the performance of calculateSchemaDiff on larg-ish databases - since we are doing the full diff for every change.

If the LLM is submitting a series of changes in quick succession, would it cause any issues? Should we cancel previous processing since the last one will always win?

We can also explore an approach similar to our undo/redo stack, where we can add-on each change to the SchemaChangesSummary instead of doing the full diff every time.

PR #21001, opened 2026-01-23
Fixing some unit test failure and mis-stubs

Benjin:
## Description

Fixing a few spots where tests were not stubbed correctly, which was causing intermittent unit test failures

## Code Changes Checklist


allancascante:
Can we add this as an instruction for copilot in AGENTS.md under unit test folder?

Benjin:
@allancascante

> Can we add this as an instruction for copilot in AGENTS.md under unit test folder?

There's already one; it does a pretty good job following those instructions, but not perfect, so they can certainly be improved.  My bigger question is why these weren't caught when people were running the tests offline before checking them in, or when the pipelines would run.  Felt like these broke somewhat randomly.

PR #21010, opened 2026-01-26
Profiler details panel

allancascante:
## Description

Used copilot to generate the details panel for the profiler, so when a user clicks on a row in the grid, a panel will be loaded with the details, there will be two sections; one to show any text from the trace and displayed on a read-only editor with SQL syntax highlighting, including open in editor and copy to clipboard buttons; and another tab displaying the details of the event in the grid showing two columns one with the name and one with the value.

image

image

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/profiler/profilerWebviewController.ts
+++ b/extensions/mssql/src/profiler/profilerWebviewController.ts
@@ -465,8 +465,7 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
         // Handle export to CSV request from webview
         // Generate CSV from the RingBuffer (source of truth) rather than from grid data
         // This ensures ALL events in the session buffer are exported
-        this.registerReducer("exportToCsv", (state, payload: { suggestedFileName: string }) => {
-            // Generate CSV from buffer if we have a session
+        this.onNotification(ProfilerNotifications.ExportToCsv, async () => {
             if (
                 this._currentSession &&
                 this._currentSession.events.size > 0 &&
@@ -474,27 +473,27 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
             ) {
                 const allEvents = this._currentSession.events.getAllRows();
                 const viewConfig = this.state.viewConfig;
+
+                // Generate suggested file name with timestamp
+                const sessionName =
+                    this._currentSession.sessionName || LocProfiler.defaultExportFileName;
+                const timestamp = generateExportTimestamp();
+                const suggestedFileName = `${sessionName}_${timestamp}`;
+
                 if (this._eventHandlers.onExportToCsv) {
-                    // Request stream and write CSV (async, but reducer returns synchronously)
-                    void this._eventHandlers
-                        .onExportToCsv(payload.suggestedFileName)
-                        .then(async (stream) => {
-                            if (stream) {
-                                await this.generateCsvFromEvents(stream, allEvents, viewConfig);
-                                stream.end(); // Close the stream to trigger 'finish' event
-                            }
-                        })
-                        .catch((error) => {
-                            const errorMessage =
-                                error instanceof Error ? error.message : String(error);
-                            console.error("Failed to export profiler session to CSV:", error);
-                            void vscode.window.showErrorMessage(
-                                LocProfiler.exportFailed(errorMessage),
-                            );
-                        });
+                    try {
+                        const stream = await this._eventHandlers.onExportToCsv(suggestedFileName);
+                        if (stream) {
+                            await this.generateCsvFromEvents(stream, allEvents, viewConfig);
+                            stream.end();
+                        }
+                    } catch (error) {
+                        const errorMessage = error instanceof Error ? error.message : String(error);
+                        console.error("Failed to export profiler session to CSV:", error);
+                        void vscode.window.showErrorMessage(LocProfiler.exportFailed(errorMessage));
+                    }
                 }
             }
-            return state;
         });
     }
 
--- a/extensions/mssql/src/reactviews/pages/Profiler/profiler.tsx
+++ b/extensions/mssql/src/reactviews/pages/Profiler/profiler.tsx
@@ -31,7 +31,6 @@ import {
     RowsRemovedParams,
 } from "../../../sharedInterfaces/profiler";
 import { ColorThemeKind } from "../../../sharedInterfaces/webview";
-import { locConstants } from "../../common/locConstants";
 import { useVscodeWebview } from "../../common/vscodeWebviewProvider";
 import "@slickgrid-universal/common/dist/styles/css/slickgrid-theme-default.css";
 
@@ -118,7 +117,6 @@ export const Profiler: React.FC = () => {
     const selectedEvent = useProfilerSelector((s) => s.selectedEvent);
     const isReadOnly = useProfilerSelector((s) => s.isReadOnly ?? false);
     const xelFileName = useProfilerSelector((s) => s.xelFileName);
-    const sessionName = useProfilerSelector((s) => s.sessionName);
 
     const {
         pauseResume,
@@ -524,20 +522,11 @@ export const Profiler: React.FC = () => {
     );
     /**
      * Handle export to CSV request.
-     * Sends export request to extension with suggested filename.
-     * The extension generates CSV from the session's RingBuffer (source of truth)
-     * to ensure ALL events are exported, not just those loaded in the grid.
+     * The extension generates the filename and CSV from the session's RingBuffer.
      */
     const handleExportToCsv = useCallback(() => {
-        // Generate suggested file name with timestamp (YYYY-MM-DD-HH-mm-ss)
-        const timestamp = new Date().toISOString().slice(0, 19).replace(/[:T]/g, "-");
-        const suggestedFileName = sessionName
-            ? `${sessionName}_${timestamp}`
-            : `${locConstants.profiler.defaultExportFileName}_${timestamp}`;
-
-        // Send to extension host - extension will generate CSV from ring buffer
-        exportToCsv(suggestedFileName);
-    }, [sessionName, exportToCsv]);
+        exportToCsv();
+    }, [exportToCsv]);
 
     return (
         <div className={classes.profilerContainer}>
--- a/extensions/mssql/src/reactviews/pages/Profiler/profilerStateProvider.tsx
+++ b/extensions/mssql/src/reactviews/pages/Profiler/profilerStateProvider.tsx
@@ -43,7 +43,7 @@ export interface ProfilerRpcMethods {
     /** Close the embedded details panel */
     closeDetailsPanel: () => void;
     /** Export events to CSV file */
-    exportToCsv: (suggestedFileName: string) => void;
+    exportToCsv: () => void;
 }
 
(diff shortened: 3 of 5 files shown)


extensions/mssql/src/profiler/profilerSession.ts
```diff
@@ -566,12 +566,22 @@ export class ProfilerSession {
         const eventNumber =
             event.eventNumber ?? eventNumberFromValues ?? this._eventNumberCounter++;
 
+        // Extract text data from various possible field names used by XEvents
+        // Common field names: sql_text, statement, batch_text, options_text, query_text
+        const textData =
```
lewis-sanchez, line 571: The field names in event can only be set one at a time right? For example, if "sql_text" had a value then the other field names wouldn't have any values?

extensions/mssql/src/reactviews/pages/Profiler/profilerDetailsPanel.tsx
```diff
@@ -0,0 +1,402 @@
...
+                    </Tab>
+                    <Tab
+                        value={DetailsPanelTab.Details}
+                        aria-label={loc.detailsTabAriaLabel}
+                        aria-selected={activeTab === DetailsPanelTab.Details}>
+                        {loc.detailsTab}
+                    </Tab>
+                </TabList>
+
+                <Toolbar className={classes.toolbar} aria-label="Details panel actions">
```
lewis-sanchez: Localize the aria-label here

extensions/mssql/docs/profiler-embedded-details-panel.md
```diff
@@ -0,0 +1,396 @@
+# Profiler Embedded Details Panel Implementation
+
+## Overview
```
Benjin: This file should not be checked in

extensions/mssql/src/profiler/profilerSession.ts
Benjin: Good catch. Should this be a priority (where the first one found is used) or a string.join across multiple fields?

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -275,6 +276,86 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
...
+            "openInEditor",
+            async (state, payload: { textData: string; eventName?: string }) => {
+                await this.openTextInEditor(payload.textData);
+                return state;
+            },
+        );
+
+        // Handle Copy to Clipboard request from embedded details panel
+        this.registerReducer("copyToClipboard", async (state, payload: { text: string }) => {
+            await vscode.env.clipboard.writeText(payload.text);
```
Benjin: @aasimkhan30 I remember you fought against copying to clipboard a fair bit in query results; is this the best way to do it?

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -275,6 +276,86 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
                 return state;
             },
         );
+
+        // Handle row selection from webview - update state with selected event details
+        this.registerReducer("selectRow", (state, payload: { rowId: string }) => {
+            const selectedEvent = this.handleRowSelection(payload.rowId);
+            return {
```
Benjin: Rather than returning a new state object, just set the `selectedEvent` property

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -275,6 +276,86 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
...
+                ...state,
+                selectedEvent,
+            };
+        });
+
+        // Handle Open in Editor request from embedded details panel
+        this.registerReducer(
+            "openInEditor",
+            async (state, payload: { textData: string; eventName?: string }) => {
+                await this.openTextInEditor(payload.textData);
```
Benjin: Reducers are intended for actions that will change state (and therefore trigger a re-render of the frontened).  `openInEditor` and `copyToClipboard` don't edit the state at all.

I think you want `this.onNotification()`

extensions/mssql/src/profiler/ringBuffer.ts
```diff
@@ -227,4 +227,20 @@ export class RingBuffer<T extends IndexedRow> {
         const bufferIndex = (this._head + index) % this._capacity;
         return this._rows[bufferIndex];
     }
+
+    /**
+     * Finds a row by its ID.
+     * @param id - The unique ID of the row to find
+     * @returns The row with the matching ID, or undefined if not found
+     */
+    findById(id: string): T | undefined {
```
Benjin, line 248: this could be pretty slow... let's not change it yet, but maybe we should be look into mapping from ID to row (and prune map contents as they're overwritten in the ring buffer)?

extensions/mssql/src/profiler/profilerSession.ts
allancascante: checking ADS code this should be a priority. So meaning if sql_text is set we use that only.

extensions/mssql/src/profiler/ringBuffer.ts
allancascante: I did the change to use an index

extensions/mssql/src/profiler/profilerWebviewController.ts
allancascante: this is the standard VS Code clipboard API and is used throughout the codebase (table designer, schema designer, table explorer all use it the same way), Query results needs the SQL Tools Service because the data lives server-side (large result sets, pagination). The profiler's case is simpler — the text is already available client-side in the details panel.

extensions/sql-database-projects/test/projectController.test.ts
```diff
@@ -21,1447 +21,2441 @@ import * as constants from "../src/common/constants";
...
-import { FolderNode, FileNode } from '../src/models/tree/fileFolderTreeItem';
-import { BaseProjectTreeItem } from '../src/models/tree/baseTreeItem';
-import { AddDatabaseReferenceDialog } from '../src/dialogs/addDatabaseReferenceDialog';
-import { IDacpacReferenceSettings } from '../src/models/IDatabaseReferenceSettings';
-import { CreateProjectFromDatabaseDialog } from '../src/dialogs/createProjectFromDatabaseDialog';
-import { ImportDataModel } from '../src/models/api/import';
-import { EntryType, ItemType, SqlTargetPlatform } from 'sqldbproj';
-import { FileProjectEntry } from '../src/models/projectEntry';
-import { SystemDatabase } from 'mssql';
+import { SqlDatabaseProjectTreeViewProvider } from "../src/controllers/databaseProjectTreeViewProvider";
```
lewis-sanchez, line 24: Are these formatting changes? The file is in the sql-database-projects directory and not in mssql. Did you have to make changes here for the profiler feature?

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -275,6 +276,84 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
                 return state;
             },
         );
+
+        // Handle row selection from webview - update state with selected event details
+        this.registerReducer("selectRow", (state, payload: { rowId: string }) => {
+            state.selectedEvent = this.handleRowSelection(payload.rowId);
+            return state;
```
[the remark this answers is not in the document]
Benjin: This is incorrect.  Both are "correct", but simply setting properties on the existing state object is (slightly) more efficient

extensions/sql-database-projects/test/projectController.test.ts
allancascante: linter made updates, just reverted them to main

extensions/mssql/src/reactviews/common/locConstants.ts
```diff
@@ -2086,6 +2086,37 @@ export class LocConstants {
...
+
+            // Details panel
+            detailsPanel: {
+                textTab: l10n.t("Text"),
+                detailsTab: l10n.t("Details"),
+                openInEditor: l10n.t("Open in Editor"),
+                copy: l10n.t("Copy"),
+                maximize: l10n.t("Maximize"),
+                restore: l10n.t("Restore"),
+                close: l10n.t("Close"),
```
Benjin: Things like "copy" and "close" probably already exist in the common bundle.  Check there before adding duplicates

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -307,6 +461,41 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
                 selectedEvent: undefined,
             };
         });
+
+        // Handle export to CSV request from webview
+        // Generate CSV from the RingBuffer (source of truth) rather than from grid data
+        // This ensures ALL events in the session buffer are exported
+        this.registerReducer("exportToCsv", (state, payload: { suggestedFileName: string }) => {
```
Benjin: since state isn't changing, this shouldn't be a reducer.  Probably should be `onNotification()`.  By being a reducer, this can trigger unnecessary re-renders.

extensions/mssql/src/reactviews/pages/Profiler/profiler.tsx
```diff
@@ -517,6 +522,22 @@ export const Profiler: React.FC = () => {
         },
         [handleRowSelection],
     );
+    /**
+     * Handle export to CSV request.
+     * Sends export request to extension with suggested filename.
+     * The extension generates CSV from the session's RingBuffer (source of truth)
+     * to ensure ALL events are exported, not just those loaded in the grid.
+     */
+    const handleExportToCsv = useCallback(() => {
```
Benjin, line 527: Why do this in React?  Seems like this could be done entirely in the webview controller, which would reduce the size of the webview bundle -> faster loading.

extensions/mssql/src/reactviews/pages/Profiler/profiler.tsx
```diff
@@ -517,6 +522,22 @@ export const Profiler: React.FC = () => {
...
+        // Generate suggested file name with timestamp (YYYY-MM-DD-HH-mm-ss)
+        const timestamp = new Date().toISOString().slice(0, 19).replace(/[:T]/g, "-");
```
Benjin: It's good practice to avoid regexes unless they're necessary due to perf and debugging concerns.  This one's simple enough that it's fine, but just a note for the future.

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -465,36 +465,35 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
...
-                        });
+                    try {
+                        const stream = await this._eventHandlers.onExportToCsv(suggestedFileName);
+                        if (stream) {
+                            await this.generateCsvFromEvents(stream, allEvents, viewConfig);
+                            stream.end();
+                        }
+                    } catch (error) {
+                        const errorMessage = error instanceof Error ? error.message : String(error);
+                        console.error("Failed to export profiler session to CSV:", error);
```
Benjin, line 492: this.logger.error

PR #21020, opened 2026-01-27
Adding copilot-instructions.md

allancascante:
## Description

Adding copilot-instructions.md referencing existing AGENT.md files.

## Code Changes Checklist

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,70 @@
+# GitHub Copilot – Repository Instructions (vscode-mssql)
+
+These instructions apply to Copilot Chat and the Copilot coding agent when working in this repository.
+**Goal:** keep changes consistent with our build, test, and review practices; and ensure the coding agent leverages the most relevant `AGENTS.md` for the area it is modifying.
+
+---
+
+## How to use AGENTS.md in this repo
+
+- **Always consult the nearest `AGENTS.md` when planning or editing files.**
+  - If you're changing files inside `extensions/<name>/…`, load and follow `extensions/<name>/AGENTS.md` for that subtree.
+  - If multiple `AGENTS.md` files exist in parent/child folders, **the closest one to the file being changed wins**. Prefer the most specific guidance and avoid conflicting rules.
+  - If no subtree file exists, fall back to the repository root `AGENTS.md` (if present).
+    _Rationale: AGENTS.md hosts agent‑oriented steps (setup, build, test, review) and is designed to be read by coding agents._
+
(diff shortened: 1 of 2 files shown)


.github/copilot-instructions.md
aasimkhan30, line 1: Are you sure we need these instructions? According to docs this should already happen. 

https://code.visualstudio.com/docs/copilot/customization/custom-instructions#_use-an-agentsmd-file

allancascante: I tried with my local vs code chat window and those are not being picked consistently; to make it work I had to add these. In the copilot chat, I clicked the configuration and click 'Generate Chat Instructions'. I think the trick is what it mentions on that doc: 'You can define custom instructions for all agents in an AGENTS.md Markdown file at the root(s) of the workspace. VS Code applies the instructions in this file automatically to all chat requests within this workspace.' Given the changes we did to the repo moving AGENTS.md into a sub folder caused it to not be picked up. Also AGENTS.md don't have: applyTo

When I asked in copilot chat in main branch why those are not being applied consistently, it mentioned that is missing; adding the instructions allow for copilot chat to read those instructions consistently.

aasimkhan30: I think they are not enabled by default. Did you try enabling it in the settings?

https://code.visualstudio.com/docs/copilot/customization/custom-instructions#_use-multiple-agentsmd-files-experimental
image

aasimkhan30: If the setting works, maybe we can just add it to .vscode/settings.json to enable nested agents for everyone.

extensions/mssql/AGENTS.md
```diff
@@ -299,3 +300,4 @@ When reviewing webview-related PRs, verify:
 - [ ] `queueMicrotask` used for non-visual immediate execution
 - [ ] No `setTimeout` during webview initialization/startup
 - [ ] Consider hidden/backgrounded webview behavior for any timing-sensitive code
+- [ ] No separate style files (.css) added, views use buildStyle at the end of the file
```
aasimkhan30, line 303: Are you referring to `makeStyles` here? I don’t see a buildStyle API in the repo or in Fluent UI v9. 
Also, while makeStyles should be the default, disallowing .css files entirely feels a bit restrictive. There are cases where we need to override Fluent UI internal styles to better match VS Code theming or style non-React libraries (SlickGrid), which aren’t always achievable via tokens or makeStyles.
Perhaps this guideline could be framed as a preference with documented exceptions, and also call out known limitations of JS-based styling (e.g. complex selectors, pseudo-elements, or Fluent internals not exposed via props).

PR #21043, opened 2026-01-29
[Application QuickStart] [Backend] DAB Config file generation

manujoseph85:
## Description

This PR adds support for DAB config file generation from DAB settings

- Added dabConfigFileBuilder to build the DAB json config file
  - DabRuntimeConfig captures runtime settings including what API types are enabled
  - DabEntityOutput will capture the entity endpoints depending on the operation selection and advanced entity configuration
- Added DAB request handlers to SchemaDesignerWebviewController to handle DAB config file generation
  - Dab.GenerateConfigRequest
  - Dab.OpenConfigInEditorNotification
  - Dab.CopyConfigNotification
- Added DabDefinitionsPanel to show the generated config to the user. Panel will be shown when user clicks 'Generate Config' button
  - We need to add a new language parameter to DesignerDefinitionPane since the panel had 'sql' hardcoded. DAB config file will be json
- dabConfigFileBuilder.test.ts and dabService.test.ts has unit tests for DAB config file generation

Also added MCP as an API type option in DAB toolbar. This is the new API type that was added to the specs.

**Capture of the config generation and config panel**

DabConfigGeneration

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/test/unit/schemaDesignerWebviewController.test.ts
+++ b/extensions/mssql/test/unit/schemaDesignerWebviewController.test.ts
@@ -12,6 +12,7 @@ import * as jsonRpc from "vscode-jsonrpc/node";
 import { SchemaDesignerWebviewController } from "../../src/schemaDesigner/schemaDesignerWebviewController";
 import VscodeWrapper from "../../src/controllers/vscodeWrapper";
 import { SchemaDesigner } from "../../src/sharedInterfaces/schemaDesigner";
+import { Dab } from "../../src/sharedInterfaces/dab";
 import { TreeNodeInfo } from "../../src/objectExplorer/nodes/treeNodeInfo";
 import MainController from "../../src/controllers/mainController";
 import {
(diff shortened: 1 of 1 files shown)


extensions/mssql/l10n/bundle.l10n.json
```diff
@@ -746,8 +746,8 @@
     "Read": "Read",
     "Update": "Update",
     "Entity name used in API routes": "Entity name used in API routes",
-    "View Config": "View Config",
-    "Generate & Run": "Generate & Run",
+    "Generate Config": "Generate Config",
```
aasimkhan30, line 749: View Config?

extensions/mssql/src/reactviews/pages/SchemaDesigner/dab/dabToolbar.tsx
```diff
@@ -114,15 +121,16 @@ export function DabToolbar() {
                         appearance="subtle"
                         icon={<FluentIcons.DocumentCopy16Regular />}
                         size="small"
-                        title={locConstants.schemaDesigner.viewConfig}>
-                        {locConstants.schemaDesigner.viewConfig}
+                        title={locConstants.schemaDesigner.generateConfig}
+                        onClick={() => void generateDabConfig()}>
```
aasimkhan30, line 125: Do you know any reason why we are not live updating dab config? It doesn't seems very compute expensive?

aasimkhan30:
Looks good. I have a few questions but the PR is fine to check-in.

extensions/mssql/src/reactviews/pages/SchemaDesigner/dab/dabToolbar.tsx
manujoseph85: Mainly because we want the user to be able to edit the config after its generated. But this part is still undecided, we can totally make it 'View config' and update it live.

extensions/mssql/l10n/bundle.l10n.json
manujoseph85: Currently, we are generating only when the user clicks the button. But we could change this in the future.

PR #21074, opened 2026-02-02
Profiler start session on load

allancascante:
## Description

When starting a profiler session, user must select a template and session to start; added also a way to detect if the session is for Azure to filter the templates that can be selected to only those available for Azure. Also added an error if the connection is for Fabric to let the user know it is not supported. Also, for Azure display the database names if the connection doesn't include one so the user can pick the database to be used when starting the session, fixed an error with the Azure templates and added logic to start the session selected if it exists, and to create it if it doesn't.

## Code Changes Checklist


extensions/mssql/src/profiler/profilerController.ts
```diff
@@ -55,25 +61,55 @@ export class ProfilerController {
         );
 
         try {
+            // Check server type and handle accordingly
+            const serverTypes = getServerTypes(connectionProfile);
+            this._logger.verbose(`Server types detected: ${serverTypes.join(", ")}`);
+
+            // Determine engine type based on server type
+            this._currentEngineType = serverTypes.includes(ServerType.Azure)
```
aasimkhan30, line 69: Please move this to some sort of shared util.

extensions/mssql/src/profiler/profilerController.ts
```diff
@@ -88,6 +124,87 @@ export class ProfilerController {
     // Private Methods
     // ============================================================
 
+    /**
+     * Checks if a database is a system database.
+     * @param databaseName - The name of the database to check
+     * @returns true if the database is a system database
+     */
+    private isSystemDatabase(databaseName: string | undefined): boolean {
+        if (!databaseName) {
```
aasimkhan30, line 133: Nice to be in some shared utility.

extensions/mssql/src/profiler/profilerController.ts
```diff
@@ -88,6 +124,87 @@ export class ProfilerController {
...
+        return SYSTEM_DATABASES.includes(databaseName.toLowerCase());
+    }
+
+    /**
+     * Ensures a user database is selected for Azure SQL connections.
+     * If no database or a system database is selected, prompts the user to select one.
+     * @param connectionProfile - The connection profile to check/update
+     * @returns The connection profile with a user database, or undefined if cancelled
+     */
+    private async ensureAzureDatabaseSelected(
```
aasimkhan30, line 145: It will be nice to do some sort of plug and play database selector. I know we are already doing this at multiple place.

extensions/mssql/src/profiler/profilerController.ts
allancascante: https://github.com/microsoft/vscode-mssql/issues/21114

aasimkhan30 (approved):

PR #21104, opened 2026-02-05
Added Flat File Dialog

laurenastrid1:

## Description

Added the flat file dialog from ADS

flatFile

image
image

## Code Changes Checklist

-   [X] New or updated **unit tests** added

### DIFF
--- a/localization/xliff/vscode-mssql.xlf
+++ b/localization/xliff/vscode-mssql.xlf
@@ -3194,11 +3194,11 @@
     <trans-unit id="++CODE++ac93cb35e24190764d9718b44e0c847f9e06a22105b20e4f8ca8ef96ed959775">
       <source xml:lang="en">Location (Workspace)</source>
     </trans-unit>
-    <trans-unit id="++CODE++d837c7bb1f3941a813c25b07944c661318596b93902745a07c356400c2450fd7">
-      <source xml:lang="en">Looking for Azure Data Studio key bindings, like F5 to execute queries?</source>
-    </trans-unit>
     <trans-unit id="++CODE++1aa90f6c7d9e7a123d71457bdbe1f733cbe0170eb978723ecf13717f64403cda">
       <source xml:lang="en">Location of the file to be imported</source>
+    </trans-unit>
+    <trans-unit id="++CODE++d837c7bb1f3941a813c25b07944c661318596b93902745a07c356400c2450fd7">
+      <source xml:lang="en">Looking for Azure Data Studio key bindings, like F5 to execute queries?</source>
     </trans-unit>
     <trans-unit id="++CODE++53f13ae99ed53bd346eb8e1c8cefb7ef8260683b50401caf101360967ea052aa">
       <source xml:lang="en">MCP</source>


extensions/mssql/src/controllers/mainController.ts
```diff
@@ -1651,6 +1669,38 @@ export default class MainController implements vscode.Disposable {
...
 
+        this._context.subscriptions.push(
+            vscode.commands.registerCommand(
+                Constants.cmdFlatFileImport,
+                async (node: ConnectionNode) => {
+                    const connectionUri = this.connectionManager.getUriForConnection(
+                        node.connectionProfile,
+                    );
+
+                    const databases = await this.connectionManager.listDatabases(connectionUri);
```
Benjin, line 1433: By preloading the databases asynchronously, there's a risk that this call takes a while and then the dialog takes a long time to open.  Probably better to pass in the connection Uri to the FFI controller, and have that display a (hopefully brief) loading state as it tries to fetch the databases.

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
```
Benjin, line 4: `flatFileImportWebviewController.ts` for consistency

extensions/mssql/src/controllers/flatFileImportController.ts
```diff
@@ -0,0 +1,411 @@
...
+/**
+ * Controller for the Add Firewall Rule dialog
+ */
+export class FlatFileImportController extends FormWebviewController<
+    FlatFileImportFormState,
+    FlatFileImportState,
+    FlatFileImportFormItemSpec,
+    FlatFileImportReducers
+> {
+    public readonly IMPORT_FILE_TYPES = ["csv", "txt"];
```
Benjin: Can be a `const` array outside of the class so it's only defined once.

extensions/mssql/src/controllers/flatFileImportController.ts
```diff
@@ -0,0 +1,411 @@
...
+    // Default form action ; this is used as part of the form reducer to handle default logic
+    private baseFormActionReducer = this["_reducerHandlers"].get("formAction");
```
Benjin: Maybe clarify in the ocmment that this is being saved because you're overriding the `formAction` reducer below?

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
...
+export class FlatFileImportController extends FormWebviewController<
+    FlatFileImportFormState,
+    FlatFileImportState,
+    FlatFileImportFormItemSpec,
+    FlatFileImportReducers
+> {
+    public readonly IMPORT_FILE_TYPES = ["csv", "txt"];
+    // Default form action ; this is used as part of the form reducer to handle default logic
+    private baseFormActionReducer = this["_reducerHandlers"].get("formAction");
+    constructor(
```
Benjin, line 46: Nit: leave a blank line before block definitions (functions, constructors) for readability.

extensions/mssql/src/controllers/flatFileImportController.ts
```diff
@@ -0,0 +1,411 @@
...
+            {
+                title: Loc.FlatFileImport.flatFileImportTitle,
+                viewColumn: vscode.ViewColumn.One,
+                iconPath: {
+                    light: vscode.Uri.joinPath(context.extensionUri, "media", "database_light.svg"),
+                    dark: vscode.Uri.joinPath(context.extensionUri, "media", "database_dark.svg"),
+                },
+            },
+        );
+        void this.initialize();
```
Benjin: If we're doing some sort of async intialization, there should be a flag that other parts can check to confirm that initialization is complete.  See how addFirewallRuleWebviewController.ts does it (Deferred object, thenable with catch, etc.)

extensions/mssql/src/controllers/flatFileImportController.ts
```diff
@@ -0,0 +1,411 @@
...
+    public readonly IMPORT_FILE_TYPES = ["csv", "txt"];
+    // Default form action ; this is used as part of the form reducer to handle default logic
+    private baseFormActionReducer = this["_reducerHandlers"].get("formAction");
+    constructor(
+        context: vscode.ExtensionContext,
+        vscodeWrapper: VscodeWrapper,
+        private client: SqlToolsServiceClient,
+        private connectionManager: ConnectionManager,
+        private provider: FlatFileProvider,
+        private node: ConnectionNode,
```
Benjin: I think this should take in a connection profile or connection ID instead.  Coupling this to an OE node reduces our flexibility in the future.

extensions/mssql/src/controllers/flatFileImportController.ts
```diff
@@ -0,0 +1,411 @@
...
+        void this.initialize();
+    }
+
+    /**
+     * Initialize the controller
+     */
+    private async initialize(): Promise<void> {
+        // Set database names for dropdown
+        this.state.serverName = this.node.connectionProfile.server;
+        this.state.formState.databaseName = ObjectExplorerUtils.getDatabaseName(this.node);
```
Benjin: Instead of using the node, I think we can take in a connection profile and (optionally) database name instead.  More flexibility with invoking this dialog that way.

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
...
+            value: db,
+        }));
+
+        // If database name is not set, set it to the first database in the list
+        if (!this.state.formState.databaseName) {
+            this.state.formState.databaseName = this.databases[0];
+        }
+
+        // Load schemas for the default/initially selected database
+        await this.handleLoadSchemas();
```
Benjin, line 126: What happens if this call throws?  Check out how other dialogs with async initialization handle that.

extensions/mssql/src/controllers/flatFileImportController.ts
```diff
@@ -0,0 +1,411 @@
...
+        }
+
+        // Load schemas for the default/initially selected database
+        await this.handleLoadSchemas();
+
+        this.registerRpcHandlers();
+        this.state.loadState = ApiStatus.Loaded;
+        this.updateState();
+
+        sendActionEvent(TelemetryViews.FlatFile, TelemetryActions.StartFlatFile);
```
Benjin: Maybe "initialize" is a better action name?  "Start" could be interpreted as starting an import.

extensions/mssql/src/controllers/flatFileImportController.ts
```diff
@@ -0,0 +1,411 @@
...
+     * Register reducers for handling actions from the webview
+     */
+    private registerRpcHandlers(): void {
+        this.registerReducer("formAction", async (state, payload) => {
+            // Reload db schemas if database name changed
+            if (payload.event.propertyName === "databaseName") {
+                void this.handleLoadSchemas();
+            }
+            // Call the default form action reducer to handle form state updates and validation
+            return this.baseFormActionReducer(state, payload);
```
Benjin: Check out `afterSetFormProperty()` in formWebviewController.  You should be able to define that in this extending class instead of overwriting the original formAction reducer.

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
...
+            const params: ProseDiscoveryParams = {
+                filePath: payload.filePath,
+                tableName: payload.tableName,
+                schemaName: payload.schemaName,
+            };
+            try {
+                state.tablePreview = await this.provider.sendProseDiscoveryRequest(params);
+                state.tablePreviewStatus = ApiStatus.Loaded;
+            } catch (error) {
+                state.errorMessage = Loc.FlatFileImport.fetchTablePreviewError;
```
Benjin, line 149: Would the types of errors you get here be useful to capture in telemetry and/or logging?

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
...
+                });
+                // Check result for errors
+                if (!insertDataResult.result.success) {
+                    throw new Error(insertDataResult.result.errorMessage);
+                }
+                state.importDataStatus = ApiStatus.Loaded;
+
+                sendActionEvent(TelemetryViews.FlatFile, TelemetryActions.ImportFile);
+            } catch (error) {
+                state.errorMessage = Loc.FlatFileImport.importFailed;
```
Benjin, line 214: Worth logging as well

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
...
+                    database: state.formState.databaseName,
+                });
+                const connectionString = await this.connectionManager.getConnectionString(
+                    connDetails,
+                    true,
+                    true,
+                );
+
+                // Set other default params for the import request
+                const batchSize = 1000; // default batch size
```
Benjin, line 187: Think we should add this as a config option? `mssql.flatFileImport.batchSize` or something

extensions/mssql/src/controllers/flatFileImportController.ts
```diff
@@ -0,0 +1,411 @@
...
+
+            try {
+                // Get connection string with password to perform the data import
+                const connDetails = this.connectionManager.createConnectionDetails({
+                    ...this.node.connectionProfile,
+                    database: state.formState.databaseName,
+                });
+                const connectionString = await this.connectionManager.getConnectionString(
+                    connDetails,
+                    true,
```
Benjin: add inline comments here so it's obvious what the `true`s are

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
...
+                    }
+                }
+
+                // Send import request
+                const insertDataResult = await this.provider.sendInsertDataRequest({
+                    connectionString: connectionString,
+                    batchSize: batchSize,
+                    azureAccessToken: azureAccessToken,
+                });
+                // Check result for errors
```
Benjin, line 206: nit: blank lines between blocks like this

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
...
+                for (const colChange of state.columnChanges) {
+                    const colChangeResult =
+                        await this.provider.sendChangeColumnSettingsRequest(colChange);
+                    if (colChangeResult.result.success === false) {
+                        throw new Error(colChangeResult.result.errorMessage);
+                    }
+                }
+
+                // Send import request
+                const insertDataResult = await this.provider.sendInsertDataRequest({
```
Benjin, line 200: Does this call just not return until all data has been completed?  No 10%/20%/etc updates?

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
...
+                state.errorMessage = Loc.FlatFileImport.importFailed;
+                state.fullErrorMessage = error.message;
+                state.importDataStatus = ApiStatus.Error;
+
+                sendErrorEvent(TelemetryViews.FlatFile, TelemetryActions.ImportFile, error, false);
+            }
+
+            return state;
+        });
+        this.registerReducer("openVSCodeFileBrowser", async (state, _payload) => {
```
Benjin, line 225: Nit: blank line between blocks

extensions/mssql/src/controllers/flatFileImportController.ts
```diff
@@ -0,0 +1,411 @@
...
+                state.importDataStatus = ApiStatus.NotStarted;
+                state.currentStep = FlatFileStepType.ColumnChanges;
+            } else if (payload.resetType === FlatFileStepType.ColumnChanges) {
+                state.columnChanges = [];
+                state.currentStep = FlatFileStepType.TablePreview;
+            } else if (payload.resetType === FlatFileStepType.TablePreview) {
+                state.tablePreviewStatus = ApiStatus.Loading;
+                state.tablePreview = undefined;
+                state.currentStep = FlatFileStepType.Form;
+            } else {
```
Benjin: Can you make this a switch statement so that each step (including FlatFileStepType.Form) is more explicit?

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
...
+                canSelectMany: false,
+                openLabel: Loc.FlatFileImport.selectFileToImport,
+                filters: {
+                    [Loc.FlatFileImport.importFileTypes]: this.IMPORT_FILE_TYPES,
+                },
+            });
+
+            // If no file selected, set form error.
+            if (!selectedFilePath) {
+                if (!state.formErrors.includes("flatFilePath")) {
```
Benjin, line 238: I see formErrors is an array, but I'm only spotting "flatFilePath" being added to it?  Where else is it being used for multiple types of errors?

Also, if it's just getting error _types_ added to it (rather than the full messages), maybe `formErrorTypes` would be more intuitive, or for the type to be a union of known strings instead of any string?

extensions/mssql/src/reactviews/pages/FlatFileImport/flatFileSummary.tsx
```diff
@@ -0,0 +1,261 @@
...
+    tokens,
+} from "@fluentui/react-components";
+import { locConstants } from "../../common/locConstants";
+import { FlatFileContext } from "./flatFileStateProvider";
+import { FlatFileHeader } from "./flatFileHeader";
+import { ApiStatus } from "../../../sharedInterfaces/webview";
+import { Checkmark20Regular, Dismiss20Regular } from "@fluentui/react-icons";
+import { FlatFileStepType } from "../../../sharedInterfaces/flatFileImport";
+
+const useStyles = makeStyles({
```
allancascante: move the styles to the bottom

extensions/mssql/src/controllers/flatFileImportController.ts
```diff
@@ -0,0 +1,411 @@
...
+            `;
+        const result = await this.client.sendRequest(
+            new RequestType<
+                { ownerUri: string; queryString: string },
+                SimpleExecuteResult,
+                void,
+                void
+            >("query/simpleexecute"),
+            {
+                ownerUri: this.node.sessionId,
```
Benjin: If the use of a node gets removed, I think you can grab this from the connection manager?

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -0,0 +1,411 @@
...
+            tableSchemaComponent.options = schemas.map((schema) => ({
+                displayName: schema,
+                value: schema,
+            }));
+            this.state.formState.tableSchema = schemas.includes(defaultSchema)
+                ? defaultSchema
+                : schemas[0];
+
+            this.state.schemaLoadStatus = ApiStatus.Loaded;
+        } catch (error) {
```
Benjin, line 464: logging and/or telemetry?

extensions/mssql/src/copilot/tools/listSchemasTool.ts
```diff
@@ -114,3 +93,24 @@ export class ListSchemasTool extends ToolBase<ListSchemasToolParams> {
         return { invocationMessage, confirmationMessages };
     }
 }
+
+export function getSchemaNamesFromResult(result: SimpleExecuteResult): string[] {
```
Benjin, line 102: Since this is now exported, can you add docstrings for what sort of query this parses the result for?

extensions/mssql/src/models/contracts/flatFile.ts
```diff
@@ -0,0 +1,119 @@
...
+    schemaName?: string;
+    fileType?: string;
+}
+
+export interface ProseDiscoveryResponse {
+    dataPreview: string[][];
+    columnInfo: ColumnInfo[];
+}
+
+export namespace ProseDiscoveryRequest {
```
Benjin, line 42: Docstring to say what this request does, since most devs won't know what "Prose" is

extensions/mssql/src/sharedInterfaces/telemetry.ts
```diff
@@ -42,6 +42,8 @@ export enum TelemetryViews {
     AzureDataStudioMigration = "AzureDataStudioMigration",
     FileBrowser = "FileBrowser",
     Backup = "Backup",
+    SQLOps = "SQLOps",
```
Benjin: Maybe we can pick a better name for all the SQLOps stuff.  That's just an old name for Azure Data Studio (Project Carbon -> SQL Ops Studio -> Azure Data Studio), so we probably shouldn't have it in modern code.

extensions/mssql/src/sqlOps/sqlOpsClient.ts
```diff
@@ -0,0 +1,190 @@
...
+    ClientOptions,
+    CustomOutputChannel,
+    SqlOpsDataClient,
+    Events,
+    LogLevel,
+} from "./clientInterfaces";
+
+export const flatFileConfigFile = "flatFileConfig.json";
+
+export class SqlOpsClient {
```
Benjin: Echoing the other comment about "SqlOps" being an old name for Azure Data Studio.  Can we just call this "FlatFileImportClient" since that's all it's used for (and likely all it will be used for)?  If you update the name, don't forget to also change the folder name.

extensions/mssql/src/sqlOps/sqlOpsClient.ts
```diff
@@ -0,0 +1,190 @@
...
+    private statusView: vscode.StatusBarItem;
```
Benjin: Is this file mostly ripped from ADS?  If so, can you add a permalink to the original file in case we need to reference it?

extensions/mssql/src/controllers/mainController.ts
```diff
@@ -845,6 +851,19 @@ export default class MainController implements vscode.Disposable {
         // Init CodeAdapter for use when user response to questions is needed
         this._prompter = new CodeAdapter(this._vscodeWrapper);
 
+        // Initialize SQL Data ops client
+        const sqlOpsClient = new SqlOpsClient(this._vscodeWrapper.outputChannel);
```
Benjin: Since this is starting up a separate process and performing a download, it feels like the sort of thing that should be done only if a user opens the FlatFile Import dialog.  Can we instead move this setup code to, say, a static function in sqlOpsClient.ts, only call that when the FFI dialog is launched, and have a loading spinner that tells the user what it's doing (downloading the extra files and starting the helper)?  It'd also have to make sure that it only launches the side-process once (instead of once per FFI dialog launch).

Since this may be a larger change, we can do it in a follow-up PR (but if we do, make an issue and stick it in Feb).

extensions/mssql/test/unit/sqlOpsClient.test.ts
```diff
@@ -0,0 +1,155 @@
...
+suite("SqlOpsClient", () => {
+    let sandbox: sinon.SinonSandbox;
+    let mockOutputChannel: vscode.OutputChannel;
+    let client: SqlOpsClient;
+    let mockContext: vscode.ExtensionContext;
+
+    setup(() => {
+        sandbox = sinon.createSandbox();
+
+        mockOutputChannel = {
```
Benjin: I think we have a prebuilt stub for this already.  Check stubVscodeWrapper()

extensions/mssql/test/unit/sqlOpsClient.test.ts
```diff
@@ -0,0 +1,155 @@
...
+            append: sandbox.stub(),
+            appendLine: sandbox.stub(),
+            clear: sandbox.stub(),
+            show: sandbox.stub(),
+            hide: sandbox.stub(),
+        } as unknown as vscode.OutputChannel;
+
+        client = new SqlOpsClient(mockOutputChannel);
+
+        mockContext = {
```
Benjin: Same here: stubExtensionContext()

extensions/mssql/test/unit/sqlOpsClient.test.ts
```diff
@@ -0,0 +1,155 @@
...
+        const startStub = sandbox
+            .stub(SqlOpsDataClient.prototype, "start")
+            .returns({ dispose: () => {} });
+        const onReadyStub = sandbox.stub(SqlOpsDataClient.prototype, "onReady").resolves();
+        const registerFeaturesStub = sandbox.stub(SqlOpsDataClient.prototype, "registerFeatures");
+
+        const result = await client.startFlatFileService(mockContext);
+
+        expect(result).to.be.instanceOf(SqlOpsDataClient);
+        expect(serverProviderStub.calledOnce).to.be.true;
```
Benjin: use sinon-chai instead of `(calledOnce).to.be.true`

extensions/mssql/test/unit/sqlOpsClient.test.ts
```diff
@@ -0,0 +1,155 @@
...
+        } as any);
+
+        const method = (client as any).generateServerOptions.bind(client);
+        const opts = method(serverPath, mockContext);
+
+        expect(opts.command).to.equal(serverPath);
+        expect(opts.args).to.include("--log-dir");
+        expect(opts.args).to.include("/mock/log");
+        expect(opts.transport).to.exist;
+        expect(spyConfig.calledWith("mssql")).to.be.true;
```
Benjin: sinon-chai

extensions/mssql/test/unit/sqlOpsClient.test.ts
```diff
@@ -0,0 +1,155 @@
...
+        handler("log_emitted", 3, "This is a warning message");
+        expect(
+            (mockOutputChannel.appendLine as sinon.SinonStub).calledWithMatch(
+                /This is a warning message/,
+            ),
+        ).to.be.true;
+        // LOG_EMITTED (simulate info, should not append)
+        (mockOutputChannel.appendLine as sinon.SinonStub).resetHistory();
+        handler("log_emitted", 1, "This is info message");
+        expect((mockOutputChannel.appendLine as sinon.SinonStub).called).to.be.false;
```
Benjin: sinon-chai

extensions/mssql/package.nls.json
```diff
@@ -271,5 +271,6 @@
     "mssql.openChangelog": "Open Changelog",
     "mssql.openAzureDataStudioMigration": "Migrate from Azure Data Studio",
     "mssql.showChangelogOnUpdate": "Show changelog after extension updates",
-    "mssql.backupDatabase": "Backup Database (Preview)"
+    "mssql.backupDatabase": "Backup Database (Preview)",
+    "mssql.flatFileImport": "Import (Preview)"
```
aasimkhan30: Import flat file? 
Import can be a little confusing as it coincides with import export wizard.
https://learn.microsoft.com/en-us/sql/integration-services/import-export-data/start-the-sql-server-import-and-export-wizard?view=sql-server-ver17

extensions/mssql/flatFileConfig.json
```diff
@@ -0,0 +1,16 @@
+{
+	"downloadUrl": "https://sqlopsextensions.blob.core.windows.net/extensions/import/service/{#version#}/flatfileimportservice-{#fileName#}",
```
aasimkhan30: Just convert it to .ts file and put this code. Let's not pollute the root. Look at config.ts file for downloading sts.

Benjin: Uh oh.  We're gonna have to host this somewhere else (STS github), 'cause this storage account is getting shut down with the ADS retirement.  Please do _not_ check this PR in without this URL getting updated to a different download location, 'cause I feel like this is going to be an easy one ot miss later.

extensions/mssql/src/sqlOps/sqlOpsClient.ts
aasimkhan30, line 1: can we extract common code out of service client and this for downloading and status? We should have one plug and play component. As this just dups a lot of functionality.

extensions/mssql/package.json
```diff
@@ -2414,18 +2431,24 @@
...
+                            "type": "object",
+                            "properties": {
                                 "table": {
                                     "description": "Table reference for get_table (and within edits).",
                                     "type": "object",
                                     "properties": {
-                                        "schema": { "type": "string" },
-                                        "name": { "type": "string" },
-                                        "id": { "type": "string" }
+                                        "schema": {
```
Benjin, line 2455: Did you change anything in this chunk of reformats?  If not, does the PR build pass with these unrelated changes reverted, or does the linter require these in this PR?

extensions/mssql/package.json
```diff
@@ -161,6 +161,7 @@
         "@azure/ms-rest-azure-env": "^2.0.0",
         "@azure/msal-common": "^14.14.0",
         "@azure/msal-node": "^2.12.0",
+        "@microsoft/ads-service-downloader": "^1.2.1",
```
Benjin: Can you double-check to see how we did the download for STS, and if that's at all reusable?

Benjin (changes requested):

extensions/mssql/src/reactviews/pages/FlatFileImport/flatFilePreviewTable.tsx
aasimkhan30, line 1: Please reduce the padding for table rows. There is too much top and bottom padding per row.

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
laurenastrid1: no :(

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
laurenastrid1: this is a prop of the form webivew controller, used to update error messages in form items. it's used in the form action reducer in form webview controller

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
laurenastrid1: if it throws, the schema form component will be set to have placeholder text saying "No schemas found". The user will not be able to proceed, because a schema selection is required, and schema won't be defined in the form state.

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -96,21 +118,13 @@ export class FlatFileImportController extends FormWebviewController<
         this.state.loadState = ApiStatus.Loaded;
         this.updateState();
 
-        sendActionEvent(TelemetryViews.FlatFile, TelemetryActions.StartFlatFile);
+        sendActionEvent(TelemetryViews.FlatFile, TelemetryActions.InitializeFlatFile);
```
Benjin: Just use the existing "Initialize" action so it's shared between all the dialog initializations.  Sorry, should've thought of that before

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -226,30 +250,36 @@ export class FlatFileImportController extends FormWebviewController<
...
+                    state.tablePreview = undefined;
+                    state.formState = {
+                        databaseName: state.formState.databaseName,
+                        flatFilePath: "",
+                        tableName: "",
+                        tableSchema: defaultSchema,
+                    };
+                    state.formErrors = [];
+                    state.currentStep = FlatFileStepType.Form;
+                    break;
```
Benjin, line 291: missing default case: log and emit error telemetry for unexpected step type

extensions/mssql/src/languageservice/serviceDownloadProvider.ts
```diff
@@ -28,16 +29,27 @@ export default class ServiceDownloadProvider {
...
      * Returns the download url for given platform
      */
     public getDownloadFileName(platform: Runtime): string {
-        let fileNamesJson = this._config.getSqlToolsConfigValue("downloadFileNames");
+        let fileNamesJson: any;
+        if (this._downloadType === DownloadType.SqlToolsService) {
+            fileNamesJson = this._config.getSqlToolsConfigValue("downloadFileNames");
+        } else {
+            fileNamesJson = this._config.getFlatFileConfigValue("downloadFileNames");
+        }
```
Benjin, line 55: Defensive programming: make sure that the catch-all case is intentional.  in this case, we fallback to downloading FFI, which doesn't seem right.  We should either fallback to logging and emitting telemetry or to default to the most "default" choice, which would be STS.  Let's log and emit here

extensions/mssql/src/languageservice/server.ts
```diff
@@ -31,7 +31,13 @@ export default class ServerProvider {
 
         // Otherwise, search the specified folder.
         if (this._config !== undefined) {
-            let executableFiles: string[] = this._config.getSqlToolsExecutableFiles();
+            let executableFiles: string[];
+            if (this._downloadProvider.type() === DownloadType.SqlToolsService) {
+                executableFiles = this._config.getSqlToolsExecutableFiles();
+            } else {
+                executableFiles = this._config.getFlatFileExecutableFiles();
+            }
```
Benjin, line 39: See comments elsewhere about the base case choice (log, emit telemetry)

extensions/mssql/src/languageservice/serviceDownloadProvider.ts
Benjin: Same for elsewhere

extensions/mssql/src/models/contracts/flatFile.ts
```diff
@@ -40,21 +45,6 @@ export namespace ProseDiscoveryRequest {
...
-/**
- * GetColumnInfoRequest
- */
-export interface GetColumnInfoParams {}
-
-export interface GetColumnInfoResponse {
-    columnInfo: ColumnInfo[];
-}
-
-export namespace GetColumnInfoRequest {
```
Benjin: Is this request no longer necessary?

extensions/mssql/src/reactviews/pages/FlatFileImport/flatFileForm.tsx
```diff
@@ -62,11 +62,11 @@ const useStyles = makeStyles({
 export const FlatFileForm: React.FC = () => {
     const classes = useStyles();
     const context = useContext(FlatFileContext);
-    const state = context?.state;
+    const state = useFlatFileSelector((s) => s);
```
Benjin: use specific selectors instead of a generic `(s) => s` pattern to minimize React redraws.

```
const formComponents = useFlatFileSelector((s) => s.formComponents);
```

Benjin: Same for state.formState

extensions/mssql/src/reactviews/pages/FlatFileImport/flatFileHeader.tsx
```diff
@@ -51,7 +52,7 @@ interface HeaderProps {
 export const FlatFileHeader: React.FC<HeaderProps> = ({ headerText, stepText }) => {
     const classes = useStyles();
     const context = useContext(FlatFileContext);
-    const state = context?.state;
+    const state = useFlatFileSelector((s) => s);
```
Benjin: specific selector

Benjin: Check the other files, too

extensions/mssql/src/models/contracts/flatFile.ts
laurenastrid1: yeah, i wasn't using it anywhere

extensions/mssql/src/controllers/flatFileImportWebviewController.ts
```diff
@@ -278,6 +278,17 @@ export class FlatFileImportWebviewController extends FormWebviewController<
...
                     break;
+                default:
+                    const errorMessage = `Unknown reset type: ${payload.resetType}`;
+                    this.logger.error(errorMessage);
+                    state.loadState = ApiStatus.Error;
+                    sendErrorEvent(
+                        TelemetryViews.FlatFile,
+                        TelemetryActions.ResetState,
+                        new Error(errorMessage),
+                        false,
```
Benjin: We should include the error message here.  If it isn't, then we'll never know what resetType was missed!

extensions/mssql/src/languageservice/serviceDownloadProvider.ts
```diff
@@ -72,8 +76,12 @@ export default class ServiceDownloadProvider {
         let versionFromConfig: string;
         if (this._downloadType === DownloadType.SqlToolsService) {
             versionFromConfig = this._config.getSqlToolsPackageVersion();
-        } else {
+        } else if (this._downloadType === DownloadType.FlatFileService) {
             versionFromConfig = this._config.getFlatFilePackageVersion();
+        } else {
+            const errorMessage = `Unsupported download type: ${this._downloadType}`;
+            this._logger.appendLine(`[ERROR] ${errorMessage}`);
+            throw new Error(errorMessage);
```
Benjin, line 84: what handles this error once it's thrown?  does it eventually make its way to telemetry getting emitted?

Benjin: Same q with above

extensions/mssql/src/languageservice/serviceDownloadProvider.ts
```diff
@@ -46,8 +46,12 @@ export default class ServiceDownloadProvider {
         let fileNamesJson: any;
         if (this._downloadType === DownloadType.SqlToolsService) {
             fileNamesJson = this._config.getSqlToolsConfigValue("downloadFileNames");
-        } else {
+        } else if (this._downloadType === DownloadType.FlatFileService) {
             fileNamesJson = this._config.getFlatFileConfigValue("downloadFileNames");
+        } else {
+            const errorMessage = `Unsupported download type: ${this._downloadType}`;
+            this._logger.appendLine(`[ERROR] ${errorMessage}`);
```
Benjin, line 53: use logger.error instead to make sure it's written with the correct criticality level.

extensions/mssql/src/reactviews/pages/FlatFileImport/flatFileHeader.tsx
```diff
@@ -52,9 +51,8 @@ interface HeaderProps {
 export const FlatFileHeader: React.FC<HeaderProps> = ({ headerText, stepText }) => {
     const classes = useStyles();
     const context = useContext(FlatFileContext);
-    const state = useFlatFileSelector((s) => s);
 
-    if (!context || !state) return;
+    if (!context) return null;
```
Benjin, line 55: Doesn't using `null` instead of `undefined` anger the linter gods?

extensions/mssql/src/languageservice/serviceDownloadProvider.ts
laurenastrid1: this logger doesn't have logger.error, it's  a different interface. if we need to replace it with the other logger type i'll do it in a follow up pr

extensions/mssql/src/languageservice/serviceDownloadProvider.ts
laurenastrid1: pops up with an error message!

extensions/mssql/src/controllers/mainController.ts
```diff
@@ -867,6 +873,14 @@ export default class MainController implements vscode.Disposable {
         // Init CodeAdapter for use when user response to questions is needed
         this._prompter = new CodeAdapter(this._vscodeWrapper);
 
+        // Initialize flat file client
+        managerInstance.onRegisteredApi<FlatFileProvider>(ApiType.FlatFileProvider)((provider) => {
+            this.flatFileProvider = provider;
+            void Promise.resolve(true);
+        });
+        this.flatFileClient = new FlatFileClient(this._vscodeWrapper);
+        await this.flatFileClient.startFlatFileService(this._context);
```
Benjin: Since the FFI service is 40 MB, it should be downloaded and started only once the dialog is opened.  Make sure that it isn't downloaded every time a user opens the FFI dialog, too

Benjin (dismissed):

Benjin (approved):

PR #21107, opened 2026-02-05
Data workspace extension code cleanup

ssreerama:

## Description

In this PR:
- Removes ADS related dialog files
- Removes azdata api calls
- cleanup package.json that has ads commands
- Removed the ADS‑only Manage Projects feature, as it is not supported in VS Code. If we decide to enable this for VS Code in the future, we can restore any reusable pieces, though the Manage Projects flow would need to be rewritten regardless.  

## Code Changes Checklist

### DIFF
--- a/extensions/data-workspace/src/common/dataWorkspaceExtension.ts
+++ b/extensions/data-workspace/src/common/dataWorkspaceExtension.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as vscode from "vscode";
-import { IExtension, IProjectType } from "dataworkspace";
+import { IExtension } from "dataworkspace";
 import { WorkspaceService } from "../services/workspaceService";
 import { defaultProjectSaveLocation } from "./projectLocationHelper";
 import { createNewProjectWithQuickpick } from "../dialogs/newProjectQuickpick";
@@ -43,7 +43,7 @@ export class DataWorkspaceExtension implements IExtension {
     return this.workspaceService.validateWorkspace();
   }
 
-  openSpecificProjectNewProjectDialog(_projectType: IProjectType): Promise<vscode.Uri | undefined> {
+  openSpecificProjectNewProjectDialog(): Promise<vscode.Uri | undefined> {
     if (!this.workspaceService.isProjectProviderAvailable) {
       void vscode.window.showErrorMessage(noProjectProvidingExtensionsInstalled);
       return Promise.resolve(undefined);
--- a/extensions/data-workspace/src/dataworkspace.d.ts
+++ b/extensions/data-workspace/src/dataworkspace.d.ts
@@ -49,11 +49,10 @@ declare module "dataworkspace" {
     validateWorkspace(): Promise<boolean>;
 
     /**
-     * Opens the new project dialog with only the specified project type
-     * @param projectType project type to open the dialog for
+     * Opens the new project dialog
      * @returns the uri of the created the project or undefined if no project was created
      */
-    openSpecificProjectNewProjectDialog(projectType: IProjectType): Promise<vscode.Uri | undefined>;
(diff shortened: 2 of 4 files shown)


extensions/data-workspace/src/common/dataWorkspaceExtension.ts
```diff
@@ -48,7 +48,9 @@ export class DataWorkspaceExtension implements IExtension {
       void vscode.window.showErrorMessage(noProjectProvidingExtensionsInstalled);
     }
 
-    return openSpecificProjectNewProjectDialog(projectType, this.workspaceService);
+    // Use VS Code quickpick flow - specific project type filtering is not supported in quickpick
+    // The user can select the appropriate project type from the list
+    return createNewProjectWithQuickpick(this.workspaceService);
   }
```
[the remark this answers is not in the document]
ssreerama: fixed

extensions/data-workspace/src/common/dataWorkspaceExtension.ts
```diff
@@ -43,12 +43,13 @@ export class DataWorkspaceExtension implements IExtension {
     return this.workspaceService.validateWorkspace();
   }
 
-  openSpecificProjectNewProjectDialog(projectType: IProjectType): Promise<vscode.Uri | undefined> {
+  openSpecificProjectNewProjectDialog(_projectType: IProjectType): Promise<vscode.Uri | undefined> {
```
Benjin: If this isn't getting used, then it's assuming it's always a SqlProj, right?  Then if we're confident no one else has created a project type, I think we should just change the API here to remove it.  It might be a good scream test :)

extensions/data-workspace/src/services/workspaceService.ts
```diff
@@ -57,28 +56,13 @@ export class WorkspaceService implements IWorkspaceService {
   }
 
   /**
-   * Verify that a workspace is open or that if one isn't and we're running in ADS, it's ok to create a workspace and restart ADS
+   * Verify that a workspace is open. VS Code doesn't require reloading the window when creating a workspace or
+   * adding the first item to an open workspace.
    */
   async validateWorkspace(): Promise<boolean> {
```
Benjin, line 62: If this always returns true, can this function be removed entirely?

extensions/data-workspace/src/common/dataWorkspaceExtension.ts
ssreerama: for now SqlProject is the only proejctType that workspace supports as only sqlproj type is defined for vscode. I agree, this will be good scream test and if anyone sees an issue we can add the support with good explanation.

extensions/data-workspace/src/services/workspaceService.ts
ssreerama: yes, this function can be removed, but I'll do that with sql projects extension cleanup as we have several references to this function

ssreerama: Done!

extensions/data-workspace/src/common/dataWorkspaceExtension.ts
ssreerama: Done!

extensions/data-workspace/src/common/utils.ts
```diff
@@ -39,16 +38,15 @@ export interface IPackageInfo {
 
 export function getPackageInfo(packageJson: any): IPackageInfo | undefined {
```
lewis-sanchez, line 39: You might also want to add the ts ignore here too since you added one below: `// eslint-disable-next-line @typescript-eslint/no-explicit-any`

lewis-sanchez (approved):

extensions/data-workspace/src/common/utils.ts
ssreerama: Thank you for noticing it, will make a note of it, add this in my next clean up PR

PR #21120, opened 2026-02-07
Making vscodewebviewProvider2 non blocking on non-essential requests. 

aasimkhan30:
## Description

1. Optimized VscodeWebviewProvider2 startup so the first render waits only for the initial state. Theme, keybindings, EOL, localization, and load-stats are now loaded in the background. A fallback path was also added to prevent infinite blank screens if state bootstrap fails.
2. Refactored localization loading into controllers/localizationCache.ts, initializing a single in-memory cache (initializeWebviewLocCache()) during extension activation. This avoids reading localization files on every webview creation. While item 1 already ensures webviews don’t block on localization, this is a perf win and reduces redundant I/O.

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/controllers/localizationCache.ts
+++ b/extensions/mssql/src/controllers/localizationCache.ts
@@ -1,0 +1,40 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from "vscode";
+
+let localizationFileCache: string | undefined;
+let localizationFileReadPromise: Promise<string | undefined> | undefined;
+
+export function initializeWebviewLocalizationCache(): void {
+    void getLocalizationFileContentsCached();
+}
+
+export async function getLocalizationFileContentsCached(): Promise<string | undefined> {
+    if (localizationFileCache !== undefined) {
+        return localizationFileCache;
+    }
+
+    if (localizationFileReadPromise) {
+        return localizationFileReadPromise;
+    }
+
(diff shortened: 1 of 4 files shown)


extensions/mssql/src/controllers/localizationCache.ts
```diff
@@ -0,0 +1,40 @@
...
+export async function getLocalizationFileContentsCached(): Promise<string | undefined> {
+    if (localizationFileCache !== undefined) {
+        return localizationFileCache;
+    }
+
+    if (localizationFileReadPromise) {
+        return localizationFileReadPromise;
+    }
+
+    if (!vscode.l10n.uri) {
```
Benjin, line 24: Is this when loc files aren't needed because MSSQL's being run in English?  Can you add that as a comment (or if it's for some error case, say what might be causing that)?

extensions/mssql/src/reactviews/common/vscodeWebviewProvider2.tsx
```diff
@@ -120,50 +120,83 @@ export function VscodeWebviewProvider2<State, Reducers>({ children }: VscodeWebv
...
-                // Mark as initialized and emit state change
-                setIsInitialized(true);
-                emit();
             } catch (error) {
                 console.error("Bootstrap failed:", error);
-                // Still mark as initialized to prevent infinite loading
-                setIsInitialized(true);
+                // Prevent indefinite blank screen when initial state fetch fails.
+                if (stateRef.current === undefined) {
+                    stateRef.current = {} as State;
```
Benjin, line 196: This just makes it so that each individual webview needs to handle its own stateless screen, right?  (or we centrally-handle that).  Would be good to have the alternative effect captured in a comment

Benjin (approved):

Issue #21095, opened 2026-02-04
[Feature Request]: Improve SQL Profiler Filter UX with Inline Column Filters


croblesm:
### Feature Description

Redesign the SQL Profiler filtering experience to use inline, column-level filters instead of the current modal-based approach. The new design should include:

1. **Always-visible filter icons** on each column header for better discoverability
2. **Inline filter dropdowns** anchored to each column that open on click
3. **Adaptive filter types** based on column data:
   - Categorical columns (EventClass, Application) → Searchable checkbox list of distinct values
   - Numeric columns (Duration, Reads, Writes, CPU) → Operator dropdown (>, <, =, between) with value input
   - Text columns (TextData, LoginName) → Operator dropdown (contains, equals, starts with) with text input
4. **Active filter chips** displayed in a bar below the toolbar showing all applied filters with one-click removal
5. **Quick filter search** in the toolbar for fast cross-column text filtering
6. **Clear All Filters** button to reset all filters at once

> **Current Experience (Modal-based):**

Current filter modal - Single clause
<img width="1979" height="1007" alt="Image" src="https://github.com/user-attachments/assets/d47cc60d-4d5b-421f-9df2-eeb65b2a3a0f" />

Current filter modal - multiple clauses]
<img width="1979" height="1007" alt="Image" src="https://github.com/user-attachments/assets/bfaf27fa-8d00-4ac1-b078-b0c399d2bb53" />

********

> **Proposed Experience (inline-based):**

Base experience, no filters
<img width="1296" height="605" alt="Image" src="https://github.com/user-attachments/assets/d2ab8f21-a573-484a-8984-33d7bb9bb462" />

Filter dropdown
<img width="1296" height="605" alt="Image" src="https://github.com/user-attachments/assets/785eb516-e91b-42ee-92b8-c038f84e54b9" />

Single column, single value filter - Active filter chip
<img width="1296" height="605" alt="Image" src="https://github.com/user-attachments/assets/5bf261d9-d61b-4173-90cb-96bdf87576f6" />

Single column, multiple value filter - Active filter chip
<img width="1296" height="605" alt="Image" src="https://github.com/user-attachments/assets/c08b17cf-24eb-4e36-836a-bece1b84b1b8" />

Multiple columns, multiple value filter - Active filter chip
<img width="1296" height="605" alt="Image" src="https://github.com/user-attachments/assets/ae8e4582-36a4-4285-85c7-56d486223736" />

Different filter options, numeric type example
[...]

PR #21151, opened 2026-02-10
Profiler - Filtering on Columns

allancascante:
## Description

Adding support to filter at column level as requested by: https://github.com/microsoft/vscode-mssql/issues/21095

These changes support filtering by column, added the extra field for quick filter at the toolbar and the filter display under the toolbar. Also each column type changes the type of filtering provided, for text that is not TextData column, filter is categorized and can filter by selecting available options from the list that renders in the filter pop-over, for TextData filtering is by string comparison with different operators and for number columns it uses number operators. Finally, it enables sorting support 1 column at the time either ascending or descending, taking the column type into consideration.

UI Changes:

image
image
image
image
image

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/reactviews/pages/Profiler/profiler.tsx
+++ b/extensions/mssql/src/reactviews/pages/Profiler/profiler.tsx
@@ -1012,16 +1012,6 @@ export const Profiler: React.FC = () => {
                     onClear={handlePopoverClear}
                 />
             )}
-            <div id="profilerGridContainer" className={classes.profilerGridContainer}>
-                <SlickgridReact
-                    gridId="profilerGrid"
-                    columns={columns}
-                    options={gridOptions}
-                    dataset={EMPTY_DATASET}
-                    onReactGridCreated={handleReactGridCreated}
-                    onScroll={handleScroll}
-                />
-            </div>
             <PanelGroup direction="vertical" className={classes.panelGroup}>
                 <Panel ref={gridPanelRef} defaultSize={100} minSize={10}>
                     <div id="profilerGridContainer" className={classes.profilerGridContainer}>
@@ -1030,7 +1020,7 @@ export const Profiler: React.FC = () => {
                             columns={columns}
                             options={gridOptions}
                             dataset={EMPTY_DATASET}
-                            onReactGridCreated={(e) => reactGridReady(e.detail)}
+                            onReactGridCreated={handleReactGridCreated}
                             onScroll={handleScroll}
                             onClick={handleRowClick}
                             onActiveCellChanged={handleActiveCellChanged}


extensions/mssql/src/reactviews/common/locConstants.ts
```diff
@@ -1979,6 +1981,34 @@ export class LocConstants {
...
+            clearFilterTooltip: l10n.t("Clear all filters and show all events"),
+            clearFilterDisabledTooltip: l10n.t("No filter is currently active"),
+
+            // Filter operators
+            operatorEquals: l10n.t("="),
+            operatorNotEquals: l10n.t("<>"),
+            operatorLessThan: l10n.t("<"),
+            operatorLessThanOrEqual: l10n.t("<="),
+            operatorGreaterThan: l10n.t(">"),
+            operatorGreaterThanOrEqual: l10n.t(">="),
```
lewis-sanchez, line 2088: These comparison operators need to be localized? I thought these would be known already considering lots of programming languages use these.

allancascante: not sure if all languages use the same characters for these, I will leave them just in case

extensions/mssql/src/profiler/profilerConfigService.ts
```diff
@@ -80,7 +80,7 @@ export class ProfilerConfigService {
     }
 
     /**
-     * Get all available templates
+     * Get all available templates.
```
Benjin: Nit: these don't need/shouldn't have periods

extensions/mssql/src/reactviews/pages/Profiler/profilerActiveFiltersBar.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+    return (
+        <div className={classes.bar}>
+            <Text size={200} weight="semibold" className={classes.label}>
+                {loc.activeFiltersLabel}
+            </Text>
+            {badges.map((badge) => (
+                <span key={badge.field} className={classes.badge}>
+                    <Text size={200} className={classes.badgeText}>
+                        {badge.label}
```
lewis-sanchez, line 115: Is this localized?

extensions/mssql/src/profiler/profilerConfigService.ts
```diff
@@ -176,7 +176,7 @@ export class ProfilerConfigService {
         // Try each mapped event field until we find a value
         for (const mappedField of column.eventsMapped) {
             const value = this.getFieldValue(event, mappedField);
-            if (value !== null && value !== undefined && value !== "") {
+            if (value !== undefined && value !== "") {
```
Benjin: could just be `if (value) { }`

extensions/mssql/src/reactviews/pages/Profiler/profilerActiveFiltersBar.tsx
lewis-sanchez: I see. It's localized on line 96: `label: loc.filterBadge(columnName, description),`

extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -0,0 +1,644 @@
...
+        this._enabled = clauses.length > 0;
+    }
+
+    /**
+     * Clears all filter clauses and disables clause filtering.
+     * Does NOT clear quick filter — use clearQuickFilter() or clearAllFilters() for that.
+     * After clearing, all events in the buffer become visible (if no quick filter).
+     */
+    clearFilter(): void {
+        this._clauses = [];
```
Benjin: `clearAllFilters` since this affects all clause filters

extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -0,0 +1,644 @@
...
+            return this._buffer.size;
+        }
+        return this.getFilteredRows().length;
+    }
+
+    /**
+     * Sets the filter clauses and enables filtering.
+     * @param clauses - Array of filter clauses (combined with AND logic)
+     */
+    setFilter(clauses: FilterClause[]): void {
```
Benjin: `setAllFilters()` since this is a bulk operation and replaces all existing filters

extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -0,0 +1,644 @@
...
+    get filteredCount(): number {
+        if (!this.isFilterActive) {
+            return this._buffer.size;
+        }
+        return this.getFilteredRows().length;
+    }
+
+    /**
+     * Sets the filter clauses and enables filtering.
+     * @param clauses - Array of filter clauses (combined with AND logic)
```
Benjin, line 108: Add info to docstring that the "AND" is for the resulting rows, and that there's only zero or one clause per column

extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -0,0 +1,644 @@
...
+     */
```
Benjin, line 109: setClauseFilters()

extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -0,0 +1,644 @@
...
+        this._clauses = [...clauses];
+        this._enabled = clauses.length > 0;
+    }
+
+    /**
+     * Clears all filter clauses and disables clause filtering.
+     * Does NOT clear quick filter — use clearQuickFilter() or clearAllFilters() for that.
+     * After clearing, all events in the buffer become visible (if no quick filter).
+     */
+    clearFilter(): void {
```
Benjin: clearColumnFilters

extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -0,0 +1,644 @@
...
+
+    /**
+     * Clears the quick filter.
+     */
+    clearQuickFilter(): void {
+        this._quickFilter = undefined;
+    }
+
+    /**
+     * Clears both column filters and the quick filter.
```
Benjin, line 141: We should be consistent in calling these either "clause filters" or "column filters" in both code and comments.  I like "column filter" better.

extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -0,0 +1,644 @@
...
+     * This allows filtering on view-level column names instead of raw event fields.
+     */
+    setRowConverter(converter: ((row: T) => Record<string, unknown>) | undefined): void {
+        this._rowConverter = converter;
+    }
+
+    /**
+     * Enables or disables the clause filter without changing clauses.
+     */
+    setEnabled(enabled: boolean): void {
```
Benjin: enableColumnFilters().  Do we need a disableColumnFilters() as well?

Or delete if not currently used

extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -0,0 +1,644 @@
...
+
+    /**
+     * Gets converted and filtered rows for the grid, applying both column filters
+     * and quick filter, then slicing for pagination.
+     * When a row converter is set, returns converted rows (not raw buffer rows).
+     * @param startIndex - Starting index in the filtered result set
+     * @param count - Maximum number of rows to return
+     * @returns Object with rows, startIndex, and totalCount
+     */
+    getConvertedFilteredRange(
```
Benjin: What does it mean to convert these?  Is this code still used/needed?

extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -0,0 +1,644 @@
...
+    private isNullOrUndefined(value: unknown): boolean {
+        // eslint-disable-next-line eqeqeq
+        return value == undefined;
+    }
+
+    /**
+     * Evaluates equality between field value and filter value.
+     * Handles type coercion based on typeHint.
+     */
+    private evaluateEquals(
```
Benjin, line 385: IMO, let's get the values converted into TS types as soon as we get them from STS.  That should make it simpler to display them in the UI and saves us trouble of parsing type here for filters.

Benjin: For a follow-up PR in March ^  Can you create a github issue and add the link here?

extensions/mssql/src/profiler/profilerSortUtils.ts
```diff
@@ -0,0 +1,12 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// Re-export sort utilities from sharedInterfaces so existing extension-side consumers
```
Benjin: Is this re-export necessary?  Both webview and extension side should be able to access sharedInterfaces

extensions/mssql/src/profiler/profilerSortUtils.ts
```diff
@@ -0,0 +1,12 @@
...
+// continue to work without changes.
```
Benjin: Check with Aasim about including functions in sharedInterfaces

extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -0,0 +1,644 @@
...
+ * Separates filtering concerns from the ring buffer storage.
+ *
+ * Key behaviors:
+ * - When filter is active, only matching rows are returned
+ * - When filter is cleared, ALL events (including those hidden while filtering) are visible
+ * - Filtering is client-side only; does not affect the underlying buffer storage
+ *
+ * @template T - Row type that extends IndexedRow
+ */
+export class FilteredBuffer<T extends IndexedRow> {
```
Benjin, line 20: Would this make more sense as `FilteredRingBuffer extends RingBuffer`

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -426,10 +445,164 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
...
+                };
+                this.updateStatusBar();
+
+                return this.state;
+            }
+            return state;
+        });
+
+        // Handle get distinct values request from webview (scans unfiltered ring buffer)
+        this.registerReducer("getDistinctValues", (state, payload: { field: string }) => {
```
Benjin: You should be able to do "onRequest" if the webview needs to fetch data from the backend.

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -426,10 +445,164 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
...
+                    values,
+                };
+                void this.sendNotification(ProfilerNotifications.DistinctValuesAvailable, response);
+            }
+            return state;
+        });
+
+        // Handle set quick filter request from webview
+        this.registerReducer("setQuickFilter", (state, payload: { term: string }) => {
+            if (this._filteredBuffer) {
```
Benjin: Nit: invert this so that you get early-exits.  Makes code easier to read.

```
if (!this._filteredBuffer) {
    return state;
}

// now continue without indentation
```

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -426,10 +445,164 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
...
+                    return this.state;
+                }
+
                 return state;
             },
         );
 
+        // Handle apply filter request from webview (client-side only)
+        this.registerReducer("applyFilter", (state, payload: { clauses: FilterClause[] }) => {
+            if (this._filteredBuffer) {
```
Benjin: invert (see other comment)

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -426,10 +445,164 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
...
+            isFilterActive: this._filteredBuffer.isFilterActive,
+            clauseCount: this._filteredBuffer.clauses.length,
+            totalCount: this._filteredBuffer.totalCount,
+            filteredCount: this._filteredBuffer.getFilteredCount(),
+        };
+
+        await this.sendNotification(ProfilerNotifications.FilterStateChanged, params);
         // Handle export to CSV request from webview
         // Generate CSV from the RingBuffer (source of truth) rather than from grid data
         // This ensures ALL events in the session buffer are exported
```
Benjin: What's with the indentation here? looks like a merge error.  the exportToCsv() reducer is inside sendFilterStateChanged()

extensions/mssql/src/sharedInterfaces/profiler.ts
```diff
@@ -228,6 +274,139 @@ export namespace ProfilerNotifications {
...
+ * @param sortDir - The sort direction (ASC or DESC)
+ * @returns Negative if a < b, positive if a > b, 0 if equal
+ */
+export function profilerSortComparator(
+    a: Record<string, unknown>,
+    b: Record<string, unknown>,
+    sortField: string,
+    sortDir: SortDirection,
+): number {
+    const valA = a[sortField];
```
Benjin, line 371: Why not just have the caller pass in `a[sortField]` and `b[sortField]`, and simplify the function signature?

allancascante: This signature is intentional profilerSortComparator is used inside createDataViewSortFn, which returns a closure passed directly to Array.sort. The 'sort' engine only provides full raw objects, so the comparator must extract the field values itself. The caller can't pre-extract them because it doesn't control the sort of callback invocation. To simplify the signature to (valA, valB, sortDir), we'd need to: 
1. Extract a value-level comparator with the simplified signature that handles the empty-to-end semantics, type detection, and direction flip.
2. Keep the row-level wrapper (or inline it)
The net result would be two functions instead of one, with the row-level function becoming a one-liner delegating to the value-level function.

extensions/mssql/src/profiler/filteredBuffer.ts
allancascante: https://github.com/microsoft/vscode-mssql/issues/21201

extensions/mssql/src/profiler/filteredBuffer.ts
allancascante: https://github.com/microsoft/vscode-mssql/issues/21201

extensions/mssql/src/profiler/filteredBuffer.ts
allancascante: removed it

extensions/mssql/src/profiler/filteredBuffer.ts
allancascante: deleted

extensions/mssql/src/sharedInterfaces/profiler.ts
allancascante: will do it as part of: https://github.com/microsoft/vscode-mssql/issues/21201

extensions/mssql/src/profiler/profilerSortUtils.ts
allancascante: this is not really needed removing it

extensions/mssql/src/controllers/reactWebviewPanelController.ts
```diff
@@ -74,7 +74,11 @@ export class ReactWebviewPanelController<
             this._panel.onDidDispose(async () => {
                 let prompt;
                 if (this._options.showRestorePromptAfterClose) {
-                    prompt = await this.showRestorePrompt();
+                    try {
+                        prompt = await this.showRestorePrompt();
+                    } catch (e) {
+                        console.error("Error showing restore prompt:", e);
```
lewis-sanchez, line 80: Maybe we should use the logger class to log errors instead of writing them to the console. This could be done in a follow up PR.

Issue #21152, opened 2026-02-10
[Bug]: Error when moving a .sql file from within a folder to the root of the project


Benjin:
### Description

> Error when moving file from c:\Users\benjind\Source\TestWorkspace\PokedexDb\Views\vw_TypeEffectiveness.sql to PokedexDb\vw_TypeEffectiveness.sql. Error: Could not find a part of the path.

Could be upstream in STS/SqlProjects DLL

### Steps to Reproduce

1. Create a sql project
2. Add a folder
3. Add a script inside that folder
4. Try to move script to project root by dragging and dropping it on top of the Project node

Expected:
file moves to project root

Reality:
Error message:
> Error when moving file from c:\Users\benjind\Source\TestWorkspace\PokedexDb\Views\vw_TypeEffectiveness.sql to PokedexDb\vw_TypeEffectiveness.sql. Error: Could not find a part of the path.

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [x] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

Platform: Windows
MSSQL and SqlProj: latest from `main`

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #21181, opened 2026-02-12
fix: use project name from .sqlproj instead of folder name when movin…

ssreerama:
…g files to project root

# Issue: [# 21152](https://github.com/microsoft/vscode-mssql/issues/21152)
## Description

This PR adds a single line of code change that fixes the move file to root directory issue.
- changes are unnoticeable due to prettier formatting the whole file which is unavoidable.
- Added comments where the real change was done, please go through that place to review.
- Added a test 
 
## Code Changes Checklist

Before:
MoveFileToRoot_Before

After:
MoveFileToRoot_After

### DIFF
--- a/extensions/sql-database-projects/test/projectController.test.ts
+++ b/extensions/sql-database-projects/test/projectController.test.ts
@@ -2013,8 +2013,9 @@ suite("ProjectsController", function (): void {
         });
 
         test.skip("Should move a file to project root when project folder name differs from project name", async function (): Promise<void> {
-            const spy = sinon.spy(vscode.window, "showErrorMessage");
-            sinon
+            const sandbox = sinon.createSandbox();
+            const spy = sandbox.spy(vscode.window, "showErrorMessage");
+            sandbox
                 .stub(vscode.window, "showWarningMessage")
                 .returns(<any>Promise.resolve(constants.move));
 


extensions/sql-database-projects/src/controllers/projectController.ts
```diff
@@ -78,2159 +97,2999 @@ interface FileWatcherStatus {
...
+
+        // Calculate the new file path
+        let folderPath;
+        // target is the root of project, which is the .sqlproj
+        if (target.element.projectFileUri.fsPath === target.element.fileSystemUri.fsPath) {
+            // Get the project name from .sqlproj file path, not the folder name
+            folderPath = path.basename(
+                target.element.projectFileUri.fsPath,
+                constants.sqlprojExtension,
+            );
```
ssreerama, line 3048: this is the fix, used to be

			folderPath = path.basename(path.dirname(target.element.projectFileUri.fsPath!));
that gets the root folder name, but the project name should be the one we need to move the file.

extensions/sql-database-projects/test/projectController.test.ts
```diff
@@ -2011,6 +2011,57 @@ suite("ProjectsController", function (): void {
                 `The file path should not have been updated when trying to move script1.sql to proj2`,
             );
         });
+
+        test.skip("Should move a file to project root when project folder name differs from project name", async function (): Promise<void> {
```
Benjin, line 2015: Why is this test marked as `skip`?

Benjin (approved):

extensions/sql-database-projects/test/projectController.test.ts
ssreerama: the mock setup is broken, which is already fixed in cleanup PR, so i didn't want those changes again here for this test. once this is merged, I'll make this test run there.

added the test to not miss the test scenario

PR #21194, opened 2026-02-13
VBump STS

ssreerama:

## Description

This PR brings:
- Dacfx 170.3
- Microsoft.Build.sql 2.0.0 (still be old)

image

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/configurations/config.ts
+++ b/extensions/mssql/src/configurations/config.ts
@@ -7,7 +7,7 @@ export const config = {
     service: {
         downloadUrl:
             "https://github.com/Microsoft/sqltoolsservice/releases/download/{#version#}/microsoft.sqltools.servicelayer-{#fileName#}",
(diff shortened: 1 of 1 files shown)


lewis-sanchez:
Looks like the pipelines are failing.

ssreerama:
> Looks like the pipelines are failing.

This appears to be tied to a specific version. During testing, SQL Tools Service fails to load, and the same version uses the Build.Sql  latest 2.1 which has the download issue. I’ve rolled back to just the DacFx version bump and will verify whether that addresses the issue

ssreerama:
This version will be updated with under progress STS build

ssreerama:
Since STS latest version build is having some issues which causes the service fail to load. this version is good to add atleast 170.3 dacfx. if we can fix the sts release pipeline with new version, we can add sdk 2.1 later.

PR #21206, opened 2026-02-14
Fixing unit test runners.

aasimkhan30:
## Description

CI tests were intermittently failing at Electron startup when ELECTRON_RUN_AS_NODE leaked into test steps. In that mode Electron runs as Node and rejects VS Code test flags (--extensionTestsPath, etc.). 

We can clear that before running every test suite and that will fix our failures. 

However, I upgraded our test runner to use the @vscode/test-electron runners. This gets us out of having custom test runners.

Now we can run tests using @vscode/test-cli. 
That automatically unsets the flag
 https://github.com/microsoft/vscode-test-cli/blob/4b38cc99abc811d1d615bdb09e2f125c22db43a0/src/cli/platform/desktop.mts#L68

## Code Changes Checklist

### DIFF
--- a/extensions/data-workspace/package.json
+++ b/extensions/data-workspace/package.json
@@ -217,6 +217,7 @@
         "@types/vscode": "1.98.0",
         "@vscode/l10n-dev": "^0.0.35",
         "@vscode/test-cli": "^0.0.12",
+        "@vscode/test-electron": "^2.5.2",
         "@vscode/vsce": "^3.1.1",
         "chai": "^5.3.3",
         "glob": "^11.0.0",
--- a/extensions/data-workspace/yarn.lock
+++ b/extensions/data-workspace/yarn.lock
@@ -1030,6 +1030,17 @@
     supports-color "^10.2.2"
     yargs "^17.7.2"
 
+"@vscode/test-electron@^2.5.2":
+  version "2.5.2"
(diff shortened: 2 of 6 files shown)


extensions/mssql/.vscode-test.mjs
```diff
@@ -1,13 +1,28 @@
...
-export default testCli.defineConfig([
-    {
-        label: "Unit Tests",
-        files: "out/test/**/*.test.js",
-        version: "insiders",
-        mocha: {
-            ui: "tdd",
-            timeout: 6_000,
+const mocha = {
+    timeout: 30_000,
```
Benjin, line 4: Looks like this is an increase from 6s to 30s for timeouts (which was only used for activation-related tests, IIRC).  Any reason we increased all of them instead of overriding the timeout for specific longer tests?

extensions/data-workspace/package.json
```diff
@@ -215,13 +215,15 @@
         "@types/sinon": "^10.0.12",
         "@types/sinon-chai": "^4.0.0",
         "@types/vscode": "1.98.0",
-        "chai": "^5.3.3",
         "@vscode/l10n-dev": "^0.0.35",
-        "@vscode/test-cli": "^0.0.10",
+        "@vscode/test-cli": "^0.0.12",
         "@vscode/test-electron": "^2.5.2",
```
Benjin: Since we no longer have `runTest.ts` using this package directly, do we still need to reference it here?

PR #21208, opened 2026-02-16
telemetry for profiler

allancascante:
## Description

Adding telemetry to profiler functionality. 

## Code Changes Checklist


extensions/mssql/src/profiler/profilerTelemetry.ts
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Static helper that emits profiler telemetry events.
+ *
+ * **Design rules**
+ * - Every method is best-effort (`try / catch` → silent).
+ * - No PII is ever sent (no query text, no user names, no file paths).
+ * - The session GUID produced by `Utils.generateGuid()` is used as the
+ *   correlation ID – there is no separate telemetry GUID.
+ */
+export class ProfilerTelemetry {
```
Benjin: Most of these shouldn't be necessary.  `sendActionEvent()` are similar calls already shouldn't be throwing, so the try-catch wrappers aren't necessary.  That mostly just leaves hardcoding the view and action for each type, which doesn't seem worth introducing a new class for.

I think you can easily just delete this class and call `sendActionEvent()` directly.

extensions/mssql/src/profiler/profilerTelemetry.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /** The close-with-unsaved-events warning was shown. */
+    static sendCloseWarningShown(
+        sessionId: string,
+        unsavedEventsCount: number,
+        userAction: CloseWarningUserAction,
+    ): void {
+        try {
+            sendActionEvent(
```
Benjin: Is this worth capturing telemetry for?

extensions/mssql/src/profiler/profilerTelemetry.ts
```diff
@@ -0,0 +1,182 @@
...
+        } catch {
+            // best-effort
+        }
+    }
+
+    /**
+     * A column filter was applied.
+     * Only sends `column:operator` pairs – **never** the filter values.
+     */
+    static sendFilterApplied(sessionId: string, filterFields: FilterClause[]): void {
```
Benjin: Is this worth capturing telemetry for?  Seems like something that is okay just to log instead of emit

extensions/mssql/src/profiler/profilerTelemetry.ts
```diff
@@ -0,0 +1,182 @@
...
+            });
+        } catch {
+            // best-effort
+        }
+    }
+
+    /** A profiling session failed to stop. */
+    static sendSessionStopFailed(sessionId: string, errorMessage: string): void {
+        try {
+            sendActionEvent(TelemetryViews.Profiler, TelemetryActions.ProfilerSessionStopFailed, {
```
Benjin: There's a `sendErrorEvent()` call that you should be using instead.  Also, be sure to set the includeErrorMessage flag depending on whether there's any risk of PII or user data in the error message (this includes schema info)

extensions/mssql/src/sharedInterfaces/telemetry.ts
```diff
@@ -225,6 +226,14 @@ export enum TelemetryActions {
...
     ImportFile = "ImportFile",
     ResetState = "ResetState",
+    ProfilerSessionStarted = "ProfilerSession.Started",
+    ProfilerSessionFailed = "ProfilerSession.Failed",
+    ProfilerSessionStopped = "ProfilerSession.Stopped",
+    ProfilerExportDone = "ProfilerSession.Export.Done",
+    ProfilerCloseWarningShown = "ProfilerSession.Close.WarningShown",
+    ProfilerFilterApplied = "Profiler.Filter.Applied",
+    ProfilerBufferOverflow = "Profiler.Buffer.Overflow",
+    ProfilerSessionStopFailed = "ProfilerSession.Stop.Failed",
```
Benjin: We don't need "profiler." and "profilersession."; those basically are redundant with TelemetryView above for collecting events.

Benjin (changes requested):

Issue #21084, opened 2026-02-03
[Bug]: SQL Project is using deprecated request npm package


kburtram:
### Description

SQL Project extension has a dependency on `"request": "^2.88.2"`, which according to the npmjs is deprecated.  The related issue https://github.com/request/request/issues/3142 says that the library isn't using modern patterns, etc., and that a new module should be used.  We need to remove this package and move to a recommended alternative.

<img width="2721" height="1055" alt="Image" src="https://github.com/user-attachments/assets/e6e6682c-927e-4771-9e20-5c80ac9de0e8" />

### Steps to Reproduce

1. Run `yarn install` or look in package.json

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [x] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

Latest main branch

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #21256, opened 2026-02-19
Removed request library being deprecated

allancascante:
## Description

Changes to remove deprecated request library as described in https://github.com/microsoft/vscode-mssql/issues/21084

## Code Changes Checklist

### DIFF
--- a/extensions/sql-database-projects/src/common/httpClient.ts
+++ b/extensions/sql-database-projects/src/common/httpClient.ts
@@ -17,6 +17,7 @@ const DownloadTimeoutMs = 20000;
  * Class includes method for making http request
  */
 export class HttpClient {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/naming-convention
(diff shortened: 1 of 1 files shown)


extensions/sql-database-projects/src/common/httpClient.ts
```diff
@@ -57,61 +58,76 @@ export class HttpClient {
...
+                }
+            });
+
+            stream.on("error", (err: Error) => {
+                outputChannel?.appendLine(constants.downloadError);
+                writer.destroy();
+                reject(err);
+            });
+
+            stream.pipe(writer);
```
Benjin: Is this when the stream actually is read?  Maybe the `writer.on(close)` and `writer.on(error)` handlers should go before this to avoid any possibility of missing events.

extensions/sql-database-projects/src/common/httpClient.ts
Benjin, line 30: While you're here, can you also fix up these linter errors?  Just add an eslint-disable-nextline for the `any`s.

Issue #21237, opened 2026-02-18
[Bug]: Restore Target Database should be editable


kburtram:
### Description

The Target Database drop down in restore dialog should be editable, so a BAK file can be restored to a new database.

<img width="1091" height="1006" alt="Image" src="https://github.com/user-attachments/assets/1b6f9507-6457-4357-ab63-c6444974556e" />

### Steps to Reproduce

1. Open restore dialog

Expected: Target database is editable

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

### Environment Information

latest release candidate

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

kburtram:
This is actually worse than I thought.  If the target DB doesn't exist at all, you can't restore to an autogenerated name that makes sense.  For example, when restoring WWI DB, I have to put it in a system DB?

<img width="902" height="316" alt="Image" src="https://github.com/user-attachments/assets/b4561479-7755-4800-bc2d-95364b83a0c6" />

kburtram:
@laurenastrid1 can you see if it's possible to auto-generate a better default target name in this scenario?  The editable textbox is less important than a good default target name.

laurenastrid1:
@kburtram will look at this

PR #21278, opened 2026-02-19
Added editable target database field for restore

laurenastrid1:
## Description

editableTestDb

Added an editable target database field for restore. The changes consist of adding a combobox as a new form component type. This should be a fairly low risk change to take because the only feature that currently uses the combobox is restore.

Fixes https://github.com/microsoft/vscode-mssql/issues/21237

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/reactviews/common/constants.ts
+++ b/extensions/mssql/src/reactviews/common/constants.ts
@@ -9,7 +9,3 @@ export const dataTierApplicationsDocumentationUrl =
     "https://learn.microsoft.com/en-us/sql/tools/sql-database-projects/concepts/data-tier-applications/overview?view=sql-server-ver17";
 
 export const url = "URL";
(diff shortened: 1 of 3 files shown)


extensions/mssql/src/reactviews/common/constants.ts
```diff
@@ -9,3 +9,7 @@ export const dataTierApplicationsDocumentationUrl =
     "https://learn.microsoft.com/en-us/sql/tools/sql-database-projects/concepts/data-tier-applications/overview?view=sql-server-ver17";
 
 export const url = "URL";
+
+export const keydown = "keydown";
+
+export const enter = "Enter";
```
Benjin: Use the value in extensions/mssql/src/reactviews/common/keys.ts.

KeyDown event type can probably also be moved there

Issue #21102, opened 2026-02-05
Connection lost after saving


FraHolt:
Type: <b>Bug</b>

When saving a new query as a .sql file the connection information is lost.

Extension version: 1.39.0
VS Code version: Code 1.109.0 (bdd88df003631aaa0bcbe057cb0a940b80a476fa, 2026-02-04T02:01:38.288Z)
OS version: Windows_NT x64 10.0.22000
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz (8 x 2803)|
|GPU Status|2d_canvas: enabled<br>GPU0: VENDOR= 0x8086, DEVICE=0x9a49 [Intel(R) Iris(R) Xe Graphics], DRIVER_VENDOR=Intel, DRIVER_VERSION=30.0.101.1960 *ACTIVE*<br>GPU1: VENDOR= 0x8086, DEVICE=0x9a49 [Intel(R) Iris(R) Xe Graphics], DRIVER_VERSION=30.0.101.1960<br>GPU2: VENDOR= 0x1414, DEVICE=0x008c [Microsoft Basic Render Driver], DRIVER_VERSION=10.0.22000.2600<br>Machine model name: <br>Machine model version: <br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|39.79GB (13.03GB free)|
|Process Argv|--crash-reporter-id 396a76b2-2f8b-4660-a35d-fd477ed62d54|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368cf:30146710
binariesv615:30325510
nativeloc1:31344060
dwcopilot:31170013
dwoutputs:31242946
copilot_t_ci:31333650
e5gg6876:31282496
pythonrdcb7:31342333
6518g693:31436602
aj953862:31281341
v1:31427190
6abeh943:31336334
cloudbuttont:31379625
[...]

PR #21305, opened 2026-02-21
Fix query editor connection transfer on save

aasimkhan30:
## Description

Fixes: #21102 

This bug was mainly happening because we were trying to auto connect last active connection and transfer connection from untitled URI to the newly saved one. Which lead to errors or failures in transfer.

This PR does bigger refactoring to sqlDocumentService.

1. Connection/query-result state transfer during untitled Save As/rename was unreliable.
2. We were relying on close/save timing heuristics, which caused race conditions and state cleanup at the wrong time. 

3. We also hit runtime errors in two edge cases:
 - removeRecentlyUsed called with undefined profile (Cannot read properties of undefined (reading 'id'))
 - URI state transfer attempted when no query-result webview state existed (No query result state found for uri ...)
What Changed

## Code Changes Checklist


extensions/mssql/src/controllers/sqlDocumentService.ts
```diff
@@ -170,57 +184,68 @@ export default class SqlDocumentService implements vscode.Disposable {
...
-            // Avoid processing events before initialization is complete
-            return;
+    public async onWillSaveTextDocument(event: vscode.TextDocumentWillSaveEvent): Promise<void> {
+        const newDocumentSignature = getDocumentSignature(event.document);
+        const untitledDocumentWithSameSignature = vscode.workspace.textDocuments.find(
+            (doc) =>
+                doc.uri.scheme === LocalizedConstants.untitledScheme &&
+                getDocumentSignature(doc) === newDocumentSignature,
+        );
+        if (untitledDocumentWithSameSignature) {
```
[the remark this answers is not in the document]
aasimkhan30: Unfortunately there is no direct way to track saved documents through extensions api

extensions/mssql/src/controllers/sqlDocumentService.ts
```diff
@@ -243,8 +268,29 @@ export default class SqlDocumentService implements vscode.Disposable {
...
+            await this.waitForOngoingCreates();
+        }
 
+        /**
+         * Since there is no reliable way to detect if this open is a result of
+         * untitled document being saved to disk or being renamed, we wait for
+         * 0.5 seconds for those events to finish and check if open document is
+         * the new saved/renamed document. If it is, we skip auto-connecting
+         */
+        await new Promise((resolve) => setTimeout(resolve, 500));
```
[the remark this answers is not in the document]
aasimkhan30: That's not correct. onDidOpen is fired before the onWillSave, onWillRename.

extensions/mssql/src/controllers/queryRunner.ts
```diff
@@ -1229,6 +1231,33 @@ export default class QueryRunner {
...
 
     public updateQueryRunnerUri(oldUri: string, newUri: string): void {
+        if (oldUri === newUri) {
+            return;
+        }
+
+        const pendingPromise = this._uriToQueryPromiseMap.get(oldUri);
+        if (pendingPromise) {
+            this._uriToQueryPromiseMap.set(newUri, pendingPromise);
+            this._uriToQueryPromiseMap.delete(oldUri);
```
Benjin, line 1245: Can you comment this function a little more?  Namely, what each case represents (pending promise existing, query string not existing)

extensions/mssql/src/controllers/queryRunner.ts
```diff
@@ -1229,6 +1231,33 @@ export default class QueryRunner {
...
+
+        const queryString = this._uriToQueryStringMap.get(oldUri);
+        if (queryString !== undefined) {
+            this._uriToQueryStringMap.set(newUri, queryString);
+            this._uriToQueryStringMap.delete(oldUri);
+        }
+
+        if (this._isExecuting) {
+            // During rename/save while executing, notifications may arrive on either URI.
+            // Register both and clean up aliases on completion/cancel.
```
Benjin: "aliases" are the other URIs for the same document?

extensions/mssql/src/controllers/sqlDocumentService.ts
```diff
@@ -170,57 +194,68 @@ export default class SqlDocumentService implements vscode.Disposable {
...
+            this._uriBeingRenamedOrSaved.add(getUriKey(untitledDocumentWithSameSignature.uri));
+            this._newUriFromRenameOrSave.add(getUriKey(event.document.uri));
+            await this.updateUri(
+                getUriKey(untitledDocumentWithSameSignature.uri),
+                getUriKey(event.document.uri),
+            );
         }
-        let closedDocumentUri: string = getUriKey(doc.uri);
-        let closedDocumentUriScheme: string = doc.uri.scheme;
+    }
```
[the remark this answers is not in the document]
Benjin: This seems reasonable

extensions/mssql/src/controllers/sqlDocumentService.ts
```diff
@@ -170,57 +198,68 @@ export default class SqlDocumentService implements vscode.Disposable {
...
+     * and query runner state to the new URI of the document after it's saved.
+     * @param event The event representing the document that is about to be saved
      */
-    public async onDidCloseTextDocument(doc: vscode.TextDocument): Promise<void> {
-        if (this._connectionMgr === undefined || doc === undefined || doc.uri === undefined) {
-            // Avoid processing events before initialization is complete
-            return;
+    public async onWillSaveTextDocument(event: vscode.TextDocumentWillSaveEvent): Promise<any> {
+        const newDocumentSignature = getDocumentSignature(event.document);
+        const untitledDocumentWithSameSignature = vscode.workspace.textDocuments.find(
```
Benjin: add comment to the effect of "find which untitled document was saved since the event doesn't contain that data"

extensions/mssql/src/controllers/sqlDocumentService.ts
```diff
@@ -242,37 +281,58 @@ export default class SqlDocumentService implements vscode.Disposable {
...
+            return;
+        }
+
+        // set encoding to false
+        this._statusview?.languageFlavorChanged(docUri, Constants.mssqlProviderName);
+
+        /**
+         * Since there is no reliable way to detect if this open is a result of
+         * untitled document being saved to disk or being renamed, we wait for
+         * 0.5 seconds for those events to finish and check if open document is
```
Benjin: I wouldn't reference the exact value in the comment in case we need to change it

extensions/mssql/src/controllers/sqlDocumentService.ts
```diff
@@ -242,37 +281,58 @@ export default class SqlDocumentService implements vscode.Disposable {
...
+         * the new saved/renamed document. If it is, we skip auto-connecting
```
Benjin:
```suggestion
         * the new saved/renamed document. If it is, we skip auto-connecting because the save/rename handlers covers those cases
```

extensions/mssql/src/models/sqlOutputContentProvider.ts
```diff
@@ -687,50 +690,28 @@ export class SqlOutputContentProvider {
...
-     * any queries were executed from the untitled document, the queryrunner will be remapped to
-     * a new resuls uri based on the uri of the newly saved file.
-     * @param untitledUri   The URI of the untitled file
-     * @param savedUri  The URI of the file after it was saved
-     */
-    public onUntitledFileSaved(untitledUri: string, savedUri: string): void {
-        // If we don't have any query runners mapped to this uri, don't do anything
-        let untitledResultsUri = decodeURIComponent(untitledUri);
-        if (!this._queryResultsMap.has(untitledResultsUri)) {
+    private migrateThrottledUpdateUri(oldUri: string, newUri: string): void {
```
Benjin, line 697: Can you add a comment here that explains the throttling?

extensions/mssql/test/unit/queryRunner.test.ts
```diff
@@ -501,6 +502,73 @@ suite("Query Runner tests", () => {
...
+        const oldUri = "file:///old.sql";
+        const newUri = "file:///new.sql";
+        const queryRunner = createQueryRunner(oldUri, oldUri);
+
+        const deferred = {
+            promise: Promise.resolve(true),
+            resolve: sandbox.stub(),
+            reject: sandbox.stub(),
+        };
+        (queryRunner as any)._uriToQueryPromiseMap.set(oldUri, deferred as any);
```
Benjin: can all these `(queryRunner as any)`s be swapped for `queryRunner["_privateMethod"]` so that the type-checker actually catches issues here?

extensions/mssql/test/unit/sqlOutputContentProvider.test.ts
```diff
@@ -201,30 +198,88 @@ suite("SqlOutputProvider Tests using mocks", () => {
...
-        expect(mockMap.has(uri)).to.be.false;
-        expect(mockMap.get(newUri)?.queryRunner.isExecutingQuery).to.be.true;
-        expect(mockMap.size).to.equal(1);
-        mockMap.clear();
+    test("updateQueryRunnerUri should migrate throttled timers to new URI", async () => {
+        const oldUri = "file:///old-timer.sql";
+        const newUri = "file:///new-timer.sql";
+
+        const oldTimer = setTimeout(() => {}, 1000);
+        (contentProvider as any)._stateUpdateTimers.set(oldUri, oldTimer as any);
```
Benjin: same here, about `as any`

extensions/mssql/src/controllers/queryRunner.ts
aasimkhan30: Yup.

Benjin (approved):

PR #21320, opened 2026-02-23
Remove cred store to vscode secret store migration code. 

aasimkhan30:
## Description

We have been migrating connection creds for multiple releases. So it is safe to assume that we don't need this migration code as most creds have already been migrated to vscode's secret store from os credential store. 

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/credentialstore/icredentialstore.ts
+++ b/extensions/mssql/src/credentialstore/icredentialstore.ts
@@ -6,7 +6,7 @@
 // This code is originally from https://github.com/microsoft/vsts-vscode
 // License: https://github.com/Microsoft/vsts-vscode/blob/master/LICENSE.txt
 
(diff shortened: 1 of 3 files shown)


extensions/mssql/src/credentialstore/icredentialstore.ts
```diff
@@ -6,7 +6,10 @@
 // This code is originally from https://github.com/microsoft/vsts-vscode
 // License: https://github.com/Microsoft/vsts-vscode/blob/master/LICENSE.txt
 
-import { Credential } from "../models/contracts";
+export interface StoredCredential {
```
Benjin: Not that it really matters, but why not just leave this called `Credential`?

Benjin (dismissed):

allancascante (approved):

Issue #21246, opened 2026-02-18
[Bug]: Flat File Import Database field should default correctly


laurenastrid1:
### Description

<img width="1172" height="700" alt="Image" src="https://github.com/user-attachments/assets/aea02ee2-57f8-4dca-8304-38383a8637ea" />

When navigating to the Flat File Import dialog via a database node, the Database field incorrectly defaults to master instead of the database the user entered from.

Expected Behavior:
The Database field should default to the database the user entered from.

Actual Behavior:
The Database field defaults to master regardless of which database node was used to open the dialog.

### Steps to Reproduce

In the object explorer, right-click a database node
Select the Flat File Import option
Observe the Database field in the import dialog

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### If you selected "Other", please describe the affected area

Flat File import

### Environment Information

Version: Latest
OS: Windows

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #21321, opened 2026-02-23
Use correct db as default value in flat file

laurenastrid1:
## Description
Uses the correct default value in flat file database; if the user enters through a db node in OE, instead of defaulting to master, default to the database

Fixes https://github.com/microsoft/vscode-mssql/issues/21246

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/controllers/mainController.ts
+++ b/extensions/mssql/src/controllers/mainController.ts
@@ -1555,6 +1555,8 @@ export default class MainController implements vscode.Disposable {
                         return;
                     }
 
+                    const database = ObjectExplorerUtils.getDatabaseName(node);
+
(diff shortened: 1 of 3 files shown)


extensions/mssql/src/controllers/mainController.ts
```diff
@@ -1555,6 +1555,8 @@ export default class MainController implements vscode.Disposable {
                         return;
                     }
 
+                    const database = ObjectExplorerUtils.getDatabaseName(node);
```
aasimkhan30, line 1558: should we add a null case check here and default to master?

laurenastrid1: it does later on in the code :)

PR #21352, opened 2026-02-25
CodeAnalysis: PR4 - Read and Update sqlproject file with code analysis rules

ssreerama:
## Description

In this PR:
- We update the sql code analysis rules updates to the sqlproejct file
- We read the sqlproejcts file and get the code analysis properties and apply back to the dialog rules
- For both update and read we use the STS sqlproject api that took care of updating the sql proj file and reading the properties
- All action buttons action events are performing well
    - Revert - opens a user info dialog on revert to dacfx defaults
    - Cancel - opens unsaved changes dialog for user confirmation when dialog is dirty
    - Apply - applies the changes, but not closes the dialog
    - Ok - apply and closes the dialog

## Code Changes Checklist

Dialog_Load_save_Rules
image

### DIFF
--- a/extensions/mssql/src/configurations/config.ts
+++ b/extensions/mssql/src/configurations/config.ts
@@ -7,7 +7,7 @@ export const config = {
     service: {
         downloadUrl:
             "https://github.com/Microsoft/sqltoolsservice/releases/download/{#version#}/microsoft.sqltools.servicelayer-{#fileName#}",
-        version: "5.0.20260224.1",
+        version: "5.0.20260225.1",
         downloadFileNames: {
             Windows_64: "win-x64-net8.0.zip",
             Windows_ARM64: "win-arm64-net8.0.zip",
--- a/extensions/mssql/src/models/contracts/sqlProjects/sqlProjectsContracts.ts
+++ b/extensions/mssql/src/models/contracts/sqlProjects/sqlProjectsContracts.ts
@@ -70,6 +70,15 @@ export namespace SetDatabaseSchemaProviderRequest {
     >("sqlProjects/setDatabaseSchemaProvider");
 }
 
+export namespace UpdateCodeAnalysisRulesRequest {
+    export const type = new RequestType<
+        mssql.UpdateCodeAnalysisRulesParams,
+        mssql.UpdateCodeAnalysisRulesResult,
+        void,
+        void
+    >("sqlProjects/updateCodeAnalysisRules");
+}
+
 //#endregion
 
 //#region File/folder functions
--- a/extensions/mssql/src/services/sqlProjectsService.ts
+++ b/extensions/mssql/src/services/sqlProjectsService.ts
@@ -416,6 +416,16 @@ export class SqlProjectsService implements mssql.ISqlProjectsService {
     }
 
     /**
+     * Update SQL code analysis settings for a SQL project.
+     * @param params Includes projectUri (absolute path to the .sqlproj), rule overrides, and optional runSqlCodeAnalysis setting.
+     */
+    public async updateCodeAnalysisRules(
+        params: mssql.UpdateCodeAnalysisRulesParams,
+    ): Promise<mssql.UpdateCodeAnalysisRulesResult> {
+        return this._client.sendRequest(contracts.UpdateCodeAnalysisRulesRequest.type, params);
+    }
+
+    /**
      * Add a SQLCMD variable to a project
      * @param projectUri Absolute path of the project, including .sqlproj
      * @param name Name of the SQLCMD variable
--- a/extensions/mssql/test/unit/projectController.test.ts
+++ b/extensions/mssql/test/unit/projectController.test.ts
@@ -60,6 +60,7 @@ suite("ProjectController Tests", () => {
             defaultCollation: "SQL_Latin1_General_CP1_CI_AS",
             projectStyle: projectStyle,
             databaseSource: "Project",
+            runSqlCodeAnalysis: false,
         };
     }
 
--- a/extensions/mssql/typings/vscode-mssql.d.ts
+++ b/extensions/mssql/typings/vscode-mssql.d.ts
@@ -803,6 +803,14 @@ declare module "vscode-mssql" {
         ): Promise<ResultStatus>;
 
         /**
+         * Update SQL code analysis settings for a SQL project.
+         * @param params Includes project file path, rule overrides, and optional RunSqlCodeAnalysis setting.
+         */
+        updateCodeAnalysisRules(
+            params: UpdateCodeAnalysisRulesParams,
+        ): Promise<UpdateCodeAnalysisRulesResult>;
+
+        /**
          * Get the cross-platform compatibility status for a project
          * @param projectUri Absolute path of the project, including .sqlproj
          */
@@ -1408,6 +1416,18 @@ declare module "vscode-mssql" {
         rules: CodeAnalysisRuleInfo[];
     }
 
+    export interface CodeAnalysisRuleOverride {
+        ruleId: string;
+        severity: string;
+    }
+
+    export interface UpdateCodeAnalysisRulesParams extends SqlProjectParams {
+        rules: CodeAnalysisRuleOverride[];
+        runSqlCodeAnalysis?: boolean;
+    }
+
+    export interface UpdateCodeAnalysisRulesResult extends ResultStatus {}
+
     export interface ExportParams {
         databaseName: string;
         packageFilePath: string;
@@ -1823,6 +1843,14 @@ declare module "vscode-mssql" {
          * Database Schema Provider, in the format "Microsoft.Data.Tools.Schema.Sql.SqlXYZDatabaseSchemaProvider"
          */
         databaseSchemaProvider: string;
+        /**
+         * Whether SQL code analysis is enabled for the project
+         */
+        runSqlCodeAnalysis: boolean;
+        /**
+         * Serialized code analysis rule overrides (e.g. "+!SR0001;-SR0003"), or null if not set
+         */
+        sqlCodeAnalysisRules?: string;
     }
 
     export interface GetDatabaseReferencesResult extends ResultStatus {
--- a/localization/LCL/fr/vscode-mssql.xlf.lcl
+++ b/localization/LCL/fr/vscode-mssql.xlf.lcl
@@ -253,6 +253,15 @@
         </Str>
         <Disp Icon="Str" />
       </Item>
+      <Item ItemId=";bundle.++CODE++0303e182467081801f8b614862519a3b174c0ae84ff08b70e7d126080c3f4b95" ItemType="0" PsrId="308" Leaf="true">
+        <Str Cat="Text">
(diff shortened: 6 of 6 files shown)


localization/xliff/vscode-mssql.xlf
```diff
@@ -4703,6 +4709,9 @@
     <trans-unit id="++CODE++daee7606b339f3c339076fe2c9f372a3ff40c8ee896005d829c7481b64ca5303">
       <source xml:lang="en">Reset</source>
     </trans-unit>
+    <trans-unit id="++CODE++37ba78770ece89995a12bcd4cc2a30a8fe4e5338f09c5a6534ccbc4f272825de">
+      <source xml:lang="en">Reset to Defaults?</source>
+    </trans-unit>
```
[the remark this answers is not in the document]
ssreerama: I believe, they'll get added in a consecutive automerge request.

extensions/mssql/src/reactviews/pages/CodeAnalysis/codeAnalysis.tsx
```diff
@@ -205,12 +259,16 @@ export const CodeAnalysisDialog = () => {
         });
     };
 
+    const resetToDefaults = () => {
+        setLocalRules(dacfxStaticRules);
```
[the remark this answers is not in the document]
ssreerama: our intension is to reset the old snapshot, not to the default severities.

extensions/mssql/src/reactviews/pages/CodeAnalysis/codeAnalysis.tsx
```diff
@@ -205,12 +272,16 @@ export const CodeAnalysisDialog = () => {
...
+    };
```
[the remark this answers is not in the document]
ssreerama: I'll add this test in next PR which mainly focus on enable code analysis on build checkbox.

extensions/mssql/src/codeAnalysis/codeAnalysisWebViewController.ts
```diff
@@ -28,11 +31,23 @@ export class CodeAnalysisWebViewController extends ReactWebviewPanelController<
     CodeAnalysisState,
     CodeAnalysisReducers
 > {
+    private readonly _operationId: string;
+
+    /**
+     * Sends a telemetry error event scoped to this dialog's operationId.
+     */
+    private sendError(action: TelemetryActions, error: Error): void {
```
Benjin: Nit: put private functions at the bottom of the class.  Generally you go:

public fields
private fields
constructor
public methods
private methods

extensions/mssql/src/codeAnalysis/codeAnalysisWebViewController.ts
```diff
@@ -94,14 +111,142 @@ export class CodeAnalysisWebViewController extends ReactWebviewPanelController<
...
+                    this.panel.dispose();
+                }
+                // Update the baseline rules so the component's useEffect resets isDirty
+                return {
+                    ...state,
+                    rules: payload.rules,
+                    message: payload.closeAfterSave
+                        ? undefined
+                        : ({ message: Loc.rulesSaved, intent: "success" } as DialogMessageSpec),
+                };
```
Benjin: It's slightly better to just edit the existing state than to create a new state object every single time.  Here and elsewhere.

extensions/mssql/src/configurations/config.ts
```diff
@@ -7,7 +7,7 @@ export const config = {
     service: {
         downloadUrl:
             "https://github.com/Microsoft/sqltoolsservice/releases/download/{#version#}/microsoft.sqltools.servicelayer-{#fileName#}",
-        version: "5.0.20260224.1",
+        version: "5.0.20260225.1",
```
Benjin: Can you do the STS bump in a separate PR because it's potentially so impactful?  Be sure to include a screenshot of changes in that PR

Benjin: Also do the contracts updates/changes in that same PR instead of here, please!

extensions/mssql/src/publishProject/projectUtils.ts
```diff
@@ -413,3 +414,45 @@ export function validateSqlCmdVariables(sqlCmdVariables?: { [key: string]: strin
...
 }
+
+/**
+ * Parses the SqlCodeAnalysisRules property value from the .sqlproj into a map of
+ * shortRuleId -> severity override.
+ *
+ * IDs may be short ("SR0001") or fully qualified ("Microsoft.Rules.Data.SR0001").
+ *   "+!<id>"  = Error
+ *   "-<id>"   = Disabled  (also handles "-!<id>")
+ *   "<id>"    = Warning   (explicit Warning override, regardless of the rule's DacFx default)
```
Benjin, line 425: Where is this pattern defined?  If this is public info, add a link to a web page that explains this.

extensions/mssql/src/publishProject/projectUtils.ts
```diff
@@ -413,3 +414,45 @@ export function validateSqlCmdVariables(sqlCmdVariables?: { [key: string]: strin
...
+            continue;
+        }
+        let severity: string;
+        let rawId: string;
+        if (t.startsWith("+!")) {
+            severity = CodeAnalysisRuleSeverity.Error;
+            rawId = t.substring(2);
+        } else if (t.startsWith("-")) {
+            severity = CodeAnalysisRuleSeverity.Disabled;
+            rawId = t.replace(/^-!?/, "");
```
Benjin: Why use a regex here instead of a `substring()`?

extensions/mssql/src/reactviews/common/confirmationDialog.tsx
```diff
@@ -0,0 +1,135 @@
...
+        <Dialog
+            inertTrapFocus
+            open={open}
+            onOpenChange={(_e: unknown, data: { open: boolean }) => {
+                if (!data.open) {
+                    onClose?.();
+                }
+            }}>
+            {surface}
+        </Dialog>
```
Benjin: Why is there both a controlled and uncontrolled version of this?  Can we just always have a controlled version because it'll be hosted?

extensions/mssql/test/unit/codeAnalysisWebViewController.test.ts
```diff
@@ -134,35 +171,49 @@ suite("CodeAnalysisWebViewController Tests", () => {
...
         );
 
-        expect(createWebviewPanelStub).to.have.been.calledOnce;
-        expect(createWebviewPanelStub).to.have.been.calledWithMatch(
+        expect(createWebviewPanelStub, "createWebviewPanel should be called once").to.have.been
+            .calledOnce;
+        expect(
+            createWebviewPanelStub,
+            "createWebviewPanel should be called with correct args",
+        ).to.have.been.calledWithMatch(
```
Benjin: Can you go through all the test cases here and strip out the useless assertions?  We have no reason to check things like createWebviewPanel being called, its args, its title.

Only test/validate _functional_ things that are specifically related to this class.

extensions/mssql/src/reactviews/pages/CodeAnalysis/codeAnalysis.tsx
```diff
@@ -268,93 +339,103 @@ export const CodeAnalysisDialog = () => {
...
-                                                                data.optionValue ?? rule.severity,
+                            {groupedRuleEntries.map(([category, categoryRules]) => {
+                                const categoryCheckedState = getCategoryCheckedState(categoryRules);
+                                return (
+                                    <Fragment key={category}>
+                                        {/* Category row */}
+                                        <TableRow>
+                                            <TableCell
+                                                className={`${styles.groupHeaderCell} ${styles.categoryHeaderCellClickable}`}
+                                                onDoubleClick={() =>
```
Benjin, line 351: Is the trigger explicitly instructed to be double-clicking?  I'm not necessarily opposed to it, but I don't think we deal with double-clicks elsewhere in the UI, so it may not be easily discovered.

extensions/mssql/src/configurations/config.ts
ssreerama: absolutely! will separate them

extensions/mssql/src/codeAnalysis/codeAnalysisWebViewController.ts
ssreerama: used to update the existing state and return it, but copilot suggested to return a new state. will revert it how it used to be.

extensions/mssql/src/publishProject/projectUtils.ts
ssreerama: I'll check if we have any public doc about pattern logic. But this is how we save the code analysis rules in the sqlproj and I added the commentby extracting the logic form the sqlproj pattern.

extensions/mssql/src/reactviews/pages/CodeAnalysis/codeAnalysis.tsx
ssreerama: No, this isn’t explicitly called out as a requirement. I added it because it seemed applicable given the number of categories, and currently the only way to collapse them is via the expand/collapse icons. I’ll double‑check with the PM, and if it’s not intended, we shouldn’t add it, I can remove this in a follow‑up PR.

extensions/mssql/src/codeAnalysis/codeAnalysisWebViewController.ts
ssreerama: done, moved any private methods that are on top to the bottom

extensions/mssql/src/configurations/config.ts
ssreerama: Done, removed the STS vbump and contract changes that are coming with the version into separate PR.

extensions/mssql/src/reactviews/common/confirmationDialog.tsx
ssreerama: Sure, the rever button click was created as a trigger event. but now changed to controlled version like other actions.

extensions/mssql/src/publishProject/projectUtils.ts
```diff
@@ -413,3 +414,45 @@ export function validateSqlCmdVariables(sqlCmdVariables?: { [key: string]: strin
...
+ * Parses the SqlCodeAnalysisRules property value from the .sqlproj into a map of
+ * shortRuleId -> severity override.
+ *
+ * IDs may be short ("SR0001") or fully qualified ("Microsoft.Rules.Data.SR0001").
+ *   "+!<id>"  = Error
+ *   "-<id>"   = Disabled  (also handles "-!<id>")
+ *   "<id>"    = Warning   (explicit Warning override, regardless of the rule's DacFx default)
+ * */
+export function parseSqlprojRuleOverrides(rulesString: string): Map<string, string> {
+    const overrides = new Map<string, string>();
```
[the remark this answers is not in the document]
ssreerama: will do this improvement in follow-up PR

aasimkhan30 (approved):

Issue #21201, opened 2026-02-13
[Bug]: FilteredBuffer improvements


allancascante:
### Description

Got varios PR comments to work on: 

Make filtered buffer work over event data with parsed types instead of applying types, every time a filter is added.
https://github.com/microsoft/vscode-mssql/pull/21151#discussion_r2805730799

Make filteredBuffer extend ringBuffer:
https://github.com/microsoft/vscode-mssql/pull/21151#discussion_r2805754199

Simplify function signature:
https://github.com/microsoft/vscode-mssql/pull/21151#discussion_r2806047535

### Steps to Reproduce

Code improvements

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

code in main

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [x] I want to work on this issue

PR #21384, opened 2026-02-27
Profiler filtering types updates

allancascante:
## Description

Changes to address comments from: https://github.com/microsoft/vscode-mssql/issues/21201. Make filtered buffer work over event data with parsed types instead of applying types, every time a filter is added. And improvements to sorting function signature.
 
## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/profiler/profilerWebviewController.ts
+++ b/extensions/mssql/src/profiler/profilerWebviewController.ts
@@ -1062,8 +1062,8 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
         const viewRef = view;
         // Use the typed row converter so that column values are pre-coerced
         // to their proper runtime types (number, Date, string) based on the
-        // column's declared ColumnDataType / FilterType.  This lets the
-        // FilteredBuffer compare values directly without re-parsing.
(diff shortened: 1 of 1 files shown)


extensions/mssql/src/profiler/filteredBuffer.ts
```diff
@@ -399,19 +411,66 @@ export class FilteredBuffer<T extends IndexedRow> extends RingBuffer<T> {
...
+            return value;
+        }
+
+        switch (typeHint) {
+            case FilterTypeHint.Number: {
+                if (typeof value === "number") {
+                    return value;
+                }
+                const num = this.parseNumber(value);
+                return num !== undefined ? num : value;
```
kburtram, line 450: Should we log anything if parsing fails in these conversion blocks and/or return undefined?  Or returning the original value makes sense here?

allancascante: the code falls back to sort any of this as strings if it cannot be parsed. UI should not allow to enter any data that is not the right type for the column. This class doesn't have logging; we can discuss if we want to add logging to these scenarios in a separate PR.

PR #21417, opened 2026-03-02
Query Profiler database filter

allancascante:
## Description

Adding option to launch from Object Explorer at database level, Extended Events are supported at server level for on-prem SQL servers; to allow filtering at database level has to be done on UI only. Adding a default filter in the UI for the selected database from which the Query Profiler is launch. 

Since this is a filter at UI level, support is maintained to remove the filter if needed to allow user to see events for other databases. 

image

image

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/AGENTS.md
+++ b/extensions/mssql/AGENTS.md
@@ -65,6 +65,8 @@ yarn lint src/ test/
 
 #### Unit Tests
 
+See [test/unit/AGENTS.md](test/unit/AGENTS.md) for unit testing conventions and patterns.
+
 ```bash
 # Unit tests require VS Code download and cannot run in sandboxed environments
 # This is expected behavior - tests work in CI with proper VS Code setup


extensions/mssql/package.json
Benjin, line 1: Can you revert this and maybe try commiting the same change through github's online editor?  I want to see if that avoids whatever issue is causing all the lines to show up as changed.

Benjin: Or you can probably merge in after https://github.com/microsoft/vscode-mssql/pull/21502

Either way, we should stop accidentally thrashing the json newlines back and forth between Windows and Mac.

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -816,6 +819,32 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
...
+                enabled: true,
+                clauses: [
+                    {
+                        field: "DatabaseName",
+                        operator: FilterOperator.In,
+                        values: [databaseName],
+                    },
+                ],
+            },
+        };
```
Benjin: We should edit the existing object, not allocate a new one like this.  Slightly more performant, and easier to read intention.

extensions/mssql/test/unit/profiler/profilerController.test.ts
Benjin: Can you stub these using sinon calls instead of manually (converting through `unknown`)?

`let connectionManager: sinon.SinonStubbedInstance<ConnectionManager>;`
and
`connectionManager = sandbox.createStubInstance(ConnectionManager);`

This way, we can avoid doing type-risky things later like:

`(mockConnectionManager.connect as sinon.SinonStub).resolves(false);`

extensions/mssql/package.json
```diff
@@ -578,6 +578,11 @@
                     "when": "view == objectExplorer && viewItem =~ /\\btype=(disconnectedServer|Server)\\b/",
                     "group": "1a_MSSQL_profiler@1"
                 },
+                {
+                    "command": "mssql.profiler.launchFromDatabase",
+                    "when": "view == objectExplorer && viewItem =~ /\\btype=(Database)\\b/",
```
aasimkhan30, line 579: Do we need to show it on server level databases

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -361,10 +364,8 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
 
         // Handle session selection from webview
         this.registerReducer("selectSession", (state, payload: { sessionId: string }) => {
-            return {
-                ...state,
-                selectedSessionId: payload.sessionId,
-            };
+            state.selectedSessionId = payload.sessionId;
+            return state;
```
aasimkhan30, line 368: Hopefully, this doesn't lead to race conditions across reducers. 

Imagine reducer 1 gets called first, modifies the state, in the meanwhile reducer 2 modifies the state and returns before reducer 1. It might lead to reducer2's modification getting overridden. 

Not saying it is happening here but it might happen occasionally. 

I need to look into this.

extensions/mssql/package.json
allancascante: yes this correct if launched from server level there will be no filtering

extensions/mssql/src/profiler/profilerWebviewController.ts
allancascante: @Benjin what do you think about @aasimkhan30 comment here?

Benjin: TLDR: nothing to change as part of this feature, IMO.

We're doing exactly this almost everywhere in our webviews.  Since this reducer runs synchronously and Node executes JS on a single thread outside of the event loop, this isn't exposing itself to anything more what the RPCs naturally have.  For async reducers, there'd be a possibility, but I don't think we've seen indications of this being an issue users are hitting.  I think we can worry about needing a parallel state-tree-lock structure if it does.

Benjin: I chatted with @aasimkhan30 offline as well.  He was just wondering out loud, not saying there was an issue with your PR.

extensions/mssql/src/profiler/profilerWebviewController.ts
```diff
@@ -816,6 +799,30 @@ export class ProfilerWebviewController extends ReactWebviewPanelController<
...
+        this._initialDatabaseFilter = databaseName;
+
+        // Also update the initial state so the UI shows filter is active from the start.
+        // Use FilterOperator.In (not Equals) because the DatabaseName column is Categorical,
+        // and the column filter popover only recognizes the In operator for Categorical columns.
+        this.state.filterState = {
+            enabled: true,
+            clauses: [
+                {
+                    field: "DatabaseName",
```
Benjin: Unless `field` is a string union type, `DatabaseName` should be a const and reused.

extensions/mssql/test/unit/profiler/profilerController.test.ts
```diff
@@ -20,56 +20,61 @@ chai.use(sinonChai);
...
+        clear: sandbox.stub(),
+    } as unknown as vscode.OutputChannel;
+    Object.defineProperty(mock, "outputChannel", {
+        get: () => mockOutputChannel,
+        configurable: true,
+    });
+    mock.getConfiguration.returns({
+        get: sandbox.stub().returns(10000),
+    } as unknown as vscode.WorkspaceConfiguration);
+    return mock;
```
Benjin: Is `utils.stubVscodeWrapper()` not usable for some reason?  I don't think that does config, but you can add just that mock instead of redoing the full thing.

extensions/mssql/test/unit/profiler/profilerController.test.ts
```diff
@@ -78,23 +83,26 @@ function createMockConnectionManager(sandbox: sinon.SinonSandbox): ConnectionMan
...
-    return {
-        connectionStore,
-        connectionUI,
-        connect: sandbox.stub().resolves(true),
-        disconnect: sandbox.stub().resolves(),
-        getConnectionCredentials: sandbox.stub().returns({}),
-    } as unknown as ConnectionManager;
+    Object.defineProperty(mock, "connectionStore", {
+        get: () => connectionStore,
+        configurable: true,
```
Benjin: what does `configurable` do/is it needed?

`sandbox.stub(connectionManager, "connectionStore").get(() => connectionStore)` should be enough

allancascante: removed the configurable you are right this is not needed.

Issue #20880, opened 2026-01-05
[Bug]: Always Encrypted still does not work properly


antoniotbtech:
### Description

This issue was previously reported on #1656 but this was not fully fixed.

Database is in azure with column encryption and the keys are in keyvault. There are no security enclaves and no Entra. Access to the database is made with username and password

### Steps to Reproduce

1. Create new database connection.
2. Select Parameters
3. Put the server name
4. Authentication Type: "SQL Login"
5. Fill username/password
6. Select database name
7. On advance settings set "Always Encrypted" to Enabled
8. Connect and do a select query on a table with encrypted columns
9. I get asked to "Sign in to a new account"

<img width="475" height="76" alt="Image" src="https://github.com/user-attachments/assets/3826a895-aaa0-4e1d-b9ed-334a5783ebda" />

When I click the option "Sign in to a new account" nothing happens and I see on the messages:

Started executing query at  Line 1
Query failed: Object reference not set to an instance of an object.
Total execution time: 00:00:02.545

If I select a wrong account, same issue:

Started executing query at  Line 1
Query failed: Object reference not set to an instance of an object.
Total execution time: 00:00:01.714

And after selecting the wrong account that query tab no longer asks me to login anymore because I assume the system cache the wrong credentials.

Also if by any chance on that sign in popup I press escape or click outside it will fail any subsequent queries with an error:

<img width="703" height="187" alt="Image" src="https://github.com/user-attachments/assets/5da58961-277b-488e-a54c-122f5844237f" />

I see the stop button on top but pressing it does not do anything.

I think the extension should also have a way of adding the azure accounts beforehand without the need of doing a query. Similar with what we do in azure data studio.

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [x] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

VS Code version: 1.107.1
Extension version: 1.38.0
Operation System: Windows 11
Connection Type: Azure SQL Server with SQL Login using Column encryption with Keyvault.

[...]

PR #19687, opened 2025-07-03
Fixing always encrypted column results for AKV keys. 

aasimkhan30:
# Pull Request Template – vscode-mssql

## Description

This PR fixes #1656

## Code Changes Checklist

- [ ] New or updated **unit tests** added
- [ ] All existing tests pass (`npm run test`)
- [ ] Code follows [contributing guidelines](https://github.com/microsoft/vscode-mssql/blob/main/CONTRIBUTING.md)
- [ ] Telemetry/logging updated if relevant
- [ ] No regressions or UX breakage

PR #21432, opened 2026-03-03
Fix the 'Add account' for AlwaysEncrypted sign-in

kburtram:
## Fix: Handle user cancellation in Key Vault security token flow

Fix bug https://github.com/microsoft/vscode-mssql/issues/20880 [Bug]: Always Encrypted still does not work properly

When using Always Encrypted with Key Vault, the extension prompts the user to select an Azure account and tenant via quick picks. If the user pressed **ESC** to dismiss either quick pick, the cancellation was not handled gracefully — the unresolved/rejected promise propagated to STS as a null response, resulting in:

```
Query failed: Object reference not set to an instance of an object.
```

This PR wraps `handleSecurityTokenRequest` in a try/catch so that when the user cancels the account or tenant selection, the extension returns an empty token response instead of letting a null reference reach STS. This produces a meaningful authentication error rather than a cryptic `NullReferenceException`.

Additionally, added unit tests for `createAccountQuickPickItems` covering:
- All accounts are listed with a trailing "Sign in" item
- The current account is labeled with `(Current Account)`
- An empty account list returns only the sign-in item

This is a follow-up for PR https://github.com/microsoft/vscode-mssql/pull/19687/changes#diff-0e6c1887b7d85ff49b178dd14cd29623eb657a1c48ba51bfcdd9b052ec6aeb32

### DIFF
--- a/extensions/mssql/src/controllers/connectionManager.ts
+++ b/extensions/mssql/src/controllers/connectionManager.ts
@@ -2123,6 +2123,12 @@ export default class ConnectionManager {
         return accountItems;
     }
 
+    /*
+     * Shows a quick pick to select an account. Returns the selected account, null if "Sign in to Azure" was selected,
+     * or undefined if the quick pick was dismissed.
+     * @params items The quick pick items to show
+     * @returns The selected account, null if "Sign in to Azure" was selected, or undefined if the quick pick was dismissed
+     */
     private async showAccountQuickPick(
         items: AccountQuickPickItem[],
     ): Promise<IAccount | null | undefined> {


kburtram:
@aasimkhan30 could you take a look when you have a chance.  This is a follow-up to https://github.com/microsoft/vscode-mssql/pull/19687/changes#diff-0e6c1887b7d85ff49b178dd14cd29623eb657a1c48ba51bfcdd9b052ec6aeb32 to handle the "Add account" option in the quick pick.

extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -2036,33 +2036,43 @@ export default class ConnectionManager {
...
+                    };
+                } else {
+                    this._keyVaultTokenCache.delete(JSON.stringify(params));
+                }
             }
-        }
-        const account = await this.selectAccount();
-        const tenant = await this.selectTenantId(account);
+            const account = await this.selectAccount();
+            const tenant = await this.selectTenantId(account);
```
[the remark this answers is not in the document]
kburtram: The code was structured this way to avoid the onHide handler resolve from overriding the resolve returned with null - since null is used to signal "add account" vs. ESC pressed.

extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -2036,33 +2036,43 @@ export default class ConnectionManager {
...
+            };
+        } catch (error) {
+            this._logger.error(`Security token request failed: ${getErrorMessage(error)}`);
+            // Return empty response rather than letting the error propagate
+            // to STS as a null reference
+            return {
+                accountKey: "",
+                token: "",
+            };
+        }
```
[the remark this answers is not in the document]
kburtram: The ESC path is causing other problems in STS, whether returning a null or this "dummy" object.  This was done to avoid a null ref in the SQL Driver key request callback, but even with this stub the call still fails when a valid account is not returned.  That leaves a lower-priority bug in that path, that needs to be fixed in the that repo, but the return value here doesn't affect that.

extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -2036,33 +2036,43 @@ export default class ConnectionManager {
...
-            getCloudProviderSettings(account.key.providerId).settings.azureKeyVaultResource,
-        );
+            const token = await this.azureController.getAccountSecurityToken(
+                account,
+                tenant,
+                getCloudProviderSettings(account.key.providerId).settings.azureKeyVaultResource,
+            );
 
-        this._keyVaultTokenCache.set(JSON.stringify(params), token);
+            this._keyVaultTokenCache.set(JSON.stringify(params), token);
```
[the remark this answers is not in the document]
kburtram: This PR just indented the existing code due to new nesting.  The code commented on here isn't causing problems so I'll leave as-is since it's unrelated to this bug fix.

extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -2074,6 +2084,16 @@ export default class ConnectionManager {
         const quickPickItems = this.createAccountQuickPickItems(accounts, currentAccountId);
         const selectedAccount = await this.showAccountQuickPick(quickPickItems);
 
+        // eslint-disable-next-line no-restricted-syntax
+        if (selectedAccount === null) {
```
aasimkhan30, line 2088: are we getting null or undefined here? Maybe we can fix this.showAccountQuickPick to return undefined.

extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -2105,28 +2125,39 @@ export default class ConnectionManager {
...
                     if (!selectedItem) {
                         resolve(undefined);
                         return;
                     }
 
                     const account = selectedItem.account;
-                    quickPick.dispose();
-                    resolve(account);
+                    // Return null to signal "sign in" was selected (account is undefined on that item)
+                    resolve(account ?? null); // eslint-disable-line no-restricted-syntax
```
aasimkhan30, line 2153: Should we just return undefined?

extensions/mssql/src/controllers/connectionManager.ts
kburtram: this was done intentionally to use `null` to indicate the "Add Account" item was selected, vs. `undefined` which is what is returned when the Quick Pick is escaped.  The alternative would be to add a more complex return type that had a flag to indicate why the account was undefined.

This was Claude's idea to overload undefined/null this way.  I wasn't convinced it was a great idea, but making a more complex return value also didn't seem great.

extensions/mssql/src/controllers/connectionManager.ts
kburtram: See previous comment.  Returning undefined doesn't distinguish between different paths to get to undefined, and only one path requires doing a sign-in.

aasimkhan30 (dismissed):

PR #21449, opened 2026-03-04
[Fix] Adding default sdk version incase config inspect returns null

ssreerama:
## Description

Pipeline generated VSIX file does not returns the config defaults (investigating what could be the reason). 
- Adding a default fallback version and comments to make it sync with the release version.
- Could not repro this in local run but pipeline generated vsix repro it.

## Code Changes Checklist

Before: Was able to repro with the 1.5.8version install
image

After: (locally, always working without the fallback version)
image

### DIFF
--- a/extensions/sql-database-projects/src/tools/netcoreTool.ts
+++ b/extensions/sql-database-projects/src/tools/netcoreTool.ts
@@ -17,6 +17,7 @@ import {
     DotnetInstallationConfirmation,
     NetCoreSupportedVersionInstallationConfirmation,
     UpdateDotnetLocation,
+    microsoftBuildSqlVersionKey,
 } from "../common/constants";
 import * as utils from "../common/utils";
 import { ShellCommandOptions, ShellExecutionHelper } from "./shellExecutionHelper";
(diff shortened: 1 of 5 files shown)


extensions/sql-database-projects/src/tools/netcoreTool.ts
```diff
@@ -50,9 +51,13 @@ export function getMicrosoftBuildSqlVersion(microsoftBuildSqlVersionKey: string)
     if (configured && semver.valid(configured)) {
         return configured;
     }
-    // Fall back to the default registered in package.json
-    const defaultValue = config.inspect<string>(microsoftBuildSqlVersionKey)?.defaultValue ?? "";
-    return defaultValue;
+
+    // Read the default value from package.json
+    const extension = vscode.extensions.getExtension(sqldbproj.extension.vsCodeName);
```
Benjin: I don't love this method as a work-around.  Reading config should be trivial, so let's just do this correctly instead.

extensions/sql-database-projects/src/tools/netcoreTool.ts
Benjin: Why does this method need to take in a config string?  Is it ever anything different from `sqlDatabaseProjects.microsoftBuildSqlVersion`?

extensions/sql-database-projects/src/tools/netcoreTool.ts
ssreerama: done, as per the discussion i have added the fallback version constant and comments for future understanding.

extensions/sql-database-projects/src/tools/netcoreTool.ts
ssreerama: user can set any version from vscode settings, so we are getting the version from the config file and then if the version is not semver then we fall back to default (constant now)
image

Issue #21414, opened 2026-03-02
[Bug]: Extract dacpac not working when user doesn't have LIST DATABASES


amthomas46:
### Description

<img width="1110" height="799" alt="Image" src="https://github.com/user-attachments/assets/034a4714-67ec-462a-89d6-37aaaa8a7b39" />
I have an existing connection to an azure sql database, so the db name is coming through server. but it wont let me select a database, nor will it let me click execute

It may be worth noting that I don't have access to the entire server, just one database, so I don't have another option (like to create a connection to the server).

### Steps to Reproduce

See above

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [x] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

1.40
macos

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

Benjin:
I think the query to get the list of databases from the server is failing silently because she doesn't have permission there.

Two improvements we can make here:
1. If the database property for the connection is set, add that to the dropdown list immediately since we know it's valid.
2. Add a visual spinner to indicate to the user that it's attempting to load the list of databases still (because a sleeping database can still take a moment to respond).

If my hunch is correct, this bug should be solved by #1.  #2 would just be a smoother experience for users.

Benjin:
This has been resolved for a while now; please feel free to reach out again if you're still encountering it!

PR #21491, opened 2026-03-06
Fix to load default database in dacpac dialog when user lacks permissions

allancascante:
## Description

In case user lack LIST ANY DATABASE permission (#21414) default to the connection database in dacpac operations. Also added support to show loading on database to improve user experience on 'slow' connections when trying to pull databases to show and fallback on errors to the default connection database. 

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/controllers/dacpacDialogWebviewController.ts
+++ b/extensions/mssql/src/controllers/dacpacDialogWebviewController.ts
@@ -948,6 +948,13 @@ export class DacpacDialogWebviewController extends ReactWebviewPanelController<
             return { isValid: false, errorMessage };
         }
 
+        // Check if the database matches the one from the active connection.
+        // If the user lacks permissions to list databases but is connected to a specific database,
+        // we should trust the connection's database and skip the server-side existence check.
+        const isConnectionDatabase =
+            this.state.databaseName &&
+            databaseName.toLowerCase() === this.state.databaseName.toLowerCase();
+
         // Check if database exists
         try {
             const result = await this.connectionManager.client.sendRequest(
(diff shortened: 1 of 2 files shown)


extensions/mssql/src/reactviews/pages/DacpacDialog/TargetDatabaseSection.tsx
```diff
@@ -113,19 +116,23 @@ export const TargetDatabaseSection = ({
...
-                        onOptionSelect={(_, data) => setDatabaseName(data.optionText || "")}
-                        disabled={isOperationInProgress || !ownerUri}
-                        aria-label={locConstants.dacpacDialog.databaseNameLabel}>
-                        {availableDatabases.map((db) => (
-                            <Option key={db} value={db}>
-                                {db}
-                            </Option>
-                        ))}
-                    </Dropdown>
+                    {isLoadingDatabases ? (
```
Benjin: I think we have a few possible states:
1. there's a database in the connection and we're querying the server for the rest of the databases
    * selectable dropdown and spinner
3. there's no database in the connection and we're querying the server for the databases
    * Disabled/empty dropdown visible with spinner
5. we've finished querying the database and have the list of databases loaded
    * Selectable dropdown, no spinner

Looks like this current code shows either a spinner or a dropdown but never both?

extensions/mssql/src/controllers/dacpacDialogWebviewController.ts
```diff
@@ -596,24 +596,50 @@ export class DacpacDialogWebviewController extends ReactWebviewPanelController<
...
-    private async listDatabases(ownerUri: string): Promise<{ databases: string[] }> {
+    private async listDatabases(
+        ownerUri: string,
+    ): Promise<{ databases: string[]; errorMessage?: string }> {
+        const systemDatabases = ["master", "tempdb", "model", "msdb"];
+        const stateDatabaseName = this.state.databaseName;
+
+        let errorMessage: string | undefined;
         try {
             const result = await this.connectionManager.client.sendRequest(
```
Benjin: This should just be `connectionManager.listDatabases()`.  No need to reach directly into the client.

extensions/mssql/src/reactviews/pages/DacpacDialog/TargetDatabaseSection.tsx
allancascante: made changes to populate the dropdown with the default database from the connection or the selected node for the OE from which is launched (if the diaglog is launched from a db node), only if the user clicks on the database dropdown we attempt to load the databases from the connection, in that case it shows the spinner and then updates the drop down with the found databases.

Issue #21373, opened 2026-02-26
[Bug]: SQL Notebook output is truncated to 50 rows without any warning


kburtram:
### Description

The query results get truncated at 50 rows in SQL Notebooks.  This seems like a pretty significant limitation.  But if the rows are going to be truncated there should be a message that says that it isn't the complete resultset from the query.

<img width="1829" height="1240" alt="Image" src="https://github.com/user-attachments/assets/6ee6b48d-6732-4603-ae94-4e8b584aeb4d" />

### Steps to Reproduce

1. Run a query with more than 50 rows 

Expected: All the rows are returned or the UI says it's a partial resultset

Actual: 50 rows are returned without an UI saying it's incomplete results

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

Latest

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

croblesm:
Interesting, I was not able to repro this issue. It works as expected for me:
<img width="1302" height="974" alt="Image" src="https://github.com/user-attachments/assets/0a174bd8-ae20-432a-9f0a-3accef0415d3" />

lewis-sanchez:
I'm not able to repro this bug either. It's working as expected for me too.

<img width="825" height="929" alt="Image" src="https://github.com/user-attachments/assets/3d1706dd-4cc5-439e-8b37-4c130263a0b2" />

PR #21506, opened 2026-03-06
Show warning when SQL notebook result set is incomplete

lewis-sanchez:
## Description

This PR closes https://github.com/microsoft/vscode-mssql/issues/21373

This PR shows a warning message that notifies the user when the result set in a SQL notebook is incomplete.

Here's an example of the warning message that will appear when the result set is incomplete:
image

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/test/unit/notebooks/sqlNotebookController.test.ts
+++ b/extensions/mssql/test/unit/notebooks/sqlNotebookController.test.ts
@@ -369,6 +369,88 @@ suite("SqlNotebookController", () => {
 
             expect(mockExecution.end).to.have.been.calledWith(false, sinon.match.number);
         });
+
+        test("shows truncation warning when result set is incomplete", async () => {
+            mockNotebookConnMgr.executeQueryString.resolves(
+                makeQueryResult({
+                    batches: [
+                        {
+                            batchSummary: makeBatchSummary(),
+                            messages: [],
+                            resultSets: [
+                                {
+                                    columnInfo: [makeColumn("id", "int")],
+                                    rows: [
+                                        [{ displayValue: "1", isNull: false }],
+                                        [{ displayValue: "2", isNull: false }],
+                                    ],
+                                    rowCount: 1000, // More rows exist than were returned
+                                },
+                            ],
+                            hasError: false,
+                        },
+                    ],
+                }),
+            );
+
+            const notebook = makeNotebook([{ text: "SELECT * FROM LargeTable" }]);
+            const cells = notebook.getCells();
+
+            await mockController.executeHandler(cells, notebook, mockController);
+
+            expect(mockExecution.replaceOutput).to.have.been.calledOnce;
+            const outputs = mockExecution.replaceOutput.firstCall.args[0];
+            expect(outputs).to.have.lengthOf(2);
+
+            // First output should be the truncation warning
+            const warningOutput = outputs[0];
+            expect(warningOutput.items[0].mime).to.equal("text/plain");
+            const warningText = new TextDecoder().decode(warningOutput.items[0].data);
+            expect(warningText).to.include("Warning: Result set is incomplete");
+            expect(warningText).to.include("2"); // Actual rows returned
+            expect(warningText).to.include("1000"); // Total rows available
+
+            expect(mockExecution.end).to.have.been.calledWith(true, sinon.match.number);
+        });
+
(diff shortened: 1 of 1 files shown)


extensions/mssql/src/notebooks/sqlNotebookController.ts
```diff
@@ -519,6 +519,20 @@ export class SqlNotebookController implements vscode.Disposable {
...
+                            vscode.NotebookCellOutputItem.text(
+                                LocalizedConstants.Notebooks.resultSetTruncated(
+                                    rs.rows.length,
+                                    rs.rowCount,
+                                ),
+                                MIME_TEXT_PLAIN,
+                            ),
+                        ]),
+                    );
+                }
```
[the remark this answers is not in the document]
lewis-sanchez: Unit tests were added in the last commit.

extensions/mssql/l10n/bundle.l10n.json
```diff
@@ -2002,6 +2002,13 @@
     },
     "(Command completed successfully)": "(Command completed successfully)",
     "(0 rows)": "(0 rows)",
+    "Warning: Result set is incomplete. Showing {0} of {1} rows. The full result set could not be loaded./{0} is the number of rows actually returned{1} is the total number of rows expected": {
```
ssreerama, line 2005: is the "/" intentional "/{0} is the number"

lewis-sanchez: That's auto generated. This is what I wrote and it doesn't have a "/":
image

ssreerama: my bad, this is an auto gen file!

extensions/mssql/src/notebooks/sqlNotebookController.ts
```diff
@@ -519,6 +519,20 @@ export class SqlNotebookController implements vscode.Disposable {
                     continue;
                 }
 
+                if (rs.rows.length < rs.rowCount) {
```
Benjin, line 522: What causes this to be the case?  If it's a normal flow (e.g. setting, hardcoded limit) that's truncating the number of rows, then the message should let them know that (and which setting to change).

extensions/mssql/test/unit/notebooks/sqlNotebookController.test.ts
```diff
@@ -369,6 +369,88 @@ suite("SqlNotebookController", () => {
...
+
+            // First output should be the truncation warning
+            const warningOutput = outputs[0];
+            expect(warningOutput.items[0].mime).to.equal("text/plain");
+            const warningText = new TextDecoder().decode(warningOutput.items[0].data);
+            expect(warningText).to.include("Warning: Result set is incomplete");
+            expect(warningText).to.include("2"); // Actual rows returned
+            expect(warningText).to.include("1000"); // Total rows available
+
+            expect(mockExecution.end).to.have.been.calledWith(true, sinon.match.number);
```
Benjin, line 413: Out of curiosity, what's "sinon.match.number"?

extensions/mssql/l10n/bundle.l10n.json
lewis-sanchez: It's cool. The *.xlf file is also auto generated.

extensions/mssql/test/unit/notebooks/sqlNotebookController.test.ts
lewis-sanchez: It's the sinon equivalent of TypeMoq's `It.isAnyNumber()`, basically just says that the function should be called with a number without the need to specify an exact number.

extensions/mssql/src/notebooks/sqlNotebookController.ts
lewis-sanchez: This isn't caused by a user setting. The notebook result grid renderer runs inside VS Code's sandboxed notebook output iframe, which means it cannot make RPC calls back to the extension host to lazily fetch additional rows (unlike the regular query result grid, which uses virtualized/on-demand loading). All rows must be pre-fetched and serialized into the cell output JSON at execution time.

The rowCount value comes from STS, which reads all rows from the database and reports the true total. The extension then fetches those rows via query/subset requests in 500-row pages. rs.rows.length < rs.rowCount would be true if the fetch is incomplete.

This is an architectural constraint of VS Code's notebook renderer sandboxing, and not a configurable setting.

Issue #21442, opened 2026-03-04
[Bug]: Can't use flat file import with Entra authentication


itallackpure:
### Description

When using the flat file import against a server with Entra authentication, I'm unable to import the file due to the following error:

> System.InvalidOperationException: Cannot set the AccessToken property if 'UserID', 'UID', 'Password', or 'PWD' has been specified in connection string. at Microsoft.Data.SqlClient.SqlConnection.CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessToken(SqlConnectionString connectionOptions) at Microsoft.Data.SqlClient.SqlConnection.set_AccessToken(String value) at Microsoft.SqlServer.Prose.Import.BcpProcess.CreateTableAndInsertDataIntoDb(String connectionString, Int32 batchSize, String azureAccessToken)

It seems like the connection string shouldn't contain User ID, UID, Password, PWD, or Authentication when an access token is also being provided, since SqlConnection doesn't allow both at the same time.

Is there a way to make this work?

### Steps to Reproduce

1. Connect to the database using Microsoft Entra ID authentication

<img width="601" height="564" alt="Image" src="https://github.com/user-attachments/assets/126087cb-19e1-43cd-a92e-83838021906a" />

2. Try to import a flat file and get error

<img width="1562" height="537" alt="Image" src="https://github.com/user-attachments/assets/6cc0cb3b-0276-48ec-a4ee-0e4d70f2cf9e" />

This is the value in the "Connection String" options:

<img width="554" height="315" alt="Image" src="https://github.com/user-attachments/assets/a6074add-556e-4c8c-ae15-528adb21a058" />

### Affected Area

- [x] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

### If you selected "Other", please describe the affected area

Flat file import

### Environment Information

VS Code: 1.109.5
MSSQL extension: 1.40.0
Operating system: Windows 11
Database type: Azure SQL managed instance

### Confirmation

[...]

PR #21513, opened 2026-03-06
Fix flat file import with Entra authentication (#21442)

kburtram:
Fixes https://github.com/microsoft/vscode-mssql/issues/21442. When using Entra auth, the connection string contains User ID and Authentication properties that conflict with setting AccessToken on SqlConnection. Strip conflicting auth properties from the connection string and refresh the token before sending the import request.


kburtram:
@laurenastrid1 could you please take a look at this PR when you have a moment?  Thanks!

laurenastrid1:
Thanks for fixing this bug!

PR #21523, opened 2026-03-09
Remove sql auth provider flag

aasimkhan30:
## Description

The provider has been enabled by default for a long time, and the extension now effectively depends on that mode. In particular, when the SQL Authentication Provider is disabled, the extension no longer implements the token refresh contract expected by STS (language service), which can lead to broken or inconsistent authentication behavior.

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/azure/msal/msalAzureController.ts
+++ b/extensions/mssql/src/azure/msal/msalAzureController.ts
@@ -115,8 +115,6 @@ export class MsalAzureController extends AzureController {
         this.initialized.resolve();
     }
 
-    public init(): void {}
-
     public async loadTokenCache(): Promise<void> {
         await this.clearOldCacheIfExists();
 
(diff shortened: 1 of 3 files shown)


extensions/mssql/src/azure/msal/msalAzureController.ts
```diff
@@ -116,12 +115,7 @@ export class MsalAzureController extends AzureController {
         this.initialized.resolve();
     }
 
-    public init(): void {
-        // Since this setting is only applicable to MSAL, we can enable it safely only for MSAL Controller
-        if (getEnableSqlAuthenticationProviderConfig()) {
-            this._isSqlAuthProviderEnabled = true;
-        }
-    }
+    public init(): void {}
```
Benjin: Can we just kill this instead?  Remove from base-class, too, since I'm guessing that's why that's still here as an empty function

PR #21575, opened 2026-03-11
Adding a flag to skip localization on build

allancascante:
## Description

Local builds could be a lot faster if we can skip localization in cases where we don't really need them. 

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/scripts/build.js
+++ b/extensions/mssql/scripts/build.js
@@ -1,10 +1,15 @@
 const { execSync } = require("child_process");
 
 const isProd = process.argv.includes("--prod");
+const skipLocalization = process.argv.includes("--skip-localization");
 const prodArg = isProd ? "--prod" : "";
 
 try {
-    execSync("yarn build:prepare", { stdio: "inherit" });
+    if (skipLocalization) {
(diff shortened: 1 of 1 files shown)


aasimkhan30:
@allancascante are you running into frequent yarn builds during agentic work? 

We already have steps to incrementally build the mssql extension in AGENTS.md file. 
 
https://github.com/microsoft/vscode-mssql/blob/cd694734f7f64fba64de64957c7c06a92f8406a5/extensions/mssql/AGENTS.md#L36-L53

allancascante:
> Can you convince me more that this is an issue? With yarn run watch, incremental build speed is very fast and doesn't redo localization.

I normally work various feature branches in parallel. Would you please explain what would be the workflow in case I have to switch to another branch while having yarn watch running? In my experience the flow would be make updates->manually clean out folder->build->run automated tests->execute->manually validate. From what I have seen yarn watch doesn't execute the unit tests as those are updated, so how could I make sure the tests are executed against the latest changes? In particular the current test execution with yarn test run against compiled .js version of the tests that are in out/ folder that are not updated in yarn watch.

PR #21597, opened 2026-03-12
Fix infinite loop on error in user group fetch request

laurenastrid1:
## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/azure/utils.ts
+++ b/extensions/mssql/src/azure/utils.ts
@@ -158,6 +158,7 @@ export async function fetchUserGroups(userId: string): Promise<UserGroup[]> {
             nextUrl = response.data["@odata.nextLink"];
         } catch (error) {
             console.error("Error fetching user groups:", error);
+            break; // Exit the loop on error
         }
     }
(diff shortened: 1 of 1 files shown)


extensions/mssql/src/azure/utils.ts
```diff
@@ -158,6 +158,7 @@ export async function fetchUserGroups(userId: string): Promise<UserGroup[]> {
             nextUrl = response.data["@odata.nextLink"];
         } catch (error) {
             console.error("Error fetching user groups:", error);
+            break; // Exit the loop on error
```
kburtram, line 161: I'm not familiar with the response here; but presumably response.data["@odata.nextLink"] won't work in the next iteration?  

Or in other words, if `response.data["@odata.nextLink"]` may work to get groups, should we set nextUrl in the catch block and try again?

Benjin: Two things that we need to handle here:
1. if getting the response throws
2. if parsing the response throws

Probably need multiple try-catch blocks in here

Issue #17992, opened 2024-08-26
Integrate with .Net Install Tool to use shared runtime


kburtram:
Investigate whether extension can integrate with .Net Install Tool extension (https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.vscode-dotnet-runtime) to use shared runtime for STS dependency.  The goal would be to reduce the size of the STS archive so that initial download/install time would be reduced by not including an isolated runtime in the archive.

PR #21631, opened 2026-03-14
Bundling sqltoolsserivce with extension.

aasimkhan30:
## Description
Fixes: #17992

1. Packaging is now split into two modes:
   - Offline packages are OS-specific and bundle a self-contained SQL Tools Service, so they do not depend on the .NET runtime extension.
   - The online package bundles the portable/framework-independent SQL Tools Service and depends on `ms-dotnettools.vscode-dotnet-runtime`.
   image

2. On activation, the extension now resolves SQL Tools Service in this order:
   - Use a platform-specific self-contained STS if present (offline VSIX scenario).
   - Otherwise use an existing portable STS if present and launch it using dotnet runtime from dotnet extension.
   - Otherwise download the portable STS and launch it using dotnet runtime from dotnet extension.
   - Download platform specific self-contained STS (Temporary fallback will remove when most people switch to portable).

3. When the portable STS is used, the extension acquires a .NET runtime via `dotnet.acquire` from `ms-dotnettools.vscode-dotnet-runtime`.

4. If portable STS download fails or .NET runtime acquisition fails, the user is shown an error with guidance to install the offline VSIX.
image

5. Fix downloading logger to make the download status more readable.
image

### Tasks left.
- [x] I had to refresh azure token cache to connect to Entra auth server. Need to investigate more on it.
- [x] Clean up error messages. 
- [x] Clean up service downloader util class
- [x] Clean up other services like flatfileimport to use the runtime instead of downloading their own runtime. 

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/l10n/bundle.l10n.json
+++ b/extensions/mssql/l10n/bundle.l10n.json
@@ -3370,7 +3370,7 @@
         "message": "The SQL Server extension couldn't start because its required background service failed to launch. Install the offline VSIX for your operating system, or check your network connection and try again. Details: {0}",
         "comment": ["{0} is the error message"]
     },
-    "Download Offline VSIX": "Download Offline VSIX",
+    "Download offline VSIX": "Download offline VSIX",
     "Copy link": "Copy link",
     "Link copied to clipboard": "Link copied to clipboard",
     "The {0} service has crashed. Details: {1}/{0} is the service name{1} is the error message": {
--- a/localization/xliff/vscode-mssql.xlf
+++ b/localization/xliff/vscode-mssql.xlf
@@ -1959,8 +1959,8 @@
     <trans-unit id="++CODE++831c69f02fd7c9c62450695c18f11c56b34ae3ababb1670737507a5cc01404b4">
       <source xml:lang="en">Don&apos;t show this again</source>
     </trans-unit>
-    <trans-unit id="++CODE++a0bc2e7b8d5d900226fa1ada66859fe900386a7a07112ef81b2bf296bfc75b79">
-      <source xml:lang="en">Download Offline VSIX</source>
+    <trans-unit id="++CODE++ade5b866b0e6b83f736f3f1e46c0448a8ed129ae76c552d26cd62d610f1eb0e5">
+      <source xml:lang="en">Download offline VSIX</source>
     </trans-unit>
     <trans-unit id="++CODE++342f82d48698a8da781e445917d1fc3b3b2426d771d0a96b949f0b42b428ab6f">
       <source xml:lang="en">Download the keymap extension to automatically import key mappings from Azure Data Studio. These changes will show in your keyboard shortcut preferences.</source>


.github/workflows/build-and-test.yml
```diff
@@ -113,25 +113,25 @@ jobs:
         continue-on-error: true
         run: |
           cd extensions/mssql
-          vsce package
+          yarn package
```
aasimkhan30: Using our custom packing script to bundle the portable runtime.

extensions/mssql/src/configurations/config.ts
```diff
@@ -5,22 +5,24 @@
...
             Windows_ARM64: "win-arm64-net8.0.zip",
             OSX: "osx-x64-net8.0.tar.gz",
             OSX_ARM64: "osx-arm64-net8.0.tar.gz",
             Linux: "linux-x64-net8.0.tar.gz",
             Linux_ARM64: "linux-arm64-net8.0.tar.gz",
+            Portable: "portable-net8.0.zip",
         },
         installDir: "./sqltoolsservice/{#version#}/{#platform#}",
         executableFiles: [
+            "MicrosoftSqlToolsServiceLayer.dll",
```
aasimkhan30: DLL have a  higher priority over .exe or .sh files as portable STS just has dlls.

extensions/mssql/package.json
```diff
@@ -43,7 +43,8 @@
     "main": "./dist/extension",
     "l10n": "./l10n",
     "extensionDependencies": [
-        "vscode.sql"
+        "vscode.sql",
+        "ms-dotnettools.vscode-dotnet-runtime"
```
aasimkhan30: We need a hard dependency on runtime extension as launching without it should fail extension activation.

extensions/mssql/src/languageservice/serviceclient.ts
```diff
@@ -173,219 +170,203 @@ export default class SqlToolsServiceClient {
...
-                        }
-                    }
-                    if (!overridePath) {
-                        console.warn(
-                            `Could not find valid SQL Tools Service EXE from ${JSON.stringify(exeFiles)} at ${stsRootPath}, falling back to config`,
-                        );
-                    }
-                }
+                await launchServer(sts_folder_override, Runtime.Portable);
+                return;
```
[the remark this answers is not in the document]
aasimkhan30: Intentional.

Benjin:
> When the portable STS is used, the extension acquires a .NET 8 runtime via dotnet.acquire from ms-dotnettools.vscode-dotnet-runtime.

This can be updated to .NET 10, right?

.github/workflows/build-and-test.yml
```diff
@@ -248,6 +239,46 @@ jobs:
             echo "dataworkspace_vsix_change_icon=$no_change_icon" >> $GITHUB_ENV
           fi
 
+      - name: Keymap - Calculate vsix file sizes
```
Benjin, line 242: Unrelated change, but okay to leave.  IMO, this script is large enough that we should probably check it in as a helper script and call it with the extension name as a parameter, so I'll refactor that later.

extensions/mssql/src/constants/locConstants.ts
```diff
@@ -3328,3 +3328,29 @@ export class RestoreDatabase {
...
+        "A required .NET runtime could not be found or installed.",
+    );
+    public static unableToStartService = (errorMessage: string) =>
+        l10n.t({
+            message:
+                "The SQL Server extension couldn't start because its required background service failed to launch. Install the offline VSIX for your operating system, or check your network connection and try again. Details: {0}",
+            args: [errorMessage],
+            comment: ["{0} is the error message"],
+        });
+    public static downloadOfflineVsix = l10n.t("Download Offline VSIX");
```
Benjin: I _think_ we're supposed to use sentence casing for buttons, so `Download offline VSIX`

extensions/mssql/src/controllers/mainController.ts
```diff
@@ -2636,13 +2636,7 @@ export default class MainController implements vscode.Disposable {
...
                 return;
             }
-            if (!self.canRunV2Command()) {
-                // Notify the user that this is not supported on this version
-                await this._vscodeWrapper.showErrorMessage(
-                    LocalizedConstants.macSierraRequiredErrorMessage,
-                );
-                return;
-            }
+
```
Benjin, line 2639: This reminds me: does this PR completely remove the concept of v1 STS?  If not, good opportunity to.

extensions/mssql/src/languageservice/serviceclient.ts
```diff
@@ -446,72 +466,141 @@ export default class SqlToolsServiceClient {
...
-                    this._logPath,
-                    "sqltools.log",
-                ),
-            );
+    private sendServiceLaunchTelemetry(
+        launchType: ServiceLaunchType,
+        serviceRuntime: Runtime,
+        platformInfo: PlatformInformation,
+    ): void {
+        sendActionEvent(SERVICE_LAUNCH_TELEMETRY_VIEW, TelemetryActions.ServiceStarted, {
```
Benjin, line 477: Can we also print/log?  It's nice to have confirmation of the location that STS is running from _after_ all the initialization/installation prints.  (if that isn't already happening elsewhere)

extensions/mssql/src/languageservice/serviceInstallerUtil.ts
```diff
@@ -87,53 +88,35 @@ let downloadProvider = new ServiceDownloadProvider(
...
-            resolve(downloadProvider.getOrMakeInstallDirectory(runtime));
-        }
-    });
+    try {
+        await serverProvider.downloadAndGetServerInstallFolder(runtime);
+    } catch (error) {
+        logger.error(`Error installing service for runtime ${runtime}: ${error.message}`);
+        throw error;
+    }
+    logger.verbose(`Service installation complete for runtime: ${runtime}`);
```
Benjin: include path?

extensions/mssql/src/controllers/mainController.ts
aasimkhan30: Yup, all references to v1 is removed.

PR #21642, opened 2026-03-16
Add Table Designer Integration to Edit Data

lewis-sanchez:
## Description

This PR adds a context menu option to the Edit Data grid, so that users can open Table Designer to change the table structure.

Open Table Designer

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/reactviews/pages/TableExplorer/TableDataGrid.css
+++ b/extensions/mssql/src/reactviews/pages/TableExplorer/TableDataGrid.css
@@ -135,71 +135,85 @@
     color: var(--vscode-list-activeSelectionForeground) !important;
 }
 
-/* VS Code-style context menu */
-#tableExplorerGrid .slick-context-menu {
+/* VS Code-style context menu
+   Note: SlickGrid appends the context menu to document.body (not inside #tableExplorerGrid),
+   so we target it directly via .slick-context-menu. */
+.slick-context-menu {
     background-color: var(--vscode-menu-background) !important;
     border: 1px solid var(--vscode-menu-border) !important;
-    border-radius: 5px !important;
+    border-radius: 4px !important;
     box-shadow: 0 2px 8px var(--vscode-widget-shadow) !important;
     padding: 4px 0 !important;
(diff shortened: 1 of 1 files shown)


aasimkhan30:
quick question, does the context menu work when we click in the blank space below the grid?

ghiscoding:
> quick question, does the context menu work when we click in the blank space below the grid?

hello as you may know I'm the maintainer of Slickgrid-React/Universal and after looking at the code, the answer is: **No** it doesn't work on empty section because it looks for a `.slick-cell` and then the plugin looks for cell/row numbers to retrieve data context... But, I'm not exactly sure why the original author (which isn't me) of SlickGrid added this condition, I could probably change that in the future and simply return `{ cell: -1, row: -1 }` to say that it's outside the cell range. 

https://github.com/ghiscoding/slickgrid-universal/blob/3b8b49c5e78e1beb61a4fd34d0b792571069efc3/packages/common/src/core/slickGrid.ts#L6056-L6068

```ts
protected handleContextMenu(e: Event & { target: HTMLElement }): void {
    const cell = e.target.closest('.slick-cell');
    if (!cell) {
      return;
    }

    // are we editing, if so cancel context menu
    if (this.activeCellNode === cell && this.currentEditor !== null) {
      return;
    }

    this.triggerEvent(this.onContextMenu, {}, e);
}
```

So yes I can probably fix the "open context menu from anywhere" in Slickgrid-React but that would only be applied to new major v10.x which I've pushed very recently (2 weeks ago) but I only support latest major. So, I would encourage you to migrate to v10.x eventually, there's significant improvements to accessibility (a11y) and other new features. 

See https://ghiscoding.gitbook.io/slickgrid-react/migrations/migration-to-10.x

ghiscoding:
You can now open the Context Menu from anywhere with Slickgrid-React/Universal [v10.3.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v10.3.0) and I also created a new Fluent UI Theme with a set of Fluent icons as well. I'll see if I can open a PR sometime this week for all of that... 😃 

Cheers

PR #21652, opened 2026-03-17
Add Schema Designer Integration to Edit Data

lewis-sanchez:
## Description

This PR integrates Schema Designer into Edit Data by adding a "View Table Diagram" button that opens schema designer with the table being edited and its related tables in the designer view.

Open Schema Designer from Edit Data

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/reactviews/pages/SchemaDesigner/toolbar/filterTablesButton.tsx
+++ b/extensions/mssql/src/reactviews/pages/SchemaDesigner/toolbar/filterTablesButton.tsx
@@ -234,6 +234,8 @@ export function FilterTablesButton() {
         initialFilterConsumedRef.current = false;
 
         let cancelled = false;
+        let retries = 0;
+        const MAX_RETRIES = 300;
         const applyFilter = () => {
             if (cancelled) {
                 return;
@@ -245,8 +247,12 @@ export function FilterTablesButton() {
                 setSelectedTables([...initialFilterTables]);
                 setShowTableRelationships(true);
                 context.resetView();
-            } else {
+            } else if (retries < MAX_RETRIES) {
+                retries++;
                 requestAnimationFrame(applyFilter);
+            } else {
+                // Max retries reached; mark filter as consumed so loadTables() can proceed
+                initialFilterConsumedRef.current = true;
             }
         };
         requestAnimationFrame(applyFilter);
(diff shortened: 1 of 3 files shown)


extensions/mssql/src/schemaDesigner/schemaDesignerWebviewController.ts
```diff
@@ -116,6 +116,18 @@ export class SchemaDesignerWebviewController extends ReactWebviewPanelController
         this.setupConfigurationListener();
     }
 
+    /**
+     * Sets the initial filter tables for the Schema Designer.
+     * When set, the FilterTablesButton will apply this filter after initialization.
+     * @param tables Array of fully qualified table names (e.g., ["dbo.Students"])
+     */
+    public setInitialFilterTables(tables: string[]): void {
+        this.updateState({
```
allancascante, line 125: I got comments about mutating the state from @Benjin the consensus is that in controllers it is safe to update the value no need to create a new instance every time we have to chnage the state.

allancascante (approved):
I left an small comment on mutating state on controller

extensions/mssql/src/schemaDesigner/schemaDesignerWebviewController.ts
manujoseph85, line 125: Check whether, if the schema designer is already open and you are trying to set filter with this, it is applied correctly. There was an issue with updating the view for example when the schema designer was already open

extensions/mssql/src/controllers/mainController.ts
```diff
@@ -1832,6 +1832,27 @@ export default class MainController implements vscode.Disposable {
...
+                        await SchemaDesignerWebviewManager.getInstance().getSchemaDesigner(
+                            this._context,
+                            this._vscodeWrapper,
+                            this,
+                            this.schemaDesignerService,
+                            databaseName,
+                            node,
+                        );
+
+                    schemaDesigner.setInitialFilterTables([filterTable]);
```
manujoseph85, line 1849: Do you anticipate the needs for filtering multiple tables in the future?

lewis-sanchez: No, I don't anticipate the need to filter for multiple tables initially. Schema Designer previously didn't even have a way to specify an initial filter and would just show all tables for a given schema.

That's not to say that a user can't filter by more tables though. The filter dialog supports that: 
image

extensions/mssql/src/schemaDesigner/schemaDesignerWebviewController.ts
lewis-sanchez: Filters are applied correctly.

Issue #21659, opened 2026-03-18
Edit Data filters are only applied to prefetched rows


FlavioAttolni:
### Description

When I use the “Edit Data” feature on tables with a lot of data, I can't specify a condition in the WHERE clause. The available filters only work on the visible data and the visible page, which makes them unusable.

Original text:

> Quando utilizzo l'edit data su tabelle con tanti dati non ho modo di specificare una condizione in where. I filtri che ci sono funzionano solo per i dati visibili e per pagina visibile rendendoli inutilizzabili. 

### Steps to Reproduce

1. Connessione al Database
2. Click Destro sulla tabella 
3. Edit Data

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [x] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

Ultima Versione per Windows

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

kburtram:
@lewis-sanchez could you please take a look and confirm? As the user noted, the filters are not expected to work only on prefetched results.

lewis-sanchez:
@kburtram, it's a known limitation and something that we want to address before Edit Data goes GA.

PR #21681, opened 2026-03-19
Load and Filter Data using User Specified Queries in Edit Data

lewis-sanchez:
## Description

This PR adds the ability for users to specify their own custom queries and to filter beyond what the in memory slickgrid filters currently do.

The embedded query editor features IntelliSense suggestions:
image

image

Long running queries can now be cancelled too:
Table Exp - Cancel Query

The result set correctly renders with the custom query result set:
image

image

## Code Changes Checklist


extensions/mssql/l10n/bundle.l10n.json
```diff
@@ -1263,6 +1263,10 @@
     "Close Script Pane": "Close Script Pane",
     "Modify Table": "Modify Table",
     "View Table Diagram": "View Table Diagram",
+    "Show SQL Pane": "Show SQL Pane",
+    "Script Changes": "Script Changes",
+    "Table Query": "Table Query",
```
laurenastrid1, line 1244: i feel like "Query Table" fits with show sql pane and "Script Changes" better instead of the other way around, but up to you

laurenastrid1 (dismissed):

ghiscoding:
I noticed that when you click on the "Run" script button, it reloads the entire top window, is that really expected? With Slickgrid-React you could just update the `dataset` variable and that would update the grid. Unless you really need to change all the buttons and grid? ... anyway that's just something I've observed in your animated gif. 

Actually I wonder, how would the user know that there's a context menu on that button? Will there be some kind of indicator? If not, then it might be hard for users to know that context menu exists (it's not very common on a button).

extensions/mssql/src/tableExplorer/tableExplorerWebViewController.ts
```diff
@@ -1254,6 +1286,141 @@ export class TableExplorerWebViewController extends ReactWebviewPanelController<
...
+                }
+
+                endActivity.endFailed(
+                    new Error("Failed to run custom table query"),
+                    true /* includeErrorMessage */,
+                    undefined /* errorCode */,
+                    undefined /* errorType */,
+                    {
+                        elapsedTime: (Date.now() - startTime).toString(),
+                        operationId: this.operationId,
```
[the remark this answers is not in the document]
lewis-sanchez: Error messages could potentially contain PII which is why telemetry error messages are generic. The original error messages are logged on line 1391-1392 via `this.logger.error(...)`, so users can share logs for diagnostics.

extensions/mssql/l10n/bundle.l10n.json
lewis-sanchez: I'll keep as is for now.

ghiscoding:
Fluent has Split Button, that might be better than a context menu? Just an idea... it does take a little extra width but it's probably better for a11y

image

lewis-sanchez:
> Fluent has Split Button, that might be better than a context menu? Just an idea... it does take a little extra width but it's probably better for a11y
> 
> image

I'm using a split button now. Thanks for taking the time to look at my PR and for the suggestion @ghiscoding!
image

extensions/mssql/src/tableExplorer/tableExplorerWebViewController.ts
```diff
@@ -214,13 +228,44 @@ export class TableExplorerWebViewController extends ReactWebviewPanelController<
...
+     * Builds a default SELECT query based on the current result set columns and table metadata.
+     * Format: SELECT TOP {currentRowCount} [col1], [col2], ... FROM [schema].[table]
+     */
+    private buildDefaultSelectQuery(): string {
+        const columns = this.state.resultSet?.columnInfo;
+        if (!columns || columns.length === 0) {
+            return "";
+        }
+
+        const columnList = columns.map((col) => `[${col.name.replace(/\]/g, "]]")}]`).join(", ");
```
manujoseph85: Export the escaping here as a util function and reuse it. There could be an already existing util function we could reuse as well

extensions/mssql/src/tableExplorer/tableExplorerWebViewController.ts
```diff
@@ -1254,6 +1300,173 @@ export class TableExplorerWebViewController extends ReactWebviewPanelController<
...
+                    return state;
+                }
+            }
+
+            state.loadStatus = ApiStatus.Loading;
+            this.updateState();
+
+            try {
+                // Dispose the current edit session
+                await this._tableExplorerService.dispose(state.ownerUri);
```
manujoseph85, line 1674: Re-initializing disposed service is an anti-pattern in general. Can we create a new service with the query and use it?

lewis-sanchez: The TableExplorerService isn't being disposed. The `dispose(ownerUri)` and `initialize(ownerUri, ...)` are RPC messages to STS that manage edit sessions, not the service itself.

extensions/mssql/src/webviews/common/locConstants.ts
```diff
@@ -1920,6 +1920,10 @@ export class LocConstants {
             closeScriptPane: l10n.t("Close Script Pane"),
             modifyTable: l10n.t("Modify Table"),
             viewTableDiagram: l10n.t("View Table Diagram"),
+            showSqlPane: l10n.t("Show SQL Pane"),
```
aasimkhan30, line 1950: Please make sure the label changes to "Hide SQL Pane" when it is open.

lewis-sanchez: Done

extensions/mssql/src/tableExplorer/tableExplorerWebViewController.ts
```diff
@@ -1254,6 +1301,174 @@ export class TableExplorerWebViewController extends ReactWebviewPanelController<
...
+                    undefined /* errorType */,
+                    {
+                        elapsedTime: (Date.now() - startTime).toString(),
+                        operationId: this.operationId,
+                    },
+                );
+
+                vscode.window.showErrorMessage(
+                    LocConstants.TableExplorer.failedToRunTableQuery(getErrorMessage(error)),
+                );
```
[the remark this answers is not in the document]
lewis-sanchez: There's a logger that is capturing error logs. Users can share those to assist with diagnosing issues because it gives them a chance to sanitize what they share by removing anything that could be considered identifiable.

kburtram:
The Table Query editor currently lacks some basic language service functionality. For example, auto-complete is limited, and pressing Tab moves focus out of the editor instead of inserting indentation or accepting suggestions. If we’re going to use an embedded editor here, it should behave much more like the regular query editor for core editing interactions.

kburtram:
Filtering with a where clause didn't work for me.

Data
image

Filtered
image

kburtram:
It's hanging for me like below.  Should there be an option to cancel the query?

image

kburtram:
It might be a bug in Schema Designer - but I wonder why the graph layout is like this?  Shouldn't the tables be closer to together, and possibly zoomed in a little more since there's a lot of unused space?

image

kburtram (dismissed):
I encountered a few bugs while testing.  Could you please take a look?

lewis-sanchez:
@copilot which of the md files do you first look at when you're doing a code review?

lewis-sanchez:
@copilot I just wanted an answer to my question not a draft PR.

ghiscoding:
hey guys, so I'm just curious, is this PR related to issue #21659? 
Also how does it communicate from frontend to backend? is it using SQL directly or is it using an API like OData? If so, have you seen Slickgrid-React [OData Example](https://ghiscoding.github.io/slickgrid-react-demos/#/example05), because using or creating a new Backend Service for Slickgrid-React might be useful to fix issue #21659 and I could maybe help with that

lewis-sanchez:
> hey guys, so I'm just curious, is this PR related to issue #21659? Also how does it communicate from frontend to backend? is it using SQL directly or is it using an API like OData? If so, have you seen Slickgrid-React [OData Example](https://ghiscoding.github.io/slickgrid-react-demos/#/example05), because using or creating a new Backend Service for Slickgrid-React might be useful to fix issue #21659 and I could maybe help with that

Hi @ghiscoding, yeah, it's related to that. The MSSQL extension uses the Edit Data Service found in the SQL Tools Service repo (https://github.com/microsoft/sqltoolsservice). Communication happens in two layers — the React webview communicates with the VS Code extension via JSON RPC over postMessage(), and the extension communicates with the SQL Tools Service process via JSON RPC over stdio. The backend executes direct SQL commands (INSERT/UPDATE/DELETE) against the database — there's no OData involved. That said, thanks for pointing out the Slickgrid-React OData example — I'll take a look at it.

ghiscoding:
> yeah, it's related to that. The MSSQL extension uses the Edit Data Service found in the SQL Tools Service repo (https://github.com/microsoft/sqltoolsservice). Communication happens in two layers — the React webview communicates with the VS Code extension via JSON RPC over postMessage(), and the extension communicates with the SQL Tools Service process via JSON RPC over stdio. The backend executes direct SQL commands (INSERT/UPDATE/DELETE) against the database — there's no OData involved. That said, thanks for pointing out the Slickgrid-React OData example — I'll take a look at it.

in that case we could create a new SlickGrid Backend Service (I currently have OData & GraphQL) that could produce SQL queries instead of OData, it would then use the SQL query to fetch backend data. That would be an optional package that you can install and it would work similar to the OData Example but using SQL. I'm sure copilot could help to create a new package to do just that. This would allow you to use the grid features like Filters/Sorts/Pagination and produce SQL query that you would then sent to your backend server (again it's the same principle as OData but using SQL instead)... basically the service would just give you an updated SQL query depending on what the user clicks in the grid (Filters/Sorts/...) 

EDIT 

there you go, an implementation of a new optional Backend Service for SlickGrid assisted by copilot. Just look at the SQL query being produced by the grid in the blue section above the grid. I think this would resolve issue #21659 and fix your use case too!? Again the idea is that the grid drives the backend service which updates the SQL query string. In other words, the backend service only does 1 thing, which is to give you a query string which is then sent to the backend server (which is what I do with my other 2 services OData / GraphQL and most users just love this approach). If that works for you, I can probably release this in the coming days. Also note that StartsWith/EndsWith can also be typed as "John*" or "*Doe" in any text filters (default is Contains)

EDIT 2

this is now available in v10.4.0 as a new optional package `@slickgrid-universal/sql`, you can see new [SQL](https://ghiscoding.gitbook.io/slickgrid-react/backend-services/sql-backend) docs for some more info. You can also see the new live [Example 51 - SQL Backend Service](https://ghiscoding.github.io/slickgrid-react-demos/#/example52)

msedge_ceFMDl89g7

ghiscoding:
@lewis-sanchez I seriously think that the new SQL Backend Service that I created for SlickGrid, what I described in previous comment, could be useful to your project. Are you intending to give it a try in a future PR and/or give it a try? It would be nice to get feedback for this new feature. Thanks

lewis-sanchez:
> @lewis-sanchez I seriously think that the new SQL Backend Service that I created for SlickGrid, what I described in previous comment, could be useful to your project. Are you intending to give it a try in a future PR and/or give it a try? It would be nice to get feedback for this new feature. Thanks

I'll try it out in the future. In the meantime, the custom query editor is the direction that we wanted to take the Table Explorer feature in.

kburtram:
@lewis-sanchez I'm testing this feature in the dev branch.  Here are some functional issues I ran into.  If you can't address in this PR, please create follow-up issues to fix subsequently.

1. Focus not positioned into editor window when showing the "Table View" pane.  If user shows the query, it seems likely they'll want to edit the query so moving cursor focus there would be better UX.

2. SQL isn't formatted similar to how we show generated SQL elsewhere.  Specifically the column list is very long and scroll off the screen.  It's not clear whether we need the same format as SELECT 1000, but we should line break columns at some point at least.

image

3. Formatting commands are not available in command palette

image

4. Changing TOP in SQL doesn't change the "Total rows to fetch" setting in grid.  It seems like these are setting the same value, so ideally changes to one would propogate to other.

image


5. Opening a Data Editor for one table and then rewriting query to a different table breaks grid.  It stays loading like this forever.  I'd expect the grid to update, and document title to change.

image

6. Adding a duplicate column in SQL statement breaks grid

image

7. Adding a syntax error to SQL breaks grid 
8. Syntax errors are not highlighted with syntax error diagnostic markers

image

kburtram (changes requested):
Please fix some of the functional issues I mentioned in previous comment. Please perform more extensive testing on the editor component to bring it closer to functionality parity with the normal SQL editor for basic editing scenarios.

lewis-sanchez:
> @lewis-sanchez I'm testing this feature in the dev branch. Here are some functional issues I ran into. If you can't address in this PR, please create follow-up issues to fix subsequently.
> 
> 1. Focus not positioned into editor window when showing the "Table View" pane.  If user shows the query, it seems likely they'll want to edit the query so moving cursor focus there would be better UX.
> 2. SQL isn't formatted similar to how we show generated SQL elsewhere.  Specifically the column list is very long and scroll off the screen.  It's not clear whether we need the same format as SELECT 1000, but we should line break columns at some point at least.
> 
> image
> 3. Formatting commands are not available in command palette
> 
> image
> 4. Changing TOP in SQL doesn't change the "Total rows to fetch" setting in grid.  It seems like these are setting the same value, so ideally changes to one would propogate to other.
> 
> image
> 5. Opening a Data Editor for one table and then rewriting query to a different table breaks grid.  It stays loading like this forever.  I'd expect the grid to update, and document title to change.
> 
> image
> 6. Adding a duplicate column in SQL statement breaks grid
> 
> image
> 7. Adding a syntax error to SQL breaks grid
> 8. Syntax errors are not highlighted with syntax error diagnostic markers
> 
> image

Hi @kburtram,

Here are the latest changes based on how you outlined them.

1. When the user opens the query editor, the focus is set on the editor.
image
2. Table column names are listed on separate lines.
image
3. Formatting commands are now available

Format Document

4. Changing TOP in SQL changes the "Total rows to fetch" setting in the grid.

Total Rows To Fetch Syncs with Query TOP


5. Opening a data editor for one table and then rewriting query to a different table isn't supported. I can open an issue for this, but not even the original ADS implementation supported this as shown here:
image

I did add a similar error notification in case users try to do this, so the experience matches.
image

6. Adding a duplicate column name is something that isn't supported either. I can open an issue to track this for a future enhancement. ADS would show an error message
image

And the MSSQL extension now shows a similar one and it no longer gets stuck like it used to:
image

7. Syntax errors in SQL no longer break the grid:
image

8. Syntax errors are now highlighted with syntax error diagnostic markers
image

image

ghiscoding:
also FYI, number 6 and duplicate fields, please note that SlickGrid throws an error when finding duplicate `id` in its column definition. 

The error is `[SlickGrid DataView] Each data element must implement a unique 'id' property`

It checks only the `id`, but the `field` prop can be duplicated, it's just the `id` since it's used for grouping and other things that expects a single item reference

laurenastrid1:
it would be nice to preserve (Ctrl+Shift+E) for running the query. Also, can we add padding between the "No data available" and the edge of the webview?: 

image

PR #21683, opened 2026-03-19
feat(results-grid): add Copy Column Name to column header context menu

### LINKED ISSUE #21632: [Feature Request]: Copy just header / column name from the result set
### Feature Description

It would be great shortcut to have a possibility to copy a header / column name from the result set (under mouse right click).

### Problem and Motivation

When executing query or procecedure you get know what result set it produce. Sometime I need to refer a column name in the next query.
Now, I can copy some results to clipboard using "... with headers" and extract column name then or run sp_help and find a column name on the list.
With direct copy a header name would be much faster.

### Related Area

- [ ] Query editor
- [x] Query results panel
- [ ] Object Explorer
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] SQL Database projects
- [ ] Query Plan Visualizer

### If you selected "Other", please describe the area

_No response_

### Confirmation

- [x] I have searched existing feature requests and couldn't find a match
- [ ] I want to help implement this feature

kburtram:
## Add "Copy Column Name" to column header context menu

Fixes #21632

### What this does

Adds a **Copy Column Name** item to the right-click context menu on query results grid column headers. Clicking it copies the raw column name to the clipboard.

### Why

Users frequently need to reference column names from result sets in subsequent queries. Until now the only way to get a column name was to copy data "with headers" and extract it, or run `sp_help`. This adds the obvious shortcut that was missing.

### Changes

- Added "Copy Column Name" menu item to the column header context menu in the query results grid
- Uses `navigator.clipboard.writeText` following the same clipboard pattern used elsewhere in the webview
- Added localization string for the menu item label
- Added unit tests for the new action

### Testing

- New unit test verifies correct column name is copied for a given column
- Existing context menu and copy tests pass unchanged
- Manual verification: right-click header → Copy Column Name → paste confirms raw column name

image


kburtram:
I wonder if this should be "Copy Column Name" or "Copy Header" to match the other menu text?


*FWIW, here's how Claude answers this question:*

>> The existing cell context menu items use "Headers" (plural, all columns): "Copy with Headers", "Copy All Headers". Those refer to the full header row as a concept.
But this new item is on a single column header's own right-click menu. You're copying that one column's name. So the semantics are different.
"Copy Header" is ambiguous — header of what? The whole header row? This cell's header? "Copy Column Name" is unambiguous and it's exactly what the user gets on their clipboard.
I'd stick with "Copy Column Name". It's more specific, matches the user's mental model (they're grabbing a column name to paste into a query), and avoids confusion with the existing plural "Headers" items. The original issue even uses the phrase "copy a header / column name" — the user themselves reached for "column name" as the clarifying term.

extensions/mssql/src/reactviews/pages/QueryResult/table/plugins/headerFilter.plugin.ts
```diff
@@ -174,6 +174,9 @@ export class HeaderMenu<T extends Slick.SlickData> {
                     onSubmit: (newWidth: number) => this.resizeColumn(column.id!, newWidth),
                 });
                 break;
+            case HeaderContextMenuAction.CopyColumnName:
+                await navigator.clipboard.writeText(column.name ?? "");
```
aasimkhan30: Should we escape it or not?

kburtram: I take the copilot suggestion to use the tooltip and fallback to the name; and add the try catch.  That seems reasonable to me.

aasimkhan30: Sorry, I meant should we add square brackets? [columnname]. Since people are using it in queries. 

As for copying, so far we use vscode api on the extension side to do that. We send a rpc notification to the extension and extension copies it for us.

kburtram: Yeah, from the copilot comment it seemed to imply that the tooltip was escaped; but it appears the same as the display string.  I can add a '[' ']'. 

Any reason to send an RPC to the extension to copy the string vs. navigator.clipboard.writeText(textToCopy);?

kburtram: Updated to use RPC to extension for copying to clipboard and added escaping.  I tested with this table and everything copied matching the SELECT TOP 1000 format.

```SQL

CREATE TABLE dbo.CopyColumnNameEscapeTest
(
    [Id] int NOT NULL IDENTITY(1,1),
    [NormalName] nvarchar(50) NULL,
    [First Name] nvarchar(50) NULL,
    [Last Name] nvarchar(50) NULL,
    [Order] nvarchar(50) NULL,
    [Select] nvarchar(50) NULL,
    [From] nvarchar(50) NULL,
    [Group] nvarchar(50) NULL,
    [Table] nvarchar(50) NULL,
    [Column-With-Dash] nvarchar(50) NULL,
    [Column.With.Dot] nvarchar(50) NULL,
    [Column/With/Slash] nvarchar(50) NULL,
    [Column,With,Comma] nvarchar(50) NULL,
    [Column:With:Colon] nvarchar(50) NULL,
    [Column;With;Semicolon] nvarchar(50) NULL,
    [Column?With?Question] nvarchar(50) NULL,
    [Column#With#Hash] nvarchar(50) NULL,
    [$Amount] decimal(10,2) NULL,
    [@LikeVariable] nvarchar(50) NULL,
    [#LikeTempTable] nvarchar(50) NULL,
    [123StartsWithNumber] nvarchar(50) NULL,
    [Name With  Multiple  Spaces] nvarchar(50) NULL,
    [TrailingSpace ] nvarchar(50) NULL,
    [ leadingSpace] nvarchar(50) NULL,
    [Brack]]et] nvarchar(50) NULL,
    [Quote"Name] nvarchar(50) NULL,
    [Apostrophe'Name] nvarchar(50) NULL,
    [Percent%Name] nvarchar(50) NULL,
    [Under_score] nvarchar(50) NULL,

    -- Localized / Unicode examples
    [München] nvarchar(50) NULL,
    [Ångström] nvarchar(50) NULL,
    [Español] nvarchar(50) NULL,
    [Français] nvarchar(50) NULL,
    [São Paulo] nvarchar(50) NULL,
    [İstanbul] nvarchar(50) NULL,
    [Καλημέρα] nvarchar(50) NULL,
    [Добрый день] nvarchar(50) NULL,
    [中文列名] nvarchar(50) NULL,
    [日本語列] nvarchar(50) NULL,
    [한국어열] nvarchar(50) NULL,
    [العربية] nvarchar(50) NULL,

    CONSTRAINT [PK_CopyColumnNameEscapeTest] PRIMARY KEY ([Id])
);
```

kburtram:
I think the wrapper to do the RPC is failing in the unit tests.  I'll take a look.

PR #21691, opened 2026-03-20
Adding global logging singleton to the extension which use vscode's better LogOutputChannel

aasimkhan30:
## Description

The old logger redirects logs to logger2 for now to reduce the change surface.  I will create follow up PR to fix old logger usage. 

As logger2 is a global logger, I am expecting a lot of code cleanup to remove passing of logger class between classes.

1. Provides color highlights. 

| Before | After |
| ------ | ------| 
|  image | image | 

2. Provides inbuilt log level picker. 
image

3. Global so we don't have to keep passing vscode wrapper or Logger between classes. 

4. Added logging to SqlDocumentService which didn't have any logging before. 

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/test/unit/tableDesigner.test.ts
+++ b/extensions/mssql/test/unit/tableDesigner.test.ts
@@ -41,6 +41,18 @@ suite("TableDesignerWebviewController tests", () => {
         stubUserSurvey(sandbox);
 
         mockVscodeWrapper = sandbox.createStubInstance(VscodeWrapper);
+        const outputChannel: vscode.OutputChannel = {
+            name: "MSSQL",
+            append: sandbox.stub(),
+            appendLine: sandbox.stub(),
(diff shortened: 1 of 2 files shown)


extensions/mssql/src/models/logger2.ts
```diff
@@ -0,0 +1,328 @@
...
+import * as vscode from "vscode";
+import * as Constants from "../constants/constants";
+import * as Utils from "./utils";
+
+export const logger2OutputChannelName = Constants.outputChannelName;
+
+/**
+ * VS Code log-channel-style logger with support for MSSQL tracing and PII settings.
+ */
+export interface ILogger2 {
```
kburtram, line 15: any reason to call it ILogger2 vs. just making this the new ILogger?  Since it's not a public interface I don't think we have back compat concerns.

PR #21742, opened 2026-03-25
Fix Windows npm bootstrap script for windows. 

aasimkhan30:
## Description
switching to launc npm.cmd with cmd instead of directly which leads to script not launching. 

## Code Changes Checklist

### DIFF
--- a/scripts/bootstrap-install.mjs
+++ b/scripts/bootstrap-install.mjs
@@ -4,7 +4,6 @@ import path from "node:path";
 import process from "node:process";
 import { workspaceTargets } from "./workspace-targets.mjs";
 
-const npmCommand = process.platform === "win32" ? "npm.cmd" : "npm";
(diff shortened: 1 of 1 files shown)


scripts/bootstrap-install.mjs
```diff
@@ -19,11 +18,18 @@ function installDirectory(directory) {
 
     console.log(`\n> bootstrap install :: ${directory}`);
 
-    const result = spawnSync(npmCommand, installArgs, {
+    const command = process.platform === "win32" ? "cmd.exe" : "npm";
```
Benjin, line 21: Windows can't run `npm` directly without needing `npm.cmd` (or `cmd npm.cmd`)?

aasimkhan30: I only have npm.cmd or npm.ps1 in my env.

PR #21743, opened 2026-03-25
Fixing legacy peer dependencies and npm audits

aasimkhan30:
## Description

Fixed a lot of peer legacy dependencies.
1. Updated to react 19 to match slick grid-reacts version 19
2. Updated ads.slickgrid to remove reference to unnecessary  jQuery events library.
3. Updated playwright tests
4. Remove unnecessary mocha and custom test runners from each file as we have switched to vscode/test-cli
5. All the code changes are compile issues from react 19 upgrade.

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/reactviews/pages/Profiler/profiler.tsx
+++ b/extensions/mssql/src/reactviews/pages/Profiler/profiler.tsx
@@ -446,9 +446,8 @@ export const Profiler: React.FC = () => {
 
             // Force re-render of headers after subscribing to the event
             // This is needed because headers may have already been rendered before we subscribed
-            const currentColumns = grid.getColumns();
-            if (currentColumns.length > 0) {
-                grid.setColumns(currentColumns);
+            if (grid.getVisibleColumns().length > 0) {
+                grid.updateColumns();
             }
         }
     }
@@ -469,7 +468,7 @@ export const Profiler: React.FC = () => {
             return;
         }
 
-        const gridColumns = grid.getColumns();
+        const gridColumns = grid.getVisibleColumns();
         gridColumns.forEach((column) => {
             const headerCell = headerContainer.querySelector(
                 `.slick-header-column[data-id="${column.id}"]`,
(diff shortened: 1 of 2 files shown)


extensions/mssql/package.json
```diff
@@ -118,16 +118,18 @@
...
+        "react": "^19.2.0",
+        "react-dom": "^19.2.0",
         "react-markdown": "^9.0.1",
         "react-resizable": "^3.0.5",
         "react-resizable-panels": "^3.0.2",
         "sinon": "^14.0.0",
         "sinon-chai": "^4.0.1",
-        "slickgrid": "github:Microsoft/SlickGrid.ADS#2.3.49",
+        "slickgrid": "github:Microsoft/SlickGrid.ADS#v2.3.51",
+        "slickgrid-react": "10.3.0",
```
ghiscoding, line 132: hi there, it's nice to see you're upgrading to latest Slickgrid-React, however there might be some breaking changes that impact your project, for your use case `grid.getColumns()` and `setColumns()` now include hidden columns as well (which wasn't the case before and is different compared to original SlickGrid). I think that is the only change that might (or not) impact your project

it might be better to ask copilot to do a quick pass over the project to avoid any issues by following the migration guide:
https://ghiscoding.gitbook.io/slickgrid-react/migrations/migration-to-10.x

oh wait, that's in a different package, the mssql package uses Slickgrid-React as it as a dependency and is still with v9.x

aasimkhan30: Thanks! I fixed it in the latest commit.

ghiscoding: looking good... maybe 1 last question, what's the reason to move Slickgrid-React (SlickGrid too) from a "dependencies" to a "devDependencies", is that really intended? is that an optional install?

aasimkhan30: Packages listed under dependencies are required by the extension at runtime and are included in the VSIX by the packaging tool (vsce).

Webview code where (slickgrid-react is), however, is bundled separately using esbuild and then included in the VSIX as a compiled artifact. Because of this, webview-specific dependencies are declared as devDependencies, to make vsce not package them again into the extension.

PR #21744, opened 2026-03-25
Resolve `spawn EINVAL` error in "npm run watch" on Windows

lewis-sanchez:
## Description

### Problem

Running npm run watch (or any workspace script) from the repo root fails on some Windows setups with:

```
> npm run watch
spawn EINVAL
```

#### What's going on

`scripts/workspaces.mjs` hardcodes `npm.cmd` as the npm executable on Windows:

```js
const npmCommand = process.platform === "win32" ? "npm.cmd" : "npm";
```

This doesn't work for every Windows npm installation. Depending on how Node/npm was installed (nvm-windows, fnm, winget, Scoop, etc.), the actual npm wrapper might be npm.cmd, npm.ps1, or just npm on PATH. If only npm.ps1 exists, spawn() throws EINVAL because it can't find npm.cmd.

There's a second issue too — .cmd and .ps1 files are shell scripts, not native executables. Calling spawn() with them on Windows without shell: true is unreliable (spawnSync handles this internally, but spawn does not).

## Fix

Two changes to `scripts/workspaces.mjs`:

1. Use "npm" unconditionally instead of guessing the file extension — let the OS resolve whichever wrapper is installed.
2. Pass shell: true on Windows so the shell can find and execute script-based wrappers like npm.cmd or npm.ps1.

## Code Changes Checklist

### DIFF
--- a/scripts/workspaces.mjs
+++ b/scripts/workspaces.mjs
@@ -3,7 +3,8 @@ import path from "node:path";
 import process from "node:process";
 import { supportedActions, workspaceTargets } from "./workspace-targets.mjs";
 
-const npmCommand = process.platform === "win32" ? "npm.cmd" : "npm";
+const npmCommand = "npm";
+const spawnOptions = { shell: process.platform === "win32" };
 
 function parseTargetValue(flag, value) {
     if (!value || value.trim().length === 0 || value.startsWith("-")) {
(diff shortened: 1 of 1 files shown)


lewis-sanchez:
It didn't work for me. I ran into the problem. We can hop on a call too, if you'd like to see it before I made changes.

Benjin:
@aasimkhan30 I also hit this issue on Windows and fixed it in the same way locally.

PR #21783, opened 2026-03-30
[Cleanup] Consolidate typings, remove root typings folder and ADS dependencies

ssreerama:
## Description

This PR:
- Consolidates the typing files that exists as duplicated ones for ADS usage
- Since all extensions live in same repo, we can use the typings directly form the mssql extension
- The root typings/ folder and all 5 files it contained (vscode-mssql.d.ts, azdata.d.ts, azdata.proposed.d.ts, mssql.d.ts, [dataworkspace.d.ts]), these were ADS-era artifacts no longer applicable to a standalone VS Code extension
- moved the schema compare enums to enums.ts centralized file.
- All tests are validated 

**How Things Are Linked Now**
- mssql:  owns [vscode-mssql.d.ts as the single canonical public API surface for the extension
- sql-database-projects:  references the data-workspace and mssql typings directly via relative paths; enums previously sourced from the mssql public API are now defined locally as plain enums
- data-workspace: references only the mssql typings; its own dataworkspace.d.ts is self-contained with no external type registry dependencies

## Code Changes Checklist


extensions/data-workspace/src/dataworkspace.d.ts
```diff
@@ -4,8 +4,10 @@
  *--------------------------------------------------------------------------------------------*/
 
 declare module "dataworkspace" {
-  import * as azdata from "azdata";
   import * as vscode from "vscode";
+
+  export type IconPath = string | vscode.Uri | { light: string | vscode.Uri; dark: string | vscode.Uri } | vscode.ThemeIcon;
```
Benjin: Is this already in the vscode-mssql package?  We can probably pull it from there.

extensions/data-workspace/tsconfig.json
```diff
@@ -14,13 +14,6 @@
...
     },
-    "include": [
-        "src/**/*",
-        "test/**/*",
-        "../../typings/vscode-mssql.d.ts",
-        "../../typings/dataworkspace.d.ts",
-        "../../typings/azdata.d.ts",
-        "../../typings/azdata.proposed.d.ts"
-    ],
+    "include": ["src/**/*", "test/**/*", "../mssql/typings/vscode-mssql.d.ts"],
```
Benjin, line 17: this doesn't need to reference its own `dataworkspace.d.ts` anymore?

extensions/mssql/src/enums.ts
```diff
@@ -43,3 +43,54 @@ export enum CodeAnalysisRuleSeverity {
     Warning = "Warning",
     Disabled = "Disabled",
 }
+
+/**
+ * The type of update action for a schema comparison difference.
+ * Must be kept in-sync with SchemaUpdateAction in SQL Tools Service.
+ */
+export enum SchemaUpdateAction {
```
Benjin, line 51: Thanks for consolidating all of these, Sai!

extensions/mssql/typings/vscode-mssql.d.ts
```diff
@@ -422,6 +422,42 @@ declare module "vscode-mssql" {
         containerName: string | undefined;
     }
 
+    /**
+     * Specifies the mode in which a task should be executed.
+     * Numeric values match the DacFx TaskExecutionMode enum.
+     */
+    export type TaskExecutionMode = number;
```
Benjin, line 429: Is this the best way to do this?  Or should we duplicate the regular enum, but with a `const` keyword?

extensions/mssql/typings/vscode-mssql.d.ts
```diff
@@ -2719,6 +2776,20 @@ declare module "vscode-mssql" {
         onDidChangeUriOwnership: vscode.Event<void>;
     }
 
+    /** The scripting operation to perform on a database object. */
+    export enum ScriptOperation {
```
Benjin, line 2780: Does this need to be `const`?

extensions/sql-database-projects/src/common/enums.ts
```diff
@@ -39,3 +39,39 @@ export enum TaskExecutionMode {
     script = 1,
     executeAndScript = 2,
 }
+
+/**
+ * Specifies the target folder structure when extracting a project from a database.
+ * Must be kept in sync with ExtractTarget in vscode-mssql.d.ts.
+ */
+export enum ExtractTarget {
```
Benjin, line 51: Why is `AuthenticationType` const, but the other two not?

extensions/data-workspace/tsconfig.json
ssreerama: No, src/**/* in the [include] already covers dataworkspace.d.ts automatically. It was never needed as an explicit entry

extensions/sql-database-projects/src/common/enums.ts
ssreerama: AuthenticationType is used in production code and typescript inlines thea as literals at compile time, where as other enums are numeric and passed as arguments to mssql/STS method calls at runtime. so they need to exist as real JS objects.

extensions/mssql/typings/vscode-mssql.d.ts
ssreerama: No, there is a runtime enum exists already and making this consts will be conflicted with both.

ssreerama: It is confusing to understand when to use the const and when to not use it.

Here is the clear explanation 

Rule of thumb for this codebase:

Use const enum when the enum is only ever used as a type (values are inlined at compile time, no runtime object needed).

Use regular enum when the enum needs to exist at runtime — e.g., iterating over it, or when a separate file declares its own copy of the same enum and passes values across module boundaries.

The specific design restriction here:

vscode-mssql.d.ts is a declaration file (ambient module). Any enum declared here is available to external consumers like sql-database-projects.

That extension has its own runtime copies of enums (e.g. TaskExecutionMode, ExtractTarget in enums.ts) because it can't rely on imports from vscode-mssql at runtime — it accesses the service via the extension API.

If you declare a const enum in the .d.ts, TypeScript will refuse to assign a regular enum value to it, even if the numeric values are identical. That breaks the bridge.

extensions/mssql/typings/vscode-mssql.d.ts
ssreerama: Defining this as a const (or duplicating the enum values) conflicts with the enum signatures available at runtime, which is a known limitation of the current design. To make this work, we only need the enum type declared in vscode-mssql.d.ts; the real enum implementation lives in enums.ts, and this avoids any runtime mismatch errors

PR #21808, opened 2026-04-03
Fixing logic for checking existing sts installation

aasimkhan30:
## Description

Previously, we only checked whether the STS folder existed before attempting to launch STS from it. That meant the check could still pass even if the folder was empty, and the launch would then fail later.

This PR updates that logic to verify that the actual service DLLs/executables are present before attempting to launch STS.

This didn't cause any visible failures, since our fallback logic would still recover by downloading the OS-specific STS package. However, this change helps avoid unnecessary exceptions and makes the startup path more robust.

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/test/unit/download.test.ts
+++ b/extensions/mssql/test/unit/download.test.ts
@@ -27,6 +27,27 @@ interface IFixture {
     decompressResult: Promise<void>;
 }
 
+async function writeRequiredServiceFiles(
+    installDirectory: string,
+    runtime: Runtime,
+): Promise<void> {
+    const fileExtension =
+        runtime === Runtime.Portable
+            ? ".dll"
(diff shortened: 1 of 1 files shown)


extensions/mssql/src/languageservice/serviceExecutablePaths.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * The service executables required for a SQL Tools Service installation to be considered runnable.
+ */
+export const requiredServiceExecutablePrefixes = [
+    "MicrosoftSqlToolsServiceLayer",
+    "SqlToolsResourceProviderService",
+] as const;
+
+export type ServiceExecutablePrefix = (typeof requiredServiceExecutablePrefixes)[number];
```
Benjin, line 17: This seems a bit clunky; why not just make it a string enum?

PR #21828, opened 2026-04-06
Adding Active Directory Default support

Benjin:
## Description

Adding support for [authentication via `ActiveDirectoryDefault` mode](https://learn.microsoft.com/en-us/sql/connect/ado-net/sql/azure-active-directory-authentication?view=sql-server-ver17#using-default-authentication), which instructs the MDS driver to look for a credential on its own in a number of different places, including Az powershell, Az CLI, a shared token cache, and environment credentials (which covers client ID + secret/cert and username + password).

To test, install the Az CLI, then run `az login` to sign in with an account.

image

To do:
- [x] update with SQL Tools Service build that includes https://github.com/microsoft/sqltoolsservice/pull/2646

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/controllers/connectionManager.ts
+++ b/extensions/mssql/src/controllers/connectionManager.ts
@@ -1333,6 +1333,8 @@ export default class ConnectionManager {
                 connectionSource: connectionSource,
             },
             undefined,
+            credentials,
+            undefined,
             true, // include call stack
         );
 
@@ -1404,7 +1406,7 @@ export default class ConnectionManager {
              */
             connectionActivity.endFailed(
                 error,
-                false, // Do not include error message as it might contain sensitive info
+                false, // includeErrorMessage
             );
             return false;
         }
--- a/extensions/mssql/src/controllers/queryRunner.ts
+++ b/extensions/mssql/src/controllers/queryRunner.ts
@@ -238,10 +238,12 @@ export default class QueryRunner {
         const cancelQueryActivity = startActivity(
             TelemetryViews.QueryEditor,
             TelemetryActions.CancelQuery,
-            undefined,
-            undefined,
-            undefined,
-            true, // Include call stack
+            undefined, // correlationId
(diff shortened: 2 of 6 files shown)


extensions/mssql/src/constants/locConstants.ts
```diff
@@ -902,7 +903,7 @@ export class ConnectionDialog {
     public static unsupportedAuthType(authenticationType: string) {
         return l10n.t({
             message:
-                "Unsupported authentication type in connection string: {0}. Only SQL Login, Integrated, and Azure MFA authentication are supported.",
+                "Unsupported authentication type in connection string: {0}. Only SQL Login, Integrated, Azure MFA, and Active Directory Default authentication are supported.",
```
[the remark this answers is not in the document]
Benjin: This is fine for connection strings

extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -925,9 +926,24 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
     async updateItemVisibility() {
         let hiddenProperties: (keyof IConnectionDialogProfile)[] = [];
 
+        if (
+            this.state.connectionProfile.authenticationType !== AuthenticationType.SqlLogin &&
+            this.state.connectionProfile.authenticationType !==
+                AuthenticationType.ActiveDirectoryDefault
+        ) {
+            hiddenProperties.push("user");
```
aasimkhan30, line 934: Do we need to add some checks in prepareConnectionInfo in connectionManager for this? Probably not but just want to make sure.

Benjin: I don't think there's anything we'd really want to add there.  STS will clear the userId field if the auth type is set to an incompatible value, so we're covered down there if nothing else.

PR #21863, opened 2026-04-10
feat: use new Slickgrid-React Fluent Theme & icons

ghiscoding:
## Description

I'm the maintainer of Slickgrid-React and since you use my lib in your project, I decided to create a new Fluent Theme and icons. I added a few Fluent icons as CSS classes which are shown below, if you need more icons just contact me (or you could create CSS classes yourself but it's a bit more work). I also created a new `base-fluent-grid-options.ts` file that I've put in the `src\webviews\pages` folder to avoid duplicating code, feel free to take over the PR, move the file and do whatever changes you want (this is just a template). Also for the icons, I decided to use the `.fi` prefix for icons (not to conflict with `.fui-Component`), though that could also be renamed if you wish, just say the word... 

### Notes

I did not and cannot directly test the UI since I do not have any MSSQL DB to test with. So if you could please test the UI before merging the PR. I assume that the new Slickgrid Fluent Theme would just be a plug and play but I can't be sure since I can't directly test the UI. Also note that because of the new Slickgrid Fluent Theme, all of the `.mdi` icons were replaced by the new fluent icons `.fi` as shown below (so none of the `.mdi` icons are available in this new theme, if you want more fluent icons, let me know, it's just a matter of copying the SVG path in the SASS SVG util)

You can also see a live demo via the Stackblitz below, I created a new set of demos using the Fluent React UI

[Open in StackBlitz](https://stackblitz.com/github/ghiscoding/slickgrid-react-demos/tree/main/react-fluent-ui)

cc @lewis-sanchez

### new Fluent Theme demos

Fluent SlickGrid

### available Fluent icons as CSS classes

image

## Code Changes Checklist


aasimkhan30:
Thanks for making the fix. It seems that header menu in tableDataGrid doesn't have the fluent icons.
I think baseFluentGridOption.headerMenu needs to be added there.

aasimkhan30:
Unfortunately, `autoScrollOnColumnResize` doesn’t work well for our grids because they end very close to the edge of the webview.
Since it relies on mousemove and drag events, we stop receiving those events once the cursor moves beyond the webview boundary.
To reproduce this in your demo, try resizing the browser window so the grid sits close to the edge, similar to this:

image

ghiscoding:
@aasimkhan30  hmm that's odds, the base grid option already have the fluent icons defined 

but wait, I think it's because you are redefining it also in your own grid option on that grid and when we spread an object, the last object property wins over any previous declarations.

Could you try with 

```ts
headerMenu: {
   ...baseFluentGridOption.headerMenu,

   // v10+ prefers hideCommands over hide*Command flags.
   hideCommands: ["freeze-columns"],
}
```

but for a better future proof merge, I think we could maybe use the internal fork of `node-extend` (which works the same as jQuery extend), so I think that if we pass true as the first argument it does a merge

> deep
Type: [Boolean](http://api.jquery.com/Types/#Boolean)
If true, the merge becomes recursive (aka. deep copy). Passing false for this argument is not supported.

```ts
extend<GridOption>(true, {} as GridOption, baseFluentGridOption, gridOptions) as GridOption;
```
or in your file, more like this
```ts
import { extend } from 'slickgrid-react';

const gridOptions = { /*...*/ };
setOptions(
  extend<GridOption>(true, {} as GridOption, baseFluentGridOption, gridOptions) as GridOption
);
```
if you could give that a try, otherwise, I'll give it a try later tonight after working hours

ghiscoding:
``

> Unfortunately, `autoScrollOnColumnResize` doesn’t work well for our grids because they end very close to the edge of the webview.
> Since it relies on mousemove and drag events, we stop receiving those events once the cursor moves beyond the webview boundary.
> To reproduce this in your demo, try resizing the browser window so the grid sits close to the edge, similar to this:

hmm that's a bit more tricky it seems, didn't think about that browser edge. I can ask copilot if it finds a way to make it work with that use case too.

aasimkhan30 (approved):

ghiscoding:
@aasimkhan30 are you going to do the changes that I suggested above in my previous comment https://github.com/microsoft/vscode-mssql/pull/21863#issuecomment-4282882735

or should I make a PR? 
Because the header menu missing fluent icon wasn't fix, I think, right?

aasimkhan30:
Merging this PR in. Thanks for the fixes @ghiscoding. Will have a follow up PR on fixing the header icons for table explorer.

aasimkhan30:
> @aasimkhan30 are you going to do the changes that I suggested above in my previous comment [#21863 (comment)](https://github.com/microsoft/vscode-mssql/pull/21863#issuecomment-4282882735)
> 
> or should I make a PR? Because the header menu missing fluent icon wasn't fix, I think, right?

@ghiscoding  I will take care of the fixes. I want to do some refactor to combine some inconsistencies across our slickgrid-react usage and I will take care of the missing icons in that PR

Issue #21252, opened 2026-02-18
[Bug]: Connecting to Fabric SQL instance in browser gives token error


kburtram:
### Description

Connecting to a Fabric SQL DB in the Fabric Browse control in the connection dialog gives me an error about token needing refresh.

This message does not provide a method to refresh the token, and I don't see this option in the UI.  The connection experience should handle token refresh transparently to the user.

<img width="617" height="209" alt="Image" src="https://github.com/user-attachments/assets/83f3985b-bdc3-4e66-b186-2da869762d8f" />

### Steps to Reproduce

1. Try to connect to Fabric SQL in browser with expired token.

Expected: Token is renewed automatically

```
log.ts:460   ERR [Extension Host] Error: Could not find account from MSAL Cache.
error @ log.ts:460
console.ts:139 [Extension Host] Error: Could not find account from MSAL Cache.
fMs @ console.ts:139
log.ts:460   ERR [Extension Host] Failed to acquireTokenSilent
error @ log.ts:460
console.ts:139 [Extension Host] Failed to acquireTokenSilent
fMs @ console.ts:139
log.ts:460   ERR [Extension Host] Failed to silently acquire token, not InteractionRequiredAuthError: no_account_in_silent_request: Please pass an account object, silent flow is not supported without account information
error @ log.ts:460
console.ts:139 [Extension Host] Failed to silently acquire token, not InteractionRequiredAuthError: no_account_in_silent_request: Please pass an account object, silent flow is not supported without account information
fMs @ console.ts:139
notificationsAlerts.ts:42 mssql: Error validating Entra authentication token; you may need to refresh your token.
```

### Affected Area

- [x] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

Latest release build, Windows

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #21898, opened 2026-04-14
Improving token refresh messaging and logic

Benjin:
## Description

Addresses https://github.com/microsoft/vscode-mssql/issues/21252

Adds an actionable button to refresh the Entra MFA token when prompted and improves the error message with manual instructions if that fails.

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
+++ b/extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
@@ -51,7 +51,7 @@ import { sendActionEvent, sendErrorEvent, startActivity } from "../telemetry/tel
 import { ApiStatus } from "../sharedInterfaces/webview";
 import { AzureController } from "../azure/azureController";
 import { AzureSubscription } from "@microsoft/vscode-azext-azureauth";
-import { ConnectionDetails, IConnectionInfo } from "vscode-mssql";
+import { ConnectionDetails, IConnectionInfo, IToken } from "vscode-mssql";
 import MainController from "../controllers/mainController";
 import { ObjectExplorerProvider } from "../objectExplorer/objectExplorerProvider";
 import { UserSurvey } from "../nps/userSurvey";
@@ -1670,7 +1670,9 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
     }
 
     private async getAzureActionButtons(): Promise<FormItemActionButton[]> {
+        const self = this;
         const actionButtons: FormItemActionButton[] = [];
+
         actionButtons.push({
             label: Loc.signIn,
             id: "azureSignIn",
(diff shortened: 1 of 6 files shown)


extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -1744,15 +1746,67 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
...
+
             if (account) {
                 let isTokenExpired = false;
 
+                async function refreshToken(): Promise<IToken | undefined> {
+                    const account = (
+                        await self._mainController.azureAccountService.getAccounts()
+                    ).find(
+                        (account) =>
+                            account.displayInfo.userId === self.state.connectionProfile.accountId,
```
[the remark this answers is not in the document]
Benjin: This is good feedback to investigate for later, but not necessary here.

extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -1744,15 +1746,67 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
...
+                        self.logger.error(
+                            `Account not found when attempting token refresh: ${self.state.connectionProfile.email} (${self.state.connectionProfile.accountId})`,
+                        );
+                    }
+
+                    return undefined;
+                }
+
                 try {
+                    // Check if token is expired or expiring soon...
```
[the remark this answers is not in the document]
Benjin: The underlying code will return false if the token expires in the next two minutes, so the comment is accurate.

extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
```diff
@@ -1744,15 +1746,67 @@ export class ConnectionDialogWebviewController extends FormWebviewController<
...
+                    if (account) {
+                        try {
+                            const token =
+                                await self._mainController.azureAccountService.getAccountSecurityToken(
+                                    account,
+                                    undefined,
+                                );
+
+                            if (AzureController.isTokenValid(token.token, token.expiresOn)) {
+                                self.vscodeWrapper.showInformationMessage(
```
aasimkhan30, line 1770: Do we need to show this info message. Info messages need to be manually cleared and I am worried if we spam users with these.

aasimkhan30 (approved):

extensions/mssql/src/connectionconfig/connectionDialogWebviewController.ts
aasimkhan30: There will be a single copy of the message but it still need to be manually cleared.

Benjin: I think it's okay here because it's in response to a user explicitly clicking a "refresh token" button.  It's not showing up any time the token is silently refreshed.

PR #21905, opened 2026-04-14
Dacpac Operations - Implementing message warning if list databases fails

allancascante:
## Description

In dacpac operations if list databases fails, a warning message is displayed.

image

New label:

image

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/controllers/dacpacDialogWebviewController.ts
+++ b/extensions/mssql/src/controllers/dacpacDialogWebviewController.ts
@@ -28,6 +28,7 @@ import {
     ConnectionMatcher,
 } from "../models/utils";
 import { UserSurvey } from "../nps/userSurvey";
+import { isSystemDatabase } from "../utils/databaseUtils";
 import { getErrorMessage } from "../utils/utils";
 
 // File extension constants
@@ -599,7 +600,6 @@ export class DacpacDialogWebviewController extends WebviewPanelController<
     private async listDatabases(
         ownerUri: string,
     ): Promise<{ databases: string[]; errorMessage?: string }> {
-        const systemDatabases = ["master", "tempdb", "model", "msdb"];
         const stateDatabaseName = this.state.databaseName;
 
         let errorMessage: string | undefined;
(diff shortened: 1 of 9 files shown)


Benjin:
I think this message is pretty easy to miss with the current styling and location.  Can you have it use the [MessageBar](https://storybooks.fluentui.dev/react/?path=/docs/components-messagebar--docs#intent
) component instead, and put it above the server selection (but below the operation selection)?

extensions/mssql/src/webviews/common/locConstants.ts
```diff
@@ -1903,6 +1903,9 @@ export class LocConstants {
             operationFailed: l10n.t("Operation failed"),
             unexpectedError: l10n.t("An unexpected error occurred"),
             failedToLoadDatabases: l10n.t("Failed to load databases"),
+            databasesCannotBeLoadedDueToPermissions: l10n.t(
+                "Unable to retrieve the list of databases. You may not have permission to list databases on this server. If your connection specifies a database, it will be preselected.",
```
Benjin, line 1907: If the user can't list all the databases but they can still connect directly to a database, is it possible for a dacFx operation to succeed?

I'm trying to understand if it's helpful or confusing to a user to still pre-select a database (would be helpful if DacFx could still succeed, confusing if it will always fail)

laurenastrid1:
> I think this message is pretty easy to miss with the current styling and location. Can you have it use the [MessageBar](https://storybooks.fluentui.dev/react/?path=/docs/components-messagebar--docs#intent) component instead, and put it above the server selection (but below the operation selection)?

Also, can we make it clearer that the message is a warning instead of an error? maybe adding a warning icon next to the message will signify this better

extensions/mssql/src/controllers/dacpacDialogWebviewController.ts
```diff
@@ -596,24 +596,53 @@ export class DacpacDialogWebviewController extends WebviewPanelController<
     /**
      * Lists databases on the connected server
      */
-    private async listDatabases(ownerUri: string): Promise<{ databases: string[] }> {
+    private async listDatabases(
+        ownerUri: string,
+    ): Promise<{ databases: string[]; errorMessage?: string }> {
+        const systemDatabases = ["master", "tempdb", "model", "msdb"];
```
laurenastrid1: Can we use the constant for system databases in constants.ts

extensions/mssql/src/webviews/common/locConstants.ts
allancascante: list databases is only needed for listing the databases in the form; user might still be able to use the default database to perform the operation

Benjin (approved):

PR #21944, opened 2026-04-17
Isolating tests in PR pipeline for quicker rerun

Benjin:
## Description

## Code Changes Checklist

### DIFF
--- a/build/PullRequest.yml
+++ b/build/PullRequest.yml
@@ -28,21 +28,42 @@ stages:
 
           - template: ./templates/build-database-management-keymap.yml@self
 
+          - pwsh: |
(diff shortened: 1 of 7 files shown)


Benjin:
Closing out this experiment.  I was hoping that we could easily carry the `npm restore` and `build` artifacts forward so that reruning the tests would be slimmer, but my assessment is:

1.  it's not as trivial as I hoped to get the de/rehydration of artifacts working, and
2. even if it were, the cost of de/rehydration and spinning up extra containers would negate any benefits anyway.

The better solution is to make sure that the unit tests _all_ use proper mocks (instead of relying on STS activation), and to run activation-based tests to a different suite.

PR #21974, opened 2026-04-21
Fix profiler details grid selection and tab switching

### LINKED ISSUE #21972: [Bug]: Query Profiler Details table doesn't support cell selection
### Description

The Query Profiler event Details table doesn't properly support cell selection.  The values are selected as text across cells and then when copied the values are formatted correct.

For example, here is when selection appears like across cells.

Image

And when copy/pasted the selection appears like this, with label/value split across lines, and partial text value for beginning or ending cells in selection.

Image

For reference, here is how selection works in ADS SQL Profiler with proper cell level selection.

Image

And the values copy/paste with Label/Value formatted on the same line, with the complete text for all selected cells.

Image

### Steps to Reproduce

1. Run the profiler and select the event details

### Affected Area

- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

_No response_

### Environment Information

Latest

### Confirmation

kburtram:
Fixes https://github.com/microsoft/vscode-mssql/issues/21972 - Replace the profiler Details tab's plain label/value rows with a small SlickGrid so the panel supports real cell range selection and copy behavior instead of browser text selection across cells.

Why this is the right fix:
- matches the ADS/SlickGrid interaction model users expect for profiler details
- centralizes the details-grid row, column, and option setup in a dedicated helper so the behavior is easier to reason about and test
- keeps localized Property/Value headers and profiler-specific cell styling alongside the new grid implementation
- recreates the column definitions whenever the Details tab is shown again because slickgrid-react mutates the column array on unmount; reusing the old array caused the blank grid after switching from Text back to Details
- adds focused unit coverage for the details-grid config so the selection and rendering assumptions stay locked in

Verification:
- npm run build:webviews
- npm run lint
- npx vscode-test --label "Unit Tests" --run out/test/unit/profiler/profilerDetailsGridConfig.test.js --code-version stable

### DIFF
--- a/extensions/mssql/src/webviews/common/FluentSlickGrid/fluentGridOptions.ts
+++ b/extensions/mssql/src/webviews/common/FluentSlickGrid/fluentGridOptions.ts
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { GridOption } from "slickgrid-react";
-import "./fluentSlickGrid.css";
 
 export const baseFluentGridOption: GridOption = {
     contextMenu: {
--- a/extensions/mssql/src/webviews/pages/Profiler/profilerDetailsGridConfig.ts
+++ b/extensions/mssql/src/webviews/pages/Profiler/profilerDetailsGridConfig.ts
@@ -6,6 +6,7 @@
 import { Column, GridOption } from "slickgrid-react";
 import { ProfilerEventProperty } from "../../../sharedInterfaces/profiler";
 import { ColorThemeKind } from "../../../sharedInterfaces/webview";
+import { baseFluentGridOption } from "../../common/FluentSlickGrid/fluentGridOptions";
 
 export interface ProfilerDetailsGridRow {
     id: number;
@@ -63,6 +64,7 @@ export function getProfilerDetailsGridColumns(propertyLabel: string, valueLabel:
 
 export function getProfilerDetailsGridOptions(themeKind: ColorThemeKind): GridOption {
     return {
+        ...baseFluentGridOption,
         autoFitColumnsOnFirstLoad: false,
         autoResize: {
             container: `#${PROFILER_DETAILS_GRID_CONTAINER_ID}`,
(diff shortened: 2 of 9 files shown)


extensions/mssql/src/webviews/pages/Profiler/profilerDetailsGridConfig.ts
```diff
@@ -0,0 +1,96 @@
...
+            cssClass: "profiler-details-value-cell",
+            excludeFromColumnPicker: true,
+            excludeFromGridMenu: true,
+            excludeFromHeaderMenu: true,
+        },
+    ];
+}
+
+export function getProfilerDetailsGridOptions(themeKind: ColorThemeKind): GridOption {
+    return {
```
aasimkhan30, line 66: Can please base fluent grid options from common folder for consistent theming and styling.

kburtram: Thanks!  I've updated, let me know if this is doesn't match the pattern you have in mind for adding new slickgrid controls.

extensions/mssql/src/webviews/common/FluentSlickGrid/fluentGridOptions.ts
```diff
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { GridOption } from "slickgrid-react";
-import "./fluentSlickGrid.css";
```
kburtram, line 7: I moved the fluentSlickGrid.css import out of fluentGridOptions.ts because baseFluentGridOption is now imported by profilerDetailsGridConfig, which is also imported by unit tests. Keeping the CSS side-effect import in the shared options module caused the extension-host test runner to fail when resolving the CSS import.

The shared CSS still lives in the same common location; it’s now imported by the actual webview consumers (profiler.tsx and TableDataGrid.tsx) instead. That keeps the runtime styling unchanged while making the shared options helper reusable and test-safe.

aasimkhan30 (approved):

lewis-sanchez (approved):

PR #22076, opened 2026-05-07
Speed up pre-commit localization check

aasimkhan30:
## Summary
- Scope pre-commit localization extraction to staged files by adding a `--precommit` mode to `scripts/localization-extract.js`.
- Keep existing full localization behavior unchanged for normal `npm run localization` usage.

## Performance

Measured on Windows/PowerShell from the repository root with no staged localization inputs.

| Scenario | Command / hook step | Before | After |
| --- | --- | ---: | ---: |
| Localization in pre-commit | `npm run localization` vs `node scripts/localization-extract.js --precommit` | 22.64s  | 0.68s |
| Full pre-commit hook | `.husky/pre-commit` | ~25.46s | 3.10s |
| CRLF check | `node scripts/ensure-crlf.js` | 0.52s | unchanged, already staged-file scoped |
| Lint | `npm run precommit` / `lint-staged --quiet` | 2.30s | unchanged, already staged-file scoped |

## Validation
- `node scripts/localization-extract.js --precommit`
- `.husky/pre-commit` via Git Bash `sh`
- `npx prettier --check scripts/localization-extract.js`
- `git diff --check`

### DIFF
--- a/scripts/localization-extract.js
+++ b/scripts/localization-extract.js
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 const vscodel10n = require("@vscode/l10n-dev");
+const { execFileSync } = require("child_process");
 const fs = require("fs").promises;
 const path = require("path");
 const logger = require("./terminal-logger");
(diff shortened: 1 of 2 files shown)


scripts/localization-extract.js
```diff
@@ -19,6 +20,56 @@ const EXTENSION_CONFIG = {
...
+    ],
+};
+
+const LOCALIZATION_SCRIPT_INPUTS = [
+    "scripts/localization-extract.js",
+    "scripts/file-utils.js",
+    "scripts/terminal-logger.js",
+];
+
+function getStagedFiles() {
```
Benjin, line 41: How is the linter being limited to just staged files?  something that could be shared between the two?

Benjin (approved):

PR #22196, opened 2026-05-26
Fix: Publish using a saved publish profile does not use stored values for SCLCMD Variables

### LINKED ISSUE #22030: [Bug]: Publish using a saved publish profile does not use stored values for SCLCMD Variables
### Description

The values stored for SCLCMD Variables in a saved publish profile is not used when trying to publish a fabric warehouse project. The schema compare is using the default value for the variable and not the value in the publish profile.

This works in version 1.41.0 but not in 1.42.1

### Steps to Reproduce

1. Select Publish on a database project that has some SCLCMD variables 
2. Select a publish profile that has a different value for the variables
3. The default value is still applied

### Affected Area

- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [x] SQL Database projects
- [ ] Query Plan Visualizer

_No response_

### Environment Information

VS Code: 1.117.0
SQL Server extension: 1.42.1
SQL Database projects: 1.5.9

### Confirmation

ssreerama:
## Description

Loading a publish profile in the publish dialog silently discarded all SQLCMD variable values, showing only the project defaults instead of the profile values.

**Root Cause**
Visual Studio saves .publish.xml files with a UTF-8 BOM character at the start of the file.
The dependency bump in [#21793](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) upgraded @xmldom/xmldom from 0.8.x to 0.9.x, which introduced strict XML parsing — it now rejects files where the XML declaration is not at the very first character.
The BOM shifts the declaration by one position, causing the parser to throw silently and return no variables.
This worked in 1.41.0 and broke in 1.42.1.

**Fix**

Strip the UTF-8 BOM before passing the profile content to the XML parser.
Added test coverage to the existing publish profile load test to simulate a BOM-prefixed file, which would have caught this regression.

## Code Changes Checklist

ReadSQLCMD_issue
ReadSQLCMD_issue_After


extensions/mssql/src/publishProject/projectUtils.ts
```diff
@@ -254,7 +254,12 @@ export function readSqlCmdVariables(profileText: string): { [key: string]: strin
 
     try {
         const parser = new DOMParser();
-        const xmlDoc = parser.parseFromString(profileText, "application/xml");
+        // Strip UTF-8 BOM (\uFEFF) if present — xmldom requires the XML declaration to be
+        // at position 0 and throws if a BOM character precedes it.
```
Benjin, line 259: I don't see anything in the linked issue (https://github.com/microsoft/vscode-mssql/issues/22030) that implies the issue the user's hitting is related to BOM.  Also, if BOM is causing the parser to throw, then wouldn't it fail to read _anything_ from the profiler, not just sqlcmd variables?

ssreerama: We read profile.xml for targetDb, connectionString string (using Regex to get these single elements) but for the sqlcmd variables we are using the dom xml parsing to get the multi variable, that is the reason the step is failing at read sqlcmd variables.

I found the issue related to BOM only after debugging it. customer noticed the variables are not applied in script/sc while publishing, but it is not even loading the UI in the first place with the udpated values from the profile.xml

extensions/mssql/src/publishProject/projectUtils.ts
```diff
@@ -254,7 +254,14 @@ export function readSqlCmdVariables(profileText: string): { [key: string]: strin
...
+        // Also prepend an XML declaration if absent (e.g. hand-authored profiles) to ensure consistent UTF-8 parsing.
+        let normalizedXml = profileText.replace(/^\uFEFF/, "");
+        if (!normalizedXml.trimStart().startsWith("<?xml")) {
+            normalizedXml = '<?xml version="1.0" encoding="utf-8"?>\n' + normalizedXml;
```
Benjin: instead of `\n`, we should use `os.EOL`

Benjin (dismissed):

extensions/mssql/src/publishProject/projectUtils.ts
ssreerama: Done!

Benjin (approved):

PR #22289, opened 2026-06-04
Fix macOS CI EINVAL error in sql-database-projects tests

ssreerama:
## Problem

The `sql-database-projects` extension tests fail on macOS CI runners with:

```
Error: listen EINVAL: invalid argument /Users/runner/work/1/s/vscode-mssql/extensions/sql-database-projects/.vscode-test/user-data/1.12-main.sock
```

macOS has a hard 103-character limit on Unix domain socket paths. The default `.vscode-test` user-data directory path on the CI runner is 107 characters — 4 over the limit.

The long extension directory name `sql-database-projects` (22 chars vs `mssql`'s 5 chars) pushes the path over the limit when the workspace is checked out at the ADO runner's standard path (`/Users/runner/work/1/s/vscode-mssql/`).

The failure became visible after `continueOnError: true` was removed from the sqlproj pipeline steps. (not true though, we have several successful builds after that too)

## Fix

1. Added `version: "insiders"` to match the `mssql` extension pattern
2. Redirect `--user-data-dir` to a short OS temp path via `fs.mkdtempSync`, keeping the socket path well under 103 chars regardless of workspace depth

## Testing

Buddy build / CI pipeline validation.

### DIFF
--- a/extensions/sql-database-projects/.vscode-test.mjs
+++ b/extensions/sql-database-projects/.vscode-test.mjs
@@ -1,15 +1,23 @@
 import { defineConfig } from "@vscode/test-cli";
 import { createMochaConfig, defaultCoverageConfig } from "../../scripts/vscode-test-config.mjs";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
 
 const mocha = createMochaConfig({
     timeout: 30_000,
 });
 
+// Use a short temp user-data-dir to avoid macOS's 103-char Unix socket path limit.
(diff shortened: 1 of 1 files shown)


caohai (approved):
Would love to know the delta causing this issue to happen today but not yesterday.

ssreerama:
> Would love to know the delta causing this issue to happen today but not yesterday.

latest vscode version has similar regression that few people called out today, since we are experiencing the same: https://github.com/microsoft/vscode/issues/319752

PR #22320, opened 2026-06-10
Replace deprecated AutoRest dependency with native TypeScript OpenAPI-to-SQL generator

ssreerama:
## Description

Removes the AutoRest external tool dependency from the SQL Database Projects extension and replaces it with a self-contained TypeScript implementation that generates SQL DDL files directly from an OpenAPI v2/v3 spec — no Node.js, npm, or any external process required on the user's machine.

## Code Changes Checklist


Benjin (changes requested):
I have questions about this PR; let's chat.

PR #22340, opened 2026-06-15
Schema Compare: surface platform badges and affected constraints in UI

roramsay:
## Description

Adds Microsoft Fabric Warehouse (`SqlDwUnified`) support to the Schema Compare webview:

- **"Affected constraints" banner** — rendered above the diff editor via a client-side filter on the existing `DiffEntry.Children`, matching DacFx full type names ending in `PrimaryKeyConstraint` / `ForeignKeyConstraint` / `UniqueConstraint` / `CheckConstraint` / `DefaultConstraint`.
- Localization strings added in `extensions/mssql/l10n/bundle.l10n.json` and `localization/xliff/vscode-mssql.xlf`.

Before:
Screenshot 2026-08-03 145119

After (Warehouse):
vscode after warehouse
Table Diffs now include the ALTER statement for constraints, which the line above is intended to draw attention to:
image

After SqlDb:
image

Pairs with the companion sqltoolsservice PR microsoft/sqltoolsservice#2771 (Fabric Warehouse constraint handling). The constraint banner works against the current STS build (`DiffEntry.Children` already populates for Fabric constraint diffs).

Diff for STS version upgrade (2 files, the second is just tests):
Screenshot 2026-08-05 105250

## Code Changes Checklist


roramsay:
@microsoft-github-policy-service agree company="Microsoft"

roramsay:
Added the three webview-level test scenarios from the spec (Feature 1847587 §Test design) in commit 1d4898c74:

Extracted `CompareDiffEditor.tsx`'s local helpers into a new `compareDiffEditorUtils.ts` module so the spec's three webview cases can be covered as pure-function unit tests without spinning up a React render harness for the component itself: `isConstraintObjectType`, `formatChildName`, `groupConstraintChildrenByAction`, `getAggregatedScript`, and the `CONSTRAINT_OBJECT_TYPE_SUFFIXES` list. `CompareDiffEditor.tsx` now consumes those helpers via named import; **behaviour is unchanged**.

New test suite (`extensions/mssql/test/unit/compareDiffEditorUtils.test.ts`) covers the three cases called out in the spec plus broader edge coverage:

- **compare diff editor filters `DiffEntry.Children` by constraint object-type suffix** (PK / FK / UNIQUE / CHECK / DEFAULT); column-change and index children are dropped so they do not double-count in the banner; substring vs `endsWith` guard.
- **compare diff editor renders matched constraint children grouped by `SchemaUpdateAction`** (Add / Change / Delete) for the *Constraints added / changed / dropped* banner, with source-name → target-name → `child.name` fallback for Drop-only diffs.
- **compare diff editor aggregated script walks `DiffEntry.Children`** and includes constraint `ALTER` scripts under the parent table in the right order so the Monaco diff editor shows the full picture after the sqltoolsservice `CreateDiffEntry` SqlDwUnified carve-out preserves them.

Plus defensive paths: `null` / `undefined` diff, empty children, missing scripts, `getSourceScript=false` (target pane).

**Local run:** all suites green; mssql `build:extension:typecheck` clean; `eslint` clean. 3191 tests pass total; the 5 unrelated pre-existing failures (PublishChangesDialogButton, SchemaDesigner copilot entry, Initialization, SqlToolsMcp bridge) are not touched by this change.

Linked sqltoolsservice PR with the contract + unit + integration tests: microsoft/sqltoolsservice#2771.

aasimkhan30:
@roramsay, it will be helpful if you can add before/after screenshots of the ux changes.

Benjin:
+1 to @aasimkhan30   Before/after and/or comparison screenshots between the SQL DB and Warehouse experiences - whatever's relevant to highlighting the changes you're making.

Benjin:
In the screenshots, looks like the Source and Target input widths are messed up after your changes?  Can you see what's going on there?

Benjin:
Also, the PR title says it's "surfacing platform badges", but I think those have been removed?  Probs should be updated.

extensions/mssql/src/webviews/pages/SchemaCompare/components/compareDiffEditorUtils.ts
```diff
@@ -0,0 +1,108 @@
...
+    diff: mssql.DiffEntry | undefined | null,
+    getSourceScript: boolean,
+): string => {
+    if (!diff) {
+        return "";
+    }
+    let script = "";
+    const diffScript = getSourceScript ? diff.sourceScript : diff.targetScript;
+    if (diffScript) {
+        script += diffScript + "\n\n";
```
Benjin, line 100: Better to us `os.EOL` than `\n` due to minor differences between UNIX and Windows plats

extensions/mssql/test/unit/compareDiffEditorUtils.test.ts
```diff
@@ -0,0 +1,336 @@
...
+
+            const aggregated = getAggregatedScript(diff, true);
+
+            expect(aggregated).to.contain(parentScript);
+            expect(aggregated).to.contain(pkScript);
+            expect(aggregated).to.contain(uqScript);
+            // Parent must precede its children so the executed script creates the table
+            // before adding constraints to it.
+            expect(aggregated.indexOf(parentScript)).to.be.lessThan(aggregated.indexOf(pkScript));
+            expect(aggregated.indexOf(pkScript)).to.be.lessThan(aggregated.indexOf(uqScript));
```
Benjin: Out of curiosity, does this particular check matter, or just that `uqScript` is (also) after `parentScript`?

Benjin (dismissed):

extensions/mssql/src/webviews/pages/SchemaCompare/components/compareDiffEditorUtils.ts
roramsay: I had no idea that was a thing.  I'll make the change.

roramsay: compareDiffEditorUtils is a webview module bundled for the browser so Node's os isn't available.  It fails both the webview typecheck and the esbuild bundle. There's no OS/platform concept in the webview sandbox, and this string just feeds the Monaco diff editor (which normalizes EOL itself), so I've kept \n

Benjin (approved):

PR #22348, opened 2026-06-17
Adding connection and group ordering

Benjin:
## Description

Addresses https://github.com/microsoft/vscode-mssql/issues/22461

Adds support for determining ordering of connections and groups by setting an `order` property in the config JSON.  Ordering rules are:

1. Groups always come before connections
2. Items with `order` set always come before items with it unset
3. Items with order set are sorted from lowest to highest
4. Items without `order` set or with equal `order` values are sorted alphabetically (case-insensitive)

Drag-and-drop in the UI is _not_ supported because VS Code only supports insertion-based drag-and-drop for linear list, not trees.

## Code Changes Checklist


Benjin (commented):
Add a telemetry event when reading from the config the first time if it's detected.

ssreerama (approved):

PR #22351, opened 2026-06-19
Add Move to Schema refactoring for SQL Projects

ssreerama:
## Summary

This PR adds **Move to Schema** refactoring support for SQL Projects in VS Code, allowing users to move database objects (tables, views, stored procedures, etc.) from one schema to another directly from the editor.

## Feature Overview

When working inside a SQL Project file, users can right-click on any schema-qualified object name and choose **"Move to Schema..."** from the context menu. A schema picker appears with all schemas defined in the project. After selecting a target schema, VS Code opens a **refactor preview panel** showing all the files that will change, with the ability to review and confirm before applying.

## What Changes

**Schema Migration Logic**
- Every reference to the moved object across all files in the project is updated to use the new schema name — not just the file where the object is defined.
- A `.refactorlog` entry is automatically written so that deployment tools (SqlPackage, SSDT) can track the schema move and generate the correct rename script when deploying.

**Preview and Confirmation**
- All proposed changes are shown in a VS Code diff preview panel before anything is applied.
- SQL file changes and the refactorlog update are all shown checked by default, giving the user full visibility and control.

**Validations and Guardrails**
- Moving memory-optimized tables is blocked with a clear error message, as this is not supported.
- If an object with the same name already exists in the target schema, the operation is blocked to prevent collisions.
- Schema objects themselves cannot be renamed using the standard Rename refactoring — that operation is now correctly blocked with an appropriate message.

## Testing

Unit tests cover the key validation scenarios: memory-optimized tables and name collision detection.
3_MoveToSchemaDone


aasimkhan30 (dismissed):

extensions/mssql/src/constants/locConstants.ts
```diff
@@ -4159,7 +4159,32 @@ export class SqlSymbolRename {
...
+            : l10n.t("Select the target schema");
+    public static noSchemasFound = l10n.t("No schemas were found in the project.");
+    public static noMovableSymbolAtCursor = l10n.t(
+        "No object that can be moved to another schema was found at the cursor.",
+    );
+    public static moveToSchemaRequestFailed = (message: string): string =>
+        l10n.t("Move to Schema request failed: {0}", message);
+    public static resolveRefactorLogFailed = (message: string): string =>
+        l10n.t("Failed to resolve the refactor log for this file: {0}", message);
+    public static previewLabel = (targetSchema: string): string =>
```
lewis-sanchez, line 4185: The preview label doesn't need a `(Preview)` tag, right?

lewis-sanchez (approved):

extensions/mssql/src/constants/locConstants.ts
ssreerama: Ah, this isn’t a feature preview tag, it refers to the preview changes shown in the VS Code preview window. I’ll rename it to something clearer to avoid confusion.

PR #22360, opened 2026-06-23
Bump @vscode/test-cli to 0.0.15 and @vscode/test-electron to 3.0.0

aasimkhan30:
## Description

Updates the VS Code test tooling across the extensions and sample to pick up the fix for **Windows Insiders re-downloading on every test run** even when the version hadn't changed (microsoft/vscode-test#343, shipped in `@vscode/test-electron@3.0.0`). The root cause was `product.json` moving under the `versioned-resources` folder, which broke version detection on Windows.

Version changes:

| Package | `@vscode/test-cli` | `@vscode/test-electron` |
|---|---|---|
| `extensions/mssql` | `0.0.12` → `0.0.15` | `2.5.2` → `3.0.0` |
| `extensions/sql-database-projects` | `0.0.12` → `0.0.15` | `2.5.2` → `3.0.0` |
| `extensions/data-workspace` | `0.0.12` → `0.0.15` | `2.5.2` → `3.0.0` |
| `samples/connection-sharing-sample` | `0.0.10` → `0.0.15` | `2.5.2` → `3.0.0` |

The `test-cli` bumps are dependency/security maintenance only (no API changes). `test-electron@3.0.0`'s sole breaking change is raising the minimum to **Node 22**, which is a non-issue since this repo already requires **Node 24+** (`minSupportedNodeMajor = 24` in `scripts/workspaces.mjs`).

## Code Changes Checklist


extensions/mssql/test/unit/localization.test.ts
```diff
@@ -7,8 +7,44 @@ import { expect } from "chai";
...
 import * as path from "path";
 import * as fs from "fs/promises";
+import * as sinon from "sinon";
+import * as vscode from "vscode";
+import { resetLoggerDefaultChannelForTest } from "../../src/models/logger";
 
 suite("Localization Tests", () => {
+    let sandbox: sinon.SinonSandbox;
+
+    setup(() => {
```
Benjin, line 18: Are these changes supposed to be part of this PR?

extensions/mssql/test/unit/localization.test.ts
```diff
@@ -44,7 +80,7 @@ suite("Localization Tests", () => {
         expect(testLocalizationConstant).to.equal("prueba");
     });
 
-    teardown(async () => {
+    suiteTeardown(async () => {
         await setLocLang("en");
     });
 });
```
Benjin, line 67: Can you move this the file with shared stubs and have the stub for vscodeWrapper consume this?

Benjin (dismissed):

extensions/mssql/test/unit/localization.test.ts
aasimkhan30: Solves an intermittent test failure in ado pipeline

aasimkhan30: Yes. This was added to cover an intermittent ADO pipeline failure where the logger default channel could be initialized during localization tests. I pushed b8dde0ed to keep that fix but move the output-channel stub setup into the shared test stubs so it is not a one-off in this test.

extensions/mssql/test/unit/localization.test.ts
aasimkhan30: Done in b8dde0ed. Added createLogOutputChannelStub in test/unit/stubs.ts and updated stubVscodeWrapper to consume it; localization.test.ts now uses the shared helper and keeps the per-test localization reset.

kburtram (approved):

ssreerama (approved):

PR #22399, opened 2026-06-27
Correcting line ending normalization

Benjin:
Hopefully this resolves the issue of package.json conflicts that claim every line was edited.

Only `scripts/ensure-crlf.js` has content changes


scripts/ensure-crlf.js
```diff
@@ -66,12 +66,21 @@ function main() {
         return;
     }
 
+    const eligibleFiles = files.filter(
```
Benjin, line 69: These are the only real changes in this PR

aasimkhan30 (approved):

PR #22404, opened 2026-06-29
Show specific error messages for unsupported rename targets

### LINKED ISSUE #22388: [Bug]: Rename (F2) opens on SQL keywords and then fails with a misleading project error
### Description

Pressing F2 (Rename) while the cursor is on a SQL **keyword** — e.g. `PRIMARY`, `KEY`, `INT`, `NOT`, `NULL`, `CREATE`, `TABLE` — incorrectly opens the inline rename box as if the keyword were a renameable symbol. After confirming a new name, it fails with the generic error: "Rename is only supported for SQL files that are part of an open SQL project. Open the project in the Database Projects panel first." — even though the file is part of an open project.

Two problems:

1. The rename box should not appear for keywords at all. Rename only accepts whether the cursor is on a "word", with no check that the token is an actual database object identifier (table, view, column, etc.) rather than a T-SQL keyword.
2. When it does fail, the error message is misleading (it talks about the project not being open, which is unrelated to the real reason — there is no renameable symbol at the cursor).

Expected:

- Rename should not be offered when the cursor is on a SQL keyword (the inline rename box should not appear).
- If a rename is attempted on something that isn't a renameable object, the message should say so (e.g. "No renameable object found at the cursor."), not the project-membership message.

(Screenshot available showing the rename box opening on `PRIMARY KEY` and the resulting generic error — can be attached to this issue.)

### Steps to Reproduce

1. Open a SQL project (in the Database Projects panel and as a workspace folder).
2. Open a file, e.g. `CREATE TABLE [Sales].[Sales] ( [Id] INT NOT NULL PRIMARY KEY )`.
3. Place the cursor on the `PRIMARY` keyword and press F2.
4. Observe: the inline rename box opens (as if it were a renameable symbol).
5. Enter a new name and confirm.
6. Observe: it fails with "Rename is only supported for SQL files that are part of an open SQL project. Open the project in the Database Projects panel first.", even though the project is open.

### Affected Area

- SQL Database projects

### Environment Information

- VS Code Version: (please fill in)
- MSSQL Extension Version: (please fill in)
- Operating System: macOS
- Connection Type: N/A (SQL project / offline)

### Confirmation

ssreerama:
Add errorMessage field to SqlSymbolRenameResponse. When STS returns a response with errorMessage set, throw that message directly so the user sees the actual reason (keyword, schema, server connection, symbol not found) instead of the generic 'not a project file' fallback.

## Description

## Code Changes Checklist


ssreerama:
Other PR already have these updated version, closing this one

PR #22416, opened 2026-07-02
Add shared extension toolkit for DI and extension infrastructure

aasimkhan30:
## Description

Adds a shared `extension-toolkit` package with VS Code-style DI primitives and an extension context service, then wires it into the mssql, SQL Database Projects, and Data Workspace extensions.

This keeps activation readable in each extension while giving us a common foundation for future shared extension infrastructure and DI.

### Toolkit layering

The toolkit has three explicit entry points with a one-way dependency direction:

- `extension-toolkit/base` contains portable primitives and cannot depend on the VS Code extension host.
- `extension-toolkit/vscode` may depend on `base`, the VS Code API, and VS Code-dependent libraries. Shared helpers such as telemetry integrations belong here.
- `extension-toolkit/vscode/testing` contains test fakes and may depend on either production layer, but production code cannot import it.

These boundaries keep the base primitives usable outside the extension host and prevent test-only behavior from entering the shipped runtime.

## Code Changes Checklist


eslint/custom-rules/extension-toolkit-layering.js
```diff
@@ -0,0 +1,96 @@
...
+        return {
+            ImportDeclaration(node) {
+                const source = node.source.value;
+                if (typeof source !== "string") {
+                    return;
+                }
+
+                const resolvedSource = resolveImportPath(source);
+
+                if (isBaseFile() && isVscodeImport(source, resolvedSource)) {
```
Benjin, line 87: Can you add comments that list out the intended limitations (and maybe "why" if it's not super obvious)?

For example, I'm trying to figure out if this would allow me to write shared helpers that invoke anything in vscode, like a shared telemetry package that would use [@vscode/extension-telemetry](https://www.npmjs.com/package/@vscode/extension-telemetry?activeTab=dependencies), but this message indicates that wouldn't be supported and I'm not immediately sure why not.

Benjin: That sort of info would probably also be good to capture in a README at `packages/extension-toolkit` and the PR description, too.

aasimkhan30: Yes—shared helpers that use VS Code APIs or VS Code-dependent libraries, such as extension telemetry, are supported and belong in extension-toolkit/vscode. Only extension-toolkit/base must remain extension-host independent. I clarified the intended dependency direction and rationale in readme

Benjin (dismissed):

caohai (approved):

PR #22449, opened 2026-07-08
Cleaning up noise from test run output

Benjin:
## Description

Cleans up several classes of test output pollution:
* missing stubs for side-effect calls that weren't defined
* (intentionally) unawaited promises that would complete after the test had already finished and cleaned up
* slow docker unit tests due to hardcoded timer instead of stub
* missing stubs for tests that incidentally involve calls to read connections/groups from config

## Code Changes Checklist


aasimkhan30:
@benjin, can you configure the runner to not show a line entry for passed tests? Adds a lot of scrolling to the GitHub runs.

aasimkhan30:
Seems like line endings for loc files are messed up.

aasimkhan30 (changes requested):
Requesting changes because this PR currently contains 140,430 additions and 140,430 deletions across 20 translated XLF files. The entire localization diff disappears when EOL whitespace is ignored, so these are line-ending-only changes unrelated to the PR.

[#22584](https://github.com/microsoft/vscode-mssql/pull/22584) has now merged the repository's canonical EOL policy into `main`. Please update this branch and restore the localization files from the updated base:

```bash
git fetch origin
git merge origin/main
git restore --source=origin/main -- localization/xliff/*.xlf
git commit -am "Remove line-ending-only localization changes"
```

Then verify that `git diff --ignore-space-at-eol origin/main...HEAD` contains the intended changes only and that the normalization/localization checks pass.

Benjin:
@aasimkhan30 

> Can you configure the runner to not show a line entry for passed tests? Adds a lot of scrolling to the GitHub runs.

I actually prefer to have the positive confirmation for passed tests.  We'll sometimes see bulk skips when a setup/teardown fails or there's a weird mock leak, so IMO it's nice to have that very visible differentiation when that happens.  Thoughts?

aasimkhan30:
I think we still do get total summary of the runs (passed, skipped, etc). Moreover, we have 3k+ tests and the list is not readable anyway.

aasimkhan30 (approved):

Issue #22124, opened 2026-05-15
[Bug]: Intellisense availiability in SQL Notebook is inconsistent


jmetape:
### Description

Intellisense does not give a list of database objects in a schema, even after performing a Refresh Intellisense Cache. I have a .sql file open in another editor window, connected to the same server and Intellisense works just fine. 

### Steps to Reproduce

1. Rt click on server connection and go to New SQL Notebook (Preview) 
2. Update intellisense cache
3. Begin typing in a cell, database objects in intellisense don't appear
4. Open a .sql file connected to same server, intellisense has the expected database objects

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [x] Other (please describe below)

### If you selected "Other", please describe the affected area

SQL Notebook

### Environment Information

I experience on both Mac and Windows machines, details:

**MAC:**
VS code info:
Version: 1.120.0 (Universal)
Commit: 0958016b2af9f09bb4257e0df4a95e2f90590f9f
Date: 2026-05-12T20:17:22Z
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: [14.2.231.22-electron.0](https://14.2.231.22-electron.0)
OS: Darwin arm64 25.3.0

MSSql Extension version:
1.42.2

Operating System:
MacOS Tahoe, Version 26.3.1 (a) (25D771280a)
COnnection Type:
SQL Server, windows authentication

**WINDOWS** 
VS code info:
Version: 1.117.0 (user setup)
Commit: 10c8e557c8b9f9ed0a87f61f1c9a44bde731c409
Date: 2026-04-21T16:12:14-07:00
Electron: 39.8.7
ElectronBuildId: 13841579
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

MSSql extension version:
1.42.2
Operating System:
Windows 11

[...]

PR #22525, opened 2026-07-22
Fix SQL notebook IntelliSense loss on save and reconnect (#22124)

kburtram:
Cell IntelliSense registrations now wait for the real connection/complete result instead of trusting the connect request acknowledgement, so failed cell binds are retried instead of silently memoized (saved notebooks previously ended up with keyword-only completions). Cell connects also use the live execution-connection credentials so they carry refreshed auth tokens and the actual database.

Saving an untitled notebook replaces its NotebookDocument, which previously disposed the notebook's connection mid-save. The connection manager is now parked when a connected untitled notebook closes and adopted by the file-based notebook that replaces it (matched by cell content within the save-transition window, guarded by tab presence and open recency), covering close-then-open, open-then-close, and lingering-document event orderings. Adoption re-registers the new cell URIs with STS and refreshes UI state.

Additional fixes from review: per-cell STS connections are released on notebook disconnect/close instead of leaking one server session per cell; stale cell connects that complete after a connection change are disconnected; completion expectations are identity-guarded against cross-cancellation between overlapping registrations; re-parking a reused untitled URI no longer leaks the superseded manager.

## Description

## Code Changes Checklist

### DIFF
--- a/extensions/mssql/src/controllers/connectionManager.ts
+++ b/extensions/mssql/src/controllers/connectionManager.ts
@@ -481,6 +481,45 @@ export default class ConnectionManager {
     }
 
     /**
+     * Registers interest in the `connection/complete` notification for a URI and returns a
+     * promise that resolves with the completion params. The `connection/connect` request only
+     * acknowledges that a connection attempt started; the actual outcome arrives later via
+     * this notification. Callers that send ConnectionRequest directly (e.g. notebook cell
+     * IntelliSense registration) use this to observe the real result — call BEFORE sending
+     * the request, and pair with {@link cancelConnectionCompleteExpectation} on timeout so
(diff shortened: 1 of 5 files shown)


kburtram:
This is a proposed fix for https://github.com/microsoft/vscode-mssql/issues/22124

aasimkhan30 (approved):

lewis-sanchez (approved):

kburtram:
This fixes the issue where saved notebooks don't get intellisense; but it doesn't resolve the issue where saving an untitled notebook loses the active connection.  I'll fix that as a separate PR so that I can take this one, which fixes the core bug reported in the issue.  The "auto-reconnect" issue can be worked-around by manually reconnecting after "save as".

PR #22583, opened 2026-07-31
Normalize PR line endings automatically

aasimkhan30:
## Summary

- queue line-ending normalization for PRs that change JSON, XML, or XLF files
- perform normalization in a trusted `workflow_run` and push the result to same-repository PR branches
- store JSON and XML with normalized LF blobs in Git while preserving the repository's CRLF checkout policy
- preserve XLF files as CRLF blobs because they are generated and checked in by OneLocBuild
- skip fork PRs because their branches are not writable by this repository

The workflow uses `LINE_ENDING_BOT_TOKEN` when configured and otherwise falls back to `GITHUB_TOKEN`. A bot token is recommended so the normalization commit triggers normal CI without requiring workflow approval.

## Validation

- both workflow files parse as YAML
- both workflow files pass Prettier
- all 83 tracked JSON and XML files are stored with LF line endings
- all 33 tracked XLF files are stored with CRLF line endings
- existing-file changes were verified with `git diff --ignore-space-at-eol` to contain no content changes


aasimkhan30:
Closing this approach while we investigate the repository's line-ending history and bot-specific behavior more deeply.

Issue #22549, opened 2026-07-27
[Bug]: When DAB GHCP deselects a table, it does not deselect all columns in that table


yoleichen:
### Description

<img width="1594" height="953" alt="Image" src="https://github.com/user-attachments/assets/bc533ac5-f51c-46e2-8fd4-eb72c3e52d3c" />

When user uses DAB GHCP to deselect a table, it does not automatically deselect all columns in that table  

### Steps to Reproduce

1. Open DAB 
2. Run this command: Remove all entities from the dbo schema and keep only the <table name> entity. Make it a read-only endpoint -- I only need the Read permission enabled.
3. Observe that the other table entities got deselected, but not the columns underneath them

### Affected Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [ ] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [x] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### Environment Information

VS Code Extension: latest

### Confirmation

- [x] I have searched existing issues and couldn't find a match
- [ ] I want to work on this issue

PR #22600, opened 2026-08-01
Fix: DAB GHCP does not deselect columns when table is deselected (#22549)

jibin7jose:
### Description
This PR fixes issue #22549 where deselecting a table using the GitHub Copilot (GHCP) integration for Data API builder (DAB) did not automatically deselect all of the columns in that table.

### Fix Details
The fix aligns the backend `applyDabToolChange` logic for the `"set_only_enabled_entities"` and `"set_all_entities_enabled"` operations with the frontend UI logic. It now maps through the entity columns and correctly updates their `isExposed` state to match the table's new `isEnabled` state.

### DIFF
--- a/extensions/mssql/src/webviews/pages/SchemaDesigner/schemaDesignerRpcHandlers.ts
+++ b/extensions/mssql/src/webviews/pages/SchemaDesigner/schemaDesignerRpcHandlers.ts
@@ -2042,18 +2042,32 @@ function applyDabToolChange(
                 selectedEntityIds.add(resolvedEntity.entity.id);
             }
 
-            config.entities = config.entities.map((entity) => ({
-                ...entity,
-                isEnabled: selectedEntityIds.has(entity.id),
-            }));
+            config.entities = config.entities.map((entity) => {
+                const isEnabled = selectedEntityIds.has(entity.id);
+                return {
+                    ...entity,
+                    isEnabled,
+                    columns: entity.columns.map((column) => ({
+                        ...column,
+                        isExposed: isEnabled,
+                    })),
+                };
+            });
             return { success: true };
         }
 
         case "set_all_entities_enabled": {
-            config.entities = config.entities.map((entity) => ({
-                ...entity,
-                isEnabled: change.isEnabled ? entity.isSupported : false,
-            }));
+            config.entities = config.entities.map((entity) => {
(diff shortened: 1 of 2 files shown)


jibin7jose:
The CI pipeline is failing on the create-or-update-comment step with a 403 error because this PR is from a fork. However, all the actual unit tests passed successfully.

extensions/mssql/src/webviews/pages/SchemaDesigner/schemaDesignerRpcHandlers.ts
```diff
@@ -2042,18 +2042,32 @@ function applyDabToolChange(
...
-                isEnabled: selectedEntityIds.has(entity.id),
-            }));
+            config.entities = config.entities.map((entity) => {
+                const isEnabled = selectedEntityIds.has(entity.id);
+                return {
+                    ...entity,
+                    isEnabled,
+                    columns: entity.columns.map((column) => ({
+                        ...column,
+                        isExposed: isEnabled,
```
aasimkhan30, line 2052: This modifies the column exposure settings in already enabled entities and sets it to enabled.

extensions/mssql/src/webviews/pages/SchemaDesigner/schemaDesignerRpcHandlers.ts
```diff
@@ -2042,18 +2042,32 @@ function applyDabToolChange(
...
-                isEnabled: change.isEnabled ? entity.isSupported : false,
-            }));
+            config.entities = config.entities.map((entity) => {
+                const isEnabled = change.isEnabled ? entity.isSupported : false;
+                return {
+                    ...entity,
+                    isEnabled,
+                    columns: entity.columns.map((column) => ({
+                        ...column,
+                        isExposed: isEnabled,
```
aasimkhan30, line 2067: I wonder if we should retain the old exposure settings here.

aasimkhan30:
Thanks for working on this issue. This PR does identify certain issues contributing to the bug.
However, the proposed fix links entity enablement to saved column visibility. As a result, it can overwrite intentional column exclusions. The fix also applies only to bulk operations. Individual entity operations might continue to behave differently.
I am addressing this issue in #22257 as part of a broader update to the entity and column state model. That work will apply consistent behavior to all operation paths and add the necessary regression tests.

PR #22630, opened 2026-08-05
Move to Schema: relocate the SQL file into its new schema folder

ssreerama:
## What

When a user runs **Refactor... → Move to Schema...** on a SQL object in a SQL project, the `.sql` file that defines the object is now relocated into the folder for its new schema, and the `.sqlproj` is updated to match.

For example, moving `dbo.table1` to schema `sss` moves `dbo/Tables/table1.sql` → `sss/Tables/table1.sql`.

## How it works

1. STS returns the SQL text edits plus the `.refactorlog` content, along with the defining file's URI and its element type.
2. A `WorkspaceEdit` carries those edits and is shown in VS Code's refactor preview (Apply / Discard).
3. On Apply, the destination is computed from the target schema and the object's element type (`SqlTable` → `Tables`, `SqlView` → `Views`, and so on).
4. A second `WorkspaceEdit.renameFile` performs the move. VS Code creates any missing folders and carries unsaved editor content to the new location.
5. `moveSqlObjectScript(..., metadataOnly: true)` updates the `.sqlproj` `<Build Include>` entries without touching the filesystem.

### Why VS Code moves the file rather than STS

After Apply, the text edits live in VS Code's in-memory documents — the files are dirty, not yet saved. If STS performed the move via `File.Move`, it would relocate the stale on-disk content while VS Code still held a dirty editor bound to the old path. Saving would then recreate the file at the original location, leaving two copies with the `.sqlproj` pointing at the wrong one.

`WorkspaceEdit.renameFile` avoids this: VS Code moves the file, carries the unsaved buffer with it, keeps the editor tab attached, and preserves undo. `metadataOnly: true` exists so STS handles only the XML.

The move is skipped when the file sits outside the project directory, isn't organised by schema folders, or is already at the destination.

## Also in this PR

- Clicking **Discard** in the refactor preview no longer shows an error. `applyEdit` returns `false` for a deliberate cancel just as it does for a real failure, so the previous code surfaced *"Failed to apply the Move to Schema changes..."* to users who had simply backed out. This is now logged at debug level, and the unused `applyEditFailed` string was removed.
- Changelog updates
- STS version bump to `6.0.20260804.2`.
image

## Testing

Unit tests in `sqlSymbolRenameProvider.test.ts` cover the move planning, the skip conditions, the `.sqlproj` metadata call, and the failure paths. `tsc --noEmit` and `eslint` are clean on the changed files.


extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
```diff
@@ -222,7 +272,24 @@ export class SqlMoveToSchemaProvider implements vscode.CodeActionProvider {
...
+                    targetSchema,
+                    schemas,
+                    response.definitionFileUri,
+                    response.elementType,
+                );
+                // Refresh the project tree to reflect the SQL text edits, .sqlproj changes,
+                // and (when the file moved) the new file location. Skip only when the move
+                // failed — the user already saw the error message in that case.
+                if (moveResult !== FileMoveResult.Failed) {
+                    await vscode.commands.executeCommand("dataworkspace.refresh");
```
Benjin: Does this need to be awaited?  Probably can be `void`

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
```diff
@@ -222,7 +272,24 @@ export class SqlMoveToSchemaProvider implements vscode.CodeActionProvider {
                 isRefactoring: true,
             });
             if (!applied) {
-                void vscode.window.showErrorMessage(loc.applyEditFailed);
+                // `applyEdit` returns false when the user clicks Discard in the refactor preview,
+                // which is a deliberate action — not an error worth surfacing to the user.
+                logger.debug("Move to Schema edits were not applied (discarded or rejected).");
+            } else {
+                const moveResult = await this.moveFileToNewSchemaFolder(
```
Benjin: Want to make sure that I understand the full flow correctly...  Is it doing this:

1. Updates the file contents and other references (the `SqlMoveToSchemaRequest` call above)
2. Early exit if file doesn't need moving
3. Moves the file on disk
4. Updates the .sqlproj with the new file location

Then most of those steps can probably live within their own helper function to keep the top level function clean and concise.

I think it'd be good to organize the code around these explicit steps with comments to guide the reader, e.g.

```
function applyMove() {
  // 1. Update scripts using Language Service; also shows
  ...
  
  // 2. Check if file should be moved on disk, exit if not
  ...
  
  // 3. Move file using VS Code API to maintain document state
  ...
  
  // 4. Update .sqlproj file
  ...
}
```

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
ssreerama: Since nothing depends on it a void is sufficient, but if the dataworspace is not registered yet, the refresh will be rejected, so adding a try catch around to catch that.

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
```diff
@@ -234,4 +301,155 @@ export class SqlMoveToSchemaProvider implements vscode.CodeActionProvider {
...
+        if (!hasSchemaPrefix && !detectedFolder) {
+            // File is at the project root and object type is unknown — skip file move.
+            return undefined;
+        }
+
+        if (hasSchemaPrefix && !schemas.some((s) => s.toLowerCase() === currentSchemaLower)) {
+            // Top-level folder is not a known project schema — skip to avoid moving
+            // non-schema-organised files (e.g. misc/tables/x.sql) unexpectedly.
+            return undefined;
+        }
```
Benjin, line 475: I suspect we can probably know a lot of these early-exit situations before calling down into this function (maybe the info from earlier steps)

ssreerama: They can't move before the early (STS) call, all four exits need definitionFileUri/elementType from the response, and planSchemaFolderMove also falls back to the trigger document's URI when definitionFileUri is absent, so we can't tell which path applies until it arrives.

I originally placed these check after the apply and once we are ready yo move, may be we can move single check for the file exists beforehand using path.

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
ssreerama: Mostly accurate (small correction is SqlMoveSchemaRequest just returns the edited content + updated refactorlog, vscode perform the edits)

I have separated the single method applyMove into helper methods based on the workflow it follows and added clear comments on each step that they are representing.

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
```diff
@@ -189,37 +190,136 @@ export class SqlMoveToSchemaProvider implements vscode.CodeActionProvider {
...
+
+            // 5. Relocate the definition file into its new schema folder and update the .sqlproj.
+            const moveResult = plan
+                ? await this.moveFileToNewSchemaFolder(refactorTarget, plan)
+                : FileMoveResult.Skipped;
+
+            // 6. Refresh the project tree to reflect the SQL text edits, .sqlproj changes, and
+            //    (when the file moved) the new file location. Skip only when the move failed —
+            //    the user already saw the error message in that case.
+            if (moveResult !== FileMoveResult.Failed) {
```
Benjin: Should this be...?

```
if (moveResult !== FileMoveResult.Failed && moveResult !== FileMoveResult.Skipped)
```

Benjin (dismissed):

Benjin (approved):

PR #22633, opened 2026-08-05
Select environment proxies by request protocol

aasimkhan30:
## Summary

- Selects environment proxy settings according to the request URL protocol.
- Prefers `HTTPS_PROXY` for HTTPS requests while retaining the standard `HTTP_PROXY` fallback.
- Prevents HTTP requests from incorrectly using `HTTPS_PROXY` and adds regression coverage for each selection path.

## Validation

- TypeScript compilation
- Focused HTTP client tests
- ESLint and Prettier

### DIFF
--- a/packages/extension-toolkit/src/base/http/httpClient.ts
+++ b/packages/extension-toolkit/src/base/http/httpClient.ts
@@ -250,7 +250,7 @@ export class HttpClient {
             validateStatus: () => true,
         };
 
-        const proxy = this.loadProxyConfig();
+        const proxy = this.loadProxyConfig(requestUrl);
 
         if (proxy) {
             this.logger?.debug(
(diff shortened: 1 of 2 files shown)


aasimkhan30:
### Bug location

The bug was in the previous environment-proxy loading path in `HttpClient`:

- `setupConfigAndProxyForRequest()` called `loadProxyConfig()` without passing the request URL.
- `loadEnvironmentProxyValue()` then selected `HTTP_PROXY` / `http_proxy` first and only checked `HTTPS_PROXY` / `https_proxy` as a fallback, regardless of the destination protocol.

This meant an HTTPS request could ignore its dedicated `HTTPS_PROXY` when `HTTP_PROXY` was also set, while an HTTP request could incorrectly use `HTTPS_PROXY` when no HTTP proxy was configured.

The updated flow parses the request URL once and selects environment variables by its protocol: HTTP uses only the HTTP proxy variables, while HTTPS prefers the HTTPS proxy variables and retains the standard HTTP proxy fallback.

Benjin (approved):

PR #22634, opened 2026-08-05
Honor proxy bypass lists

aasimkhan30:
## Summary

- Bypasses proxy agent creation for requests matched by `NO_PROXY`, `no_proxy`, or VS Code's `http.noProxy` setting.
- Supports exact hosts, domain wildcards, optional ports, and the global wildcard.
- Forces matched requests to stay direct instead of falling back to Axios environment proxy discovery.

## Validation

- TypeScript compilation
- Focused HTTP client tests
- ESLint and Prettier


packages/extension-toolkit/src/base/http/httpClient.ts
```diff
@@ -495,6 +516,51 @@ interface ProxyAgentOptions {
...
+            ruleHost = ruleEndpoint.hostname;
+            rulePort = ruleEndpoint.port || undefined;
+        } catch {
+            return false;
+        }
+    } else {
+        const match = rule.match(/^(.+):(\d+)$/);
+        ruleHost = match?.[1] ?? rule;
+        rulePort = match?.[2];
+    }
```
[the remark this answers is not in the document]
aasimkhan30: Thanks for flagging this. Node’s WHATWG URL implementation retains brackets in `URL.hostname` (`new URL("https://[::1]:8443").hostname === "[::1]"`), and the existing regex captures `[::1]` and `8443`, so bracketed IPv6 rules already compare correctly. I added a regression test in 0ece1004a covering both the matching port and a mismatched port.

packages/extension-toolkit/src/base/http/httpClient.ts
```diff
@@ -284,6 +287,12 @@ export class HttpClient {
             validateStatus: () => true,
         };
 
+        if (this.shouldBypassProxy(requestUrl)) {
+            this.logger?.debug("Request URL matched the proxy bypass list.");
```
Benjin: Might be good to include the request URL in the log message so users can verify that only acceptable URLs are bypassing the proxy?

packages/extension-toolkit/test/httpClient.test.js
```diff
@@ -394,6 +404,118 @@ describe("HttpClient", () => {
             assert.equal(proxy, undefined);
         });
 
+        for (const source of ["NO_PROXY", "no_proxy", "VS Code http.noProxy"]) {
```
Benjin: I don't hate it, but a bit funky... Why not stick the loop inside the test? instead of generating multiple tests?

Benjin: I feel like all these new tests would be better with all cases in a consistent format.  WIth all the possible patterns that the no-proxy pattern matcher supports, it'd be easier to quickly scan an array of cases to make sure everything's covered than separate test methods that have boilerplate.

Something like:

```
it("Proxy bypasses as correct", {} => {
const cases = [{pattern, input, output}, ... ];

for (case of cases) {
  // test case
}
});
```

packages/extension-toolkit/test/httpClient.test.js
```diff
@@ -9,12 +9,20 @@ const { PassThrough, Writable } = require("node:stream");
 const { afterEach, beforeEach, describe, it, mock } = require("node:test");
```
Benjin: Not particularly important, but any reason we aren't using the same test ontology/libs as the rest of the extension?

`it()` and `assert()` here, but everywhere else uses `test()` and `expect()`

aasimkhan30: Maybe some weirdness from agents.

packages/extension-toolkit/src/base/http/httpClient.ts
aasimkhan30: Addressed in 350e16fc5. The debug message now includes the request origin (for example, https://api.example.com) so users can identify the bypassed endpoint without exposing credentials, paths, query values, or fragments. Added regression coverage for the sanitized log output.

packages/extension-toolkit/test/httpClient.test.js
aasimkhan30: Addressed in 350e16fc5. The source variants now run inside one table-driven test rather than generating separate tests.

aasimkhan30: Addressed in 350e16fc5. Consolidated the bypass scenarios into a single case table covering environment/config sources, precedence, plain and wildcard domains, ports, IPv6, and the global wildcard.

packages/extension-toolkit/test/httpClient.test.js
aasimkhan30: Kept 
ode:test and 
ode:assert intentionally because this is the standalone packages/extension-toolkit test suite, and the existing file and neighboring toolkit tests consistently use those APIs. The MSSQL extension unit tests use Mocha/Chai, but converting this package would be a broader unrelated change. The new bypass coverage now follows the existing toolkit test conventions.

aasimkhan30: Follow-up: aligned the file with the extension naming in 0a7a46df9 by using the native `suite()` and `test()` exports from `node:test`. The package keeps its existing Node runner and `node:assert`, avoiding a new test dependency while matching the established suite/test structure.

Benjin (approved):

PR #22635, opened 2026-08-05
Redact secrets from proxy warnings

aasimkhan30:
## Summary

- Redacts proxy URL credentials and query or fragment values before warnings reach UI or logs.
- Sanitizes parser error details that repeat the configured proxy value.
- Adds regression coverage ensuring proxy secrets are absent from both warning channels.

## Validation

- TypeScript compilation
- Focused HTTP client tests
- ESLint and Prettier

### DIFF
--- a/packages/extension-toolkit/src/base/http/httpClient.ts
+++ b/packages/extension-toolkit/src/base/http/httpClient.ts
@@ -498,6 +498,17 @@ interface ProxyAgentOptions {
     rejectUnauthorized: boolean;
 }
 
+/**
+ * Redacts credentials and query or fragment values from proxy URL-like strings.
+ *
+ * Examples:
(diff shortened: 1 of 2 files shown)


packages/extension-toolkit/src/base/http/httpClient.ts
```diff
@@ -495,6 +498,29 @@ interface ProxyAgentOptions {
     rejectUnauthorized: boolean;
 }
 
+function redactProxySecrets(proxy: string): string {
```
Benjin, line 512: Can you add a docstring that gives examples of proxy formats that this is able to redact?

Benjin (dismissed):

Benjin (approved):

PR #22648, opened 2026-08-07
Add a Restore command to the SQL project tree

### LINKED ISSUE #22645: Add a Restore command to the SQL project tree
Custom code analysis rules come from NuGet analyzer packages, which have to be restored before the rules are discoverable. There is no restore command in the SQL Database Projects extension today, so the only way to restore is to drop to a terminal and run `dotnet restore` by hand.

## Proposed work

Add `sqlDatabaseProjects.restore` alongside the existing build commands, so users can restore packages without leaving the projects view.

`extensions/sql-database-projects/package.json` currently orders the project context menu as:

| Group | Command |
| --- | --- |
| `1_dbProjectsFirst@1` | `sqlDatabaseProjects.build` |
| `1_dbProjectsFirst@2` | `sqlDatabaseProjects.buildWithCodeAnalysis` |
| `1_dbProjectsFirst@3` | `sqlDatabaseProjects.publish` |

Restore should slot in next to Build.

Implementation follows the existing `buildProject` pattern in `projectController.ts`, which builds a `vscode.Task` with a `ProcessExecution` of `dotnet`, so a `dotnet restore` task can reuse the same machinery (including `NetCoreTool.verifyNetCoreInstallation()`).

## Acceptance criteria

- [ ] A Restore command is available on the project node and runs `dotnet restore` for that project
- [ ] Restore output is surfaced the same way build output is, and failures are reported to the user
- [ ] All new user-facing strings are localized

ssreerama:
## Problem

Custom code analysis rules only become discoverable once the analyzer package they come from has been restored. There is no restore command in the SQL Database Projects extension today, so the only way to restore is to drop to a terminal and run `dotnet restore` by hand or Build the project.

## Change

Adds a `Restore Packages` command to the project node, between **Build With Code Analysis** and **Publish**:

| Group | Command |
| --- | --- |
| `1_dbProjectsFirst@1` | Build |
| `1_dbProjectsFirst@2` | Build With Code Analysis |
| **`1_dbProjectsFirst@3`** | **Restore Packages** ← new |

It runs `dotnet restore <project>.sqlproj` as a `vscode.Task`, the same mechanism Build uses. Like its siblings it is hidden from the Command Palette (`"when": "false"`), since it needs a tree node argument.

image


Benjin:
One question for me to have additional context: what sorts of nuget packages does a sqlproj need to restore?  Are the system dacpacs via nuget?

extensions/sql-database-projects/src/common/constants.ts
```diff
@@ -364,6 +365,7 @@ export const confirmCreateProjectWithBuildTaskDialogName = l10n.t(
 );
 export const buildTaskName = l10n.t("Build");
 export const buildWithCodeAnalysisTaskName = l10n.t("Build with Code Analysis");
+export const restoreTaskName = l10n.t("Restore");
```
Benjin: Since it may not be clear what's getting "restored", maybe "Restore NuGet packages" would be better?

extensions/sql-database-projects/src/common/constants.ts
```diff
@@ -538,6 +540,14 @@ export function errorFindingBuildFilesLocation(err: any) {
 export function projBuildFailed(errorMessage: string) {
     return l10n.t("Build failed. Check output pane for more details. {0}", errorMessage);
 }
+/**
+ * @param errorMessage omitted when dotnet reported the failure itself, since the details are
+ * already in the task terminal.
+ */
+export function projRestoreFailed(errorMessage?: string) {
+    const message = l10n.t("Restore failed. Check the terminal output for more details.");
```
Benjin: Same here:

```suggestion
    const message = l10n.t("Restore NuGet packages failed. Check the terminal output for more details.");
```

ssreerama:
> One question for me to have additional context: what sorts of nuget packages does a sqlproj need to restore? Are the system dacpacs via nuget?

For sqlprojects, we restore the below nuget packages types on each build.

- system dacpacs
- Microsoft.Build.Sql
- Custom code analysis rules package (new one, since this is independent on whether the project has been built or not to show the rules, restoring it would avoid loading stale rules unknowingly)

extensions/sql-database-projects/src/common/constants.ts
ssreerama: Checking with Drew about this Option naming :)

dzsquared: "Restore Packages" for the menu, "Restore NuGet packages " in messages/logging

Benjin (approved):

PR #22679, opened 2026-08-11
Update Query Results view icon

aasimkhan30:
## Summary
- replace the Query Results view icon with a simplified table glyph
- use a shared icon for the panel container, webview view, and query-result editor
- use `currentColor` so the icon follows VS Code theme styling

| Location |Before | After |
| ------- | -------|-------|
| Query results view | image |  image |
| Editor toolbar |  image | image |
| query results panel |  image |   image |

## Validation
- parsed the extension manifest and SVG successfully
- ran `npm run build:prepare` from `extensions/mssql`
- ran `git diff --check`

Related to #22680


extensions/mssql/media/revealQueryResult.svg
Benjin, line 1: You didn't like the old one?  I feel like the header bar with a consistent grid feels more like "query results" and the new one feels more like a layout table.

Benjin (approved):

PR #22705, opened 2026-08-11
Add Move to Schema command for database project files

### LINKED ISSUE #22387: [Feature Request]: Add "Move to Schema…" to the object context menu in the Database Projects tree
## Feature Description

Add a "Move to Schema…" context menu (right-click) action on object nodes in the Database Projects tree. Today, Move to Schema is only available from the editor (Refactor… / code action) when a SQL file is open and the cursor is on the object. It would be helpful to also invoke it directly from the project tree by right-clicking an object (table, view, stored procedure, function, etc.) and choosing "Move to Schema…", which would show the same schema picker and refactor preview.

## Problem and Motivation

Currently, to move an object to another schema I have to open its `.sql` file, place the cursor on the object, and use the Refactor menu. When working in the Database Projects tree, it's more natural to act on the object node directly without opening the file first. Adding the action to the tree's right-click menu would:

- Make the feature discoverable where users already manage project objects.
- Save the open-file + position-cursor steps.
- Match the way other object actions are exposed in the tree.

The action should reuse the existing Move to Schema flow (schema picker + refactor preview + refactorlog update), just triggered from the tree node instead of the editor cursor.

## Related Area

- SQL Database projects

## Confirmation

- [x] I have searched existing feature requests and couldn't find a match
- [ ] I want to help implement this feature

ssreerama:
This PR:
- Add a SQL Database Projects file context-menu entry for Move to Schema.
- Forward database project file paths to the MSSQL Move to Schema provider.
- Resolve the target SQL object position in the provider for file-based entry points and calling the existing Move Schema logic
- Added unit coverage.

image

Sql_file_moveToSchema_opiton

### DIFF
--- a/extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
+++ b/extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
@@ -280,6 +280,10 @@ export class SqlMoveToSchemaProvider implements vscode.CodeActionProvider {
      * and string literals so statement detection ignores non-executable text.
      */
     private tokenizeMoveToSchemaSql(text: string): MoveToSchemaSqlToken[] {
+        // Match only syntax that affects CREATE/ALTER object detection: whitespace and
+        // comments/strings to skip, quoted or bare identifiers to keep, and dots for
+        // schema-qualified names. The sticky flag keeps token positions exact while the
+        // loop below advances over punctuation this parser does not care about.
         const tokenPattern =
             /\s+|--[^\r\n]*|\/\*[\s\S]*?\*\/|'(?:''|[^'])*'|\[(?:[^\]]|\]\])+\]|"(?:""|[^"])*"|`(?:``|[^`])*`|[A-Za-z0-9_$#@]+|\./y;
         const tokens: MoveToSchemaSqlToken[] = [];
--- a/extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
+++ b/extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
@@ -202,7 +202,7 @@ export class SqlMoveToSchemaProvider implements vscode.CodeActionProvider {
     // NOTE: This region exists only to bridge the tree-entry (file path only) flow into the
     // existing Move-to-Schema document+position flow. If STS exposes an API that resolves the
     // movable object position from a file path, replace this entire region with that STS call.
-
+    // TODO Task: https://github.com/microsoft/vscode-mssql/issues/22709
     /**
      * Starts Move to Schema from a Database Projects tree file path.
      */
--- a/extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
+++ b/extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
@@ -67,6 +67,11 @@ interface SchemaFolderMovePlan {
     newAbsUri: vscode.Uri;
 }
 
+interface MoveToSchemaSqlToken {
+    text: string;
+    start: number;
+}
+
 /**
  * Surfaces a "Move to Schema..." action under the editor's **Refactor...** menu for SQL files in a
  * SQL project.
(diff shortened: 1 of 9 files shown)


extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
```diff
@@ -189,6 +198,116 @@ export class SqlMoveToSchemaProvider implements vscode.CodeActionProvider {
...
+        return undefined;
+    }
+
+    /**
+     * Tokenizes SQL text into identifiers and separators while skipping whitespace, comments,
+     * and string literals so statement detection ignores non-executable text.
+     */
+    private tokenizeMoveToSchemaSql(text: string): MoveToSchemaSqlToken[] {
+        const tokenPattern =
+            /\s+|--[^\r\n]*|\/\*[\s\S]*?\*\/|'(?:''|[^'])*'|\[(?:[^\]]|\]\])+\]|"(?:""|[^"])*"|`(?:``|[^`])*`|[A-Za-z0-9_$#@]+|\./y;
```
ssreerama, line 288: Tried using an alternate approach to not use the regex, but that would add a mini sql lexer here which would look like a huge change for this PR. If ok to go with that route, I can make the changes. But, I felt if we can use the scriptdom parsing to get the object name back we can remove the whole parsing logic here.

aasimkhan30 (dismissed):

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
aasimkhan30: Please add comment on the regex.

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
```diff
@@ -189,6 +198,116 @@ export class SqlMoveToSchemaProvider implements vscode.CodeActionProvider {
         await this.applyMove(document, position, selected.label, schemas);
     }
 
+    //#region Tree Entry Local Parsing (replaceable)
```
ssreerama, line 201: the whole region is replaceable if STS can give the sql object name from the parser, which script Dom already have.  But, needed STS changes which can be done in another release cycle not at the ask mode time.

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
ssreerama: Done, Thank you

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
```diff
@@ -189,6 +198,120 @@ export class SqlMoveToSchemaProvider implements vscode.CodeActionProvider {
...
+        }
+
+        return undefined;
+    }
+
+    /**
+     * Tokenizes SQL text into identifiers and separators while skipping whitespace, comments,
+     * and string literals so statement detection ignores non-executable text.
+     */
+    private tokenizeMoveToSchemaSql(text: string): MoveToSchemaSqlToken[] {
```
Benjin, line 282: I don't see any unit tests that cover this method.  Please add several, especially since we're (temporarily) relying on a complicated REGEX.

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
Benjin: This is just to provide a quality-of-life improvement, correct?  Might be better just to hold this fix back from this release so that we can do it correctly (using STS/ScriptDom)

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
ssreerama: Right, Added the test coverage for the regex method and also created a task for the scriptDom replacement and added as a todo comment in the code.

extensions/mssql/src/languageservice/sqlMoveToSchemaProvider.ts
ssreerama: Had a chat with Drew and this can be moved to other cycle as is not very urgent one.

have moved the issue to next cycle, updating the PR to draft (will close it and open a new one if required)

Benjin:
After discussing offline, we're going with the scriptdom-based design and will take it in a following release.

PR #22741, opened 2026-08-14
edit data: retain changes after failed saves

### LINKED ISSUE #22735: When adding a row fails in Edit Data, the row can't be attempted to be re-added
Repro:
1. Create two tables, e.g. Types (ID, Name) and Pokemon (ID, Name, Type1, Type2)
2. Create a PK on Types and FK on Pokemon
3. Without adding any types, attempt to add a new Pokemon using Edit Data
4. Observe error message:

> Failed to save changes: The UPDATE statement conflicted with the FOREIGN KEY constraint "FK__Pokemon__Primary__5BE2A6F2". The conflict occurred in database "TestDb", table "dbo.Types", column 'TypeID'.
> The statement has been terminated.

Expected:
Row is still displayed as "dirty", option to save row still is enabled

Reality:
Option to save row is disabled, row/cell states display as saved

Benjin:
## Summary
Keeps pending Edit Data changes available when a save fails, allowing users to correct the row and retry without re-entering their work. Save operations also wait for pending edits before committing.

Fixes #22735

## Screenshots
**Before:**

_Add a screenshot showing a failed insert displayed as saved with retry disabled here._

**After:**

_Add a screenshot showing the failed insert still pending and retry available here._


Benjin:
This isn't close to fixing the issue.  I suspect this would require deeper changes at the STS layer.

PR #22748, opened 2026-08-14
Edit Data: clear restored cell changes

### LINKED ISSUE #22736: When editing a cell to change the value, then editing it back to the original value, the cell still displays dirty state

Benjin:
## Summary
Clears Edit Data dirty state when a cell is changed back to its original value while preserving other pending edits. Pending cell operations remain consistent across saves, reverts, and session changes.

Fixes #22736

## Screenshots
**Before:**

_Add a screenshot showing a restored cell still marked as modified here._

**After:**

_Add a screenshot showing the restored cell and row in a clean state here._


Benjin:
Superseded by #22888, which redesigns the fix against the latest Edit Data/Table Explorer contracts on `main`.

PR #22752, opened 2026-08-17
Query Editor: add completion sound notifications

Benjin:
## Summary
- add a default-off setting to play a chime after full query execution completes `query.playCompletionSound`
- support an optional custom WAV path `query.completionSoundFile` (limited to 400 KB)

## Platform behavior
- macOS: uses the built-in `/usr/bin/afplay`
- Windows: uses PowerShell with the native .NET `System.Media.SoundPlayer`
- Linux: tries common WAV-capable players (`pw-play`, `paplay`, `aplay`, `ffplay`, `mpv`, `cvlc`)

Linux playback is "best effort" because installed audio players differ so widely across distributions.

## Bundled sound attribution
The default chime is converted from [`Information_Block.ogg`](https://github.com/akx/Notifications/blob/master/OGG/Information_Block.ogg) and uses its [Creative Commons Public Domain Mark 1.0](https://creativecommons.org/publicdomain/mark/1.0/) option.

Fixes #22733

Issue #22733, opened 2026-08-13
[Feature Request]: Alerts upon query execution


AndCely:
### Feature Description

Add an option to enable custom sound notifications or system alerts upon query execution completion in the Query Editor.

### Problem and Motivation

When working with large database scripts or running heavy `UPDATE`/`INSERT` queries that take several minutes to process, developers often switch to other windows to continue working. Currently, there is no direct, native sound notification or customizable alert when a query finish executing in VS Code (especially when executing a selected block of text).

The current experience requires manually monitoring the editor to see if the execution has completed, which hurts productivity and workflow efficiency. While workaround scripts or system terminal hacks exist, they either require editing the SQL code directly (e.g., adding `PRINT CHAR(7);`) or using external command-line tools like `sqlcmd`.

It would be extremely helpful to have a built-in extension setting (or a toggle in the Query Editor toolbar) to play a customizable sound or display a native system notification when a query or selected block finishes running.

### Related Area

- [ ] Connection dialog (SQL Server | Azure browse/Fabric browse)
- [x] Query editor
- [ ] Query results panel
- [ ] Object Explorer
- [ ] GitHub Copilot integration
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] Local SQL Server Container provisioning
- [ ] SQL database in Fabric provisioning
- [ ] DACPAC/BACPAC export/import
- [ ] SQL Database projects
- [ ] Query Plan Visualizer
- [ ] Other (please describe below)

### If you selected "Other", please describe the area

*No response*

### Confirmation

- [x] I have searched existing feature requests and couldn't find a match
- [ ] I want to help implement this feature

PR #22755, opened 2026-08-17
Query Editor: add completion sound notifications - <audio> element

Benjin:
## Summary

Alternative prototype for #22733 that plays query-completion sounds through an invisible HTML `<audio>` element in the query results webview instead of launching native platform players.

- adds a default-off completion sound setting and optional custom MP3 path
- validates custom paths in the extension host and embeds MP3s up to 200 KiB as `data:audio/mpeg` URLs
- generates a small copyright-free PCM WAV chime in memory for empty, invalid, missing, unreadable, oversized, or browser-undecodable files
- sends playback only after the server reports the full query execution complete, not for individual batches or cleanup
- supports both the results pane and document-tab results webviews
- stops playback at five seconds using both an armed timeout and the media playback clock

## Why data URLs

Direct `file://` access is blocked by webview isolation. `webview.asWebviewUri()` would require granting the webview access to the selected file's parent directory. A bounded data URL grants the webview only the selected audio bytes and avoids broadening `localResourceRoots`.

## Reviewer notes

This approach is platform-independent and makes future WAV/OGG support mostly a validation/MIME-type change. MIDI support still depends on Chromium/Electron codec behavior. The main tradeoffs versus the native-player prototype in #22752 are webview lifecycle/autoplay dependency and base64/message overhead, bounded here by the 200 KiB limit.

## Validation

- `npm run build -- --target mssql`
- `npm run package -- --target mssql --online`
- extension and webview TypeScript builds
- targeted ESLint for all changed TypeScript/TSX files
- 51 focused tests passed across query completion, webview routing, and audio source generation

Repository-wide lint remains blocked by pre-existing CRLF/Prettier errors in unrelated files.


Benjin:
Really wanted this simple design to work, but Chromium requires interaction with the webview before audio elements can be played, which doesn't work well for a notification.

Closing in favor of https://github.com/microsoft/vscode-mssql/pull/22752

PR #22772, opened 2026-08-20
Move the Projects workspace implementation into the MSSQL extension

aasimkhan30:
## What this does

Moves the Projects workspace **implementation** out of the **data-workspace** extension and into **mssql**.

data-workspace keeps its manifest exactly as it is today — the view container, view, commands, menus and settings are all still contributed by it — but it no longer ships an implementation. mssql supplies the code behind those contributions and registers the handlers at runtime.

## Why

The Projects view, project provider framework and workspace commands were implemented in a separate extension whose only consumer was SQL Database Projects. Consolidating the implementation into mssql removes a cross-extension API hop and one of the extra API surfaces we want to retire, without changing anything users see.

## What changed

**mssql** gains the implementation in `src/dataWorkspace/`:

- workspace service, project provider registry, Projects tree data provider, dialogs and path helpers
- registers the `projects.*` / `dataworkspace.*` command handlers and the tree data provider
- `activate()` additionally returns `dataWorkspace`, the same API shape project extensions already consumed
- activates on `onView:dataworkspace.views.main`, so the tree provider is registered when the view is revealed

It registers **nothing** when the data-workspace extension is not installed — there is nothing to back when its contributions are absent.

**data-workspace** keeps its `contributes` block byte-identical to `main`. Only the implementation goes: its `src`, tests, dependencies and `test` script are removed, leaving an activation stub.

**sql-database-projects** is unchanged apart from resolving the workspace API from mssql instead of data-workspace, and its typings reference:

```diff
- vscode.extensions.getExtension(dataWorkspaceExtensionId)!.exports
+ vscode.extensions.getExtension(mssqlExtensionId)!.exports.dataWorkspace
```

It still depends on both extensions: data-workspace for the contributions it uses, mssql for the API behind them.

**Localization:** the moved strings convert from `vscode-nls` to `vscode.l10n` in a `DataWorkspace` class in `locConstants.ts`, matching the rest of mssql. They were never in an extraction pipeline before, so they become localizable for the first time (`vscode-nls` is no longer a dependency).

## User-visible impact

None, and deliberately so. Because data-workspace's contributions are unchanged, everything VS Code keys to those identifiers is preserved — a hidden Projects view, a view moved to another container, custom keybindings, and existing settings values including `dataworkspace.excludedProjects` and `projects.defaultProjectSaveLocation`.

Verified against `main`:

| | main | this branch |
|---|---|---|
| commands | 149 | 149 (identical ids, titles, categories) |
| settings | 143 | 143 |
| views | 5 | 5 |
| activity bar containers | 2 | 2 |

## Verification

- `npm run build` — all 5 targets
- `npm run build:extension:typecheck` — mssql and sql-database-projects
- `npm run lint` — no new findings (the sql-database-projects warnings are pre-existing on `main`; identical counts on both)

## Notes for reviewers

- The bulk of the diff is file moves; `git diff -M` renders them as renames.
- `ExtensionActivationError` and `WorkspacePlaceholder` are now parameterized, because they referenced helpers and constants that cannot live in `locConstants`. The former's call site previously passed a raw error object into a string slot — it now passes `getErrorMessage(err)`.
- `IconPathHelper` and four icons are deleted: it only populated `folder` and `refresh` icons that nothing read, which was already the case on `main`, and the two `Open_existing` icons belonged to a dialog that does not exist in the VS Code port.
- data-workspace's `test` steps are removed from CI and the release pipeline, since it no longer has a test suite.


extensions/sql-database-projects/src/common/utils.ts
```diff
@@ -294,8 +294,8 @@ export function getSqlProjectsInWorkspace(): Promise<vscode.Uri[]> {
 }
 
 export function getDataWorkspaceExtensionApi(): dataworkspace.IExtension {
-    const extension = vscode.extensions.getExtension(dataWorkspaceExtensionId)!;
-    return extension.exports;
+    const extension = vscode.extensions.getExtension(mssqlExtensionId)!;
```
aasimkhan30, line 297: Probably the most important bit.

extensions/sql-database-projects/src/common/utils.ts
```diff
@@ -294,8 +294,8 @@ export function getSqlProjectsInWorkspace(): Promise<vscode.Uri[]> {
 }
 
 export function getDataWorkspaceExtensionApi(): dataworkspace.IExtension {
```
Benjin, line 296: Minor, but probably could be renamed to `getDataWorkspaceApi since the extension no longer exists.

Benjin:
What's your long-term thinking for the functionality organziation?

Given that SQL Database Projects is the only consumer of the data workspace API, why put the data workspace code in MSSQL instead of SQL Database Projects?  Only reasons I can think of immediately are:

* SQL Database Projects will also eventually get folded into the MSSQL extension, or
* we're expecting to have other MSSQL-related extensions use the data workspace API

caohai (approved):

PR #22788, opened 2026-08-21
Fix SQL Server query result and schema workflows

KarunakarKotha09:
Fixes three SQL Server extension usability issues:\n\n- Hide the invalid Copy as IN clause action for multi-column selections.\n- Exclude the current schema from the Move to Schema picker.\n- Hide the unsupported drag-replace handle in the read-only Beta Results Grid.\n\nValidated with build, lint, focused tests, and online packaging.


aasimkhan30:
Please keep the PRs focused on one issue at a time and coordinate with the person who the issue is assigned to. Thanks!

PR #22822, opened 2026-08-25
Import performance harness under tools/perftest

kburtram:
## Summary

Imports the standalone `kburtram/perftest` performance harness into this repository under `tools/perftest` and adds lightweight root npm commands for working with it.

This is a tooling-only change. It does not enable performance scenarios by default, change extension runtime behavior, or add the harness to the normal vscode-mssql bootstrap/build graph.

The diff is large because it brings over the complete harness as a self-contained tool (218 files under `tools/perftest`) and applies the repository's normal formatting. Most of the change is the source import and mechanical relocation work.

## What is included

- Five npm workspaces:
  - `packages/observability-contracts`
  - `packages/perf-contracts`
  - `packages/perftest-cli`
  - `packages/perftest-inproc`
  - `extensions/mssql-perf-driver`
- The performance CLI, scenario registry, runners, reporters, central-store support, and unit tests.
- SQL provisioning, seed, and XEvents assets.
- SQL fixtures under `workspaces/perf`.
- Thirty example JSONC configurations.
- A Grafana dashboard and implementation documentation.
- Historical `IMPLEMENTATION_PLAN.md` and `PROGRESS.md` files, retained and clearly marked as frozen historical context.
- `UPSTREAM.md`, which records the source and future synchronization policy.

## Repository integration

The root `package.json` now exposes these convenience commands:

- `npm run perftest`
- `npm run perftest:install`
- `npm run perftest:build`
- `npm run perftest:test`
- `npm run perftest:typecheck`

These commands delegate into `tools/perftest` with `npm --prefix`. The harness is intentionally not part of the normal root workspace/bootstrap path because its CLI includes the native `better-sqlite3` dependency. Keeping it opt-in avoids adding that dependency and its build cost to every contributor and CI job.

The root `.gitignore` also contains narrow exceptions for `tools/perftest/examples/**`; these are needed because the repository's broad `examples` rule would otherwise omit the tracked example configurations.

## Import provenance

- Source repository: `kburtram/perftest`
- Source commit: `860365ff65f9bbb4e102309860004d8e9bdf36f4`
- Imported from tracked files only, including the two local source commits that were ahead of its remote.

Generated reports, local run data, build outputs, dependencies, databases, and other machine-local artifacts were not imported. The old nested `.github/workflows/perf-nightly.yml` was also excluded because GitHub does not discover workflows nested below the repository-level `.github/workflows` directory.

## Relocation fixes

The import includes the small changes needed for the new in-repository location:

[...]

PR #22827, opened 2026-08-26
Add unified core observability and performance instrumentation

kburtram:
## Summary

This PR introduces the core observability substrate for vscode-mssql and moves the existing `PERF_MODE` instrumentation onto it. It is the first product extraction from the larger `dev/refactor` tree and is intentionally based directly on `main`; it is not stacked on the perftest tooling PR.

The change provides one privacy-safe event/span model for extension diagnostics, performance markers, bounded live-tail consumption, and durable session JSONL output. It also adds a focused first set of lifecycle probes so activation, SQL Tools Service startup, connection, query, command, and webview behavior can be measured consistently.

Related tooling: #22822. That PR is useful for end-to-end validation, but this branch has no build-time or runtime dependency on `tools/perftest` and can merge independently.

### Unified diagnostics contracts and core

- Adds extension-owned diagnostic event, span, trace, classification, sink, health, and session contracts.
- Adds trace/span identity and cross-plane link support.
- Makes span completion idempotent so duplicate success/failure paths cannot emit multiple terminal records.
- Supports safe sink registration, replacement, disposal, and isolated sink failures.
- Uses a small runtime event-to-feature map so ordinary extension activation does not load the large generated registry snapshot.

### Privacy and redaction

- Classifies attributes before they reach sinks.
- Defaults unknown non-performance attributes to a closed/redacted classification.
- Prevents provider error messages from becoming diagnostic metadata; failures use bounded reason categories instead.
- Keeps SQL text, result values, connection strings, passwords, tokens, and other sensitive values out of the diagnostic envelope.
- Adds on-disk privacy tests for the session JSONL sink.

### Sinks and compatibility

- Adds a bounded live-tail sink with exact dropped-record/gap accounting.
- Adds a session JSONL sink with privacy-safe health reporting and deterministic cleanup.
- Preserves the existing performance marker wire format through a compatibility sink/facade.
- Refactors `Perf`/`PERF_MODE` instrumentation to emit through the unified diagnostics core while retaining gated behavior when performance mode is disabled.

### Initial instrumentation

- Extension activation begin/end.
- Command invocation.
- SQL Tools Service spawn, ready, and failure lifecycle.
- Connection begin, retry, ready, and safe failure phases.
- Query submission, completion, and immediate failure.
- Webview render marks, including race-free enablement and queued early marks.
- Query Results grid render milestones.

### Test and development seams

- Adds a performance-only extension state API reserved for later driver health checks.
- Adds `PERF_DOTNET_PATH` as a controlled local-validation seam for selecting a patched .NET runtime.
- Adds an exact generated observability registry snapshot plus conformance tests. The generated file is validation input; normal activation does not import it.

### Review hardening

- Pairs query submission/completion markers across document, current-statement, and Quick Query execution paths, including request failures.
- Treats non-2xx marker POST responses as delivery failures and honors the harness's exact partial-rejection count.
- Restores a bounded, honestly labeled fallback for paint markers when a hidden webview throttles `requestAnimationFrame`.
- Makes repeated sink registration idempotent, disposes replacement/timed sinks safely, and bounds elevated-capture durations.
- Resets queued delivery and per-subscriber gap state at live-tail subscription boundaries so a new subscriber receives its snapshot plus only new events.
- Keeps the runtime privacy map complete for all emitted literal marker attributes and expands syntax-aware conformance checks across marker names, registry features, attributes, and direct diagnostics calls.
- Removes the redundant `mssql.sts.pid` event; the existing `mssql.sts.ready` marker remains the single PID-bearing lifecycle signal.

## Important design choices

- This is infrastructure plus a coherent first group of probes, not the complete Debug Console feature.
- Product code owns its diagnostic contracts. The external performance harness may verify them, but the extension does not import generated build output from that tool.
- Diagnostics fail closed: unknown attributes are not assumed safe, error text is not copied into events, and sink health errors do not expose payload content.
- Performance-mode compatibility is preserved so existing harness scenarios continue to work while future diagnostics can share the same core.
- The branch is directly mergeable to `main`; no unmerged PR is used as its base.

## Intentionally out of scope

- Debug Console controller or React UI.
- Performance History and in-product self-test UI.
- Generic capture/replay repositories and full session-bundle management.
- Central diagnostics upload.
- SQL Data Plane and STS v2 endpoint wrappers.
- Metadata service, native SQL language service, and AI completion.
- Query Studio, OE v2, Vector, Spatial, and their feature-specific diagnostic pages.

Those pieces will be extracted in later direct-to-main PRs after their foundations are accepted.

## Validation

- [x] Fresh root `npm install` through the configured Microsoft npm proxy; no intended lockfile changes.
- [x] Full repository `npm run build` (Extension Toolkit, MSSQL extension/typecheck/bundle/webviews, SQL Database Projects, and Data Workspace).
- [x] Final MSSQL `npm run build:extension`.
- [x] MSSQL `npm run lint`.
- [x] Final focused observability + Query Runner suites passed, including real HTTP rejection accounting, all query entry-point marker paths, and live-tail unsubscribe/resubscribe boundaries.
- [x] Related controller, query, service-client, webview, extension API, and main-controller suites: 330 passed, 2 pending.
- [x] Complete MSSQL unit selection excluding two independently reproduced current-`main` Windows path-expectation suites passed with exit code 0.
- [x] Full unfiltered MSSQL run reached 3,631 passed and 12 pending; its eight failures were isolated to untouched current-`main` Windows path assertions (one `Utility Tests - Path handling` case and seven `QueryCompletionSoundService` cases).
- [x] Built and patched current SQL Tools Service `main` (`e833cc23`) into the extension's Windows and Portable targets.
- [x] Perftest build/typecheck passed. Registry vendor-sync passed; all non-live-store harness suites passed (the optional central-store integration requires a separately provisioned localhost SQL instance).
- [x] Final real VS Code `ext-normal-activation` run `2026-08-26T21-14-03Z_2381ee87` passed against the final review-fix tree and patched STS build.
- [x] Online MSSQL packaging passed using the locally patched STS payload, producing a 212.66 MiB validation VSIX. Its duplicated local Windows/Portable Debug payload is intentionally not representative of release artifact size.

## Commit structure

1. `perf: PERF_MODE-gated harness instrumentation (activation/command/STS markers, perf-only state API)`
2. `perf: connection/query markers + webview mark bridge`
3. `perf: PERF_DOTNET_PATH runtime fallback + race-free webview mark enablement`
4. `diag: add unified observability core and privacy-safe sinks`


extensions/mssql/src/controllers/queryRunner.ts
```diff
@@ -434,12 +436,17 @@ export default class QueryRunner {
                     );
                 }
             }, Constants.stsImmediateActivityTimeout);
+            Perf.marker("mssql.query.submit", "begin");
```
kburtram: **Should-fix.** This is the only execute path that emits `mssql.query.submit`; `runStatement` (`QueryExecuteStatementRequest`, :366) and `runQueryString` (`QueryExecuteStringRequest`, :501) skip it, yet all three complete through `handleQueryComplete`, which emits `mssql.query.complete` (:551). Ctrl+F5 / quick query therefore produce an `end` with no `begin`: the registry pairs `submit ↔ complete`, the harness's `mssql.query.toComplete` metric is silently absent ("not both observed"), and `lintCorrelation` reports unmatched pairs. Emit the `submit` (and the failure-path `complete { hasError, errorClass }`) in all three, e.g. via a small private helper.

(source: Fable)

extensions/mssql/src/webviews/common/perfMarks.ts
```diff
@@ -0,0 +1,101 @@
...
+ * requestAnimationFrame, design §17.5) — the honest "visually complete"
+ * moment. Runs unconditionally so pre-enablement marks still carry
+ * paint-accurate timestamps; the cost is two coalesced rAF callbacks on
+ * completion events only.
+ */
+export function perfMarkAfterNextPaint(
+    name: string,
+    attrs?: { [key: string]: string | number | boolean | null },
+): void {
+    requestAnimationFrame(() => {
```
kburtram, line 114: **Should-fix / question.** `dev/refactor`'s version (`perfMarkAfterNextPaintComputed`) had a 500 ms fallback with `rafThrottled: true`, with a comment documenting the incident: "Hidden/backgrounded webviews get their rAF throttled to a standstill (VS Code suspends hidden views) … BOOT-4: warmup reps lost editorInteractive entirely to this." This extraction dropped it. With a bare double rAF, `mssql.resultsGrid.renderComplete` never fires if the results panel is not visible when the query completes (background tab, minimised window) → the harness's `waitForMarker` times out → invalid rep. Was the drop deliberate? If not, restore the fallback and register `rafThrottled` on the `renderComplete` entry in #22822 (it is registered only on the Query Studio marks today).

(source: Fable)

extensions/mssql/src/webviews/pages/QueryResult/queryResultsGridView.tsx
```diff
@@ -170,6 +172,29 @@ export const QueryResultsGridView = ({
...
+        if (isExecuting !== false || gridList.length === 0) {
+            return;
+        }
+        const totalRows = gridList.reduce(
+            (total, item) =>
+                total + (resultSetSummaries?.[item.batchId]?.[item.resultId]?.rowCount ?? 0),
+            0,
+        );
+        const key = `${uri}:${gridList.length}:${totalRows}`;
+        if (lastPerfMarkKey.current === key) {
```
kburtram, line 192: Latent: the key is `uri:resultSets:totalRows`, so re-executing the same document to an identical shape never emits a second `renderComplete` (`isExecuting` toggles but the key does not). No current scenario re-runs inside one launch — each rep is a fresh VS Code — so nothing breaks today, but a "re-run warm" scenario or the Debug Console live view would silently miss the mark. Reset `lastPerfMarkKey.current = undefined` when `isExecuting === true` (or key on an execution counter).

(source: Fable)

kburtram (commented):
## Review — PR01 core observability (`39cd4409c`)

Read every non-generated line of the diff (~2.5K lines) plus the `dev/refactor` originals it was extracted from, and cross-checked the wire contract against `tools/perftest` at `f44cb8263` (#22822). Verified locally on this head: `tsgo` typecheck for the extension and the webviews, `eslint` on every changed file, the three new suites in VS Code (`--grep "Core observability|Core diagnostics redaction|Observability Contract"` → 21 passing), a fresh `build:extension-bundle`, and `git merge-tree` against the current `origin/main` (`a362cc38d`, one Juno commit newer than the PR base) — clean.

Overall this is a well-bounded extraction: the vendored registry snapshot is byte-identical to #22822's generated output, `featureFor` and the marker phases/attr names were corrected to match the registry relative to `dev/refactor` (`connection.failed` is now `instant`, `commandId` not `command`), provider error text is kept out of the envelope, and the no-sink path is a single length check. Nothing here blocks merge on its own; the items below are the gaps I would close first.

1. **`mssql.query.submit` is only emitted on the `runQuery` path** (`queryRunner.ts:439`). `runStatement` (`QueryExecuteStatementRequest`, :366) and `runQueryString` (:501) still produce `mssql.query.complete` from `handleQueryComplete`, so Ctrl+F5 / quick query yield an `end` with no `begin`. The registry pairs `submit ↔ complete`; the harness's `mssql.query.toComplete` metric goes silently absent and `lintCorrelation` reports unmatched pairs. Instrument all three paths (same failure-path `complete { hasError, errorClass }` too).
2. **`PerfModeSink` counts HTTP 4xx/5xx as delivered** (Copilot's thread at `sinks.ts:186`) — confirmed empirically: the harness answers `400 {accepted, rejected}` when any line is rejected and `401` on token mismatch, and `http.request` emits no `error` for those. Only timeouts/refused connections are accounted. Check `response.statusCode` (and use the `rejected` count from the body when present).
3. **Dropped rAF-throttle fallback in `perfMarkAfterNextPaint`** (`perfMarks.ts:98`). `dev/refactor` had a 500 ms fallback with `rafThrottled: true` and a comment documenting the incident it fixed ("BOOT-4: warmup reps lost editorInteractive entirely"). Hidden/backgrounded webviews get rAF suspended, so `mssql.resultsGrid.renderComplete` never fires if the results panel is not visible when the query completes → `waitForMarker` timeout → invalid rep. If the drop was deliberate for scope, say so; otherwise restore it (and register `rafThrottled` on the entry in #22822).
4. **`mssql.sts.pid` feature bucket** (Copilot's thread at `perfTelemetry.ts:218`) — agree; in addition it is not tagged `perfMarker`, so `PerfModeSink` ignores it, no sink in this PR receives it, and it bypasses the conformance regex. Since `mssql.sts.ready` already carries `pid` (and that is what the harness reads — `runPipeline.ts:718`), drop it or route it through `Perf.marker`.
5. **Conformance test coverage** (`observabilityContract.test.ts:59`): only `Perf.marker("…")` literals are scanned (missed `mssql.sts.pid`), only four names are checked for feature-bucket agreement, and attrs are never compared. Extend it to every emitted name and capture `type:`/attr keys — it is the only thing keeping `featureFor` honest.

- `diagnosticsCore.ts:136` (Copilot): initialise `policy` to `CAPTURE_POLICIES.redacted` to match the documented invariant at :304 — no effect today (PERF_MODE attrs are all `diagnostic.metadata`; no other sink exists), but the first non-perf sink would otherwise see everything `omitted`.
- `perfTelemetry.ts:99` (Copilot): add `reason` and `resultSets` to `ATTR_CLASSIFICATION`. Note the map is currently unreachable (consulted only when `!enabled`, and outside PERF_MODE no sink is registered), which is why a test enumerating emitted attr keys is the more durable fix.
- `queryResultsGridView.tsx:188`: `lastPerfMarkKey` suppresses `renderComplete` when the same URI re-executes to an identical shape. No current scenario re-runs inside one launch (each rep is a fresh VS Code), so it is latent — reset the key when `isExecuting` flips to `true`.
- `sinks.ts:63`: `PerfModeSink` has no `dispose()`, so `removeSink` cannot clear its flush timer (the unit test at `coreObservability.test.ts:172` leaves one that POSTs to `127.0.0.1:1` after the test ends).
- `diagnosticsCore.ts:172`: `addSink` of an already-registered instance disposes it and re-adds the disposed object; and the "sink replacement" behaviour named in the PR description has no test.
- Registry gaps to fix in #22822 and re-vendor here — attrs emitted but undeclared: `mssql.activate.end` (`failed`, `error`, `errorClass`), `mssql.sts.spawn.end` (`error`, `errorClass`, `pid`), `mssql.sts.ready` (`pid` — load-bearing, see `runPipeline.ts:718`), `mssql.connection.failed` (`error`, `errorNumber`), `mssql.query.complete` (`errorClass`), `mssql.resultsGrid.renderComplete` (`resultSets`). All six entries are `attrsComplete: false`, so per the registry conventions this is a docs gap rather than a violation — but `pid` on `sts.ready` is a real contract.
- `sinks.ts:13` (Copilot): the "non-blocking writes" comment vs `appendFileSync`/`writeFileSync` — fix the comment now (writes are timer-batched, off the emission path, but they do block the host for their duration); switch to a serialised `fs.promises` chain when PR02 gives the sink a consumer.

### C — questions / notes (no change required)

- **`mssql.perf.getState` has no caller in #22822** at `f44cb8263` (only `docs/PRODUCT_INSTRUMENTATION.md` mentions it). The description says "consumed by the harness"; today `markersQueued`/`markersDropped` are observed by nothing. Either wire the driver's success checks to it (perftest follow-up) or soften the description.
- Forward references that a reader of `main` cannot resolve: `SAFE_ERROR_CODES` (`SqlDataPlane.*`) and five of the six non-builtin `SAFE_ERROR_NAMES` (only `MissingEntraAuthAccountError` exists on main), `ROOT_BEGINNERS` (`mssql.oe.*`, `command.*.begin`, `userAction.*`), `FORWARDED_SPAN_TYPES` (`rpc.`, `webview.`, `sts.`, `sqlDataPlane.*`), the `RANK_ORDER` comment about the "vendored central-contract equality test (cls-rank/1)", "(WI-2.3)" in `SessionDiagSink`, `debug-docs/MSSQL_Debug_Console_Technical_Design.md`, "harness design §16.3 / §17.5", "Phase 4 / Phases 1-3", "M16.7". Harmless allowlists, but a one-line "reserved for <later PR>" note would save the next reviewer a search.
- Unused-but-shipped surface (`LiveTailSink`, `SessionDiagSink`, `setCaptureMode`/revert timer, rich mode, `withTrace`/`bindEntityTrace`, `sinkHealthSnapshot`): consistent with the "foundation" scope in the split plan. A fresh bundle at this head tree-shakes both sinks and does **not** include the generated registry. (Your local `dist/extension.js` from 22:02 *did* include it — 130 KB larger — it predates the final commit; not a PR defect, just do not trust that artifact.)
- `PERF_DOTNET_PATH` (`dotnetRuntimeProvider.ts:32`) is equivalent in trust to the existing environment-controlled overrides and is double-gated; only note is the 28 % patch coverage — a unit test with `PERF_MODE=1` + an inaccessible path would cover both branches.
- Cross-PR: #22822's `vendorSync.test.ts` only checks its own package; any registry edit there (items above) must be followed by re-vendoring `observabilityContract.generated.ts` here. `dev/refactor`'s vendored copy is already stale relative to both.

Copilot's five comments are all valid; I replied on each thread with severity and the concrete fix.

(source: Fable)

extensions/mssql/test/unit/observabilityContract.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    });
+
+    test("every Perf marker literal emitted by src/ is registered", function () {
+        if (!fs.existsSync(SRC_ROOT)) {
+            this.skip(); // packaged test run without sources
+        }
+        const emitted = new Set<string>();
+        for (const file of walk(SRC_ROOT)) {
+            const source = fs.readFileSync(file, "utf8");
+            for (const m of source.matchAll(/Perf\.(?:marker|begin|end|instant)\(\s*"([^"]+)"/g)) {
```
kburtram: Coverage gaps in the one test that keeps the runtime map honest: (1) only `Perf.marker("…")` literals are scanned, so `diag.emit({ type: "mssql.sts.pid" })` in `perfTelemetry.ts:216` is never checked — add a `type:\s*"(mssql\.[^"]+)"` pattern; (2) the feature-bucket test above checks four hard-coded names — check every emitted name (`featureFor(name) === explainEventName(name).entry?.feature`); (3) attrs are never compared. Six emitted markers carry attrs the registry does not declare (`activate.end`: failed/error/errorClass; `sts.spawn.end`: error/errorClass/pid; `sts.ready`: pid; `connection.failed`: error/errorNumber; `query.complete`: errorClass; `resultsGrid.renderComplete`: resultSets). All six entries are `attrsComplete: false`, so per the conventions that is a docs gap, not a violation — but `pid` on `sts.ready` is load-bearing (`runPipeline.ts:718` discovers the STS process from it) and should be declared in #22822 and re-vendored.

(source: Fable)

extensions/mssql/src/languageservice/serviceclient.ts
```diff
@@ -389,7 +391,20 @@ export default class SqlToolsServiceClient {
...
+            } catch (error) {
+                Perf.marker("mssql.sts.spawn.end", "end", {
+                    error: true,
+                    errorClass: diagnosticErrorClass(error),
+                });
+                throw error;
+            }
+            const stsPid = this.client.serverProcess?.pid;
+            Perf.setStsPid(stsPid);
+            Perf.marker("mssql.sts.spawn.end", "end", { pid: stsPid ?? null });
```
kburtram, line 406: FYI, this attr is a real cross-repo contract: `perftest-cli/src/run/runPipeline.ts:718` registers the STS process (for dotnet-counters/trace/gcDump) only when `mssql.sts.ready.attrs.pid` is a number — yet the registry entry declares no attrs. Please declare it in #22822 and re-vendor. Two nits: `pid: null` when `serverProcess` is undefined is the honest choice (no fabrication) — fine; and `Perf.setStsPid` is never refreshed if `LanguageClient` auto-restarts the server after a crash, so `getState().stsPid` can go stale.

(source: Fable)

extensions/mssql/src/diagnostics/diagnosticsCore.ts
```diff
@@ -0,0 +1,562 @@
...
+    ];
+
+    constructor() {
+        const stamp = new Date().toISOString().replace(/[-:T]/g, "").slice(0, 14);
+        this.sessionId = `sess_${stamp}_${process.pid}`;
+    }
+
+    // --- sinks ---------------------------------------------------------------
+
+    public addSink(sink: DiagnosticSink): void {
```
kburtram, line 178: Nit: `addSink(sink)` where `sink` is the instance already registered disposes it (`removeSink` → `dispose()`) and then re-adds the disposed object — for `LiveTailSink` that means listener/pending cleared. Early-return when the same instance is present. Also, the PR description lists "sink replacement" among the tested behaviours, but no test covers `addSink` replacing a same-id sink and disposing the old one.

(source: Fable)

extensions/mssql/src/diagnostics/sinks.ts
```diff
@@ -0,0 +1,542 @@
...
+    /^(rpc\.|webview\.|sts\.|sqlDataPlane\.auth\.token(?:\.|$)|sqlDataPlane\.tsNative\.query\.terminal$)/;
+
+/**
+ * Forwards marker-tagged diagnostics to the perf orchestrator's HTTP sink.
+ * Events emitted through the Perf facade (tag "perfMarker") are relayed in the
+ * exact legacy wire format; diagnostic spans matching FORWARDED_SPAN_TYPES are
+ * relayed additively (tagged diag) so harness traces carry cross-process
+ * detail. Everything else stays out of the harness contract.
+ */
+export class PerfModeSink implements DiagnosticSink {
```
kburtram, line 64: Nit: `PerfModeSink` is the one sink without `dispose()`, so `DiagnosticsCore.removeSink` cannot clear `flushTimer` (unref'd, so no hang). Add `dispose() { clearTimeout(this.flushTimer); this.flush(); }` — and call it from `coreObservability.test.ts:172`, which currently leaves a 250 ms timer that POSTs to `127.0.0.1:1` after the test has ended.

(source: Fable)

extensions/mssql/src/perf/perfTelemetry.ts
```diff
@@ -0,0 +1,246 @@
...
+        this.enabled = process.env.PERF_MODE === "1";
+        if (this.enabled) {
+            const markerUrl = process.env.PERF_MARKER_URL;
+            const token = process.env.PERF_CONTROL_TOKEN;
+            if (markerUrl && token) {
+                this.perfSink = new PerfModeSink(
+                    markerUrl,
+                    token,
+                    process.env.PERF_RUN_ID ?? "unknown-run",
+                    Number(process.env.PERF_REP_ID ?? "0"),
```
kburtram: Nit: `Number(process.env.PERF_REP_ID ?? "0")` yields `NaN` on any non-numeric value; `JSON.stringify` then emits `"repId": null`, the harness schema (`repId: integer ≥ 0`) rejects every marker with 400, and — until the non-2xx accounting in `PerfModeSink.flush` is fixed — the sink still reports `markersDropped: 0`. `Number.parseInt` + `Number.isInteger` guard, fallback 0.

(source: Fable)

extensions/mssql/src/diagnostics/redaction.ts
```diff
@@ -0,0 +1,254 @@
...
+ * every payload field passes through here.
+ */
+export function classify(
+    raw: unknown,
+    cls: DataClassification,
+    policy: CapturePolicy,
+): ClassifiedValue {
+    // Non-sensitive primitives pass through untouched (numbers/booleans carry
+    // no text payload risk for metadata classes).
+    if (ALWAYS_PLAIN.has(cls)) {
```
kburtram, line 103: Nit: the `ALWAYS_PLAIN` branch returns the raw value untruncated; `MAX_PLAIN_LENGTH` (:117) only applies in the policy-gated branch. Today every `diagnostic.metadata` attr is an enum/number, but the first free-text metadata string of any size goes to disk/harness whole. Apply the same truncation here. Separately (question): `NEVER_PLAIN` classes are tokenised even under `policy.mode === "off"` because :108 runs before :113 — intended, or should "off" omit them like everything else?

(source: Fable)

extensions/mssql/src/diagnostics/diagnosticsCore.ts
```diff
@@ -0,0 +1,562 @@
...
+     * v1 keeps it simple: one policy applied at emission covering all sinks,
+     * with PERF_MODE (synthetic data by contract) treated as full-capture.
+     */
+    public setCaptureMode(mode: CaptureMode, options?: { reason?: string; durationMs?: number }) {
+        if (this.policyRevertTimer) {
+            clearTimeout(this.policyRevertTimer);
+            this.policyRevertTimer = undefined;
+        }
+        if (mode === "full") {
+            const durationMs = Math.min(options?.durationMs ?? 15 * 60_000, 60 * 60_000);
```
kburtram: Nit (no caller in this PR): `Math.min(options?.durationMs ?? 15 * 60_000, 60 * 60_000)` accepts 0 or a negative value → the revert timer fires immediately. Clamp with a lower bound as well. Related at :518–540: `fail()` omits `timingClass: "officialSameProcess"` while `end()` sets it, so failed spans are classified differently from successful ones of the same family.

(source: Fable)

extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -1443,6 +1444,7 @@ export default class ConnectionManager {
             serverlessWakeFailedAttempts = 0,
         } = options;
 
+        Perf.marker("mssql.connection.begin", "begin");
```
kburtram: Nit: `mssql.connection.begin` fires before `prepareConnectionInfo` (:1467), which can throw or be cancelled (password / Entra prompts) — that path leaves a `begin` with neither `ready` nor `failed`. The harness pairs the last `begin` before the first `ready`, so metrics are unaffected; `lintCorrelation` would show an orphan. A `connection.failed { reason: "prepareFailed" }` in a catch around `prepareConnectionInfo` (or moving `begin` after it) closes it.

(source: Fable)

extensions/mssql/src/perf/perfApi.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Perf } from "./perfTelemetry";
+
+export const perfGetStateCommand = "mssql.perf.getState";
+
+export function registerPerfApi(context: vscode.ExtensionContext): void {
+    if (!Perf.enabled) {
+        return;
+    }
+    context.subscriptions.push(
+        vscode.commands.registerCommand(perfGetStateCommand, () => Perf.getState()),
```
kburtram, line 23: Question: nothing in #22822 at `f44cb8263` calls `mssql.perf.getState` (only `docs/PRODUCT_INSTRUMENTATION.md` describes it), so `markersQueued`/`markersDropped`/`stsPid` are currently observed by nothing and the PR description's "consumed by the harness" is aspirational. Fine to keep the seam (it is tiny and gated), but either wire the driver's success checks to it in a perftest follow-up or soften the description.

(source: Fable)

extensions/mssql/src/diagnostics/diagnosticsCore.ts
```diff
@@ -0,0 +1,562 @@
...
+    health?(): { id: string; healthy: boolean; detail: string; counters: Record<string, number> };
+}
+
+export interface DiagSpan {
+    readonly traceId: string;
+    end(status?: DiagStatus, fields?: Record<string, RawField>): void;
+    fail(error: unknown): void;
+}
+
+const SAFE_ERROR_CODES = new Set([
```
kburtram, line 78: Nit: of everything in `SAFE_ERROR_CODES` / `SAFE_ERROR_NAMES`, only the JS builtins and `MissingEntraAuthAccountError` exist on `main`; the `SqlDataPlane.*` codes and the other five names belong to later PRs. Harmless as an allowlist, but a reader of `main` will search for them — a one-line "reserved for the SQL Data Plane / Entra PRs" note (same for `ROOT_BEGINNERS` :155 and `FORWARDED_SPAN_TYPES` in sinks.ts:53) would help. Practical consequence today: every error in this PR classifies as `Error` or `UnknownError` (`ResponseError` from vscode-jsonrpc keeps `name === "Error"`), so `errorClass` carries little signal until those land.

(source: Fable)

extensions/mssql/src/perf/perfTelemetry.ts
```diff
@@ -0,0 +1,246 @@
...
+    messages: "user.text",
+    error: "diagnostic.metadata",
+    errorClass: "diagnostic.metadata",
+    errorNumber: "diagnostic.metadata",
+    failed: "diagnostic.metadata",
+    commandId: "diagnostic.metadata",
+    hasError: "diagnostic.metadata",
+    pid: "diagnostic.metadata",
+    rowCount: "diagnostic.metadata",
+};
```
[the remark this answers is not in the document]
kburtram: Valid, low severity. Missing today: `reason` (`mssql.connection.failed`, bounded enum) and `resultSets` (`mssql.resultsGrid.renderComplete`, number) — both should be `diagnostic.metadata`. Worth knowing the map is currently unreachable: it is consulted only when `!this.enabled`, and outside PERF_MODE no sink is ever registered, so `Perf.marker` returns at the `anySinkActive` check. That is exactly why the durable fix is a test: extend `observabilityContract.test.ts` to capture the attr keys of every emitted `Perf.marker`/`perfMarkAfterNextPaint` literal and assert each has an entry here (or is declared in the registry).

(source: Fable)

extensions/mssql/src/perf/perfTelemetry.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    public setStsPid(pid: number | undefined): void {
+        this.stsPid = pid;
+        if (pid !== undefined) {
+            diag.emit({
+                feature: "connection",
+                type: "mssql.sts.pid",
+                fields: { pid: { raw: pid, cls: "diagnostic.metadata" } },
+            });
```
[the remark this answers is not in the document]
kburtram: Agree — `featureFor("mssql.sts.pid")` gives `stsLifecycle`, which is what the registry says. Two further points on the same call: the event is not tagged `perfMarker`, so `PerfModeSink` ignores it and no sink in this PR receives it at all; and because it goes through `diag.emit` rather than `Perf.marker`, the conformance test's regex never sees the name. Since `mssql.sts.ready` already carries `pid` (and `perftest-cli/src/run/runPipeline.ts:718` reads it from there), I would either delete this emission or route it through `Perf.marker` so it is covered by the same contract.

(source: Fable)

extensions/mssql/src/diagnostics/diagnosticsCore.ts
```diff
@@ -0,0 +1,562 @@
...
+    return `trace_${(hint ?? "act").replace(/[^a-z0-9]/gi, "").slice(0, 12)}_${Date.now().toString(36)}_${traceCounter.toString(36)}`;
+}
+
+export class DiagnosticsCore {
+    private sinks: DiagnosticSink[] = [];
+    private seq = 0;
+    private eventCounter = 0;
+    public readonly sessionId: string;
+    private policy: CapturePolicy = CAPTURE_POLICIES.off;
+    private storePolicy: CapturePolicy = CAPTURE_POLICIES.off;
```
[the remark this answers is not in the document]
kburtram: Agree with the reading: :304–305 documents that the emission policy is `redacted` whenever the store policy is `off`, but the field initialiser says `off`, and nothing in this PR calls `setCaptureMode`. No behavioural effect today — under PERF_MODE every attr is classified `diagnostic.metadata` (ALWAYS_PLAIN, unaffected by mode) and there is no other sink — but the first non-perf sink added by PR02 would see every name/text field `omitted`. Initialise `policy = CAPTURE_POLICIES.redacted` and add a one-line test that a fresh core with a collecting sink classifies a `server.name` field as `digest`, not `omitted`.

(source: Fable)

extensions/mssql/src/diagnostics/sinks.ts
```diff
@@ -0,0 +1,542 @@
...
+                {
+                    method: "POST",
+                    headers: {
+                        "content-type": "application/x-ndjson",
+                        authorization: `Bearer ${this.token}`,
+                    },
+                    timeout: PERF_POST_TIMEOUT_MS,
+                },
+                (response) => response.resume(),
+            );
```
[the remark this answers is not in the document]
kburtram: Confirmed, and I would rank it should-fix rather than nit. The harness's `/v1/markers` handler (`controlServer.ts:189`) answers `400 {accepted, rejected}` when any NDJSON line fails schema validation and `401` on a token mismatch; `http.request` emits no `error` for either, so the batch is counted as delivered. I checked the other paths empirically: a timeout → `request.destroy()` → `ECONNRESET` error → counted; refused connection → counted; 400 → not counted. Fix: in the response callback, treat `statusCode` outside 2xx as dropped — and since the body carries `rejected`, use that count when it parses so the accounting stays exact rather than all-or-nothing. Caveat for expectations: `markersDropped` is only observable through `mssql.perf.getState`, which nothing in #22822 calls yet.

(source: Fable)

extensions/mssql/src/diagnostics/sinks.ts
```diff
@@ -0,0 +1,542 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Diagnostic sinks. Each sink is independently gated:
+ *  - PerfModeSink:    PERF_MODE=1 (harness runs) — reproduces the exact legacy
+ *                     perf-marker wire format so Phases 1-3 stay bit-compatible.
+ *  - LiveTailSink:    active while a Debug Console subscribes; bounded ring
+ *                     with exact gap accounting.
+ *  - SessionDiagSink: user-enabled Session Diag store; JSONL segment journal
+ *                     with batched, non-blocking writes.
```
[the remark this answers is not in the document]
kburtram: Agree it overstates. The writes are timer-batched (500 ms, off the emission hot path), but `appendFileSync`/`writeFileSync` do block the extension host for their duration, and the constructor does a synchronous `mkdirSync` + manifest write. Since this sink has no consumer in this PR, fixing the comment ("batched synchronous writes off the emission path") is enough now; when PR02 wires it in, a serialised `fs.promises` write chain (one in-flight append, manifest rewrite after it settles) keeps the exact-range accounting intact.

(source: Fable)

extensions/mssql/src/perf/perfTelemetry.ts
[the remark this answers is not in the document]
kburtram: Fixed in 829d4d30d. `reason`, `resultSets`, and the fallback's `rafThrottled` flag now have explicit runtime privacy classifications. The conformance suite now parses the source syntax and fails if any literal marker attribute lacks a runtime classification. (Source: GPT-5.6 Sol)

extensions/mssql/src/perf/perfTelemetry.ts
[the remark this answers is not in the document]
kburtram: Fixed in 829d4d30d by removing the redundant `mssql.sts.pid` diagnostic emission. `setStsPid` now only maintains the gated state seam, while the existing `mssql.sts.ready` marker remains the single PID-bearing lifecycle signal used by the harness. (Source: GPT-5.6 Sol)

extensions/mssql/src/diagnostics/diagnosticsCore.ts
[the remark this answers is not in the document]
kburtram: Fixed in 829d4d30d. A fresh core now uses the redacted emission policy while its store policy remains off, matching `setCaptureMode("off")`. A regression test verifies that an active collecting sink receives a digest for `server.name`, not an omitted or plaintext value. (Source: GPT-5.6 Sol)

extensions/mssql/src/diagnostics/sinks.ts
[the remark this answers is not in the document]
kburtram: Fixed in 829d4d30d. The sink now treats every non-2xx response as a delivery failure, uses a valid harness `rejected` count for exact partial accounting, counts the whole batch for malformed/foreign responses, and guards against response/request double-counting. A real local HTTP test covers a 400 response with one accepted and one rejected marker. (Source: GPT-5.6 Sol)

extensions/mssql/src/diagnostics/sinks.ts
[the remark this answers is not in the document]
kburtram: Fixed in 829d4d30d by describing the implementation accurately as batched synchronous writes off the emission path. Converting the session store to a serialized asynchronous write chain remains part of the later session-management integration, where its lifecycle and ordering can be tested as a unit. (Source: GPT-5.6 Sol)

extensions/mssql/src/controllers/queryRunner.ts
kburtram: Fixed in 829d4d30d. Shared helpers now emit `mssql.query.submit` and a failure terminal marker for `runQuery`, `runStatement`, and `runQueryString`. Unit coverage checks the submit markers on all three paths and failure completion on each request style without depending on telemetry call counts or ordering. (Source: GPT-5.6 Sol)

extensions/mssql/src/webviews/common/perfMarks.ts
kburtram: Fixed in 829d4d30d. The 500 ms hidden-webview fallback is restored, is exactly-once with the double-rAF path, and labels fallback emissions with `rafThrottled: true`; that attribute is also runtime-classified and conformance-checked. The source registry entry is intentionally left to the #22822 source-of-truth follow-up while `attrsComplete` remains false. (Source: GPT-5.6 Sol)

extensions/mssql/src/webviews/pages/QueryResult/queryResultsGridView.tsx
kburtram: Fixed in 829d4d30d. The grid now clears its last render-marker key whenever execution starts, so a subsequent execution with the same URI and result shape emits its own post-paint completion marker. (Source: GPT-5.6 Sol)

extensions/mssql/test/unit/observabilityContract.test.ts
kburtram: Fixed in 829d4d30d. The regex check was replaced with a TypeScript syntax walk covering literal `Perf.marker`, webview mark, direct `diag.emit`, and direct `diag.startSpan` emissions. It validates every discovered name, every runtime feature bucket, every literal attribute's runtime privacy classification, and closed registry attrs when `attrsComplete` becomes true. Registry documentation gaps under `attrsComplete: false` remain in #22822, which owns the source registry. (Source: GPT-5.6 Sol)

extensions/mssql/src/languageservice/serviceclient.ts
kburtram: The load-bearing `mssql.sts.ready.attrs.pid` behavior is retained and the final runtime activation scenario passed with it. I did not edit the generated snapshot independently: its registry source belongs to #22822 and still marks this attr list incomplete. I also did not add auto-restart PID plumbing here because `getState` currently has no caller and restart lifecycle instrumentation belongs with the later service-lifecycle work; the API and PR text now describe it as a reserved seam. (Source: GPT-5.6 Sol)

extensions/mssql/src/diagnostics/diagnosticsCore.ts
kburtram: Fixed in 829d4d30d. Registering the identical sink object is now an idempotent no-op; registering a different sink with the same ID still disposes and replaces the old sink. The regression test exercises both cases and verifies delivery through the replacement. (Source: GPT-5.6 Sol)

extensions/mssql/src/diagnostics/sinks.ts
kburtram: Fixed in 829d4d30d. `PerfModeSink.dispose()` clears the pending timer and flushes queued markers. The existing sink test now disposes the sink and verifies the timer is cleared, and the real HTTP response test also disposes it in `finally`. (Source: GPT-5.6 Sol)

extensions/mssql/src/perf/perfTelemetry.ts
kburtram: Fixed in 829d4d30d. `PERF_REP_ID` now accepts only a decimal digit string whose numeric value is a non-negative safe integer; every other input falls back to 0, so `NaN` can never become `null` on the wire. Tests cover missing, negative, fractional, nonnumeric, unsafe, and valid values. (Source: GPT-5.6 Sol)

extensions/mssql/src/diagnostics/redaction.ts
kburtram: Fixed in 829d4d30d. String values in `ALWAYS_PLAIN` classifications now receive the same 4,096-character bound and honest original-length metadata, with tests for both SQL text and diagnostic metadata. Tokenizing `secret`, `connection.string`, and `token` even under off mode is intentional: it is a hard never-plaintext invariant for any active non-store sink, now documented immediately above the branch and covered across all capture modes. (Source: GPT-5.6 Sol)

extensions/mssql/src/diagnostics/diagnosticsCore.ts
kburtram: Fixed in 829d4d30d. Finite full-capture durations are clamped to 1 second through 1 hour, while non-finite values use the 15-minute default. Failed spans now carry `timingClass: "officialSameProcess"`, matching successful spans, and both behaviors have regression coverage. (Source: GPT-5.6 Sol)

extensions/mssql/src/controllers/connectionManager.ts
kburtram: Fixed in 829d4d30d by moving `mssql.connection.begin` after `prepareConnectionInfo` completes. Prompt cancellation or credential/Entra preparation failure therefore cannot leave an unmatched connection interval, while the marker still brackets the actual connection attempt. (Source: GPT-5.6 Sol)

extensions/mssql/src/perf/perfApi.ts
kburtram: Addressed in 829d4d30d by softening both the API documentation and PR description: this is a reserved, PERF_MODE-gated driver-health seam, not something #22822 currently consumes. I kept the small gated command but did not create a cross-PR dependency by wiring unrelated harness code from this product PR. (Source: GPT-5.6 Sol)

extensions/mssql/src/diagnostics/diagnosticsCore.ts
kburtram: Addressed in 829d4d30d. The allowlist, root-beginner list, forwarded-span families, and session bundle seam now identify which entries are reserved for later feature PRs; stale phase/design-document references were also removed. The limited current error-class signal is expected until those bounded producers land. (Source: GPT-5.6 Sol)

kburtram:
Review round completed in `829d4d30d` (with current `main` merged first as `92078433f`). I replied to and resolved all 18 Copilot/Fable threads.

Material fixes include:

- balanced submit/complete markers for all three query entry points and their request-failure paths;
- exact non-2xx/partial-rejection accounting in the harness sink, plus timer disposal;
- the hidden-webview rAF fallback and same-shape query re-run handling;
- redacted default emission policy, bounded capture elevation, failed-span timing classification, and bounded plain metadata strings;
- idempotent sink registration/replacement;
- safe `PERF_REP_ID` parsing and removal of the redundant `mssql.sts.pid` event;
- syntax-aware marker/feature/attribute conformance checks and the missing runtime attribute classifications;
- connection measurement moved after credential/Entra preparation;
- corrected present/future-scope comments and an accurate description of the reserved perf state API.

Two cross-PR boundaries were retained deliberately: source registry documentation gaps remain with #22822 while affected entries are explicitly `attrsComplete: false`, and STS auto-restart PID refresh remains later lifecycle work because the reserved `getState` API has no current caller.

Final validation:

- extension typecheck/build, webview build, lint, and full repository build passed;
- focused observability + Query Runner selection: 71 passed;
- broad MSSQL unit selection passed with only the two independently reproduced current-main Windows path suites excluded;
- current SQL Tools Service `main` (`e833cc23`) rebuilt and patched into Windows + Portable targets;
- perftest build/typecheck and all non-live-central-store suites passed; registry vendor sync passed;
- real `ext-normal-activation` run `2026-08-26T20-45-14Z_8eb626ef`: 1/1 passed against the final tree and patched STS;
- online MSSQL packaging passed, producing the expected 212.66 MiB local validation VSIX.

(Source: GPT-5.6 Sol)

extensions/mssql/src/diagnostics/sinks.ts
```diff
@@ -0,0 +1,597 @@
...
+        this.listener = listener;
+        return {
+            snapshot: [...this.ring],
+            lastSeq: this.ring.length > 0 ? this.ring[this.ring.length - 1].seq : 0,
+        };
+    }
+
+    public unsubscribe(): void {
+        this.listener = undefined;
+    }
```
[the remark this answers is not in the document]
kburtram: Fixed in bf2fcf211. Both unsubscribe and subscriber replacement now cancel the old delivery timer and clear its pending batch plus per-subscriber gap state, while retaining the ring snapshot and lifetime health counters. The regression test deliberately overflows the first subscriber, unsubscribes before delivery, verifies the next subscriber snapshot, and then proves it receives only the new event with no inherited gap. Full build, broad unit, package, and real activation validation were repeated before push. (Source: GPT-5.6 Sol)

kburtram:
Follow-up: Copilot opened one material LiveTail subscriber-boundary thread after the first remediation push. Fixed in `bf2fcf211`: unsubscribe/subscriber replacement now clears the old timer, pending batch, and per-subscriber gap state while preserving the ring snapshot and lifetime health. Added an overflow/unsubscribe/resubscribe regression test. Repeated typecheck, lint, full monorepo build, core tests (12 passed), broad MSSQL unit selection, online MSSQL packaging, and real activation run `2026-08-26T21-14-03Z_2381ee87` (1/1 passed). The thread was replied to and resolved; PR now has 19 total review threads and zero unresolved at this point. (Source: GPT-5.6 Sol)

extensions/mssql/src/controllers/connectionManager.ts
```diff
@@ -1464,6 +1465,10 @@ export default class ConnectionManager {
 
         credentials = await this.prepareConnectionInfo(credentials, connectionActivity);
 
+        // Measure the actual connection attempt. Credential/Entra preparation
+        // may prompt, throw, or be cancelled and must not leave an orphan begin.
+        Perf.marker("mssql.connection.begin", "begin");
```
kburtram, line 1470: Moving `begin` after `prepareConnectionInfo` closes the prompt/cancel case — thanks. One orphan path remains (Copilot flagged it in a suppressed "previously missed" note that never became a thread): `cancelConnection` rejects the completion deferred with `new Error("Connection cancelled")` (:2113), and `const result = await connectionCompletePromise.promise` (:1577) is outside any try/catch, so a user-cancelled connect throws out of `connect()` with a `begin` and neither `ready` nor `failed`. Wrap that await: on rejection emit `mssql.connection.failed { error: true, reason: "cancelled" }` and rethrow (`reason` stays a bounded enum). Low severity — no harness scenario cancels — but it is the last unpaired begin in this method.

(source: Fable)

extensions/mssql/src/webviews/common/perfMarks.ts
```diff
@@ -0,0 +1,115 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Webview-side perf marks. Disabled by default: marks
+ * are emitted only after the extension controller sends PerfEnableNotification,
+ * which it does only under PERF_MODE=1. Outside perf mode every call here is
+ * an inert boolean check.
```
kburtram: Nit (also in Copilot's suppressed notes): with the fallback restored this header is now inaccurate — outside PERF_MODE `perfMark` still captures both clocks and queues up to `MAX_PENDING` (50) marks, and `perfMarkAfterNextPaint` always schedules two rAF callbacks plus a 500 ms timer. All of that is deliberate and cheap; just say so here ("bounded queue and timers; no RPC until enablement") rather than "an inert boolean check".

(source: Fable)

kburtram (commented):
## Round-2 verification — `bf2fcf211`

Checked every one of the 19 resolved threads against the fix diff (`92078433f..bf2fcf211`, 14 files, +577/−84 excluding the `main` merge) and re-ran the gates on this head: `tsgo` typecheck (extension + webviews) clean, `eslint` on all changed files clean, focused VS Code run of `coreObservability`, `coreObservabilityRedaction`, `observabilityContract`, `queryRunner` → 72 passing (32 core/redaction/contract + 40 Query Runner), fresh `build:extension-bundle` (registry still not bundled; `LiveTailSink`/`SessionDiagSink` tree-shaken), `git merge-tree` against `origin/main` at `f827fd28e` clean.

**All 19 resolutions are correctly implemented.** Notes on the ones that were more than one-liners:

- Query probes: `markQuerySubmitted`/`markQuerySubmissionFailed` on all three request paths; the four new `queryRunner.test.ts` cases cover submit on each path and the failure `complete` on `runStatement` and `runQueryString`.
- `PerfModeSink.flush`: non-2xx counted once via the `settled` guard, `rejected` from the harness body used when it is a positive integer ≤ batch size, foreign bodies count the whole batch, timeout now destroys with an error so it lands in the same accounting; real-HTTP test asserts `dropped === 1` for `{accepted:1, rejected:1}`. `dispose()` clears the timer and flushes.
- rAF fallback restored with exactly-once semantics and `rafThrottled: true` (classified at runtime, conformance-checked).
- Conformance test replaced the regex with a TS syntax walk over `Perf.marker` / `perfMark*` / `diag.emit` / `diag.startSpan`; every emitted name is checked for registration, feature bucket, and attr classification; closed (`attrsComplete: true`) entries are attr-checked too.
- Core: default emission policy `redacted` (+ test), idempotent `addSink` + same-id replacement (+ test), duration clamp `[1 s, 1 h]` (+ test), `fail()` carries `officialSameProcess` (+ test), `mssql.sts.pid` removed.
- `LiveTailSink`: subscribe/unsubscribe/dispose all go through `clearSubscriberDelivery`; pending ⊆ ring so the new subscriber's snapshot loses nothing; lifetime counters retained (+ test).
- `connection.begin` moved after `prepareConnectionInfo`; `parsePerfRepId` (+ tests); ALWAYS_PLAIN truncation (+ test); comment/scope cleanups and the `getState` wording in code and PR description.

Two small inline items below (both surfaced by Copilot as "suppressed / previously missed" notes that never became threads); neither blocks merge.

**Now that #22822 is on `main`** (`f827fd28e`), the registry source is `tools/perftest/packages/observability-contracts/src/registry/event-types.json` in this repo, and the vendored snapshot here is byte-identical to it. The attr declarations deferred in round 1 (`mssql.sts.ready.pid` — load-bearing for `runPipeline.ts:718` — plus `resultSets`, `errorClass`, `rafThrottled`, `error`/`failed`/`errorNumber`) can be made either here (edit registry → `npm run generate` → copy the generated `.ts` over `src/sharedInterfaces/observabilityContract.generated.ts`; the new `attrsComplete` conformance test then enforces them) or in a small follow-up on `main`. Follow-up is fine; just do not let `attrsComplete: true` land on those entries before the attrs are declared.

LGTM from my side once the two nits are either fixed or explicitly deferred.

(source: Fable)

extensions/mssql/src/controllers/connectionManager.ts
kburtram: Fixed in 95871a91e. The completion await now goes through a small `awaitConnectionCompletion` helper that emits `mssql.connection.failed { error: true, reason: "cancelled" }` on rejection and rethrows; `connect()`'s control flow is otherwise unchanged. Two unit tests cover the helper (rejected → marker + same error rethrown; resolved → no marker). The conformance suite still passes — `reason` was already runtime-classified.

(source: Fable)

extensions/mssql/src/webviews/common/perfMarks.ts
kburtram: Fixed in 95871a91e: header now states the actual out-of-perf-mode cost (bounded pending queue, two rAF callbacks + one 500 ms timer, no RPC until enablement).

(source: Fable)

extensions/mssql/test/unit/connectionManager.test.ts
```diff
@@ -282,6 +283,43 @@ suite("ConnectionManager Tests", () => {
...
+            } catch (error) {
+                thrown = error;
+            }
+
+            expect(thrown).to.equal(failure);
+            expect(perfMarkerStub).to.have.been.calledOnceWith(
+                "mssql.connection.failed",
+                "instant",
+                sinon.match({ error: true, reason: "cancelled" }),
+            );
```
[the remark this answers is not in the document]
kburtram: Intentional, so leaving as is. This test targets the `awaitConnectionCompletion` helper directly, not the full `connect()` flow, and the helper has exactly one side effect: emitting the terminal marker on rejection. `calledOnceWith` is what asserts that the interval is closed exactly once (no duplicate terminal marker), which is the property the diagnostics core also guarantees for spans. If the helper ever grows a second marker, updating this assertion would be the right consequence, not a brittleness problem.

(source: Fable)

aasimkhan30 (approved):

PR #22826, opened 2026-08-26
Update STS to 6.0.20260827.1 to pickup STSv2 endpoints

kburtram:
Update STS to 6.0.20260827.1 to pickup STSv2 endpoints

Changes:
https://github.com/microsoft/sqltoolsservice/commit/e833cc23a52ba9c0a34c9e777218a5fbd343e9dc Add T-SQL brace matching service API (https://github.com/microsoft/sqltoolsservice/pull/2803)
https://github.com/microsoft/sqltoolsservice/commit/160a22699548927fb82a6c75b7245163140a1d25 STS2 integration — StreamJsonRpc host, opt-in bootstrap, and real-process E2E coverage (https://github.com/microsoft/sqltoolsservice/pull/2786)
https://github.com/microsoft/sqltoolsservice/commit/9490ba1d3c0b977c18177997d2df3f4f4db9322e STS2 drivers — SQL Client production adapter and SQLite portable adapter (https://github.com/microsoft/sqltoolsservice/pull/2785)
https://github.com/microsoft/sqltoolsservice/commit/5cc26712399306b8390dbd3998990c2d09f4034f STS2 multiplexer — safe shared-stdio routing and crash containment (https://github.com/microsoft/sqltoolsservice/pull/2784)
https://github.com/microsoft/sqltoolsservice/commit/7f8dc3fab3dd6d92f79a576f2731fd3fba834572 STS2 runtime — write-ahead coordination, replay, and executable scenarios (https://github.com/microsoft/sqltoolsservice/pull/2783)
https://github.com/microsoft/sqltoolsservice/commit/988051e2f21cdae7d1a48e44c756eb76191440e7 STS2 foundation — contracts, abstractions, and deterministic core (https://github.com/microsoft/sqltoolsservice/pull/2782)
https://github.com/microsoft/sqltoolsservice/commit/3aa0d631a561e232550d30e57f5eb2bc5b5eec90 Add handling for custom batch separator in language service (https://github.com/microsoft/sqltoolsservice/pull/2802)
https://github.com/microsoft/sqltoolsservice/commit/9b5e558dec051a5c305b2f8af7e91c849145bd0e Juno: check in to JUNO/hb_9c3e0ac3-59f8-43eb-b471-ca4843782f8a_20260819205807226. (https://github.com/microsoft/sqltoolsservice/pull/2800)
[...]

PR #22832, opened 2026-08-27
Add provider-neutral SQL data plane and STS v2 backend

kburtram:
# Summary

This change adds the provider-neutral SQL Data Plane foundation and its first production binding: SQL Tools Service v2 over the extension's existing stdio JSON-RPC transport.

The branch is based directly on current `main` (`f827fd28e`, which includes the merged perftest import) and is intentionally **not stacked on PR #22827**. It is the first independently mergeable data-access prerequisite for the planned metadata service, native language service, and AI-completion work.

The existing connection, query editor, language service, and Object Explorer paths are unchanged. The new lane is experimental, disabled by default, and has no consumer in this PR that replaces established product behavior.

> **Merge prerequisite:** [PR #22826](https://github.com/microsoft/vscode-mssql/pull/22826) must merge first so the bundled SQL Tools Service is at least `6.0.20260825.2`, the first release that accepts `--enable-sts2`. This PR also version-gates the launch flag, so enabling the setting cannot crash an older bundled service while the prerequisite is pending.

## Why this is one PR

The domain contract, capability registry, STS v2 wire profile, lifecycle engine, and conformance tests form one reviewable boundary. Splitting the wire adapter from the contract would leave either an unexercised interface PR or an adapter with no stable semantics. Keeping them together gives reviewers a complete, buildable vertical slice while excluding all feature UI and higher-level consumers.

### Provider-neutral SQL Data Plane

- Domain interfaces for connection/session/query lifecycle, streamed result sets, messages, plans, typed errors, and backend availability.
- A versioned capability model with static, negotiated, and session-refined answers.
- A multi-provider registry with lazy single-flight startup, passive status, per-session provider binding, configuration fingerprinting, draining/recomposition, and explicit disposal.
- Capability checks that run before credential providers, plus policy-driven fallback suggestions and per-profile fallback memory.
- A deterministic fake backend and shared conformance suite for backend semantics.
- Serverless Azure SQL open/retry coordination behind a narrow host adapter.
- A passive `MSSQL: Show SQL Data Plane Status` command.

### STS v2 binding

- Pinned `v2/*` request, response, and notification DTOs isolated from the domain API.
- An adapter over the existing `SqlToolsServiceClient` transport; no second process or transport is introduced.
- `v2/initialize`, connection open/close, query execute/cancel/dispose, result-set/row/message/completion routing, database-context changes, fatal transport handling, and capture-control capability negotiation.
- Per-query ordered delivery lanes and credit acknowledgements only after the consumer accepts a page.
- Protocol invariants for metadata-before-rows, gapless page sequence, monotonic row offsets, one terminal, and silence after terminal.
- Bounded initialize/open/cancel/close/dispose deadlines with deterministic synthesized terminals when the service does not settle.
- Explicit backend disposal that closes sessions, releases notification subscriptions, clears state, and is idempotent.

### Result cell contract

- Shared vector and spatial metadata/encoding helpers used by streamed result pages.
- Exact decimal, temporal, binary, JSON, vector, spatial, null-bitmap, and truncation decoding tests.

### Activation and gating

- New application-scoped `mssql.sqlDataPlane.enabled` preview setting, default `false`.
- SQL Tools Service receives `--enable-sts2` only when that setting is enabled **and** the configured bundled service version supports the flag.
- The setting explains that the window must be reloaded because service launch arguments are fixed at process start.
- Backend, capability-fallback, and lifecycle-timeout settings are explicitly declared, constrained, and localized; the test-only fake backend cannot be selected as a production default.
- Registry construction is cheap and passive; it does not start a backend, connect, prompt, or resolve credentials.

### Review hardening

- Startup single flights are generation-guarded: disposal or configuration recomposition retires an in-flight factory, disposes any late-created service, and prevents stale success/failure callbacks from mutating current state.
- STS2 scalar wrappers (`int64`, `decimal`, temporal, binary, GUID, non-finite double, and provider values) decode losslessly instead of rendering as `[object Object]`; safe JSON bigint values retain exactness.
- Unknown pre-registration notifications are bounded by query and event count. Recently retired query IDs are also bounded, and late notifications are diagnosed and dropped instead of becoming permanent orphans.
- Wire query failures now expose the stable domain error code while retaining STS/server details in the structured backend/server fields.
- Immediate cancel-after-execute handles rejection without waiting for an impossible query ID; deliberate session close synthesizes `canceled`, while backend disposal synthesizes `connectionLost`.
- Initialization identifies “not enabled” only from JSON-RPC `MethodNotFound`, and timeout text distinguishes disabled, busy, and wedged service states.
- The wire DTOs are pinned to the current camelCase STS2 contract, including the acknowledged result-set identity.
- Capability-fallback prompts, actions, and automatic-route notifications are localized through the repository constants/catalog while the policy layer remains VS Code-free and unit-testable.

## Privacy and failure behavior

- SQL text, result values, passwords, access tokens, connection strings, server names, and raw provider failures are excluded from the diagnostics observer contract.
- The STS v2 observer accepts only bounded scalar protocol metadata and defaults to a no-op in this independent PR.
- Authentication/open/startup status uses stable error codes and safe messages rather than raw provider error text.
- PR #22827 can later supply an adapter from this narrow observer boundary to the accepted diagnostics core without making this PR depend on it.

## Deliberately excluded

- PR #22827's observability core, activation/performance markers, and diagnostics storage.
- Debug Console UI, session history, durable capture repositories, and replay UI.
- Metadata service/cache, native language service, and AI completions.
- Query Studio, Object Explorer v2, TypeScript-native provider, Vector UI, and Spatial UI.
- Any migration of existing query/connection consumers to the new data plane.

## Branch and merge relationship

- Base: current `main` after #22822.
- Head: `dev/karlb/query_02_sts2_data_plane`.
- Merge after #22826 so the bundled service version satisfies the STS2 launch prerequisite.
- This PR contains none of PR #22827's product changes.
- PR #22827 and this PR are feature-independent, although both touch a few composition files. Whichever merges second should merge current `main` and retain both additive registrations.
- Metadata can start from `main` after this PR merges. Generic Debug Console work can proceed separately after #22827 merges. AI completions wait for the data plane, metadata, and language-service slices.

## Validation

- `npm install` at the repository root: passed through the configured Microsoft npm proxy; no lockfile change.
- `npm run build`: passed for every repository workspace.
- `npm run build -- --target mssql`: passed.
- `npm run lint -- --target mssql`: passed with zero errors.
- Review-focused STS v2 protocol/lifecycle/privacy tests: **53 passed**.
- Review-focused SQL Data Plane registry/race tests: **15 passed**.
- Service-launch/version-gating tests: **32 passed**.
- Hosted capability-fallback decision/open tests: **13 passed**.
- Focused serverless-open and result-cell codec tests: **54 passed**.
- Full VS Code-hosted MSSQL unit run: **3,631 passed, 12 pending**; the same eight unrelated Windows path-expectation failures reproduced on current `main` (`Utility Tests - Path handling` and `QueryCompletionSoundService`). The complete inverse selection passed with coverage disabled.
- Perftest nested project: build and strict typecheck passed; ordinary non-live suite **347 passed, 16 environment-gated tests skipped**. The optional central-store environment variables were cleared because this machine's configured localhost integration database was not running.
- Current SQL Tools Service `main` (`890bbef6`) restored/published and patched into this worktree: passed.
- STS v2 service unit suite: **497 passed, 10 skipped**.
- STS v2 real stdio end-to-end suite: **7 passed**.
- Online VSIX packaging with the patched service and `--skip-service-install`: passed; produced `mssql-1.46.0.vsix` (178.97 MiB). The reduction from the earlier local package reflects current STS main making the SQLite driver opt-in.
- Commit hooks, localization extraction, Prettier/ESLint checks, and `git diff --check`: passed.

The perftest `ext-normal-activation` scenario is not an applicable independent-PR gate yet: that scenario waits for `mssql.activate.begin/end`, which are introduced by still-open PR #22827 and therefore do not exist on this branch. During the attempted run, VS Code invoked the real extension's activation path and the extension launched the patched STS process, but the harness could not observe the absent completion marker. The scenario should be rerun on current `main` after #22827 lands, or on a disposable combined validation tree.

## Test plan

- [x] Full repository build
- [x] MSSQL build/typecheck/bundles
- [x] MSSQL lint
- [x] SQL Data Plane conformance, fallback, registry, and launch tests
- [x] STS v2 wire, lifecycle, privacy, deadline, and disposal tests
- [x] Serverless-open and result-cell codec tests
- [x] Broad MSSQL unit regression selection
- [x] Perftest build/typecheck/non-live tests
- [x] Current STS v2 unit and stdio E2E suites
- [x] Patched-service online VSIX packaging


extensions/mssql/src/languageservice/serviceclient.ts
```diff
@@ -51,6 +51,13 @@ type ServiceLaunchType =
     | "portableDownloaded"
     | "platformDownloaded";
 
+/** Add the STS v2 lane flag without changing the default service launch. */
+export function configureSqlDataPlaneLaunchArgs(args: string[], enabled: boolean): void {
+    if (enabled) {
+        args.push("--enable-sts2");
```
kburtram, line 81: **Should-fix (blocking as sequenced).** `config.ts:10` on this branch still pins SQL Tools Service `6.0.20260810.1`; `--enable-sts2` first exists in STS at `160a2269` (2026-08-24, #2786). The legacy option parser treats any unknown argument as an error — `CommandOptions.cs:112` appends "Unknown argument", the `finally` at :126 prints usage and sets `ShouldExit`, and `ServiceLayer/Program.cs:33` returns before the host starts. So on `main` + this PR, a user who enables `mssql.sqlDataPlane.enabled` and reloads gets **no** SQL Tools Service (connections, IntelliSense, OE all gone) until they find the setting again. The description's validation used a locally patched service, which hides this. Options: (a) declare the hard dependency on #22826 (6.0.20260825.2) in the body and merge after it; (b) additionally gate this push on `config.service.version` (`configurations/config.ts:10`) so the setting can never launch an incompatible binary — cheap and self-protecting; and say in the setting's description which service build the preview needs.

(source: Fable)

extensions/mssql/src/services/sqlDataPlane/api.ts
```diff
@@ -0,0 +1,742 @@
...
+                base64: raw.v,
+                ...(raw.bytes !== undefined ? { byteLength: raw.bytes } : {}),
+                truncated,
+            };
+        }
+        return hint === "xml"
+            ? { kind: "xml", value: raw.v, truncated }
+            : { kind: "string", value: raw.v, truncated };
+    }
+    switch (hint) {
```
kburtram, line 759: **Should-fix.** The real service does not send bare scalars for most types: `WireValueEncoder.cs:207-275` emits `{ $t: "int64" | "decimal" | "datetime2" | "datetimeoffset" | "time" | "guid" | "binary" | "double" | "provider", v: string }` wrappers (bigint beyond 2^53, every decimal/money, every temporal, guid, varbinary, NaN/Infinity, provider fallbacks), on both the legacy `rows` and the compact shape. This switch has no wrapper branch, so `decimal` → `{ kind: "number", value: "[object Object]" }`, `datetime2` → `display: "[object Object]"`, varbinary → `hexPrefix: "[object Object]"`, guid → `"[object Object]"`. Also `bigint` values within 2^53 arrive as JSON numbers but the `number:approx` hint (`sts2Backend.ts:211`) marks them `exact: false` although nothing was lost — the doc above says exactness should be false only when the token was actually lost. Nothing calls `decodeCell` in this PR (in `dev/refactor` the grid unwraps structurally in `queryStudioGridOps.ts`), so no user-visible impact today, but this PR ships the contract and the header says this is the only decode path. Suggest: unwrap before the hint switch (`int64`/`decimal` → `number` with the string token and `exact: true`; `double` → `number` inexact; temporals → `datetime`; `binary` → `binary { base64 }`; `guid`/`provider` → `string`), and add a `sts2Backend` test that pushes a page with these wire shapes through `onWireRows` and asserts `decodeCell`/`cellDisplay` output. The conformance test at `sqlDataPlaneConformance.test.ts:393` only feeds a plain string for `number:approx`.

(source: Fable)

extensions/mssql/src/services/sts2/sts2Backend.ts
```diff
@@ -0,0 +1,1428 @@
...
+            method: string,
+            deliver: (query: Sts2Query, params: T) => void,
+        ) =>
+            this.rpc.onNotification(method, (raw) => {
+                const params = raw as T;
+                const query = this.queries.get(params.queryId);
+                if (query) {
+                    deliver(query, params);
+                } else {
+                    const buffered = this.orphanBuffer.get(params.queryId) ?? [];
```
kburtram: **Recommended.** Unbounded, and only `registerQuery` drains it. Every synthesized terminal — cancel deadline (:1260), dispose drain (:1281), protocol violation (:1348), session close (:779), fatal (:759) — ends with `unregisterQuery` while the service may still emit `rows`/`message`/`complete` for that queryId (the service parks close until the query terminal, and a cancel can race the drain). Those late notifications arrive after the id is gone, get buffered under it here, and are never released for the life of the backend. Cap the buffer (per id and total) and keep a small retired-id set (recently unregistered ids) so late events for known-terminated queries are dropped with a `protocolViolation`-style diag warning instead of retained. Test: cancel with no scripted terminal, let `completeAfterCancelMs` synthesize, then `push(queryComplete, …)` and assert nothing is retained.

(source: Fable)

extensions/mssql/src/services/sts2/sts2Backend.ts
```diff
@@ -0,0 +1,1428 @@
...
+            totalRows: this.totalRows,
+            errorCount: this.errors,
+            durationMs: Date.now() - this.startMs,
+            ...(totalRowsAffected(params.rowsAffected) !== undefined
+                ? { rowsAffected: totalRowsAffected(params.rowsAffected) }
+                : {}),
+            ...(params.error
+                ? {
+                      error: {
+                          code: params.error.code ?? STS2_ERROR_CODES.queryFailedServer,
```
kburtram: **Recommended.** This puts the wire identity (`Sts2.QueryFailed.Server`) into the domain `error.code`, while `mapOpenError` (:618) and every synthesized terminal use `DataPlaneErrorCodes.*` with the wire code under `backend.code`. `DataPlaneErrorCodes.queryFailed` (`api.ts:601`, documented as "server-originated query failure; details ride server{}") is consequently never emitted anywhere in the PR. Consumers cannot switch on a stable code for the most common failure. Suggest `code: DataPlaneErrorCodes.queryFailed, backend: { kind: "sts2-jsonrpc", code: params.error.code }, server: …`.

(source: Fable)

extensions/mssql/src/services/sqlDataPlane/sqlDataPlaneService.ts
```diff
@@ -0,0 +1,896 @@
...
+import { FakeBackend, FAKE_CAPABILITIES } from "./fakeBackend";
+import { openWithServerlessWake } from "./serverlessWake";
+import { Sts2Backend, Sts2Rpc, DEFAULT_DEADLINES, Sts2Deadlines } from "../sts2/sts2Backend";
+
+// ---------------------------------------------------------------------------
+// Settings
+// ---------------------------------------------------------------------------
+
+const SETTING_ENABLED = "mssql.sqlDataPlane.enabled";
+const SETTING_BACKEND = "mssql.sqlDataPlane.backend";
```
kburtram, line 77: **Recommended.** `mssql.sqlDataPlane.backend`, `mssql.sqlDataPlane.capabilityFallback` (`providerSuggestions.ts:25`) and the five `timeouts.*` keys are read here but not declared in `package.json` (only `enabled` is; `dev/refactor` declared `backend` and `capabilityFallback`). Undeclared keys still resolve from `settings.json`, so `"mssql.sqlDataPlane.backend": "fake"` makes the transcript backend the default for every future consumer — `alternativesFor`/`kindsSatisfying` exclude the `test` realm, `defaultBackendKind()` does not. Either declare them (with `backend` enum-restricted to `sts2-local` until ts-native lands) or refuse `fake` as a default outside tests.

(source: Fable)

extensions/mssql/src/services/sts2/sts2Backend.ts
```diff
@@ -0,0 +1,1428 @@
...
+                    protocolVersion: result.specVersion,
+                },
+            };
+        } catch (error) {
+            const timedOut =
+                error instanceof SqlDataPlaneError &&
+                error.code === DataPlaneErrorCodes.clientTimeout;
+            const code = sts2ErrorCode(error);
+            const message = error instanceof Error ? error.message : String(error);
+            const notEnabled =
```
kburtram: Nit (classification). `code === undefined` makes every non-STS2 failure read "notEnabledOnService (launch STS with --enable-sts2)" — including `SqlToolsServiceClient.sendRequest`'s "Cannot send request before the language client is initialized" (`serviceclient.ts:749`) and transport drops. And on the real service the branch cannot fire: the legacy `MessageDispatcher` silently drops requests with no handler (`MessageDispatcher.cs:329-334`, the "TODO: Return message not supported"), so a service launched without the flag never answers `v2/initialize` and the user sees the 20 s timeout with "service busy or wedged". Suggest: not-enabled only on JSON-RPC `-32601` (`ResponseError.code`), and word the timeout reason to include "or STS v2 not enabled on this service".

(source: Fable)

extensions/mssql/src/services/sts2/sts2Backend.ts
```diff
@@ -0,0 +1,1428 @@
...
+            await withDeadline(
+                this.rpc.sendRequest(STS2_METHODS.connectionClose, {
+                    connectionId: this.connectionId,
+                }),
+                this.backend.deadlines.closeMs,
+            );
+        } catch {
+            // Bounded close: proceed to closed; the active query synthesizes.
+        }
+        this.activeQuery?.synthesizeTerminal("connectionLost", "session closed");
```
kburtram: Question. On a deliberate `close()` the active query is synthesized as `connectionLost` with `outcomeCertainty: "unknown"` / `outcomeReason: "transportLost"`. Per the service (`Sts2CoreReducer.cs` `DecideConnectionClose`), close with a running query issues the cancel and parks the close reply until the query's `complete(canceled)` has gone out — so on the real wire the terminal is `canceled` and this line is a no-op; only the scripted test at `sts2Backend.test.ts:914` (close answers `{}` immediately, no terminal pushed) observes `connectionLost`. Suggest scripting that test like the service (push `complete(canceled)` before resolving close) and using `canceled` for the close-path synthesis, since the transport was not lost.

(source: Fable)

kburtram (commented):
## Review — PR02 SQL Data Plane + STS v2 binding (`571ce1a45`)

Read every line of the 22 files (+8,656) and diffed each against its `dev/refactor` original: 18 of the 22 are byte-identical (CRLF aside); the split-specific edits are confined to `sqlDataPlaneService.ts` (ts-native factory and diag-core observer removed, `registerSqlDataPlane` now constructs with the extension version, `toErrorInfo` no longer copies error text), `sts2Backend.ts` (the `diag` import replaced by the injectable `Sts2DiagnosticObserver`, `dispose()` added, open/auth error text replaced by fixed messages, `onFatal` reason no longer forwarded) and the matching test edits. Cross-checked the pinned wire against `sqltoolsservice` main `e833cc23`: high-water ack is the cumulative per-query page ordinal (`Sts2CoreReducer.cs:584-593`) as the binding assumes; `v2/initialize` advertises exactly the capability keys the binding reads (`:114-130`); `SerializeTypeHints`/`SerializeColumns` match `typeHintFor`/the column helpers field-for-field; `encrypt`/`trustServerCertificate` accept what the binding sends; `connection.close` parks its reply until the active query's `complete(canceled)` has been emitted (`DecideConnectionClose`), which is the ordering the close path relies on. Verified locally at this head: `tsgo` typecheck pass, `eslint` on every changed file pass, the 8 affected suites in VS Code → **159 passing**, and `git merge-tree` against #22827's head `95871a91e` → no conflicts (both PRs' `extension.ts`/`package.json` edits are additive and disjoint).

The substrate itself is in good shape — lane ordering, invariant checks, deadline synthesis, credential-before-capability ordering and the privacy boundary all hold up under reading and the scripted suites. The items below are what I would close before merge, ordered by consequence.

1. **Enabling the preview against the service pinned on `main` takes SQL Tools Service down** (`serviceclient.ts:57`). `config.ts:10` still pins `6.0.20260810.1`; `--enable-sts2` first exists in STS at `160a2269` (2026-08-24, #2786). The legacy parser treats any unknown argument as an error, prints usage and sets `ShouldExit` (`Microsoft.SqlTools.Hosting/Utility/CommandOptions.cs:112-129`), and `Program.Main` returns immediately (`ServiceLayer/Program.cs:33`) — so a user who flips `mssql.sqlDataPlane.enabled` and reloads loses connections, IntelliSense and Object Explorer until they find the setting again. The description's validation ran against a locally patched service, which is why it did not show. Either (a) state the hard dependency on #22826 (the 6.0.20260825.2 bump) in the body and merge after it, or better (b) also gate the flag on the configured service version so the setting can never launch an incompatible binary, and say in the setting description which service version the preview needs.
2. **`decodeCell` does not decode what the real service sends** (`api.ts:700`). STS's `WireValueEncoder` emits `{ $t: "int64" | "decimal" | "datetime2" | "datetimeoffset" | "time" | "guid" | "binary" | "double" | "provider", v: string }` wrappers for every lossy/ambiguous scalar (`WireValueEncoder.cs:207-275`), on both the legacy and compact row shapes. `decodeCell` — documented as "the ONLY place compact → CellValue happens" — has no wrapper branch, so a `decimal` cell becomes `{ kind: "number", value: "[object Object]" }`, a `datetime2` becomes `display: "[object Object]"`, binary becomes `hexPrefix: "[object Object]"`. Relatedly, `bigint` values within 2^53 arrive as JSON numbers but the `number:approx` hint marks them `exact: false` although nothing was lost. No caller in this PR (nor in `dev/refactor`, where the grid handles wrappers structurally in `queryStudioGridOps.ts`), so nothing is user-visible yet — but this PR is the one that ships the contract, and the description's "exact decimal, temporal, binary … decoding tests" is not what `queryResultCellCodec.test.ts` covers (vector, spatial, truncation, base64). Add a wrapper branch before the hint switch and one `sts2Backend` test that pushes a page with the real wire shapes through `onWireRows` + `decodeCell`.

- `sqlDataPlaneService.ts:377` (Copilot's thread): agree on the late-`create()` resurrection, and it has two more entry points — `handleConfigurationChanged` treats an entry with `startup` in flight as running (`:638`) and, with zero sessions, recomposes it, after which the late `.then` installs a backend built with the OLD settings under the NEW fingerprint; and registry `dispose()` during startup leaves a live `Sts2Backend` with its notification handlers registered on the language client. Guard on `entry.startup === thisStartup` in both `.then` and `.catch`, dispose a late service, and fail `service()` typed when the entry was torn down while awaited. The `deferCreate` hook in `sqlDataPlaneRegistry.test.ts:264` is enough to test it.
- `sts2Backend.ts:357`: the orphan buffer is unbounded and only drained by `registerQuery`. Every synthesized terminal (cancel deadline, dispose drain, protocol violation, close, fatal) `unregisterQuery`s while the service may still send `rows`/`message`/`complete` for that queryId — those arrive after the id is gone, get buffered under it, and are never released. Cap it (per-id and total) and keep a small retired-id set so late events for known-terminated queries are dropped with a diag warning instead of retained.
- `sts2Backend.ts:1217`: `onWireComplete` puts the wire code (`Sts2.QueryFailed.Server`) into the domain `error.code`, whereas `mapOpenError` and every synthesized terminal use `DataPlaneErrorCodes.*` with the wire code in `backend.code`; `DataPlaneErrorCodes.queryFailed` is consequently never emitted anywhere. Consumers cannot switch on a stable code for the most common failure.
- `sqlDataPlaneService.ts:77`: `mssql.sqlDataPlane.backend`, `.capabilityFallback` and the five `.timeouts.*` keys are read but not declared in `package.json` (`dev/refactor` declared the first two). Undeclared keys still resolve from `settings.json`, so `"mssql.sqlDataPlane.backend": "fake"` routes every future consumer to the transcript backend (the `test` realm is only excluded from *alternatives*, not from the default). Declare them (with `backend` enum-restricted to `sts2-local` for now) or reject `fake` outside tests.
- `sts2Backend.ts:325`: the "notEnabledOnService" classification fires for every error without `data.code` — including "Cannot send request before the language client is initialized" and transport drops — and on the real service the branch cannot fire at all: the legacy dispatcher silently drops unknown request methods (`MessageDispatcher.cs:329-334`), so a disabled service produces the 20 s `initializeMs` timeout and the reason reads "service busy or wedged". Detect not-enabled by JSON-RPC `-32601` if it ever arrives, and word the timeout reason to include the not-enabled possibility.

### C — questions / notes

- `sts2Backend.ts:779`: on a deliberate `close()` the synthesized status is `connectionLost` with `outcomeCertainty: "unknown"`. The service cancels first and emits `complete(canceled)` before answering the close, so on the real wire the query ends `canceled`; only the scripted test at `sts2Backend.test.ts:914` (close replies `{}` immediately, no terminal) observes `connectionLost`. Script the test like the service, and consider `canceled` for the close-path synthesis.
- `sts2Backend.ts:1236`: `cancel()` before the execute response waits on `backendQueryId`, which never settles if execute is rejected — the caller sees the full `cancelAckMs` (10 s) before `uncertain`. Race `handle.accepted` as well.
- `wire/v2.ts:163`: the binding sends `resultSetId` on `v2/query.ack` (`sts2Backend.ts:1140`) but `V2QueryAckParams` does not declare it; the PascalCase `V2WireColumn` variants are dead (the service emits camelCase only, `DriverEffectRunner.cs:747-790`) — harmless, but the module claims to be the single source of wire truth.
- `sts2Backend.ts:1400`: the sink-error path resolves completion without `sink.onComplete` (by design per the conformance test) but also skips `finishTerminal`, so `cancelDeadlineTimer` is left armed (unref'd; fires into a `terminalSent` guard).
- Coverage gaps beyond A2: no test drives `Sts2Backend.dispose()` with an active query (the dispose test opens a session only); no test for late notifications after a synthesized terminal (B2); no registry test for teardown during startup (B1); `registerSqlDataPlane`/`statusSummary` have no test at all (the command is the only user-visible surface here).
- `package.json:1764` (Copilot): agree, localize the title; `dev/refactor` had the same hard-coded string.
- Toggling `mssql.sqlDataPlane.enabled` off at runtime disposes nothing (`enabled` is not in any fingerprint) — fine given the reload requirement, just noting it is the watcher's only unhandled key.
- Forward references a reader of `main` cannot resolve: "design doc 03/04", "web addendum §3.x", "TSQ2 addendum", "AD-1/AD-2", "D-00xx", "QO-n", "FOUND-n", "BOOT-4", `sharedInterfaces/queryStudioGridOps.ts`, `debugConsole`, OE v2. Same note as on #22827 — a one-line "reserved for <later PR>" would save the next reviewer a search.
- `serviceclient.ts:699` places `--enable-sts2` between `--parallel-message-processing` and its `-limit` value; the STS2 bootstrap strips it before the legacy parser on a compatible build, so ordering is fine there — the issue is only the incompatible pinned build (A1).

### Copilot's two comments

Both valid: the localization nit is minor; the single-flight race is real and wider than described — replied on each thread with the additional entry points and a test shape.

### Merge readiness

- Blocking: A1 (sequence after #22826 or gate the flag), A2 (decode contract vs. real wire, plus its test).
- Should land here: B1–B5.
- Follow-up acceptable: C items, coverage gaps.
- Unchanged from the split plan: no consumer in this PR exercises the plane end-to-end; the first consumer PR should include the real-service run the description defers (`ext-normal-activation` after #22827, or a data-plane scenario once the harness has one).

(source: Fable)

extensions/mssql/src/services/sts2/sts2Backend.ts
```diff
@@ -0,0 +1,1428 @@
...
+    // --- control -------------------------------------------------------------
+
+    private async cancel(): Promise<CancelAck> {
+        this.cancelRequested = true;
+        const deadlineAt = Date.now() + this.backend.deadlines.cancelAckMs;
+        try {
+            // A user can cancel immediately after execute() returns, before the execute
+            // response assigns the server query id. Never send the empty placeholder: wait
+            // for this query's own id within the existing cancel-ack deadline.
+            const queryId =
```
kburtram: Nit. If `cancel()` is called before the execute response and the execute is then **rejected**, `handle.backendQueryId` never settles (only `accepted` does, as `rejected`), so the caller waits the full `cancelAckMs` (10 s) before getting `uncertain: true`. Race `this.handle.accepted` alongside `backendQueryId` and return `{ acknowledged: true }` (nothing to cancel) on rejection.

(source: Fable)

extensions/mssql/src/services/sts2/wire/v2.ts
```diff
@@ -0,0 +1,345 @@
...
+    };
+    [key: string]: unknown;
+}
+
+export interface V2QueryExecuteResult {
+    queryId: string;
+}
+
+/** High-water ack (per-page `pageSeq` also accepted by the service). */
+export interface V2QueryAckParams {
```
kburtram, line 163: Nit (contract drift). The binding sends `{ queryId, resultSetId, throughPageSeq }` on `v2/query.ack` (`sts2Backend.ts:1140-1144`) but this type does not declare `resultSetId` (the object literal is typed `unknown` at the call site, so nothing catches it). Conversely the PascalCase `V2WireColumn` variants are dead: `DriverEffectRunner.SerializeColumns` (:747-790) emits camelCase only. Harmless, but the header says this module is the single source of wire truth.

(source: Fable)

extensions/mssql/src/services/sts2/sts2Backend.ts
```diff
@@ -0,0 +1,1428 @@
...
+        return true;
+    }
+
+    /** One sink callback in flight per query — the ordered lane. */
+    private enqueue(step: () => void | Promise<void> | undefined): void {
+        this.lane = this.lane
+            .then(async () => {
+                await step();
+            })
+            .catch((error) => {
```
kburtram, line 1515: Nit. This path resolves `completion` without `sink.onComplete` (by design, per the conformance test) but bypasses `finishTerminal`, so a `cancelDeadlineTimer` armed before the sink threw stays armed (unref'd; it will fire into the `terminalSent` guard). Route through a shared "terminal bookkeeping" step or clear the timer here too.

(source: Fable)

extensions/mssql/test/unit/sts2Backend.test.ts
```diff
@@ -0,0 +1,1280 @@
...
+    });
+
+    test("concurrent starts share one initialize (single flight)", async () => {
+        const rpc = standardRpc();
+        const backend = new Sts2Backend(rpc);
+        await Promise.all([backend.start(), backend.start(), backend.start()]);
+        expect(rpc.requests.filter((r) => r.method === STS2_METHODS.initialize)).to.have.length(1);
+    });
+
+    test("happy path: ordered sink events, structured rowsAffected, one terminal", async () => {
```
kburtram, line 188: Coverage gaps worth closing in this PR: (1) no test feeds the wire's typed wrappers (`{ $t: "decimal" | "int64" | "datetime2" | "guid" | "binary", v }`) through `onWireRows` → `decodeCell` — the only `$t` shapes exercised are `truncated`, `vector`, `spatial`; (2) the disposal test at :523 opens a session but no query — `dispose()` with an active query should prove the completion settles (`connectionLost`, synthesized) and that nothing is retained in `queries`/`orphanBuffer`; (3) no test for notifications arriving after a synthesized terminal (see the orphan-buffer comment); (4) the close test at :914 scripts the service differently from how it behaves (close answers before the query terminal).

(source: Fable)

extensions/mssql/test/unit/sqlDataPlaneRegistry.test.ts
```diff
@@ -0,0 +1,499 @@
...
+    });
+
+    test("duplicate factory registration is a programming error", () => {
+        const service = makeService(new TestConfig(), testFactory("fake", { createCalls: 0 }));
+        expect(() => service.registerFactory(testFactory("fake", { createCalls: 0 }))).to.throw(
+            SqlDataPlaneError,
+        );
+    });
+
+    test("single-flight startup; failed startup is retryable", async () => {
```
kburtram, line 280: The `deferCreate` hook here is enough to pin Copilot's race: start `service.service()`, call `service.dispose()` (variant: `config.set(knob, 2); service.handleConfigurationChanged()`) **before** `releaseCreate()`, then assert the pending `service()` rejects typed, `entrySnapshots()[0].state` is not `running`, and the fake backend the factory produced was disposed. Today the first variant leaves a live backend on a disposed registry and the second installs a backend built with the old knob under the new fingerprint.

(source: Fable)

extensions/mssql/package.json
```diff
@@ -1758,6 +1758,12 @@
                 "command": "mssql.openInMssqlExtensionFromAzureResources",
                 "title": "%mssql.openInMssqlExtensionFromAzureResources%",
                 "category": "MS SQL"
+            },
+            {
+                "command": "mssql.sqlDataPlane.showStatus",
+                "title": "Show SQL Data Plane Status",
+                "category": "MS SQL",
+                "enablement": "config.mssql.sqlDataPlane.enabled"
```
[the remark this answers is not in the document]
kburtram: Agree — minor. Repo convention is `"title": "%mssql.sqlDataPlane.showStatus%"` plus a `package.nls.json` entry; `dev/refactor` carried the same hard-coded string, so this is the right place to fix it. While touching this block: the setting's `markdownDescription` should also name the SQL Tools Service build the preview requires — on the service pinned by `config.ts` today (`6.0.20260810.1`) the flag this setting adds makes the service exit at startup (see the `serviceclient.ts` comment).

(source: Fable)

extensions/mssql/src/services/sqlDataPlane/sqlDataPlaneService.ts
```diff
@@ -0,0 +1,896 @@
...
+                    return service;
+                })
+                .catch((error: unknown) => {
+                    // Failed startup is retryable: clear the single flight.
+                    entry.startup = undefined;
+                    entry.state = "failed";
+                    entry.lastError = toErrorInfo(error, kind);
+                    throw error;
+                });
+        }
```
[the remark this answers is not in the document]
kburtram: Agree, and it is wider than the dispose case. Three entry points share the root cause (the `.then`/`.catch` never check that `entry.startup` is still *their* promise):

1. `handleConfigurationChanged` (:638) treats an entry with `startup` in flight as running, so with zero sessions it calls `recomposeEntry` → `disposeEntry` clears `startup` and refreshes the fingerprint — then the late `.then` installs a backend built with the **old** settings under the **new** fingerprint, and the change is silently lost.
2. Registry `dispose()` during startup: the late backend is installed on a disposed registry and, for `sts2-local`, its `v2/*` notification handlers stay registered on the language client after deactivation.
3. The `.catch` of a late *failed* create overwrites `state`/`lastError` of an entry that was already reset to idle/disposed.

Fix shape: `const startup = factory.create(...).then(service => { if (entry.startup !== startup) { void (service as { dispose?(): unknown }).dispose?.(); throw new SqlDataPlaneError(unavailable, "backend torn down during startup", true); } … })`, the same guard in `.catch`, and after `await entry.startup` in `service()` fail typed if `entry.service` is gone. The existing `deferCreate` hook in `sqlDataPlaneRegistry.test.ts:264` is enough to test both variants (dispose vs. config change before `releaseCreate()`).

(source: Fable)

extensions/mssql/package.json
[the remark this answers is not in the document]
kburtram: Fixed in e31f95bd6. Both command registrations now use %mssql.sqlDataPlane.showStatus%; the command and all new setting descriptions are backed by package.nls.json and the regenerated XLIFF catalog. The enabled-setting text also states the minimum compatible STS version. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sqlDataPlane/sqlDataPlaneService.ts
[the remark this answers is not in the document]
kburtram: Fixed in e31f95bd6. Startup completion is now generation-guarded: only the current startup may install or fail an entry, a late-created service is disposed, and the awaiting caller receives a typed unavailable error if the entry was retired. Added separate dispose-during-startup and configuration-change-during-startup regression tests. (Source: GPT-5.6 Sol)

extensions/mssql/src/languageservice/serviceclient.ts
kburtram: Fixed defensively in e31f95bd6. The extension now adds --enable-sts2 only when the preview is enabled and the configured dotted STS version is at least 6.0.20260825.2; older bundles launch normally instead of failing on an unknown option. Version-gate tests cover older, equal, and newer builds, and the PR now explicitly requires #22826 to merge first. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sqlDataPlane/api.ts
kburtram: Fixed in e31f95bd6. decodeCell now recognizes the STS2 typed scalar envelope before applying type hints and maps int64, decimal, double, temporal, GUID, binary, and provider values to the lossless domain forms. Safe native bigint numbers retain exactness. Both codec conformance tests and an end-to-end wire-row test cover the real wrapper shapes. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sts2/sts2Backend.ts
kburtram: Fixed in e31f95bd6. Pre-registration buffering is bounded to 64 query IDs and 32 events per ID, recently retired IDs are bounded to 256, and late notifications for retired queries are diagnosed and dropped. Passive status reports the bounded counts, with regression coverage for both overflow and post-terminal notifications. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sts2/sts2Backend.ts
kburtram: Fixed in e31f95bd6. Wire query failures now expose the stable SqlDataPlane.QueryFailed domain code, retain the STS code in backend.code, and keep server details in server. A regression test asserts all three layers. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sqlDataPlane/sqlDataPlaneService.ts
kburtram: Fixed in e31f95bd6. backend, capabilityFallback, and all five timeout settings are now declared, constrained, and localized in package.json. The preview backend enum is limited to sts2-local, and production construction also rejects a test-realm default unless factories were explicitly injected by a test. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sts2/sts2Backend.ts
kburtram: Fixed in e31f95bd6. Initialization is classified as not enabled only for JSON-RPC MethodNotFound (-32601); transport and pre-initialization failures remain ordinary unavailability. Timeout text now explicitly includes an STS2-disabled service alongside busy or wedged cases, with a generic-error regression test. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sts2/sts2Backend.ts
kburtram: Fixed in e31f95bd6. Deliberate session close now synthesizes canceled if a nonconforming backend answers without the terminal, while backend disposal still uses connectionLost because the service lane is being torn down. Tests cover the real service ordering and the fallback ordering independently. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sts2/sts2Backend.ts
kburtram: Fixed in e31f95bd6. Cancel and dispose now race query-ID availability against the execute acceptance result; rejection or abort returns acknowledged immediately because no backend query exists to cancel. A regression test verifies that the cancel request is not sent and no deadline wait occurs. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sts2/wire/v2.ts
kburtram: Fixed in e31f95bd6. V2QueryAckParams now declares resultSetId, and V2WireColumn is pinned to the camelCase contract emitted by current STS2 with the dead PascalCase variants removed. Wire-column tests now exercise only the current shape. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sts2/sts2Backend.ts
kburtram: Fixed in e31f95bd6. The sink-failure terminal path now clears and undefines the cancel deadline timer before completing locally, matching the shared terminal bookkeeping. A fake-timer regression test proves no delayed callback remains. (Source: GPT-5.6 Sol)

extensions/mssql/test/unit/sts2Backend.test.ts
kburtram: Fixed in e31f95bd6. Coverage now includes typed scalar wrappers through onWireRows and decodeCell, backend disposal with an active query, notification arrival after a synthesized terminal, bounded orphan state, and service-real close ordering plus the nonconforming fallback. (Source: GPT-5.6 Sol)

extensions/mssql/test/unit/sqlDataPlaneRegistry.test.ts
kburtram: Fixed in e31f95bd6. Added both requested deferred-factory races. Registry disposal and configuration recomposition now make the pending service call reject with a typed error, prevent the entry from returning to running, and dispose the late-created backend. (Source: GPT-5.6 Sol)

kburtram:
Review round complete in e31f95bd6. All 14 Copilot/Fable threads were assessed as valid, fixed with targeted regression coverage, replied to individually, and resolved. Revalidation passed MSSQL typecheck/build/lint; focused suites 53 + 15 + 32; the complete hosted inverse regression selection; STS2 unit 497 passed/10 skipped and E2E 7 passed against STS main 890bbef6; and patched-service online VSIX packaging. PR #22826 remains the merge prerequisite for the bundled STS version, with a defensive launch-version gate included here. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sqlDataPlane/sqlDataPlaneService.ts
```diff
@@ -0,0 +1,946 @@
...
+}
+
+function toErrorInfo(error: unknown, kind: SqlBackendKind): SqlDataPlaneErrorInfo {
+    if (error instanceof SqlDataPlaneError) {
+        return {
+            code: error.code,
+            message: error.code,
+            retryable: error.retryable,
+            backend: { kind },
+        };
```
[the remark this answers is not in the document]
kburtram: Not changed. This projection feeds passive status, whose privacy contract excludes raw provider failures; SqlDataPlaneError.message can originate at a backend boundary and may contain server, account, or connection details. The stable code is deliberately used as the safe message token, while an unknown exception is replaced with the generic Provider startup failed text. Returning error.message here would undo the redaction boundary documented and tested by this PR. (Source: GPT-5.6 Sol)

extensions/mssql/src/services/sqlDataPlane/providerSuggestions.ts
```diff
@@ -0,0 +1,93 @@
...
+    const alternativeName = displayNameFor(alternative);
+    const missing = describeMissingCapabilities(check);
+    if (policy === "auto") {
+        // Visible, never silent (TSQ2 §8.2): one notification names the
+        // ACTUAL provider; diagnostics carry the same fact via backendKind.
+        interaction.notify(
+            `This connection requires ${missing}, which ${currentName} does not support. ` +
+                `Connected with ${alternativeName} instead.`,
+        );
+        return { kind: "useAlternative", alternative, automatic: true };
```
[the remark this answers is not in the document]
kburtram: Fixed in 93dea3d75. The vscode-free policy layer now passes semantic capability/provider values to the host interaction, and the VS Code adapter renders the prompt, action, and automatic-fallback notification through the dedicated LocConstants.SqlDataPlane functions. The source bundle and XLIFF catalog were regenerated; typecheck, lint, full MSSQL build, and both hosted fallback suites (13 tests) pass. (Source: GPT-5.6 Sol)

extensions/mssql/package.json
```diff
@@ -1759,6 +1759,12 @@
                 "command": "mssql.openInMssqlExtensionFromAzureResources",
                 "title": "%mssql.openInMssqlExtensionFromAzureResources%",
                 "category": "MS SQL"
+            },
+            {
+                "command": "mssql.sqlDataPlane.showStatus",
+                "title": "%mssql.sqlDataPlane.showStatus%",
+                "category": "MS SQL",
+                "enablement": "config.mssql.sqlDataPlane.enabled"
             }
```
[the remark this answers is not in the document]
kburtram: Not changed — keeping the gate. This is a preview feature that is off by default, and the repo's pattern is to gate feature commands on their own setting (`when: config.mssql.enableQueryHistoryFeature`, `config.mssql.copilot.enableSqlToolsMcp` in this same file). The diagnosis case you describe ("why isn't enabling working") happens *after* the user has turned the setting on, and `enablement` is evaluated live — the command lights up as soon as the setting flips, no reload needed. With the setting off, the only thing the dump could report is `enabled: false`, which the setting itself already shows; the trade is a permanent "SQL Data Plane" entry in every user's command palette for a lane nothing consumes yet.

(source: Fable)

extensions/mssql/src/services/sqlDataPlane/sqlDataPlaneService.ts
```diff
@@ -0,0 +1,946 @@
...
+    }
+
+    // -----------------------------------------------------------------------
+    // Status (passive: reads state only)
+    // -----------------------------------------------------------------------
+
+    availability(): DataPlaneAvailability {
+        const entry = this.entries.get(this.tryDefaultKind() ?? "sts2-local");
+        return entry?.service?.availability ?? { state: "unknown" };
+    }
```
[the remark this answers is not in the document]
kburtram: Agreed and fixed in 1b23ae134. `availability()` now resolves the configured kind first and returns `{ state: "unknown" }` when it is invalid, instead of borrowing the `sts2-local` entry's state. Regression test: a running `sts2-local` entry, the setting flipped to an unknown value, asserts `unknown` plus `normalizedBackend: INVALID(...)` in the summary.

(source: Fable)

extensions/mssql/src/sharedInterfaces/queryResultCellCodec.ts
```diff
@@ -0,0 +1,588 @@
...
+
+function bytesToSqlHex(bytes: Uint8Array): string {
+    const digits = "0123456789ABCDEF";
+    const output = new Array<string>(bytes.byteLength * 2 + 1);
+    output[0] = "0x";
+    for (let i = 0; i < bytes.byteLength; i++) {
+        output[i * 2 + 1] = digits[bytes[i] >> 4] + digits[bytes[i] & 0x0f];
+    }
+    return output.join("");
+}
```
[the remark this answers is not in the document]
kburtram: Agreed and fixed in 1b23ae134: the array is now `byteLength + 1` with contiguous writes (one two-digit element per byte). Output is unchanged — the spatial hex/`STGeomFromWKB` assertions in `queryResultCellCodec.test.ts` cover it.

(source: Fable)

extensions/mssql/src/services/sqlDataPlane/sqlDataPlaneService.ts
```diff
@@ -0,0 +1,949 @@
...
+            },
+            get backendInfo() {
+                return service.backendInfo;
+            },
+            canOpen: (params) => service.canOpen(params),
+            openSession: async (params) => {
+                const session = await service.openSession(params);
+                registry.registerSession(entry, session);
+                return session;
+            },
```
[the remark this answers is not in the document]
kburtram: Not changed in this PR — valid design observation, tracked for the first consumer PR.

The two paths are intentional today: `openSession` / `openSessionWithFallback` are the documented consumer entry points (capability tripwire before any credential provider, serverless wake, fallback memory — pinned by the `credential tripwire` and `fallback open` tests), while `service()` / `serviceForProfile()` return the raw provider view for lifecycle accounting and identity stamping. In the branch this was extracted from, every direct `view.openSession` caller opens *follow-on* sessions (metadata hydration, Query Studio secondary sessions) for a profile whose primary open already went through `openSessionWithFallback` — `serviceForProfile` exists precisely to route those to the provider the guarded path chose. Nothing in this PR calls either path.

Agree the view could cheaply apply the same pure requirement check (`evaluateRequirements(effectiveSet, requirementsFor(params))`) so the guarantee holds regardless of entry point; that belongs with the first consumer (the metadata PR), where the semantics can be tested against real callers rather than changed speculatively here.

(source: Fable)

aasimkhan30 (approved):

PR #22835, opened 2026-08-27
Add local diagnostics sessions and Debug Console

kburtram:
## Summary

Adds an opt-in, local-only diagnostics experience for investigating MSSQL extension and SQL Tools Service behavior in one place. This PR introduces durable diagnostic sessions, an authenticated STS diagnostics channel, and the generic Debug Console/Perf Test History UI.

This is PR03 in the `dev/query` extraction plan. It targets `main` directly (it is not a stacked PR) and builds on work already merged through:

- #22822: in-repository perftest tooling
- #22827: core observability, redaction, trace identity, and sink foundations
- #22832: SQL Data Plane and STS v2 foundation
- #22826: released STS `6.0.20260827.1`

Both `mssql.debugConsole.enabled` and `mssql.sessionDiag.enabled` default to `false`, so the new viewer, status item, listener, and durable capture are inert for ordinary users. Enabling the Debug Console requires a VS Code restart so STS inherits its diagnostics endpoint consistently.

### Local diagnostic sessions

- Redacted or digest capture from extension startup through shutdown.
- Explicit, reason-bearing, time-bounded elevation to full capture; it automatically returns to redacted mode.
- Bounded JSONL segment journals with manifests, exact dropped ranges, session count/age/size retention, store integrity checks, local export, and clear/open-folder commands.
- Live-tail buffering plus durable backfill for recoverable subscriber gaps.
- Session history and cross-session action trends.
- Optional low-cost rich counters for CPU, memory, event-loop delay, and span activity.

### Generic MSSQL Debug Console

- Current and saved session sources plus imported perftest runs.
- Overview KPIs and anomaly/action summaries.
- Consolidated event trace with filtering.
- Cause tree and trace-quality/correlation diagnostics.
- Cross-process waterfall.
- SQL activity view.
- Diagnostics health and store-integrity view.
- Perf summary/import and full Perf Test History browsing, comparison, waterfalls, SQL activity, artifacts, dumps, and rich diagnostics.

The webview remains a renderer: filesystem access, source indexing, queries, aggregation, capture changes, and destructive confirmations stay in the extension host.

### SQL Tools Service diagnostics

- Starts before STS so the released service inherits `STS_DIAG_URL` and a random bearer token.
- Binds only to `127.0.0.1` and accepts authenticated POST batches.
- Enforces a 4 MiB request limit by actual UTF-8 byte count.
- Tolerates malformed individual NDJSON lines without affecting product behavior.
- Normalizes STS input to classified protocol metadata before it reaches diagnostics sinks.

### Contract and lifecycle integration

- Adds registry entries for Debug Console open and session export/elevation events, regenerates the shared contract artifacts, and verifies vendor parity with `tools/perftest`.
- Enables webview marks when either `PERF_MODE` or an in-product diagnostics sink wants them; retries are bounded and exist only while capture is active.
- Keeps persisted session metadata synchronized with capture-policy changes and timed elevation expiry.
- Cleans up sinks, timers, listeners, rich-collection gates, and the STS listener during disposal/deactivation.

## Privacy and safety boundaries

- Nothing is uploaded by this PR.
- Secrets, tokens, and connection strings are never persisted, including during elevated capture.
- Default capture is off; full capture cannot be selected as a persistent setting.
- Session identifiers and manifest segment paths are validated and contained under the configured store.
- Symlinked session directories/segments and malformed manifests/events are ignored.
- Cached active-session data is explicitly invalidated after a flush before gap backfill.
- Perf History run deletion is host-confirmed and path-contained.

## Deliberately out of scope

- AI-completion-specific Debug Console pages or adapters
- Generic replay and session bundles
- Central upload
- SQL Data Plane-specific status pages
- In-process self-test execution
- Metadata service, native language service, Query Studio, OE v2, Vector, or Spatial feature code

The next critical-path extraction toward AI completions is the metadata service/cache/freshness layer, followed by native language-service integration. Feature-specific AI diagnostics can attach to this generic console after those pieces land.

## Validation

- `npm install` from the repository root: passed; no dependency or lockfile change.
- `npm run build -- --target mssql`: passed, including extension and webview typecheck/bundles.
- `npm run lint` in `extensions/mssql`: passed.
- Focused diagnostics/Perf History/contract tests: **70 passed, 0 failed**.
- Full MSSQL hosted unit suite: **3,853 passed, 12 pending, 8 failed**. The eight failures are the existing Windows-only fixture baseline from untouched code: one tilde/path-separator assertion and seven Query Completion Sound path/process assertions. No PR-owned test failed.
- Observability-contract TypeScript build: passed.
- Observability-contract Vitest: **4 files / 31 tests passed**, including generated-output and vendored-snapshot synchronization.
- Online VSIX packaging with released STS `6.0.20260827.1`: passed; **101.05 MiB, 1,420 files**.
- Pre-commit localization extraction and lint-staged checks: passed.

## Suggested review order

1. `diagnosticsManager.ts`, `sessionStore.ts`, and `stsDiagListener.ts` for lifecycle/privacy boundaries.
2. `debugConsoleWebviewController.ts` and the shared interfaces for the host/webview API.
3. `analysis.ts`, `traceFilter.ts`, and Perf History providers for query/aggregation behavior.
4. `DebugConsole/` webview pages for presentation.
5. Unit/privacy/contract tests and generated registry updates.

> **Status / prerequisites (review round 1 follow-up)**
>
> - **SQL Tools Service lane:** the service-side emitter that consumes `STS_DIAG_URL`/`STS_DIAG_TOKEN` (STS commit `43a79552`, "Add opt-in STS diagnostics spans") is on STS `dev/query`/`dev/runbook` only — it is not on STS `main` and not in the released `6.0.20260827.1`. Until it ships, the listener and env injection here are inert (loopback-only, random token, unref'd) and the SQL Tools Service lane of the console stays empty. An STS PR for it has not been opened yet.
> - **Webview localization:** host-side strings are localized; the Debug Console webview's user-facing strings are still English literals and are deferred to the refactoring pass planned before merge (policy decision for an internal diagnostics surface, applied together with the page split).


aasimkhan30 (dismissed):

extensions/mssql/src/diagnostics/sessionStore.ts
```diff
@@ -0,0 +1,484 @@
...
+
+    // --- retention -----------------------------------------------------------------
+
+    public enforceRetention(maxSessions: number, maxAgeDays: number, maxTotalBytes?: number): void {
+        const sessions = this.listLocalSessions();
+        const cutoff = Date.now() - maxAgeDays * 86_400_000;
+        const doomed = new Set(
+            sessions.filter(
+                (s, index) =>
+                    s.manifest.status !== "active" &&
```
kburtram: **Should-fix.** Retention exempts every manifest whose `status` is `"active"` (here and at :338), but only `SessionDiagSink.close()` ever writes `"closed"` (`sinks.ts:594`). A crashed extension host, an OS shutdown, or VS Code killing a slow `deactivate` leaves the manifest `"active"` forever, and such sessions are then exempt from the count, age **and** size rules for good — the store's `maxTotalMB` budget does not hold, and the PR's "session count/age/size retention" claim is only true for clean shutdowns. `enforceRetention` also has no way to tell the *current* session apart from orphans, which is why it has to exempt all of them. Fix: take the current session id (as `clearAll(exceptSessionId)` already does) and treat any other `"active"` manifest as evictable (surface it as "interrupted" in `listSources`); optionally have the sink rewrite stale actives to `"partial"` on startup. Test: two `active` manifests, one current, `maxSessions = 1` → the orphan is removed, the current one is kept.

(source: Fable)

extensions/mssql/src/extension.ts
```diff
@@ -59,6 +62,15 @@ let activation: MssqlActivation | undefined;
 
 export async function activate(context: vscode.ExtensionContext): Promise<MssqlExtensionApi> {
     initializeExtensionToolkit();
+
+    // Install diagnostics before the first activation marker so an enabled
+    // session store captures the complete activation lifecycle.
+    const diagnosticsManager = new DiagnosticsManager(context);
+    context.subscriptions.push(diagnosticsManager);
```
kburtram: **Should-fix.** `mssql.openDebugConsole` is registered only when `mssql.debugConsole.enabled` was already on at activation, but `package.json` gives the command `enablement: config.mssql.debugConsole.enabled`, and `DiagnosticsManager.updateStatusItem` reads the setting live (`diagnosticsManager.ts:244-259`, bound to the same command). So the documented flow — flip the setting, restart later — immediately produces a palette entry and a status-bar item that both fail with "command 'mssql.openDebugConsole' not found" until the restart. Register the command unconditionally (`registerDebugConsole` needs nothing that only exists at activation — `diagnosticsManager` is always constructed) and either construct the console lazily or show a "restart VS Code to finish enabling the Debug Console" message when the STS listener could not be started in this session.

(source: Fable)

extensions/mssql/src/extension.ts
```diff
@@ -139,6 +151,18 @@ class MssqlActivation {
         // Exposed for testing purposes
         vscode.commands.registerCommand("mssql.getControllerForTests", () => controller);
         registerSqlDataPlane(context);
+
+        // Start before SQL Tools Service spawns so the child inherits the
+        // diagnostics endpoint and token. The listener only accepts bounded,
+        // authenticated local batches and emits classified metadata.
```
kburtram, line 184: **Recommended (description accuracy / sequencing).** The service-side emitter that consumes `STS_DIAG_URL`/`STS_DIAG_TOKEN` is `sqltoolsservice` commit `43a79552` ("Add opt-in STS diagnostics spans"), which exists only on STS `dev/query`/`dev/runbook` — it is not on STS `main` and not in the released `6.0.20260827.1` that #22826 pinned. So today this listener has no producer; the body's "the released service inherits STS_DIAG_URL and a random bearer token" describes the env injection, not any diagnostics that arrive. Please state the dependency on the STS PR (and link it) the way #22832 did for #22826, and note that the SQL Tools Service lane of the console stays empty until it ships. Harmless otherwise: loopback-only, random token, unref'd.

(source: Fable)

extensions/mssql/src/diagnostics/stsDiagListener.ts
```diff
@@ -0,0 +1,149 @@
...
+    for (const [key, value] of Object.entries(event.fields ?? {})) {
+        if (
+            typeof value === "string" ||
+            typeof value === "number" ||
+            typeof value === "boolean" ||
+            value === null
+        ) {
+            // STS emitter contract: protocol metadata only (methods, counts,
+            // durations, type names) — classified accordingly.
+            fields[key] = { raw: value, cls: "diagnostic.metadata" };
```
kburtram, line 171: **Recommended (defense in depth).** Everything STS sends is trusted as-is: `feature` and `type` are arbitrary strings (`:135-137`), and every scalar field is classified `diagnostic.metadata`, which the redaction layer passes through **plain** in every mode and `PerfModeSink` then forwards to the harness wire. The privacy guarantee therefore rests entirely on the emitter contract in a different repo/release train. Cheap hardening here: cap field count (e.g. 32) and string length (e.g. 256 chars) with truncation accounting, restrict `feature` to a known set (`rpc`, `sqlDriver`, `objectExplorer`, else `"sts"`), and bound `type` length. The unit test (`stsDiagListener.test.ts:53`) already exercises the nested-object drop; add the oversize-string case alongside it.

(source: Fable)

extensions/mssql/src/diagnostics/sessionStore.ts
```diff
@@ -0,0 +1,484 @@
...
+                if (!fs.existsSync(file)) {
+                    if (segment.events > 0) {
+                        issues.push(
+                            `${label}: segment ${segment.file} missing (${segment.events} events)`,
+                        );
+                    }
+                    continue;
+                }
+                try {
+                    const content = fs.readFileSync(file, "utf8");
```
kburtram: **Recommended.** `validateStore()` reads every segment of every stored session fully and synchronously (`readFileSync` + `split("\n")`) on the extension-host thread, and the controller calls it on every `dc/getHealth` (`debugConsoleWebviewController.ts:382-388`). With the default 512 MB budget, opening the Health page stalls the host for seconds and churns hundreds of MB of strings. The manifest already carries `sizeBytes` and per-segment counts: verify only the trailing segment fully (that is where an interrupted write can be), check the others by `statSync().size` against the manifest, and read the boundary lines with a bounded read instead of the whole file.

(source: Fable)

kburtram (commented):
## Review — PR03 local diagnostics sessions + Debug Console (`fa79102ed`)

Read all 45 files (+15,123) and diffed every new file against its `dev/refactor` original: 18 of 29 are byte-identical (CRLF aside); the rewritten ones (`debugConsoleWebviewController.ts`, `sharedInterfaces/debugConsole.ts`, `sessionStore.ts`, `diagnosticsManager.ts`, `stsDiagListener.ts`, `state.tsx`, `pagesPerf.tsx`, `pagesMore.tsx`) lose the completions/self-test/central-upload/replay/data-plane surfaces as the split plan says, and the lines they *add* (manifest/event validators, session-id pattern, bounded request body, generation-safe disposal, capture-policy manifest sync) are correct. Cross-checked the host against `sqltoolsservice` (the STS emitter is **not** on STS `main`, see A3) and the Perf History providers against the harness on this branch (`tools/perftest`) file-by-file, plus two real run directories on this machine. Verified at this head: `tsgo` typecheck for the extension and the webviews, `eslint` on every changed file, the 8 affected suites in VS Code → **70 passing**, vendored `observabilityContract.generated.ts` byte-identical to `tools/perftest`'s generated output, and `git merge-tree` against `origin/main` `69f54be2a` (#22826/#22827/#22832 all merged) → clean.

The lifecycle/privacy core is in good shape: capture is off by default, full capture is unreachable from settings, secrets never reach the journal (canary suite), session ids and segment paths are contained, symlinked dirs/segments are skipped, and the console cleans up sinks/timers/listeners on dispose. The findings below are where it does not yet meet its own stated guarantees, then robustness/performance items, then notes. Since you plan a refactoring pass before merge, I put the structural ones under "Refactoring candidates" at the end.

1. **Real SQL text is imported as plaintext and can be exported as "redacted"** (`perfRunImport.ts:251`). `batch_text ?? statement_text` from a run's `artifacts/sql/sql-activity.jsonl` is emitted as `cls: "sql.text", handling: "plain"` with `redactedFields: 0` on the strength of a comment ("synthetic harness DB: plain by contract"). Nothing enforces that: the harness persists SQL text whenever `captureSqlText && passType === "diagnostic"` (`perftest-cli/src/collectors/sqlServerXEvents.ts:160-166`) and a run can target any server (`environment.sql.provider: "external"` in the real runs here). Any directory the user opens or imports is trusted the same way; the text reaches the webview via `sqlActivity`/`waterfall`, and `dc/export` writes the source verbatim to a file titled "redacted JSONL". This contradicts "SQL text … never enters diagnostics". Apply the normal `sql.text` policy at import (digest/omit) unless the rep proves synthetic (`sql-activity-rollup.json.sqlTextPersisted` + provider ≠ external), or keep only `object_name`/counts.
2. **Retention exempts crash-orphaned sessions forever** (`sessionStore.ts:326`). Only `SessionDiagSink.close()` writes `"closed"`; a crash/kill leaves `"active"`, and every `"active"` manifest is exempt from the count, age and size rules — so `maxTotalMB` is not a bound, and `enforceRetention` cannot tell the current session from orphans. Pass the current session id (as `clearAll` does) and treat other actives as evictable / "interrupted".
3. **`mssql.openDebugConsole` is invoked before it exists** (`extension.ts:69`). It is registered only when the setting was on at activation, but the command has `enablement: config.mssql.debugConsole.enabled` and the status-bar item reads the setting live — flipping the setting produces a palette entry and status item that fail with "command not found" until the restart the description asks for. Register unconditionally and show a restart message (or construct lazily).

- **STS lane has no producer yet** (`extension.ts:157`): the emitter is STS `43a79552`, only on STS `dev/query`/`dev/runbook`, not in `6.0.20260827.1`. State the dependency and link the STS PR; the body's "released service inherits STS_DIAG_URL" reads as if diagnostics arrive today. When it lands, bound what the listener trusts (`stsDiagListener.ts:131`): field count/length caps and a known `feature` set — every STS scalar is classified `diagnostic.metadata`, i.e. plain in the journal and forwarded to the harness wire.
- **Harness contract drift in Perf History** (verified against `tools/perftest` sources and real runs): `result.warmup` is never written by the harness (`PerfResult` has no such field; the CLI derives warmup from `run-config.snapshot.jsonc.warmupRepetitions`, `runLoader.ts:82-89`), so warmup reps count as valid samples and p50/p95/n differ from `perftest history` (`directoryProvider.ts:528`); `failureReason` is never populated — the harness records `errors[]` (`directoryProvider.ts:505`); `summary.scenarios[].skipped/reason` and `summary.connection.connectionMode` have no writer (C).
- **Run-wide wallclock KPI pools different scenario sets** (`perfHistoryService.ts:313`): a single-scenario activation run followed by a query-only run reports "-98%" with no scenario-mix caveat. Compute deltas per scenario over the intersection.
- **Unbounded synchronous reads on the extension host**: `validateStore()` reads every segment of every session on each Health open (`sessionStore.ts:402`); History loads up to 15 stored sessions' full event arrays to compute medians (`controller:553`); `markers.jsonl`/`sql-activity.jsonl` are `readFileSync`+`split` with only a per-*line* cap — soak reps put every iteration's markers in one file, and >512 MB throws `ERR_STRING_TOO_LONG` out of the RPC (`perfRunImport.ts:73`, `perfHistoryService.ts:564`); `importPerfMetrics` re-scans every `result.json` per summary request (`perfRunImport.ts:319`); `enforceRetention` (`rmSync` of up to the whole budget) runs inside `activate()` before `activate.begin` (`diagnosticsManager.ts:157`).
- **`compareReps` parses without shape validation** (`perfHistoryService.ts:570`): a `null` line in either rep throws out of the RPC instead of becoming a note.
- **`deleteRun` scope** (`directoryProvider.ts:950`): containment is correct (single segment, no `..`), but any direct child of a misconfigured `perfRunsRoot` is deletable by name — it does not require the id to be an indexed run and does not reject symlinks/junctions. Host modal + `rmSync` on a link removes the link only, so this is scope, not escape.
- **Webview robustness** (verified): none of the 34 `sendRequest(...).then()` calls has a `.catch` — any host error leaves a page on "Loading…" with an unhandled rejection; no effect guards out-of-order responses (fast source/run switches pair the wrong data); `liveEvents` in `state.tsx` has no consumer but re-renders every `useDc()` consumer at push rate (~8/s), defeating the 1 Hz throttle; the runs table's scroll position is yanked back every second while live (`pagesPerfHistory.tsx:1223`); `ph/deleteRun`/`ph/addSource` errors are dropped so a failed delete is silent (`:757`); waterfall and SQL Activity render every activity/row with no cap and the host `buildWaterfall`/`sqlActivityRows` are uncapped, while every other list is capped.
- **Localization** (Copilot's thread): accurate — zero `locConstants` imports under `pages/DebugConsole` (~400 strings) and the host has the same gap (~20 strings in the manager/controller: dialogs, quick picks, status bar, panel title). Decide once for this surface (localize both, or document the exemption for an internal diagnostics tool), not half.

### C — questions / notes

- `mssql.debugConsole.richSnapshotHeartbeat` is read (`richCollection.ts:40`) but not declared with the other nine settings.
- `dc/getPerfSummary` accepts a filesystem path from the webview (`controller:466`); no page sends it — drop the field so paths stay host-chosen.
- `buildWaterfall` `BigInt(monotonicNs)` throws on a non-numeric string that `isDiagEvent` does not validate (one bad journal line fails the whole trace), and `IRREGULAR_END` pairs with the *first* begin rather than the last-before-end (`analysis.ts:287`, `:312`).
- `DumpTab` defaults `repId` to 0 while reps are 1-based (`pagesPerfHistory.tsx:2271`); `CompareTab` hard-codes baseline reps `[0..4]`.
- Keyboard access: row selection and link-styled actions are `<tr onClick>`/`<a onClick>` without focusability (`pagesCore.tsx:716` representative).
- Percentile definitions differ from the CLI (nearest-rank here vs linear interpolation + trimmed mean at n≥10 in `regression/statistics.ts`), so in-product medians will not match `perftest history` for even n; `mergeGroup` averages medians unweighted under a `p50Ms` label. `importWarnings` filters validations by the substring `"import"`, which no harness validation name contains. Quick-compare sets `deltaPct = 0` for a zero baseline and applies a 50 ms wallclock floor to non-wallclock metrics.
- `dump(file: "result")` surfaces `environment.machineId` (= hostname), branch names and `errors[].stack` (local paths) to the webview — local-only, but worth knowing.
- Under full elevation the export dialog still says "redacted JSONL"; an unrelated settings edit during elevation silently drops the elevation without an `elevation.expired` event (`diagnosticsManager.ts:90-102`).
- Registry vs code: `sessionDiag.elevated.reason` is `identifierSensitive` in the registry and `user.text` in code — neither is plain, so harmless, but the registry note ("stored only in the local audit record") does not describe what redacted mode does with it.
- Test coverage: `perfRunImport.ts` and `perfHistoryService.ts` have no tests at all; `perfHistoryProvider.test.ts` fixtures are not harness-shaped (`warmup: false`, no `errors[]`/`environment`/`validations`), which is why the contract drift above is invisible — a fixture copied from a real run (`pr22827-runtime-validation/runs/…`) would catch it. No test covers `enforceRetention` with an orphaned active session, the console command when the setting is toggled live, or the STS listener with oversized field strings.

### Copilot's three comments

All valid: the `proc:driver` doc mismatch (nit), the byte-vs-UTF-16 cap (agree — `:79` on this head, same pattern at `perfHistoryService.ts:566`; the missing *file*-level cap is the bigger issue), and the webview localization gap (accurate; host has the same). Replied on each thread.

### Refactoring candidates (for the pass you mentioned)

1. One `useRpcQuery(request, params, deps)` hook in the webview: error state, stale-response guard, and loading placeholder in one place — removes the two most repeated defects.
2. Stream/index instead of whole-file reads: `readline` for JSONL segments and markers, manifest-derived integrity checks, a persisted per-session action summary for History.
3. Split `pagesPerfHistory.tsx` (2,481 lines) per tab and move the shared `useVirtualRows` into `common.tsx`; cap `buildWaterfall`/`sqlActivityRows` host-side.
4. Derive warmup/failure/percentiles from the harness contract module (`perf-contracts`) rather than re-implementing — the drift found here is exactly what that shared package exists to prevent.
5. Localization policy for the surface, applied to host and webview together.

### Merge readiness

- Blocking: A1 (SQL text plain/exportable), A2 (retention), A3 (command registration).
- Should land here: B items (STS dependency statement, harness contract drift, webview error/stale handling, uncapped reads/rendering, `deleteRun` scope).
- Follow-up acceptable: C items and the coverage gaps, or fold them into the refactoring pass — happy to re-review after that.

(source: Fable)

extensions/mssql/src/controllers/debugConsoleWebviewController.ts
```diff
@@ -0,0 +1,656 @@
...
+                        medianMs: Number(median.toFixed(1)),
+                        count: entry.durations.length,
+                        errors: entry.errors,
+                    });
+                    trendMap.set(actionLabel, trend);
+                }
+            };
+
+            // Stored sessions (newest first, bounded), then the live session.
+            for (const { manifest } of this.diagnostics.store.listLocalSessions().slice(0, 15)) {
```
kburtram, line 556: **Recommended.** The History page loads the *entire* event array of up to 15 stored sessions (`eventsForSource` → `loadSessionEvents` reads and JSON-parses every segment) just to compute per-action medians, and `SessionStore`'s cache keeps only 5, so every visit re-parses the evicted ones. Peak memory is the largest session plus five cached ones. Persist the action summary (label → durations/errors) in the manifest at `close()`/flush, or compute it from a bounded tail; the manifest's `eventCount`/`gapCount` already cover the row-level columns.

(source: Fable)

extensions/mssql/src/diagnostics/diagnosticsManager.ts
```diff
@@ -0,0 +1,299 @@
...
+                    this.provenance,
+                );
+                diag.addSink(this.storeSink);
+                diag.emit({
+                    feature: "sessionDiag",
+                    type: "sessionDiag.enabled",
+                    fields: { captureMode: { raw: mode, cls: "diagnostic.metadata" } },
+                });
+                const config = vscode.workspace.getConfiguration();
+                this.store.enforceRetention(
```
kburtram, line 209: **Recommended.** `enforceRetention` runs synchronously inside `enableCapture`, which the constructor calls from `activate()` *before* `mssql.activate.begin` (`extension.ts:68`): it lists every session, parses every manifest, `statSync`s segments for older manifests, and `rmSync`s directories that can total hundreds of MB — all inside the activation window this very PR instruments. Defer it (`setImmediate`/after `activate.end`) or run it from the flush timer; the current session's directory is created either way.

(source: Fable)

extensions/mssql/src/diagnostics/diagnosticsManager.ts
```diff
@@ -0,0 +1,299 @@
...
+                .update(SETTING_ENABLED, true, vscode.ConfigurationTarget.Global);
+        });
+        register("mssql.sessionDiag.disable", async () => {
+            await vscode.workspace
+                .getConfiguration()
+                .update(SETTING_ENABLED, false, vscode.ConfigurationTarget.Global);
+        });
+        register("mssql.sessionDiag.clear", async () => {
+            const sessions = this.store.listLocalSessions();
+            const choice = await vscode.window.showWarningMessage(
```
kburtram, line 275: **Recommended (convention).** Host-side user-facing strings are hard-coded English here and throughout this file and the controller: this dialog and its `"Delete all"` action, the elevate prompt/placeholder/quick-pick items (:214-223), the status-bar texts and tooltips (:263-276), the panel title `"MSSQL Debug Console"`, the delete-run dialog (`controller:231-235`), the import/export dialog titles and warnings (:443, :451, :596). Repo convention is `LocConstants` via `l10n.t` so the pre-commit extractor picks them up — Copilot flagged the webview (~400 strings, zero `locConstants` imports under `pages/DebugConsole`); the host has the same gap (~20). Decide once for this surface: localize both (the plumbing exists) or document the exemption for an internal diagnostics tool — but not half.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/pagesCore.tsx
```diff
@@ -0,0 +1,898 @@
...
+        }
+        void rpc
+            .sendRequest(DcGetTraceQualityRequest.type, {
+                sourceId: activeSourceId,
+                traceId: model.traceId,
+            })
+            .then(setQuality);
+    }, [rpc, activeSourceId, model?.traceId]);
+
+    useEffect(() => {
```
kburtram: **Recommended (two related patterns, this effect is representative).** (1) None of the 34 `sendRequest(...).then(...)` calls under `pages/DebugConsole` has a `.catch` — the host wrapper rethrows handler errors, so any host-side failure (a source deleted underneath the index, a path-containment refusal in `perfHistoryService.dump`, a read error) leaves the page on "Loading…" forever with an unhandled rejection in the console. A tiny `useRpcQuery` hook, or `.catch(e => setError(String(e)))` per effect, plus an error line instead of the placeholder, fixes all of them. (2) No effect guards against out-of-order completion: switch source A→B here and `setModel` for A's trace can land after B's list is shown; in Perf History (`pagesPerfHistory.tsx:656-717`) clicking run R1 then R2 quickly can pair R2's details with R1's series. Standard fix: `let stale = false; … .then(r => { if (!stale) set(r); }); return () => { stale = true; };`.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/state.tsx
```diff
@@ -0,0 +1,312 @@
...
+    const snapshot = useSyncExternalStore(subscribe, getSnapshot);
+    const state: DebugConsoleState | undefined =
+        snapshot && Object.keys(snapshot).length > 0 ? snapshot : undefined;
+    const [sources, setSources] = useState<DebugSource[]>([]);
+    const [activeSourceId, setActiveSourceId] = useState<string>("");
+    // Live is not a mode the user toggles — it IS the current-session source.
+    // ("" = initial state before the host pushes the live source id.)
+    const isLive = activeSourceId === "" || activeSourceId.startsWith("live:");
+    const [route, navigate] = useState<DcRoute>({ page: "overview" });
+    const [liveEvents, setLiveEvents] = useState<DiagEvent[]>([]);
```
kburtram: **Recommended.** `liveEvents` has no consumer anywhere (only this file references it), yet every `dc/livePush` (~8/s during a run) calls `setLiveEvents`, which changes the memoized context value (it is in the deps at :321) and re-renders every `useDc()` consumer — TopBar, LeftNav and the active page (the whole Perf History workbench included) at push rate, defeating the 1 Hz `dataVersion` throttle at :175-191, while holding up to 20,000 events (`LIVE_VIEW_CAP`) for nothing. Delete it (keep only the gap bookkeeping) or move it to a ref. Related: :247-253 ignores `lastSeq` from `dc/subscribeLive`; safe only because the host builds the snapshot synchronously — merge `events.filter(e => e.seq > lastSeq)` so it stays safe if the handler ever awaits.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/waterfallView.tsx
```diff
@@ -0,0 +1,754 @@
...
+                        ref={zoomer.hscrollRef}
+                        tabIndex={0}
+                        role="application"
+                        aria-label="Waterfall timeline — scroll to zoom, drag to pan, W/S/A/D keys"
+                        onScroll={zoomer.onScroll}
+                        onKeyDown={zoomer.onKeyDown}
+                        onMouseDown={zoomer.onMouseDown}
+                        onDoubleClick={zoomer.reset}>
+                        <div className="dc-wf2-inner" style={{ width: `${zoom * 100}%` }}>
+                            {lanes.map((lane) => (
```
kburtram, line 359: **Recommended.** One absolutely-positioned `div` per activity here and one `tr` per activity at :543-575, with no cap, and `windowFrac` state changes on every scroll tick (:110-114, :299-305) re-render the whole chart. Host-side `buildWaterfall` (`analysis.ts:268-401`) has no activity cap either and runs over the full source, so a trace with thousands of spans (OE expansion fan-out, an imported perf rep) means thousands of DOM nodes rebuilt per scroll frame. Cap/aggregate in `buildWaterfall` (top-N by duration plus an honest "N more" note) and only render bars intersecting the window; memoize per-lane arrays independent of `windowFrac`. Every other list in the console is capped (Trace 400, runs 500 + virtualized, perf SQL 200) — this and the SQL Activity page (`pagesMore.tsx:257-275`, host `sqlActivityRows` uncapped, selection by object identity lost on every 1 Hz refetch) are the two that are not.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/pagesPerfHistory.tsx
```diff
@@ -0,0 +1,2481 @@
...
+        }
+        const index = props.runs.findIndex((r) => r.runId === props.focusRunId);
+        if (index < 0) {
+            return;
+        }
+        const top = index * RUN_ROW_HEIGHT;
+        if (top < el.scrollTop || top > el.scrollTop + el.clientHeight - RUN_ROW_HEIGHT * 2) {
+            el.scrollTop = Math.max(0, top - el.clientHeight / 3);
+        }
+    }, [props.focusRunId, props.runs]);
```
kburtram: **Bug.** This effect depends on `props.runs`, and `runs` is replaced with a new array on every `dataVersion` bump (:640-648), which happens at 1 Hz whenever any live event flows (`state.tsx:230`). If the user has scrolled the focused run out of view, the condition at :1220 is true every second and `scrollTop` is yanked back — the runs table cannot be scrolled while the extension is doing anything. Depend on `props.focusRunId` (or the resolved index) only.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/pagesPerfHistory.tsx
```diff
@@ -0,0 +1,2481 @@
...
+                    : [...selectedRunIds, runId],
+            );
+        } else {
+            setSelectedRunIds([runId]);
+        }
+        anchorRunRef.current = runId;
+    };
+
+    const deleteRun = (runId: string) => {
+        void rpc.sendRequest(PhDeleteRunRequest.type, { sourceId, runId }).then((outcome) => {
```
kburtram: **Recommended.** Only `outcome.ok` is checked; `outcome.error` from `ph/deleteRun` is dropped, so a real delete failure (path-containment refusal, EBUSY on Windows with a report open) is silent — the run simply stays in the list after the user confirmed a modal. Same at :156-163 for `ph/addSource`, `state.tsx:127-148` drops `GapBackfillResult.reason`, and the top-bar Export (`shell.tsx:248-254`) discards `{ path, error }` while the Exports page shows them. Surface the reason.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/pagesPerfHistory.tsx
```diff
@@ -0,0 +1,2481 @@
...
+
+function DumpTab(props: {
+    sourceId: string;
+    runId: string;
+    scenarioId: string;
+    details: PerfScenarioDetails | undefined;
+}) {
+    const { rpc } = useDc();
+    const [file, setFile] = useState<"summary" | "result" | "markersHead">("result");
+    const [repId, setRepId] = useState(0);
```
kburtram: Nit (bug). `repId` defaults to `0` but reps are 1-based (`directoryProvider.ts:982` builds `rep-00` from it), so the first dump request targets a directory that never exists and the `<select value={0}>` at :2295 shows the first option while state says 0; `CompareTab` (:2331) defaults to 1. Also :2368 hard-codes baseline reps `[0..4]` rather than deriving from the run.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/pagesCore.tsx
```diff
@@ -0,0 +1,898 @@
...
+                                                {(row as GapRecord).throughSeq} ·{" "}
+                                                {(row as GapRecord).reason} —{" "}
+                                                <span className="dc-muted">
+                                                    retained in Session Diag store when capture is
+                                                    on
+                                                </span>
+                                            </td>
+                                        </tr>
+                                    ) : (
+                                        <tr
```
kburtram, line 733: Accessibility (representative). Row selection — the page's primary action here, Overview → waterfall at :215-219, run/scenario selection at `pagesPerfHistory.tsx:1276-1282`/:1399-1408 — is on `<tr onClick>`; the link-styled actions at :192-196, :264-277, :315-319, :413-419 are `<a>` with `onClick` but no `href`/`role`/`tabIndex`, so they are not focusable; waterfall bars (`waterfallView.tsx:378-394`) have no keyboard selection. Keyboard/screen-reader users cannot select an event, run, or scenario. `role="button" tabIndex={0}` + Enter/Space on rows (or a roving-tabindex table) and real `<button>`s for the actions.

(source: Fable)

extensions/mssql/src/diagnostics/analysis.ts
```diff
@@ -0,0 +1,456 @@
...
+            const stem = event.type.slice(0, -".end".length);
+            const begin = beginStack.get(`${stem}@${event.process}`);
+            if (begin) {
+                beginStack.delete(`${stem}@${event.process}`);
+                const sameProcess =
+                    begin.process === event.process &&
+                    begin.monotonicNs !== undefined &&
+                    event.monotonicNs !== undefined;
+                const durationMs = sameProcess
+                    ? Number(BigInt(event.monotonicNs!) - BigInt(begin.monotonicNs!)) / 1e6
```
kburtram: Nit (robustness). `BigInt(event.monotonicNs!)` throws on a non-numeric string, and `isDiagEvent` (`sessionStore.ts:63`) does not validate `monotonicNs` — one malformed line in a stored or imported journal makes `dc/getWaterfall` reject for the whole trace. Guard with a digit-only check and fall back to the epoch delta. Also `IRREGULAR_END` pairing (:312) takes the *first* matching begin in the trace, not the last one before the end, so a trace with two `connection.begin`s pairs the second `ready` with the first `begin` (the harness normalizer pairs last-begin-before-first-end).

(source: Fable)

extensions/mssql/src/diagnostics/richCollection.ts
```diff
@@ -0,0 +1,162 @@
...
+ * needed (per-span rich deltas cover most investigations). It emits only
+ * when mssql.debugConsole.richSnapshotHeartbeat is ALSO on (default off);
+ * the sampler still runs so span deltas and percentiles stay available.
+ */
+function heartbeatEnabled(): boolean {
+    try {
+        // eslint-disable-next-line @typescript-eslint/no-require-imports
+        const vscode = require("vscode") as typeof import("vscode");
+        return vscode.workspace
+            .getConfiguration("mssql.debugConsole")
```
kburtram, line 40: Nit. `mssql.debugConsole.richSnapshotHeartbeat` is read here but not declared in `package.json` (the other nine new settings are). Users can only reach it by hand-editing `settings.json` (with an "unknown setting" warning). Declare it or fold it into `richCollection`.

(source: Fable)

extensions/mssql/src/controllers/debugConsoleWebviewController.ts
```diff
@@ -0,0 +1,656 @@
...
+            this.diagnostics.store.registerPerfRun(
+                `perfrun:${this.perfRunCounter}`,
+                imported.label,
+                imported.events,
+            );
+            return this.listSources();
+        });
+
+        this.onRequest(DcGetPerfSummaryRequest.type, async (request) => {
+            const root =
```
kburtram: Nit (host/webview boundary). `dc/getPerfSummary` accepts a filesystem path (`perfRunsRoot`) from the webview and reads it. No page sends it today (`pagesMore.tsx:684` only mentions the setting), and the stated design is "the webview remains a renderer" with paths chosen host-side — drop the field from the request type so the setting (or the host picker) stays the only source, as `ph/addSource` already does.

(source: Fable)

extensions/mssql/src/diagnostics/perfRunImport.ts
```diff
@@ -0,0 +1,392 @@
...
+                            ? {
+                                  rowCount: {
+                                      v: row.row_count,
+                                      cls: "diagnostic.metadata" as const,
+                                      handling: "plain" as const,
+                                  },
+                              }
+                            : {}),
+                        text: text
+                            ? // Synthetic harness DB: plain by contract, truncated for sanity.
```
kburtram: **Should-fix (privacy invariant).** "Synthetic harness DB: plain by contract" is not enforced anywhere: the harness persists `batch_text`/`statement_text` whenever `captureSqlText && passType === "diagnostic"` (`perftest-cli/src/collectors/sqlServerXEvents.ts:160-166`), and a run can target any server (`environment.sql.provider: "external"` in the real runs on this machine). Any directory the user opens as a Perf History source or imports via the picker is trusted the same way. The text is emitted as `cls: "sql.text", handling: "plain"` with `redactedFields: 0`, reaches the webview through `sqlActivity`/`waterfall`, and `dc/export` (`controller:590-618`) writes the `perfrun:` source verbatim into a file whose dialog says "redacted JSONL" — so real SQL text leaves the machine labeled redacted. That contradicts the PR's "SQL text … never enters diagnostics" invariant. Apply the normal `sql.text` policy at import (digest/omit) unless the rep proves synthetic (`artifacts/sql/sql-activity-rollup.json.sqlTextPersisted` + provider ≠ external), or keep only `object_name` and the counts.

(source: Fable)

extensions/mssql/src/diagnostics/perfHistory/directoryProvider.ts
```diff
@@ -0,0 +1,1055 @@
...
+                        continue;
+                    }
+                    if (metric.official) {
+                        official[metric.name] = metric.value;
+                    } else {
+                        diagnostic[metric.name] = metric.value;
+                    }
+                }
+                const status = result.status ?? "unknown";
+                const warmup = result.warmup === true;
```
kburtram: **Recommended (harness contract drift).** `result.warmup` is never written by the harness: `PerfResult` (`perf-contracts/src/result.ts:139-154`) has no `warmup` field, `normalizeRep` does not carry it, and `runPipeline.ts:341,364` only passes warmup to the SQLite store. The CLI's own loader derives it as `repId < config.warmupRepetitions` from `run-config.snapshot.jsonc` (`perftest-cli/src/central/runLoader.ts:82-89`), and its aggregates exclude warmup reps (`sqliteStore.ts:433,547,634`). Here every rep is a valid sample: with `warmupRepetitions: 1, repetitions: 5`, p95 becomes the cold rep and `n` shows 6 where `perftest history` shows 5 (also :113, :201, :812, :881 and `perfRunImport.ts:365`). Same family: `failureReason` (:505, :535) is never populated — the harness records `errors[]` (`result.ts:131-137`); `summary.scenarios[].skipped/reason` and `summary.connection.connectionMode` (:474-475, :578-586, :623-625) have no writer anywhere in the repo. The fixtures in `perfHistoryProvider.test.ts:217-278` write `warmup: false` and omit `errors[]`, which is why none of this is visible to the tests.

(source: Fable)

extensions/mssql/src/diagnostics/perfHistory/perfHistoryService.ts
```diff
@@ -0,0 +1,770 @@
...
+                  scenarioCount: 0,
+              } as PerfHistorySource);
+        const rows = provider
+            ? provider.allRunRows().sort((a, b) => a.createdUtc.localeCompare(b.createdUtc))
+            : [];
+        const latest = rows[rows.length - 1];
+        const previous = rows.length > 1 ? rows[rows.length - 2] : undefined;
+        const failedReps = rows.reduce((sum, r) => sum + r.failedReps, 0);
+        const invalidReps = rows.reduce((sum, r) => sum + r.invalidReps, 0);
+        const deltaVsPrevPct =
```
kburtram, line 329: **Recommended.** `deltaVsPrevPct` (and the trend at :325-334) compares run-wide wallclock medians pooled across *whatever scenarios each run happened to contain* (`directoryProvider.ts:540-545`). The two real runs on this machine are single-scenario activation runs (~7-8 s); a following query-only run at ~100 ms reports "-98%" on the Summary tab with no scenario-mix caveat. Compute the delta per scenario over the intersection of the two runs (then aggregate), or drop the KPI and keep the per-scenario `latestSlower`.

(source: Fable)

extensions/mssql/src/diagnostics/perfRunImport.ts
```diff
@@ -0,0 +1,392 @@
...
+
+export function importPerfRep(repDir: string, repLabel: string): DiagEvent[] {
+    const events: DiagEvent[] = [];
+    let seq = 0;
+    const traceId = `perf_${repLabel.replace(/[^a-z0-9]/gi, "_")}`;
+
+    const markersPath = path.join(repDir, "markers.jsonl");
+    if (fs.existsSync(markersPath)) {
+        let skippedLines = 0;
+        for (const line of fs.readFileSync(markersPath, "utf8").split("\n")) {
```
kburtram: **Recommended.** Whole-file `readFileSync(...).split("\n")` on the extension-host thread here, at :177 (`sql-activity.jsonl`) and in `perfHistoryService.ts:472,564`; the only cap is per *line* (512 KiB). Soak reps write every iteration's markers into one `markers.jsonl` (`runPipeline.ts:938-946`), so hundreds of MB is realistic: the host blocks for seconds per waterfall/compare request, and above ~512 MB `readFileSync` throws `ERR_STRING_TOO_LONG` — uncaught here (the RPC rejects), swallowed into `found: false` in `richDiagnostics`. `statSync` first and refuse above a file cap with an `importWarnings` entry, or stream with `readline`. Related: `importPerfMetrics` (:319) re-scans every `result.json` under the root per `dc/getPerfSummary` request — the indexed provider exists to avoid exactly that.

(source: Fable)

extensions/mssql/src/diagnostics/perfHistory/perfHistoryService.ts
```diff
@@ -0,0 +1,770 @@
...
+                    timestampUnixNs: string;
+                    attrs?: Record<string, unknown>;
+                }> = [];
+                for (const line of fs.readFileSync(file, "utf8").split("\n")) {
+                    const trimmed = line.trim();
+                    if (!trimmed || trimmed.length > 512 * 1024) {
+                        continue;
+                    }
+                    try {
+                        markers.push(JSON.parse(trimmed));
```
kburtram: **Recommended.** Any parsed JSON is pushed here (`null`, `42`, `"x"`), then :592 `BigInt(m.timestampUnixNs)` and :594-596/:618/:655 dereference `name` outside try/catch — a single `null` line in either rep rejects the whole compare request instead of producing a note. Keep only entries with `typeof m?.name === "string" && typeof m?.timestampUnixNs === "string"` (as `importPerfRep` :85 does) and count refused lines into `notes`. `richDiagnostics` (:471-508) has the same shape: one bad line discards every snapshot parsed before it.

(source: Fable)

extensions/mssql/src/diagnostics/perfHistory/directoryProvider.ts
```diff
@@ -0,0 +1,1055 @@
...
+        };
+    }
+
+    /**
+     * Delete one run: remove the directory from disk and evict it from the
+     * index (persisted). Refuses run ids that are not direct children of the
+     * root — no path tricks.
+     */
+    public deleteRun(runId: string): { ok: boolean; error?: string } {
+        if (!/^[^/\\]+$/.test(runId) || runId.startsWith(".")) {
```
kburtram, line 1008: **Recommended (scope).** Containment is right — single segment, `..`/dot-prefix rejected, host-side modal — but "no path tricks" undersells what is deletable: any direct child of the configured root by name, whether or not it is an indexed run, and symlinked/junction children are not rejected even though the scan itself skips them (:392 `isDirectory()`), so the index and the delete surface disagree. With an application-scoped `perfRunsRoot` pointed one level too high, a confirmed click removes a non-run directory. Require `this.index.runs[runId]` to exist and refuse `lstatSync(runDir).isSymbolicLink()`. (`rmSync` on a link removes the link, not the target, so this is scope, not escape.)

(source: Fable)

extensions/mssql/test/unit/perfHistoryProvider.test.ts
```diff
@@ -0,0 +1,456 @@
...
+        expect(since.rows).to.have.length(1);
+        const paged = filterRuns(rows, { sourceId: "s", offset: 1, limit: 1 });
+        expect(paged.rows).to.have.length(1);
+        expect(paged.total).to.equal(2);
+    });
+});
+
+// --- filesystem provider -------------------------------------------------------
+
+function writeRun(
```
kburtram, line 225: Coverage. The fixtures here are not harness-shaped (`warmup: false` written explicitly, no `errors[]`/`environment`/`git`/`validations`), which hides the contract drift noted on `directoryProvider.ts:528`; a fixture copied from a real run (`C:\repos\work2\pr22827-runtime-validation\runs\2026-08-26T20-45-14Z_8eb626ef` has `summary.json`, `result.json`, `markers.jsonl`) would catch it. `perfRunImport.ts` and `perfHistoryService.ts` have no tests at all (marker import + oversized-line accounting, `sql.text` handling, `importPerfMetrics`, `dump` cap/containment, `compareReps` with malformed lines, `richDiagnostics`, `summary()` KPIs). The corrupt-run test (:298) covers unparseable JSON only — `null`/array bodies at `directoryProvider.ts:478/:510` throw outside the per-rep try and silently drop the run from the index; a non-`rep-` directory yields a `NaN` repId that breaks the :935 sort. Elsewhere: nothing exercises `enforceRetention` with an orphaned `active` session, the console command when the setting is toggled live, or the STS listener with oversized field strings.

(source: Fable)

extensions/mssql/src/sharedInterfaces/traceFilter.ts
```diff
@@ -0,0 +1,126 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Consolidated Trace filter expressions (completions-style live capture):
+ *
+ *   dur>1000      duration greater than 1000ms   (also dur>=, dur<, dur<=, 1.5s)
+ *   proc:sts      process (sts|extension|webview|sql|driver|harness|system)
+ *   feat:query    feature bucket
```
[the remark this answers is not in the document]
kburtram: Agree — nit. `driver` is a waterfall *lane* (`analysis.ts:345` puts `sts.sql.*`/`sts.smo.*` spans on a `"driver"` lane) but not a `DiagProcess`, which is probably where the alias came from. Either drop `driver` from the doc/alias table or, if a filter for the driver lane is wanted, add it as a feature/type filter rather than a process alias.

(source: Fable)

extensions/mssql/src/diagnostics/perfRunImport.ts
```diff
@@ -0,0 +1,392 @@
...
+            const trimmed = line.trim();
+            if (!trimmed) continue;
+            // Imports are untrusted: an absurd line (multi-MB attr blob,
+            // binary garbage) is skipped with accounting, never parsed into
+            // memory pressure or a crash.
+            if (trimmed.length > MAX_MARKER_LINE_BYTES) {
+                skippedLines++;
+                continue;
+            }
```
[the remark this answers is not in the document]
kburtram: Agree. On this head the check is at `perfRunImport.ts:79` (`trimmed.length > MAX_MARKER_LINE_BYTES`), and the same code-units-vs-bytes pattern is at `perfHistoryService.ts:566`; `Buffer.byteLength(trimmed, "utf8")` fixes both (or rename the constant to chars). The bigger gap in the same code path is that there is no *file*-level cap at all — `readFileSync(...).split("\n")` on soak-sized `markers.jsonl` blocks the host and throws `ERR_STRING_TOO_LONG` above ~512 MB; see the comment on :73.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/shell.tsx
```diff
@@ -0,0 +1,368 @@
...
+            { id: "query", label: "Query & Results", icon: "▶" },
+            { id: "oe", label: "Object Explorer", icon: "⌥" },
+        ],
+    },
+    {
+        group: "Session",
+        items: [
+            { id: "exports", label: "Exports", icon: "⇩" },
+            { id: "settings", label: "Settings", icon: "✲" },
+        ],
```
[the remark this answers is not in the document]
kburtram: Accurate: zero `locConstants` imports under `pages/DebugConsole` and roughly 400 hard-coded strings (JSX text + `title`/`placeholder`/`aria-label`), while `QueryResult`/`SchemaDesigner` route theirs through `webviews/common/locConstants.ts`. The host side of this PR has the same gap (~20 strings in `diagnosticsManager.ts`/`debugConsoleWebviewController.ts`: dialogs, quick picks, status bar, panel title) that the pre-commit extractor cannot see. Whether an internal diagnostics surface must be localized is a policy call for the maintainers — but it should be one decision applied to both sides, not the host in English and the webview localized or vice versa. Details in the review body.

(source: Fable)

extensions/mssql/src/sharedInterfaces/traceFilter.ts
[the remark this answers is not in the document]
kburtram: Fixed in 0d03cd2f3. `driver` is dropped from the `proc:` alias table and its doc line; the driver *lane* in the waterfall is unaffected (it is not a process).

(source: Fable)

extensions/mssql/src/diagnostics/perfRunImport.ts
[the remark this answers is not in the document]
kburtram: Fixed in 0d03cd2f3. The per-line guard now uses `Buffer.byteLength(trimmed, "utf8")` through a shared `readJsonlBounded` reader that also enforces a 64 MiB whole-file cap before anything is parsed; `perfHistoryService` (`richDiagnostics`, `compareReps`) reads through the same helper. Test: a line under the cap in UTF-16 units but over it in bytes is refused; an oversized file is refused with a reason.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/shell.tsx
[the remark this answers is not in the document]
kburtram: Partially addressed in 0d03cd2f3: every host-side user-facing string in this PR now goes through `LocConstants` (new `SessionDiag` and `DebugConsole` classes — dialogs, quick picks, status bar, panel/dialog titles) and the bundle/XLIFF were regenerated. The webview's ~400 strings are deliberately left for the refactoring pass the author has planned before merge: it is a policy decision for an internal diagnostics surface, and doing it means threading `locConstants` through eleven page files — better done once, with the page split, than as a late-review patch. Tracked in the PR body.

(source: Fable)

extensions/mssql/src/diagnostics/sessionStore.ts
kburtram: Fixed in 0d03cd2f3. `enforceRetention` now takes the current session id and protects only that session; an `active` manifest that is not current (crash/kill/OS shutdown) is evictable by count, age and size like any other, and `listSources` labels it "(interrupted)". Retention is also deferred off the activation path (see the manager thread). Tests: `debugConsoleSessionStore.test.ts` — orphaned active session evicted by count and by size budget, current session never.

(source: Fable)

extensions/mssql/src/extension.ts
kburtram: Fixed in 0d03cd2f3. `mssql.openDebugConsole` is registered unconditionally; the handler gates on the setting (info message when off) and, when the setting was enabled after activation, opens the console and shows a one-time "restart to include SQL Tools Service diagnostics" notice because the STS listener only starts before the service spawns. No more "command not found" from the palette entry or the status-bar item.

(source: Fable)

extensions/mssql/src/extension.ts
kburtram: Addressed in the PR body: it now states that the service-side emitter (STS `43a79552`, currently on STS `dev/query`/`dev/runbook`, not on STS main and not in `6.0.20260827.1`) has not shipped, that the SQL Tools Service lane of the console stays empty until it does, and that the listener/env injection here is inert but harmless. No STS PR exists yet to link.

(source: Fable)

extensions/mssql/src/diagnostics/stsDiagListener.ts
kburtram: Fixed in 0d03cd2f3. The listener bounds what it trusts: ≤32 fields (rest counted as dropped), strings ≤256 chars (truncation counted), keys ≤64 chars, `type` ≤128 chars, non-finite numbers rejected, and `feature` restricted to the emitter's buckets (`rpc`, `sqlDriver`, `objectExplorer`, `connection`, `query`, `system`) with anything else collapsing to `sts`. Accounting fields (`stsDiag.truncatedFields`/`droppedFields`) make the bound visible. Test added.

(source: Fable)

extensions/mssql/src/diagnostics/sessionStore.ts
kburtram: Fixed in 0d03cd2f3. `validateStore` no longer reads whole journals: every segment is checked by `stat` size plus bounded reads of its first line (64 KiB) and last byte (trailing newline); only the trailing segment — the one an interrupted write can tear, capped at 5,000 events by the sink — gets a full line count. The existing canary (clean journal / torn journal) still passes and a new test covers the bounded path.

(source: Fable)

extensions/mssql/src/controllers/debugConsoleWebviewController.ts
kburtram: Fixed in 0d03cd2f3. History no longer loads whole sessions: `SessionStore.historySummaryFor` computes the per-session aggregate (events, errors, action label/feature/status/duration) once and persists it as `history-summary.json` next to the manifest; the journal is read transiently without evicting the query cache, and the current session is never cached. Test covers compute-once, sidecar reuse, non-cacheable, and id containment.

(source: Fable)

extensions/mssql/src/diagnostics/diagnosticsManager.ts
kburtram: Fixed in 0d03cd2f3. Retention is scheduled 15 s after the store sink starts (unref'd, cleared on dispose, best effort) instead of running synchronously inside `activate()` before `mssql.activate.begin`. Same change also keeps an active time-bounded elevation alive across an unrelated settings edit (it ends on its own deadline with `elevation.expired`; an explicit disable still wins).

(source: Fable)

extensions/mssql/src/diagnostics/diagnosticsManager.ts
kburtram: Fixed in 0d03cd2f3. All host-side strings in the manager, controller and perf-history service now go through `LocConstants` (`SessionDiag`, `DebugConsole`); quick-pick durations are structured items rather than parsed labels; the export dialog title is honest under full capture. Bundle/XLIFF regenerated by the pre-commit extractor. The webview half is deferred to the refactoring pass (see the Copilot localization thread).

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/pagesCore.tsx
kburtram: Fixed in 0d03cd2f3. Added `useRpcQuery` in `common.tsx` (stale-response guard + error state + retry) and converted the Overview, Trace quality, Trace query, cause tree, Waterfall (list+model in one guarded query), feature pages, SQL Activity, History and Health cards to it — every page now renders an `RpcError` line with Retry instead of "Loading…" forever. `pagesPerfHistory.tsx`'s 14 effects/handlers keep their multi-state shape but each has a `stale` flag and an error path that reports to a new top-bar notice; the state provider's own RPCs have error handlers too. Zero un-handled `sendRequest` promises remain.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/state.tsx
kburtram: Fixed in 0d03cd2f3. `liveEvents` and its 20k-event buffer are gone: live pushes only bump `dataVersion` (throttled to 1 Hz); pages query the host archive. `lastSeq` is no longer relevant since the webview keeps no buffer. Backfill keeps the gap bookkeeping and now surfaces the host's reason on failure.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/waterfallView.tsx
kburtram: Fixed in 0d03cd2f3. Host `buildWaterfall` caps at 800 activities (longest kept; extent and critical path use the full set) and reports `activityCap { shown, total }`, which the page and the event table display; `sqlActivityRows` caps at 5,000 newest rows and returns `{ rows, total }` so the SQL Activity page shows the honest count and selects by `sourceEventId` instead of object identity. The chart renders only bars intersecting the visible window (plus a half-window margin), so scroll ticks re-render the on-screen slice.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/pagesPerfHistory.tsx
kburtram: Fixed in 0d03cd2f3. The keep-visible effect depends on `props.focusRunId` only; the 1 Hz `runs` replacement no longer yanks the scroll position.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/pagesPerfHistory.tsx
kburtram: Fixed in 0d03cd2f3. `ph/deleteRun`, `ph/addSource`, rescan, backfill and the top-bar Export now surface their `error`/`reason` (and rejections) in a dismissible top-bar notice; a confirmed delete that fails can no longer look like success.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/pagesPerfHistory.tsx
kburtram: Correction on my side, then fixed in 0d03cd2f3: harness reps are 0-based (`rep-00` exists in the real runs; `runPipeline.ts` numbers from 0), so the Dump tab's default of 0 was right. What was wrong is now fixed: Dump defaults to the first rep the run actually has (state is validated against the rep list), and Compare no longer hard-codes baseline reps `[0..4]` or defaults to rep 1 — it offers the run's rep ids and defaults both sides to the first passed rep.

(source: Fable)

extensions/mssql/src/webviews/pages/DebugConsole/pagesCore.tsx
kburtram: Fixed in 0d03cd2f3. Added an `activatable()` helper (tabIndex/role=button/Enter+Space) and a `LinkButton` (a real `<button>` styled as a link) in `common.tsx`; applied to every row/card/chip selection in the console pages and the perf-history tables, and to the link-styled actions. Waterfall bars stay pointer targets; the event table rows below the chart are the keyboard path to the same selection (the hint bar says so).

(source: Fable)

extensions/mssql/src/diagnostics/analysis.ts
kburtram: Fixed in 0d03cd2f3. `pairDuration` uses the monotonic clocks only when both are well-formed digit strings (else the epoch delta), `isDiagEvent` refuses malformed `monotonicNs` at load, and irregular pairs now take the LAST unpaired begin before the end (tracked per pairing), matching the harness normalizer.

(source: Fable)

extensions/mssql/src/diagnostics/richCollection.ts
kburtram: Fixed in 0d03cd2f3. `mssql.debugConsole.richSnapshotHeartbeat` is declared and localized in `package.json`/`package.nls.json` (default false).

(source: Fable)

extensions/mssql/src/controllers/debugConsoleWebviewController.ts
kburtram: Fixed in 0d03cd2f3. `dc/getPerfSummary` takes no parameters; the root comes from the setting only.

(source: Fable)

extensions/mssql/src/diagnostics/perfRunImport.ts
kburtram: Fixed in 0d03cd2f3. Imported SQL text goes through the normal `sql.text` policy in digest mode (`classify(text, "sql.text", CAPTURE_POLICIES.digest)`): grouping by digest survives, plaintext never enters diagnostics, `redactedFields` is honest, and the `synthetic` tag is gone. Test: a canary query string never appears anywhere in the imported events, and equal text yields equal digests.

(source: Fable)

extensions/mssql/src/diagnostics/perfHistory/directoryProvider.ts
kburtram: Fixed in 0d03cd2f3. Warmup is derived the way the harness loader does it — `repId < warmupRepetitions` from `run-config.snapshot.jsonc` (a small JSONC reader; comments/trailing commas) — with `result.warmup === true` kept as an override; `failureReason` falls back to `errors[0].message`; `summary.json`/`result.json` bodies that are valid JSON but not objects are guarded, and a non-`rep-NN` directory without a numeric `repId` is skipped rather than producing `NaN`. Percentiles now use the CLI's linear-interpolation definition (`percentile([1,2,3,4],50) === 2.5`); the quick-compare absolute floor applies to wallclock only and a zero baseline is `inconclusive`; the baseline scenario is picked by the requested metric; `importWarnings` includes every failed validation. The provider test fixtures are harness-shaped now (`run-config.snapshot.jsonc`, `errors[]`, no `warmup` field) and assert all of this.

(source: Fable)

extensions/mssql/src/diagnostics/perfHistory/perfHistoryService.ts
kburtram: Fixed in 0d03cd2f3. `deltaVsPrevPct` is the median of the latest run's per-scenario deltas, each against that scenario's own baseline (closest earlier run that has it) — never a pooled run-wide median across different scenario sets. The run-wide trend series stays and the type comment says what it is. `latestSlower` now applies the same >10% AND >50 ms rule as the quick-compare verdict.

(source: Fable)

extensions/mssql/src/diagnostics/perfRunImport.ts
kburtram: Fixed in 0d03cd2f3. All JSONL artifact reads (`markers.jsonl`, `sql-activity.jsonl`, in import, waterfall, SQL activity, rich diagnostics and compare) go through `readJsonlBounded`: `statSync` first, 64 MiB whole-file cap refused with a reason that surfaces as an import warning / compare note, per-line UTF-8 byte cap. `importPerfMetrics` is memoized per root on a cheap fingerprint (run names + `summary.json` mtimes), so a summary refresh re-parses only when a run appears or changes.

(source: Fable)

extensions/mssql/src/diagnostics/perfHistory/perfHistoryService.ts
kburtram: Fixed in 0d03cd2f3. `parseMarkerLines` keeps only object lines with a string `name` and a digit-string `timestampUnixNs`; `compareReps` and `richDiagnostics` use it, so a `null`/number/string line is skipped (and counted into the notes) instead of rejecting the request or discarding earlier snapshots. Test added.

(source: Fable)

extensions/mssql/src/diagnostics/perfHistory/directoryProvider.ts
kburtram: Fixed in 0d03cd2f3. `deleteRun` now requires the id to be an indexed run of this source and refuses symlinked/junction children (`lstat`), matching what the scan indexes; the doc comment says exactly that. Bundle sources also no longer get `.dc-history-index.json` written into them (`persistIndex: false`). Tests: non-indexed child refused and left in place; junction refused (when the host can create one); no index file in a bundle root.

(source: Fable)

extensions/mssql/test/unit/perfHistoryProvider.test.ts
kburtram: Fixed in 0d03cd2f3. New `perfRunImport.test.ts` (bounded reads by bytes, file cap, JSONC parsing, warmup/repId/failureReason helpers, SQL-text-never-plain, import warnings, metrics memo, marker-line validation, dump redaction) and `debugConsoleSessionStore.test.ts` (retention with orphaned actives, size budget, interrupted labels, malformed monotonic clock, bounded validation, history sidecar); provider fixtures are harness-shaped and cover the warmup/errors contract, `null` summary bodies, non-indexed/linked deletes and read-only sources; the listener test covers oversized strings, field caps and the feature allowlist.

(source: Fable)

extensions/mssql/src/diagnostics/diagnosticsCore.ts
```diff
@@ -181,6 +182,7 @@ export class DiagnosticsCore {
         }
         this.removeSink(sink.id);
         this.sinks.push(sink);
+        this.notifySinkStateChanged();
     }
```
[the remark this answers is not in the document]
kburtram: Agreed and fixed in 9cdce6323: `addSink` now detaches the old instance through a private `detachSink` (no intermediate inactive notification) and `notifySinkStateChanged` fires only when `anySinkActive` actually changed, so listeners see transitions, never churn — a second sink or a same-id replacement produces no notification at all, and `WebviewBaseController` no longer reschedules its enable timers on replacement. The unit test now covers both cases and still expects `[true, false]`.

(source: Fable)

Benjin (dismissed):

Benjin (approved):

PR #22842, opened 2026-08-28
Improve Fluent Result Grid behavior parity

aasimkhan30:
## Description

- Make Ctrl/Cmd row-number clicks add partial rows, and remove only fully selected rows.
- Make Shift row-number clicks select full rows from the active row.
- Add Ctrl/Cmd and Shift selection behavior to column headers.
- Restore cell double-click selection for all visible cells in the row.
- Split row-wide selections around hidden columns so SlickGrid accepts every range.
- Exclude hidden columns from row selection, Select All, copy, and selection payloads.
- Keep the first visible data cell active after row-wide actions.
- Fit columns on resize-handle double-click using fetched result rows.
- Allow fit-to-content to grow or shrink and persist the new width.
- Build filter-menu values from rows allowed by other active filters.
- Map Save As rows back to source rows after sorting or filtering.
- Map copy, save, summary, and persisted selections through reordered columns.
- Support selection mapping when the row-number column is disabled.
- Navigate Home, End, and Shift+Arrow through visible data columns only.
- Require exact modifiers for fallback Tab, Select All, Shift+Arrow, and Shift+F10 shortcuts.
- Copy all visible data columns when no explicit selection exists.
- Return no copy range for an empty grid.
- Publish summaries for every selected range using source-row mappings.
- Add focused tests for selection, keyboard, copy, save, summaries, filters, and autosizing.

## Validation

- Full MSSQL build passes.
- 76 Fluent Result Grid tests pass.
- Changed-file ESLint passes.
- Workspace lint is blocked by existing CRLF errors outside this change.

## Code Changes Checklist

- [x] Code follows contributing guidelines
- [x] No known regressions or UX breakage


ghiscoding:
@aasimkhan30 I'm not sure if you've noticed but I did ship Slickgrid-React [v10.10.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v10.10.0) including multi-range selections and fixes for all the issues you've opened. So should your code, especially range selections, change after getting the most recent versions? I would assume this should simplify the code in some areas. I also wonder why is the double-click to resize by content not working for you? Is it because you're using your own custom DataView provider? If so, I need to cancel the resize when that happens and I'm also curious to see if it's fixable on my side or not. 

Side note, I did catch a new problem with multi-range selection when using row selections, but I think the multi-range for cell selections is correct. You can try the multi-range in the updated [Example 48](https://[ghiscoding.github.io/slickgrid-react-demos/#/example48](https://ghiscoding.github.io/slickgrid-react-demos/#/example48) but make sure to click the checkbox "Enable multi-selection" below 1st grid (I need to make it more visible)

Anything else, feel free to open any new issues if it help to fix them natively like multi-range selections (I think it profits every user of the lib, so that one totally made sense)

aasimkhan30:
@ghiscoding, we have a 7-day cooldown period before we can use a new package version in our repo. I’ll bump the package and fix code sometime next week. I’ll also test against the latest package before filing any additional issues on your end.

extensions/mssql/src/webviews/common/FluentResultGrid/internal/fluentResultGridKeyboard.ts
```diff
@@ -195,11 +206,21 @@ export function getFluentResultGridKeyboardAction(
...
+    ) {
         return { kind: "moveFocus", forward: false };
     }
 
-    if (!event.shiftKey && event.code === "Tab") {
+    if (
+        !event.shiftKey &&
+        !isFluentResultGridMetaOrCtrlKeyPressed(event) &&
+        !event.altKey &&
+        event.code === "Tab"
```
lewis-sanchez, line 222: I don't think these key codes will ever change, but It might be good to consider declaring the strings as constants at the top. This is more of a nit, so not something you need to take.

lewis-sanchez (approved):

PR #22836, opened 2026-08-27
Add shared metadata service and persistent catalog cache

kburtram:
# Summary

This PR extracts the shared metadata foundation from `dev/refactor` as an independent, direct-to-`main` change. It gives future native language-service and AI-completion work one bounded catalog API without taking a dependency on the Debug Console PR or the later Query Studio/OE v2 feature sets.

The feature remains experimental and inert by default. Persistent metadata is not composed until `mssql.metadataCache.enabled` is explicitly enabled and the window is reloaded.

### Progressive metadata catalog

- Hydrates schemas, objects, columns, primary/unique keys, foreign keys, parameters, and descriptions over the merged SQL Data Plane abstraction.
- Builds auxiliary server- and database-level catalogs and schema-context projections.
- Supports lazy module-definition retrieval so large or sensitive definitions are not fetched as part of ordinary catalog hydration.
- Detects relevant DDL drift and governs refresh/poll behavior with bounded per-database backoff, freshness state, focus suspension, and serverless safeguards.
- Includes a small comment-aware `leadingKeyword` utility required for DDL recognition without importing Query Studio's batch splitter.

### Shared store and lifecycle

- Adds a shared `MetadataStore` with key-correct leases, server/database lifecycle management, bounded idle LRU behavior, and deterministic catalog replacement.
- Adds the host-level `MetadataStoreService` composition point used by future language and completion consumers.
- Avoids creating the persistent coordinator or maintenance timer when the feature is disabled; enabled maintenance is unref'd and disposed with the controller.

### Persistent cache

- Adds a versioned gzip codec and manifest with content hashes and validation.
- Uses contained, filesystem-safe hashed path segments rather than raw server/database identifiers.
- Handles incomplete, corrupt, incompatible, and untrusted snapshots defensively, including quarantine/fail-closed paths.
- Enforces maximum age and total-size limits with LRU maintenance.
- Supports explicit offline and safe-stale decisions while surfacing snapshot age.
- Persists structural catalog data only. It does **not** persist descriptions, module definitions, query text, result values, credentials, tokens, connection strings, or other secrets.

### Authentication/profile boundaries

- Adds stable non-secret profile fingerprints for store/cache identity.
- Defers password and access-token acquisition until a real session open requires them.
- Keeps credentials out of metadata state, diagnostics, manifests, and cache payloads.

### Commands, settings, and localization

- Adds default-off `mssql.metadataCache.enabled` and bounded age/size/offline settings.
[...]

PR #22860, opened 2026-08-31
Add native SQL language service foundation

kburtram:
# Summary

This PR extracts the native SQL language-service foundation from `dev/refactor` as the fifth independently reviewable unit in the refactor migration. It is based directly on current `main`, including the merged shared metadata service/cache work from PR #22836.

The change is intentionally a source-level foundation: it adds the pure language engine, metadata adapters, scripting support, and focused tests without registering a user-visible command, setting, provider, controller, or activation path. Query Studio and AI-completion integration will compose these APIs in later private-preview-gated PRs.

### Native SQL language core

- Public request/response and engine contracts
- SQL lexer, batch segmenter, cursor expectation parser, and structural sketch parser
- Binding, database context, name-chain, overlay, quoting, fuzzy matching, and immutable text snapshots
- Generated SQL keyword, built-in function, default-completion, snippet, and system-catalog data

### Language features

- Completion
- Diagnostics
- Hover
- Signature help
- Go to definition
- Folding ranges
- Document symbols

### Hosting and scheduling

- Native engine orchestration
- Consumer-neutral engine router
- Version-aware request scheduler and stale-result suppression
- Diagnostic journal integration using the accepted shared diagnostics contract
- Scripting host used by definition generation

### Metadata integration

- Adapter over the shared metadata-service interfaces now in `main`
- Catalog, fixture, null, and system-catalog providers
- Database-context-aware catalog views and test fixtures

### SQL scripting foundation

- Shared scripting API and writer
- CREATE TABLE, DML template, and module emitters
- Scripting service used by language definition workflows

### Observability

- Consumer-neutral `sqlLanguage.route` diagnostics events
- `sqlLanguage` feature classification in performance telemetry
- Existing observability-registry prefix contracts cover `sqlLanguage.*` and `sqlScripting.*`

## Private-preview boundary

This PR does **not** add or change:

- `package.json` commands, menus, settings, activation events, or defaults
- controller/main-extension registration
- a VS Code language-provider activation path
- Query Studio UI or bridges
- AI-completion behavior
- existing Object Explorer or language-service behavior

The new source is not reachable from the extension entry bundle in this PR (`extension-metafile.json` contains no `sqlLanguage` or `sqlScripting` inputs). A later integration PR must explicitly compose it behind the umbrella experimental/private-preview gate and the appropriate feature-specific gate before it can affect users.

## Deliberately deferred

- Query Studio bridge/facade and UI integration
- AI completions
- `SqlDocumentService` and service-client integration
- batch-execution splitting and sqlcmd preprocessing
- query execution helpers
- migration of existing public language providers to the native engine

Keeping these out makes this PR an independently testable language-library boundary rather than coupling it to a particular preview UX.

## Validation

Validation was completed after merging the latest `origin/main` (`95cf790e5`) into the branch:

- Full monorepo build: passed
- Full monorepo lint: passed
- Localization extraction/pre-commit validation: passed; no PR localization changes
- Focused hosted SQL language/scripting tests: 675 passed, 0 failed
- Focused observability-contract tests: 7 passed, 0 failed
- Broad hosted mssql unit suite: 4,710 passed, 0 failed, 10 skipped
  - Excluded only the two documented current-main Windows baseline suites: `Utility Tests - Path handling` and `QueryCompletionSoundService`
- SQL Database Projects hosted suite: 205 passed, 0 failed, 6 skipped
- Online production VSIX packaging: passed with released SQL Tools Service `6.0.20260827.1`
  - `mssql-1.46.0.vsix`: 1,542 files, 99.45 MiB
  - Package audit confirmed the new language/scripting source is not included as an extension bundle input

## Refactor sequence

This is PR05 in the extraction sequence. PR01/PR02 and the metadata-service PR04 are already in `main`; the Debug Console PR03 remains an independent UX iteration. The next extraction will add a gated consumer/integration layer rather than exposing this foundation by itself.

(Source: GPT-5.6 Sol)


extensions/mssql/src/sqlScripting/scriptWriter.ts
```diff
@@ -0,0 +1,102 @@
...
+            if (text.charCodeAt(i) === 10 /* \n */) {
+                this.line++;
+                this.character = 0;
+            } else if (text.charCodeAt(i) !== 13 /* \r counts with its \n */) {
+                this.character++;
+            }
+        }
+        this.offset += text.length;
+        return this;
+    }
```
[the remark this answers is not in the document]
kburtram: Valid in principle, narrow in practice — worth the one-line hardening when you batch fixes. Reachability check: `ScriptWriter` only ever sees emitter-controlled text (always `
`), catalog identifiers via `quoteIdentifier`, and fidelity notes already CR-folded by `sanitizeCommentText`; stored module definitions — the one input that realistically carries CR-only endings — bypass the writer entirely (`moduleEmitter` returns the stored text with a head rewrite and records no anchors). So the only live vector is an object/column NAME containing a bare CR: legal sysname, pathological in practice, but then every later anchor on that script drifts. Suggested fix: treat a lone `
` (not followed by `
`) as a newline in `append()` and mirror it in `withHeader`'s line count.

_(source: Fable)_

extensions/mssql/src/sqlScripting/scriptingService.ts
```diff
@@ -0,0 +1,390 @@
...
+                    : this.unavailable(request, info, "unsupported", [
+                          "tables support create scripting only (no stored definition to rewrite)",
+                      ]);
+            }
+            if (MODULE_KINDS.has(info.kind)) {
+                return this.scriptModule(request, info, operation);
+            }
+            return this.unavailable(request, info, "unsupported", [
+                "synonym targets are not hydrated — definition scripting unavailable",
+            ]);
```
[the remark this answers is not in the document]
kburtram: Disagree — the message is accurate as written. It says synonym *targets* are not hydrated, which is exactly the shipped state of the shared metadata model on main: synonym objects hydrate identity-only (readiness `synonyms: "lite"` — the explicit PR #22836 disposition), and `sys.synonyms.base_object_name` is deferred to a later enrichment, so there is no definition to script from. The branch is only reachable for kind `synonym` (tables and all module kinds are handled above it). If anything, tighten the wording to name the readiness fact — e.g. "synonym definitions require target hydration, which is deferred (metadata readiness: synonyms=lite)" — but no code change is needed.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/host/nativeEngine.ts
```diff
@@ -0,0 +1,932 @@
...
+                    expectationKind: { raw: "none", cls: "diagnostic.metadata" },
+                    expectationConfidence: { raw: "high", cls: "diagnostic.metadata" },
+                    itemCount: { raw: 0, cls: "diagnostic.metadata" },
+                    suppressed: { raw: "emptySpace", cls: "diagnostic.metadata" },
+                });
+                return Promise.resolve(emptyCompletionResult());
+            }
+            const target = this.locateStatement(analysis, req.text, offset);
+            if (target === undefined) {
+                // Empty document / between statements: statement start.
```
kburtram, line 298: **[A] The `locateStatement === undefined` fallback bypasses comment/string suppression — typing inside a trailing `--` comment gets the full 558-item keyword popup.**

`locateStatement` (line 233) returns `undefined` whenever the text between the last statement's end and the cursor is non-whitespace — which is exactly where trailing comments live. This route then hardcodes `context: { kind: "statementStart" }` and never calls `completionExpectationAt`/`classifyContext`, so the `LineComment`/`StringLiteral`/`SqlCmdDirective` suppression in `classifyContext` (`context.ts:245-260`) is skipped entirely. Measured against the compiled PR code (fixture provider, `trigger: "invoke"` — same path quick-suggestions take):

```
SELECT 1 -- comment ¦here        → 558 items   (should be 0, reason "comment")
SELECT 1 --¦                     → 558 items
SELECT 1 -- comm¦ent\nSELECT 2   → 4 items     (keyword-filtered on "comm" — same bypass)
SELECT 1 /* unclosed ¦           → 558 items   (unterminated block after the statement)
SELECT /* blah ¦ */ 1            → 0 items ✓   (inside a statement span → suppression works)
SELECT 'abc ¦def'                → 0 items ✓
```

With `editor.quickSuggestions` on, this means a completion popup on nearly every word typed into an end-of-line comment — the most common comment kind. Fix: before taking the statement-start fallback, run the same suppression check the main route gets (e.g. `completionExpectationAt(text, analysis.lexed.tokens, emptySketch(offset), offset)` and return `emptyCompletionResult()` when `suppressReason` is set, keeping the span fields you already emit for `expectationSuppressed`). A test in `sqlLanguageCompletionExpectation.test.ts` for "cursor in trailing line comment" and "cursor after an unterminated block comment" would pin it — the existing suite covers only in-statement comment positions, which is why this survived 682 green tests.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/data/systemObjectCatalog.ts
```diff
@@ -0,0 +1,722 @@
...
+            "memory_object_address",
+            "cacheobjtype",
+            "objtype",
+            "plan_handle",
+            "pool_id",
+        ],
+    },
+    {
+        schema: "sys",
+        name: "dm_exec_sql_text",
```
kburtram, line 133: **[C] `sql_handle`/`plan_handle` are listed as columns of these two TVFs, but they are the *input parameter*, not result columns.**

Verified against a live SQL Server 2025 — the real result columns are:

```
sys.dm_exec_sql_text(@handle):   dbid, objectid, number, encrypted, text
sys.dm_exec_query_plan(@handle): dbid, objectid, number, encrypted, query_plan
```

so completion/hover would offer `est.sql_handle` on `CROSS APPLY sys.dm_exec_sql_text(r.sql_handle) est`, which error 207s. Suggest replacing the two entries' column lists with the real ones (and `text`/`query_plan` are the ones people actually want). For what it's worth, I validated every other (object, column) pair in the curated catalog against the live server — these two are the only misses, and all 157 builtin function names are genuine.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/features/diagnostics.ts
```diff
@@ -0,0 +1,2062 @@
...
+            if (isFunctionCall) {
+                const fnName = parts[parts.length - 1].toUpperCase();
+                if (FIRST_ARG_OPAQUE_FUNCTIONS.has(fnName)) {
+                    skipNextChain = true;
+                }
+                i = j + 1;
+                continue;
+            }
+
+            const skip =
```
kburtram, line 1739: **[A] `SELECT alias = expression` gets a false 207 — and `WHERE CURRENT OF cursor` a false 207 — from the same prev-keyword skip list.**

The chain scanner skips a head only when `prev.keyword === undefined` or prev ∈ {AS, FOR, COLLATE}. Verified against the compiled engine (real catalog provider):

```
SELECT Total = SUM(C01) FROM s0.T000001;        → mssql(207) Invalid column name 'Total'
SELECT Grand = C01 + C02 FROM s0.T000001;       → mssql(207) Invalid column name 'Grand'
UPDATE s0.T000001 SET C01 = 1 WHERE CURRENT OF cur;  → mssql(207) Invalid column name 'cur'
```

`alias = expr` is everywhere in legacy T-SQL, so this squiggles ordinary scripts on every keystroke. Fix: treat a chain head whose `afterChain` is `=` in a select-list context as an alias definition (skip), and add `OF` to the skip-after-keywords set. Both need honesty-corpus entries — the suite has no alias-equals or cursor case today.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/features/diagnostics.ts
```diff
@@ -0,0 +1,2062 @@
...
+                            `(did you mean ${joined}?).`,
+                        "mssql(102)",
+                    );
+                    recovered = true;
+                    i = nextIndex;
+                    continue;
+                }
+            }
+
+            if ((current === "GROUP" || current === "ORDER") && next !== undefined) {
```
kburtram, line 797: **[A] `WITHIN GROUP (ORDER BY …)` gets a false 102 "Expected BY after GROUP." and the whole statement's binder tier is then suppressed.**

`checkSelectSyntaxRecovery` fires for any depth-0 `GROUP`/`ORDER` not followed by `BY`. In `WITHIN GROUP (ORDER BY …)` the aggregate's parens are already closed, so `GROUP` sits at depth 0 followed by `(`. Verified:

```
SELECT STRING_AGG(C01, ',') WITHIN GROUP (ORDER BY C01) FROM s0.T000001;
  → mssql(102)@0:35 Expected BY after GROUP.
```

`STRING_AGG`/`PERCENTILE_CONT`/`PERCENTILE_DISC` are everyday modern T-SQL. Root cause is shared with the `AT TIME ZONE` comment: `WITHIN` has no entry in `keywords.generated.ts`, so the recovery check can't recognize the construct. Fix: exempt `GROUP` when the previous significant token is `WITHIN` (and add `WITHIN` as a contextual keyword); add a `WITHIN GROUP` case to the honesty corpus (zero hits today).

_(source: Fable)_

extensions/mssql/src/sqlLanguage/features/diagnostics.ts
```diff
@@ -0,0 +1,2062 @@
...
+            runReported = false;
+            runWords = [];
+        };
+        resetRun();
+
+        const clauseOf = (offset: number): ClauseKind | "residue" => {
+            return innermostClause(s.sketch, offset) ?? "residue";
+        };
+
+        const thresholdOf = (clause: ClauseKind | "residue"): number => {
```
kburtram, line 874: **[A] `AT TIME ZONE` in WHERE/ON/HAVING/ORDER BY gets false 102s from the unit-run tier — `AT`/`ZONE` are missing from the keyword data.**

`AT` isn't in `keywords.generated.ts`, so it lexes as a plain identifier and forms a ≥2 "value unit" run in expression clauses (threshold 2). Verified:

```
SELECT C01 FROM s0.T000001 WHERE C02 AT TIME ZONE 'UTC' >= 1;
  → mssql(102)@0:37 Incorrect syntax near 'AT.'
  → mssql(102)@0:50 Incorrect syntax near ''UTC''.
```

The SELECT list escapes only because its threshold is 3 and `TIME` breaks the run as a type keyword. Fix: add `AT` and `ZONE` (contextual) to the keyword data or the run-breaker set — that also covers `AT TIME ZONE` inside `SELECT CAST(... AT TIME ZONE ...)`. No suite hit for `AT TIME ZONE` anywhere today.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/core/sketch/index.ts
```diff
@@ -0,0 +1,1245 @@
...
+            j = k;
+            expectSource = false;
+            continue;
+        }
+        if (w !== undefined && (JOIN_WORDS.has(w) || w === "APPLY")) {
+            j = b.next(j + 1, endExclusive);
+            expectSource = true;
+            continue;
+        }
+        if (b.punct(j) === ",") {
```
kburtram, line 344: **[A] Multi-column derived-table/VALUES alias lists — `v(a, b)` — invent a phantom table source, producing false 208s on legal SQL.**

In `parseFromSources` the comma check (`punct(j) === "," → expectSource = true`) runs before the `!expectSource` skip, and the alias column list's `(` is stepped token-by-token rather than balance-skipped, so the comma inside `v(a, b)` re-arms source expectation and `b` becomes a "table". Verified:

```
SELECT v.a FROM (VALUES (1, 2), (3, 4)) v(a, b);
  → mssql(208)@0:45 Invalid object name 'b'.
SELECT d.c1 FROM (SELECT C01, C02 FROM s0.T000001) d (c1, c2);
  → mssql(208) Invalid object name 'c2'.  +  mssql(207) Invalid column name 'c1'.
```

Note the knock-on 207: the phantom source also poisons scope, so the *real* derived-table columns stop resolving. The honesty suite covers only the single-column `v(a)` form, which dodges the bug. Fix: after a completed alias, balance-skip a following paren group; add both multi-column forms to the corpus.

_(source: Fable)_

kburtram (commented):
# Review: native SQL language service foundation (PR05)

**Bottom line:** this is a genuinely strong foundation — pure, deterministic, honest about metadata readiness, fast (measurements below), with an unusually disciplined privacy/observability posture and 682 green tests. The architecture decisions (freshness verdict as data, pinned snapshot per request, fire-and-forget hydration kicks, refusal-over-fabrication scripting) are the right ones and verified to work end-to-end against the real PR04 metadata store. What keeps it from "100% correct" is a cluster of **false-positive diagnostics on ordinary T-SQL** — the one failure class a language service cannot afford, and the honesty corpus (which is otherwise excellent) simply doesn't contain these constructs. All are small, localized fixes. Details inline; tiers below.

### What I verified (beyond reading all 59 files)
- **Gates** at `eafbbdc04` in a clean clone: emit ✔, webviews typecheck ✔, eslint 0 errors + prettier ✔ on all 67 changed files; all 15 `sqlLanguage*`/`sqlScripting*` suites + observability contract → **682 passing, 0 failing**. CI fully green at head incl. all platform launches. (Local `tsgo` reports errors only in main-side files byte-identical to `origin/main` — environmental, not this PR.)
- **End-to-end probe** over the compiled PR code: real `MetadataStore` hydrated with the 10,201-object B16 fixture → `CatalogLanguageMetadataProvider` → `NativeSqlLanguageEngine`:

| request | p50 | p95 |
|---|---|---|
| completion, FROM-target (10,201 objects) | 4.0 ms | 8.7 ms |
| completion, member access / alias columns / hover | 0.1 ms | ≤1 ms |
| completion at end of a 2,000-statement script | 0.8 ms | 35.6 ms |
| diagnostics, 2,000-statement script | 68.5 ms cold, ~0 cached | — |
| documentSymbols / folding, 2k lines | 0.3 / 1.0 ms | — |

  Malformed-input sweep (unterminated string/bracket/comment, unbalanced parens, empty doc): no throws; 5,000 nested parens: 13 ms (depth-guarded, iterative).
- **Freshness honesty, observed live:** unknown object on a `ready` catalog → `mssql(208)`; same text with `metadataFreshness: "notValidated"` → 0 diagnostics, `suppressed: { metadataNotValidated: 1 }` — exactly the api.ts contract.
- **Every false-positive claim below was reproduced against the compiled engine** with a real catalog before being posted (codes and positions in the inline comments).
- **Generated data vs a live SQL Server 2025:** every (object, column) pair in `systemObjectCatalog.ts` checked — 2 wrong entries (inline comment), all 43 objects and the other ~500 pairs exist; all 157 builtin function names are genuine.

### Tier A — false positives on ordinary SQL (fix before integration)
1. **Completion suppression bypassed after the last statement** — typing in a trailing `--` comment gets the full 558-item popup (`nativeEngine.ts:280`).
2. **`SELECT alias = expr`** and **`WHERE CURRENT OF`** → false 207 (`diagnostics.ts:1710`).
3. **`WITHIN GROUP (ORDER BY …)`** → false 102 + binder tier suppressed (`diagnostics.ts:787`).
4. **`AT TIME ZONE`** → false 102s; `AT`/`ZONE`/`WITHIN` missing from keyword data (`diagnostics.ts:856`).
5. **`v(a, b)` derived/VALUES column aliases** → phantom source, false 208 + knock-on 207 (`sketch/index.ts:337`).
6. **NBSP in pasted SQL** → identifier swallows the next keyword, false 208 (`lexer.ts:84`).
7. **`UPDATE TOP (n)`** → false `208 'TOP'`, SET clause lost (`sketch/index.ts:1042`).
8. **`db..tbl`** → false `208 'db'`, no columns (`sketch/index.ts:199`).

### Tier B — recommended
GOTO labels 2812-as-error + synonym EXEC-less 2812 (`diagnostics.ts:1147`); `IS NOT DISTINCT FROM` clause mis-split (`sketch/index.ts:686`); CREATE VIEW/PROC invisible to the overlay (`overlay.ts:107`); qualified script-created tables miss the overlay (`binder.ts:107`); INSERT/UPDATE column completion bypasses the overlay for `#temp` (`completion.ts:1014`); CS-catalog dedupe folding (`completion.ts:279`); `alias.` expand inserts `o.o.…` (`completion.ts:536`); provider full-scan + search-cap-before-filter (`catalogProvider.ts:354`); `caseSensitive ?? false` vs fail-closed resolve (`catalogProvider.ts:156`); `(version, length)` analysis cache — collision hit in practice (`nativeEngine.ts:130`); router timeout inert for sync features + no router tests (`router.ts:188`); DML placeholder `/* */` breakout (`dmlTemplateEmitter.ts:37`); opener-at-EOF drops the token (`lexer.ts:288`); GO-parity target `batchSplitter.ts` doesn't exist + corpus missing (`lexer.ts:17`); real catalog provider untested against a real snapshot (`sqlLanguageProviders.test.ts:110`).

### Tier C — notes (no inline comment)
`dm_exec_sql_text`/`dm_exec_query_plan` curated columns (inline, small); `DECLARE @x in|` drops the typed type prefix (`context.ts:370`); money literals unlexed (`$12.5`, `€100`); `SelectItem.alias` conflated with trailing dotted part; overlay folds case-insensitively even on CS catalogs; `EXECUTE AS USER = 'x'` / `DECLARE … CURSOR FOR` / `IF … ELSE` segmentation quirks (tolerant direction); caret-after-`*/` boundary semantics differ from strings (currently masked by Tier A #1 — re-check after that fix); router doesn't record the bridge outcome after a native-failure fallback; `LEN`/`DATALENGTH`/`CHARINDEX` docs omit the bigint-for-max caveat; missing common scalars (SOUNDEX, DIFFERENCE, COT/ACOS/ASIN/ATAN/ATN2, PARSENAME, TYPE_ID) degrade to keyword completion; duplicate `"hierarchyid"` default completion; every request re-lexes the full document (linear, cached per version — fine for v1; the incremental line-state hooks exist unused).

### Verified OK (highlights — each checked with file:line evidence)
Total-coverage lexing with `[a]]b]`/`"a""b"`/`N'…'''`/nested block comments; GO rules match the documented splitter shape (line-start, counts, never inside constructs); CRLF/LF/lone-CR line math and UTF-16 position math incl. surrogate pairs; iterative, depth-guarded sketch — no throw found on any malformed input; alias/CTE/temp-table/correlated-subquery scoping; catalog case rule honored in binder and all features (fold only when CI); determinism — no `localeCompare`/`Date.now`/`Math.random` anywhere in results; readiness gating throughout (T2 needs objects ready+full; lite→partial; USE-switch, cross-DB, linked-server, dynamic SQL all suppress-and-count); per-feature degradation without throwing on absent/offline metadata; no lease acquire/leak anywhere in the language layer (leases stay with the facade); completion/hover never block on the network — the only async provider read is `definition`'s lazy module fetch; scheduler version-aware cancellation and coalescing correct incl. the overlapping-run guard; every executable identifier in the scripting emitters bracket-escaped via `quoteIdentifier` (no UDT double-quoting); module CREATE↔ALTER rewrites token-level and honest (encrypted/permission/notLoaded refusals); strict-host scripting bounded and refusal-based; hover renders only catalog facts (never module text or column default/computed definitions); route/diagnostics/scripting spans carry counts and enums only — completion prefix rides as `user.text` and object parts as `object.name` through the shared redaction chokepoint, so nothing is plaintext under the default policy.

### Copilot threads
- `scriptWriter.ts:39` (CR-only): valid in principle, narrow in practice — module text bypasses the writer and fidelity notes are CR-folded, so the only vector is an identifier containing a bare CR; the one-line hardening (treat lone `\r` as a newline) is worth batching. Replied there.
- `scriptingService.ts:123` (synonym message): disagree — the message says synonym *targets* aren't hydrated, which is exactly the shipped metadata state (`synonyms: "lite"`, target hydration deferred in PR #22836). Optional wording tweak only. Replied there.

### Merge readiness
The boundary claim checks out (no commands/settings/activation paths; not reachable from the extension bundle), CI is green, and none of the findings is architectural. I'd fix the eight Tier A items (plus the honesty-corpus entries that pin them) before the integration PR makes this user-visible — they're each a few lines, and they're precisely the class of bug that erodes trust in diagnostics on day one. The B items can land here or in the follow-up; the missing-test list (GO parity corpus, real-provider suite, scale fixture, CRLF/TextSnapshot units) is the highest-leverage hardening.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/core/sketch/index.ts
```diff
@@ -0,0 +1,1245 @@
...
+            default:
+                j++;
+                break;
+        }
+    }
+    return endExclusive;
+}
+
+/** Find where the current clause ends: next clause keyword at depth 0. */
+function findClauseEnd(b: SketchBuilder, i: number, endExclusive: number): number {
```
kburtram, line 715: **[B] `IS [NOT] DISTINCT FROM` (SQL 2022) mis-splits the WHERE clause at the second `FROM` → false 208.**

`findClauseEnd` stops at any depth-0 `FROM`. Verified:

```
SELECT C01 FROM s0.T000001 WHERE C01 IS NOT DISTINCT FROM C02;
  → mssql(208) Invalid object name 'C02'.
```

Cheap fix: don't treat `FROM` as a clause boundary when the previous significant token is `DISTINCT` (the only place `DISTINCT FROM` is legal). Worth a corpus entry — SQL 2022's null-safe comparison is spreading fast through new code.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/features/diagnostics.ts
```diff
@@ -0,0 +1,2062 @@
...
+            }
+            if (runCount >= thresholdOf(clause)) {
+                flagRun(unitStart, unitEnd);
+            }
+            i = unitEndExclusive;
+        }
+        return reportedAny;
+    };
+
+    const resolveExecLessProcedureHead = (
```
kburtram, line 1165: **[B] EXEC-less-call analysis: GOTO label definitions report a 2812 *error*, and calls through procedure synonyms report a false 2812.**

Verified:

```
myLabel:
SELECT C01 FROM s0.T000001;
GOTO myLabel;
  → mssql(2812)@0:0 Could not find stored procedure 'myLabel'.
```

`label:` lexes as identifier + Unknown token and falls into the exec-less path; with objects `ready`/mode `full` it resolves `notFound` → 2812 at **error** severity — the only error-severity false positive found, on a legal control-flow construct. Separately, `resolveExecLessProcedureHead` maps a resolved **synonym** to `notFound` (line 1180-1185): a synonym aliasing a proc is callable EXEC-less, and since synonym readiness is `lite` by contract the honest verdict is `unknown`, not 2812. Fix: skip heads immediately followed by `:` (label definition), and return `unknown` for synonym resolutions; add corpus entries for both.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/core/overlay.ts
```diff
@@ -0,0 +1,184 @@
...
+                    parts: [decl.name],
+                    kind: "tableVariable",
+                    columns: decl.tableColumns ?? [],
+                    fromStatement: ordinal,
+                    batchIndex,
+                });
+            }
+        }
+
+        // CREATE TABLE — temp or script-local real table. ALTER TABLE never
```
kburtram, line 107: **[B] Objects created by the script itself via CREATE VIEW/PROC/FUNCTION are invisible to the overlay — deployment scripts get false 208s on their own objects.**

The overlay records `createdTable` and `selectInto` only; module DDL is ignored. Verified:

```
CREATE VIEW dbo.MyV AS SELECT C01 FROM s0.T000001;
GO
SELECT C01 FROM dbo.MyV;
  → mssql(208)@2:16 Invalid object name 'dbo.MyV'.
```

Migration/deployment scripts (create view → use view) are a primary editor workload, and CREATE TABLE already gets the speculative treatment, so this is an inconsistency rather than a design decision. Suggest recording module headers (`moduleHeader` is already sketched) as suppression-only overlay entries — same safe direction as `DROP TABLE`'s documented handling.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/features/completion.ts
```diff
@@ -0,0 +1,1158 @@
...
+                detail: `${writable.length} writable columns`,
+                filterText: "all",
+            },
+            "",
+            100,
+        );
+    }
+}
+
+function columnsOfParts(
```
kburtram, line 1021: **[B] INSERT column lists and UPDATE SET completion bypass the overlay — `#temp` tables and table variables get nothing.**

`addInsertColumns`/`addUpdateSetColumns` fall back to `columnsOfParts`, which calls `pinned.resolveObject` directly, unlike diagnostics/hover/definition which use the overlay-aware `resolveNameParts`. After `CREATE TABLE #tmp (a int, b int)`, `INSERT INTO #tmp (` offers nothing and misreports `columnsNotReady`. The definition suite proves the shared path handles this (`UPDATE #t SET a` navigates fine) — completion just doesn't use it. Fix: route both through `resolveNameParts`; add a `#temp` INSERT-columns completion test (the suite only covers catalog targets).

_(source: Fable)_

extensions/mssql/src/sqlLanguage/features/completion.ts
```diff
@@ -0,0 +1,1158 @@
...
+        ) {
+            start--;
+            continue;
+        }
+        break;
+    }
+    return start < offset ? text.slice(start, offset) : "";
+}
+
+function completionDedupeKey(item: Omit<SqlCompletionItem, "sortText">): string {
```
kburtram: **[B] `completionDedupeKey` lowercases labels unconditionally — case-distinct objects on a case-sensitive catalog are silently dropped.**

On a CS catalog with `dbo.Foo` and `dbo.foo`, the second table-source item is discarded by the fold-blind dedupe, while every other fold in this file honors `env.caseSensitive`. Fix: fold the key only when the catalog is case-insensitive. The CS tests cover aliases/columns but not case-distinct sibling objects — worth one assertion.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/features/completion.ts
```diff
@@ -0,0 +1,1158 @@
...
+                    {
+                        label: col.name,
+                        kind: "column",
+                        insertText: quoteIdentifier(col.name),
+                        detail: columnDetail(col.typeDisplay, col.isPrimaryKey, col.fromLabel),
+                        filterText: col.name,
+                    },
+                    prefix,
+                    col.isPrimaryKey === true ? 10 : 0,
+                );
```
kburtram, line 539: **[B] "Expand columns" offered after `alias.` inserts alias-qualified names *after the dot* — accepting it yields `o.o.OrderId, o.o.CustomerId, …`.**

`addStarExpansion` qualifies with the alias whenever the source has one, and `replaceRange` is only set when a real `*` token exists. Verified on the compiled engine at `SELECT o.` (dot-triggered):

```
insertText = "o.C01, o.C02, o.C03, …"   replaceRange = null
```

so the default replace (the empty typed prefix after the dot) produces the doubled qualifier. The `SELECT o.*` path is correct. Fix: when the expansion is offered at a bare `alias.` position, either emit unqualified column names or set `replaceRange` to cover the `alias.` prefix.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/provider/catalogProvider.ts
```diff
@@ -0,0 +1,412 @@
...
+     * degrade to F1 with an explicit fidelity note instead of a wrong claim.
+     */
+    getKeyConstraints(ref: LangObjectRef): readonly LangKeyConstraint[] | undefined {
+        if (this.snapshot.readiness.keys !== "ready") {
+            return undefined;
+        }
+        return this.snapshot.getKeyConstraints(ref.objectId);
+    }
+
+    searchObjects(query: ObjectSearchQuery): readonly LangObjectInfo[] {
```
kburtram, line 358: **[B] Provider object queries: full-catalog scan+sort per empty-prefix request, and the prefix-search cap is applied before kind/schema filtering.**

Two related issues in `queryObjects`:
1. Empty-prefix (post-`FROM `, post-`Sales.`) requests call `snapshot.listObjects(...)` — a full O(N) walk allocating `ObjectInfo` per object plus an O(N log N) `ordinalCompare` sort, per request, with no cache across keystrokes. Measured end-to-end on a 10,201-object catalog: p50 4.0 ms / p95 8.7 ms — acceptable today, but it is the per-keystroke hot path and linear in catalog size; a cached sorted view per (generation, kinds, schema) would make it O(result).
2. `snapshot.search(prefix, limit * 2)` caps at 1,200 folded-name matches across **all** kinds and only then filters by `kinds`/`schema` — a catalog with >1,200 `usp_*` procedures makes `FROM usp…` return zero table/view candidates even when matching views exist. Silent wrong-by-omission truncation; pass the kind filter into the search (or over-fetch per kind).

The latency test's catalog has 9 objects, so neither is visible in CI — a ≥5k-object completion/diagnostics fixture test would pin both.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/provider/catalogProvider.ts
```diff
@@ -0,0 +1,412 @@
...
+        // CREATE OR ALTER: SQL Server 2016 SP1+ programmability feature
+        // (design 05 §1.2: gate by server capability, not compat level).
+        const createOrAlter =
+            version !== undefined &&
+            (version.major >= 14 || (version.major === 13 && version.build >= 4001));
+        const dropIfExists = version !== undefined && version.major >= 13;
+        return {
+            currentDatabase: this.host.currentDatabase(),
+            defaultSchema: snapshot?.defaultSchema ?? "dbo",
+            caseSensitive: snapshot?.caseSensitive ?? false,
```
kburtram: **[B] `caseSensitive: snapshot?.caseSensitive ?? false` defaults open while the snapshot's `resolveName` fails closed — a failed H0 probe turns wrong-case references into false 208s.**

When the env probe failed but objects are `ready`/`full`, the features fold case-insensitively (`?? false`) yet `CatalogSnapshot.resolveName` with `caseSensitive === undefined` requires a raw match — so a wrong-case reference folds fine everywhere in the features and still resolves `notFound` → `mssql(208)` on what is actually a case-insensitive server. The two layers should agree: propagate the undefined ("unknown") state and have the diagnostics gate suppress-and-count (like `metadataNotValidated`) rather than report, until the case rule is known.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/host/nativeEngine.ts
```diff
@@ -0,0 +1,932 @@
...
+     * Pin one metadata generation for a request, decorated with the static
+     * system-object catalog fallback (provider/systemCatalogView): sys /
+     * INFORMATION_SCHEMA names resolve for every feature through the one
+     * resolution path, and live metadata always wins.
+     */
+    private pinView(): IPinnedMetadataView {
+        return withSystemObjectCatalog(this.provider.pin());
+    }
+
+    private analyze(text: string, version: number): DocumentAnalysis {
```
kburtram, line 132: **[B] `analyze` caches on `(version, text.length)` without comparing content or document identity — I hit the collision in practice within minutes of probing.**

The cache is sound only under a strict "one engine instance per document, monotonically increasing versions" invariant that nothing in this file enforces. While probing, two same-length texts issued with the same version number served **the previous document's diagnostics verbatim** (`SELECT d.c1 FROM (SELECT C01, C02 FROM s0.T000001) d (c1, c2);` results returned for a different statement). Any future host that reuses an engine across documents, resets versions, or replays a version after undo hits this silently. A cheap content guard (store the text reference and compare `cached.text === text`, or a hash) removes the trap; also worth asserting the invariant in `setProvider` (which clears `diagnosticsMemo` but retains this cache).

_(source: Fable)_

extensions/mssql/src/sqlLanguage/host/router.ts
```diff
@@ -0,0 +1,253 @@
...
+            span.fail(error);
+            throw error;
+        }
+    }
+
+    private async invokeNative<T>(
+        feature: SqlLanguageFeature,
+        invoke: (engine: SqlLanguageFeatureEngine) => Promise<T | undefined>,
+    ): Promise<{ value?: T | undefined; failed?: boolean; timedOut?: boolean }> {
+        const timeoutMs = this.options.nativeTimeoutMs ?? 5000;
```
kburtram, line 192: **[B] The native timeout/circuit-breaker race is inert for every synchronous feature — and the router has no test file at all.**

`Promise.race([nativeCall, timeout])` can only observe the timeout if the native call yields to the event loop. Every native feature except `definition` computes synchronously and returns `Promise.resolve(...)` — `diagnostics()` runs `while (pass.step()) {}` to completion — so the `setTimeout` callback cannot fire until the work already finished: the "nativeTimeout → bridge fallback" branch is unreachable for completion/hover/signature/diagnostics/folding/symbols, and a pathological document blocks the extension host for the full pass with no cutoff. Only thrown errors trip the breaker for sync features. Either document that native features must be self-bounding (and slice the pull-path diagnostics like the scheduler's `diagnosticsPass` already does), or make the engine yield between slices so the timeout is real.

There is no `sqlLanguageRouter.test.ts`: `breakAfterFailures`, the `broken` set, maturity gating, the timeout race, and `resetCircuits` are all untested — this finding would have surfaced immediately under a router test with a slow/throwing native engine.

_(source: Fable)_

extensions/mssql/src/sqlScripting/dmlTemplateEmitter.ts
```diff
@@ -0,0 +1,177 @@
...
+export interface DmlEmitOutput {
+    readonly text: string;
+    readonly anchors: readonly ScriptAnchor[];
+    readonly fidelityNotes: readonly string[];
+}
+
+function placeholder(column: LangColumn): string {
+    const nullability =
+        column.nullable === true ? ", NULL" : column.nullable === false ? ", NOT NULL" : "";
+    return `/* ${column.name} ${column.typeDisplay}${nullability} */`;
```
kburtram: **[B] DML placeholders interpolate identifiers/type names into `/* … */` without `sanitizeCommentText` — a name containing `*/` breaks out of the comment into live SQL.**

`placeholder()` emits `/* ${column.name} ${column.typeDisplay}${nullability} */` and the WHERE builder emits `` ` = /* ${column.typeDisplay} */` `` (line 168). `typeDisplay` for a UDT is `[schema].[name]` verbatim, and object/type names may legally contain `*/` — `[a*/b]` closes the comment early and the trailing text becomes live tokens inside the emitted INSERT/UPDATE template. `sanitizeCommentText` exists precisely for this (and `createTableEmitter` applies it to descriptions at lines 111/190) — these two interpolations just miss it. Same one-liner applies to the `--` refusal texts in `moduleEmitter.ts:88` and `scriptingService.ts:299` (a name containing a newline spills onto a live line). Pathological names, user-reviewed templates — but the fix is free and there's no negative fixture (no column/type containing `*/` in `sqlScripting.test.ts`).

_(source: Fable)_

extensions/mssql/src/sqlLanguage/core/lexer.ts
```diff
@@ -0,0 +1,586 @@
...
+            let i = pos + 2;
+            while (i < length && text.charCodeAt(i) !== 10 && text.charCodeAt(i) !== 13) {
+                i++;
+            }
+            tokens.push({ kind: TokenKind.LineComment, start: pos, end: i });
+            pos = i;
+            atLineStart = false;
+            continue;
+        }
+        if (ch === 47 /* / */ && text.charCodeAt(pos + 1) === 42) {
```
kburtram, line 305: **[B] A string/bracket/comment opener as the last character of the document emits no token — the total-coverage invariant breaks and suppression is lost at exactly that caret.**

The opener branches set `mode`/`constructStart` and advance past the opener; only the *resume* branch ever pushes the construct token (including unterminated-at-EOF). When the opener is the final character, the `while (pos < length)` loop exits first. Verified on the compiled lexer and engine:

```
lex("SELECT '")  → the quote at [7,8) is covered by no token
completion at "SELECT '␣"  → 558 items   (typing the opening quote of a literal at EOF)
```

The coverage invariant is this module's core contract (its own test at `sqlLanguageCore.test.ts:42-52` would fail on these inputs — they're just not in the corpus). Fix: flush the pending construct after the loop, same as the unterminated case; add `"SELECT '"`, `"FROM ["`, `"/*"` to the invariant test.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/core/lexer.ts
```diff
@@ -0,0 +1,586 @@
...
+
+export interface LexResult {
+    readonly tokens: readonly Token[];
+    /** Entry state per line; lineStates.length === lineCount. */
+    readonly lineStates: readonly LineStartState[];
+}
+
+const CODE_STATE: LineStartState = { mode: LineStartMode.Code, blockDepth: 0 };
+
+function isIdentStart(ch: number): boolean {
```
kburtram, line 84: **[A] U+00A0 (NBSP) lexes as identifier-start and swallows the following keyword — SQL pasted from the web/Teams/Word gets false 208s.**

Every `ch > 127` is identifier-start, and the whitespace branch accepts only space/tab/VT/FF, so an NBSP merges with the next word. Verified end-to-end:

```
SELECT * FROM dbo.Orders␠WHERE Status = 1;      (␠ = U+00A0)
  → mssql(208) Invalid object name 'dbo.Orders WHERE'.
```

NBSP in pasted SQL is extremely common (every web page, chat client, and Word doc), and the failure is baffling to the user (the text *looks* identical). Also internally inconsistent with GO handling: `GO_REMAINDER` uses `\s` (matches NBSP after GO) while `precededByLineWhitespaceOnly` rejects ` GO` — and the documented `^\s*GO` splitter shape would accept it, so execution would split where the language service doesn't. Fix: treat Unicode space separators (at minimum U+00A0, ideally the `\s` set minus line terminators) as whitespace in the lexer; add an NBSP fixture.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/core/sketch/index.ts
```diff
@@ -0,0 +1,1245 @@
...
+            } else if (next === "EXEC" || next === "EXECUTE") {
+                exec = parseExec(b, j + 1, endExclusive) ?? undefined;
+            } else if (next === "OUTPUT" || next === "DEFAULT") {
+                // INSERT ... OUTPUT / DEFAULT VALUES — opaque v1.
+            }
+            break;
+        }
+        case "UPDATE": {
+            kind = "update";
+            const name = readNameParts(b, i + 1, endExclusive);
```
kburtram: **[A] `UPDATE TOP (n) target SET …` reads `TOP` as the target — false `208 'Invalid object name TOP'` and the SET clause is lost.**

`readNameParts` accepts `TOP` as the target name; `name.next` then sits on `(`, so the `SET` detection fails and no `setAssignments` clause is recorded. Verified:

```
UPDATE TOP (100) dbo.Orders SET Status = 1;
  → mssql(208) Invalid object name 'TOP'.
completion at "UPDATE TOP (100) dbo.Orders SET ␣" → generic expression items, no target columns
```

DELETE already has an explicit TOP skip (lines 1086-1092); UPDATE (and INSERT) need the same. `UPDATE TOP (n)` is the standard batched-update idiom, so this is an ordinary-SQL false positive plus a completion regression in the same statement.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/core/sketch/index.ts
```diff
@@ -0,0 +1,1245 @@
...
+    let j = b.next(i, endExclusive);
+    if (!b.isNameToken(j)) {
+        return undefined;
+    }
+    const parts: string[] = [];
+    const start = b.tok(j)!.start;
+    let end = b.tok(j)!.end;
+    parts.push(b.namePart(j)!);
+    let k = j + 1;
+    for (;;) {
```
kburtram, line 199: **[A] `db..tbl` (default-schema shorthand) loses the table part — false `208 'Invalid object name db'` and no columns through the alias.**

After `db`, the second `.` isn't a name token, so `readNameParts` returns `["db", ""]` and `parseFromSources` never reads `tbl`. Verified:

```
SELECT * FROM Db..Orders;            → mssql(208) Invalid object name 'Db'.
SELECT x.␣ FROM Db..Orders x         → 0 column items
```

`server.db..table` / `db..table` shorthand is common in ad-hoc scripts. The honesty contract already treats 3-part cross-database names as opaque — the empty middle part just needs to reach that path (accept the empty part in `readNameParts` and let the binder's parts-length gate classify it) instead of truncating the chain.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/core/binder.ts
```diff
@@ -0,0 +1,336 @@
...
+    ctx: NameResolutionContext,
+): SourceResolution {
+    const fold = (value: string): string => (ctx.caseSensitive ? value : value.toLowerCase());
+    const parts = rawParts.filter((p) => p.length > 0);
+    if (parts.length === 0) {
+        return { kind: "opaque", reason: "unknownSketchRegion" };
+    }
+    const last = parts[parts.length - 1];
+    // 1. Overlay objects (#temp, @tablevar, script tables).
+    if (parts.length === 1 || last.startsWith("#")) {
```
kburtram, line 121: **[B] The overlay is consulted only for single-part names and `#temp` — schema-qualified references to script-created tables fall through to the catalog and report opaque/notFound.**

`CREATE TABLE dbo.Staging (Id int); SELECT * FROM dbo.Staging` → `resolveNameParts(["dbo","Staging"])` skips the overlay (which stored the object *with* its qualified parts, `overlay.ts:731-739`) and the catalog misses it. Unqualified `FROM Staging` works — inconsistent, and qualified DDL is the dominant style in deployment scripts. Related (same script-shape family): the overlay records only `createdTable`/`selectInto`, so CREATE VIEW/PROC/FUNCTION-then-use scripts also 208 on their own objects — see the overlay comment. Fix: match qualified parts against the overlay's stored parts before the catalog.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/core/lexer.ts
```diff
@@ -0,0 +1,586 @@
...
+ * of the input is covered by exactly one token (trivia included), spans are
+ * exact for replacement edits and anchors, and keyword-looking identifiers are
+ * lexed as identifiers WITH keyword metadata — T-SQL permits many
+ * keyword-looking identifiers, so keyword-ness is a parser/context decision.
+ *
+ * Line-start states are tracked so incremental re-lexing can resume at any
+ * line whose entry state is unchanged (design §7.1 "track line-start states").
+ *
+ * Purity: no vscode, no node APIs, no imports outside src/sqlLanguage
+ * (lint-enforced). GO recognition follows the execution splitter's rules
```
kburtram: **[B] The GO-parity contract points at `src/sql/batchSplitter.ts`, which does not exist in this PR or on main — and the promised parity corpus test is absent.**

The lexer and segmenter headers (and the core suite header, which says "EXECUTION-SPLITTER PARITY over a GO corpus") anchor the GO rules to a batch splitter that was deliberately left out of PR05, so the central guarantee — "the language service segments exactly where execution splits" — is unverifiable as shipped, and drift can't be caught (the NBSP-before-GO inconsistency in the lexer comment above is exactly that class). Suggest either landing a table-driven GO corpus now against the documented `^\s*GO` shape (cases: `GO 0`, `GO` at EOF without newline, ` GO`, `GO;`, `GO 5 -- c`, GO inside strings/comments/bracket names, `'…\nGO\n…'`) so the splitter PR must conform to it, or rewording the comments until the splitter exists.

_(source: Fable)_

extensions/mssql/test/unit/sqlLanguageProviders.test.ts
```diff
@@ -0,0 +1,199 @@
...
+        expect(salesOnly.every((o) => o.schema === "Sales")).to.equal(true);
+    });
+
+    test("databases and schemas are listed", () => {
+        expect(provider.databases()?.map((d) => d.name)).to.include("FixtureDb");
+        expect(pinned.listSchemas().map((s) => s.name)).to.deep.equal(["Sales", "dbo"]);
+    });
+});
+
+suite("sqlLanguage catalog adapter offline honesty", () => {
```
kburtram, line 133: **[B] The provider that ships — `CatalogLanguageMetadataProvider` over a real `CatalogSnapshot` — is the least-tested one.**

These tests use hand-rolled `handle` fakes (`… as unknown as MetadataCatalogHandle`), and every feature/scripting suite runs on `FixtureLanguageMetadataProvider`. No test builds a `CatalogBuilder` snapshot and asserts the real `SnapshotPinnedView` mapping: `getColumns` PK-name matching (catalogProvider.ts:282-308), `getKeyConstraints` readiness gating (347-352), `resolveObject` result mapping (252-272), `getDefinition` (405-411). That's precisely the drift risk between the language engine's fixtures and the actual PR #22836 contract — my end-to-end probe over the real store worked, but nothing in CI pins it. One suite that hydrates a `MetadataStore` + `FakeBackend` catalog and runs completion/hover/diagnostics through the real provider would close the gap (it also would have caught the empty-prefix scan and search-cap issues flagged in catalogProvider.ts).

_(source: Fable)_

extensions/mssql/src/sqlScripting/scriptWriter.ts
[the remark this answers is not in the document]
kburtram: Fixed in b66db2d8f. ScriptWriter and header anchor accounting now treat CR, LF, and CRLF correctly, with regression coverage for standalone CR and mixed endings. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlScripting/scriptingService.ts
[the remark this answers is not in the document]
kburtram: Fixed in b66db2d8f. The refusal now accurately says the synonym definition text is unavailable through this scripting engine; it no longer claims synonyms are not hydrated. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/host/nativeEngine.ts
kburtram: Fixed in b66db2d8f. Completion now checks the lexical token before the no-statement fallback and suppresses comments, strings, and SQLCMD regions. Regression tests cover trailing comments, trailing strings, and unterminated constructs at EOF. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/data/systemObjectCatalog.ts
kburtram: Fixed in b66db2d8f. Removed sql_handle and plan_handle from the two TVF result shapes and added exact-shape tests for both curated objects. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/features/diagnostics.ts
kburtram: Fixed in b66db2d8f. SELECT alias assignment heads followed by equals are skipped in select-list context, and OF is now skipped for CURRENT OF. Both legal forms have diagnostics regression cases. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/features/diagnostics.ts
kburtram: Fixed in b66db2d8f. WITHIN is contextual keyword data, WITHIN GROUP no longer triggers GROUP BY recovery, and clause scanning preserves the ordered-set expression. Covered by diagnostics tests. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/features/diagnostics.ts
kburtram: Fixed in b66db2d8f. AT and ZONE are contextual keyword entries, preventing value-run false positives without reserving them as identifiers. AT TIME ZONE is covered in the honesty corpus. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/core/sketch/index.ts
kburtram: Fixed in b66db2d8f. A derived or VALUES source now balance-skips its correlation-column alias list, so commas cannot create phantom sources. Both multi-column forms are covered. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/core/sketch/index.ts
kburtram: Fixed in b66db2d8f. Clause scanning recognizes DISTINCT FROM as an expression operator rather than a new FROM clause, with an IS NOT DISTINCT FROM regression test. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/features/diagnostics.ts
kburtram: Fixed in b66db2d8f. Label definitions are excluded from EXEC-less procedure analysis, including the lexer colon-token representation, and resolved synonyms remain unknown rather than becoming false 2812 errors. Both paths are tested. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/core/overlay.ts
kburtram: Fixed in b66db2d8f. CREATE and ALTER module headers now create suppression-only script overlay entries; CREATE OR ALTER segmentation and module-name extraction are also covered so later batches do not report false 208s. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/features/completion.ts
kburtram: Fixed in b66db2d8f. INSERT and UPDATE target-column completion now resolves through the script overlay as well as the catalog. Temp-table cases are covered by completion tests. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/features/completion.ts
kburtram: Fixed in b66db2d8f. Completion deduplication now follows the pinned catalog comparison rule, preserving case-distinct candidates on case-sensitive catalogs while retaining case-folding elsewhere. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/features/completion.ts
kburtram: Fixed in b66db2d8f. Bare alias-dot expansion now emits unqualified names because insertion begins after the dot; explicit star replacement retains qualification. Both paths are covered. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/provider/catalogProvider.ts
kburtram: Partially fixed in b66db2d8f. The correctness issue is real: prefix search now grows its window until filtered results are full or the range is exhausted, with a 1,205-object regression fixture. I am not adding an empty-prefix cache in this iteration: the measured 4.0 ms median and 8.7 ms p95 are within the interactive budget, while a cache adds generation and filter invalidation complexity without fixing incorrect behavior. That optimization should be driven by a demonstrated budget regression. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/provider/catalogProvider.ts
kburtram: Fixed in b66db2d8f. Unknown comparison rules now fail closed to exact spelling and are explicitly marked unknown; diagnostics suppress 207 and 208 metadata claims until the rule is established. Regression coverage pins both lookup and diagnostic behavior. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/host/nativeEngine.ts
kburtram: Fixed in b66db2d8f. Both document analysis and diagnostics memoization now require exact text equality in addition to version and length. A same-version, same-length collision regression test verifies the stale result cannot recur. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/host/router.ts
kburtram: Addressed in b66db2d8f at the appropriate boundary. The router already has preference, maturity, failure-breaker, reset, and failure-fallback tests in sqlLanguageCore.test.ts; I added explicit async-timeout coverage and documented that Promise racing is cooperative. I am not converting every synchronous interactive feature into an artificial yielding state machine here: diagnostics publication already uses the sliced diagnosticsPass scheduler, and the remaining synchronous paths are designed and measured to be self-bounding. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlScripting/dmlTemplateEmitter.ts
kburtram: Fixed in b66db2d8f. DML placeholders, keyed WHERE type text, module refusals, and generic unavailable labels now pass through sanitizeCommentText. Tests include malicious object, column, and type text containing comment terminators. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/core/lexer.ts
kburtram: Fixed in b66db2d8f. The lexer flushes a pending comment, string, quoted identifier, or bracketed identifier at EOF, preserving total coverage and lexical suppression. Final-character opener cases are in the coverage corpus. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/core/lexer.ts
kburtram: Fixed in b66db2d8f. The lexer recognizes Unicode horizontal space separators, including NBSP, while keeping line terminators distinct; identifier scanning excludes them. NBSP parsing and GO behavior are covered. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/core/sketch/index.ts
kburtram: Fixed in b66db2d8f. UPDATE sketching now skips TOP count forms and optional PERCENT before reading the target. UPDATE TOP diagnostics and target parsing have regression coverage. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/core/sketch/index.ts
kburtram: Fixed in b66db2d8f. Multipart-name parsing preserves the empty default-schema component, and binder arity checks use raw parts so db..table and server.db..table reach the honest cross-database or linked-server path. Tests cover the shorthand. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/core/binder.ts
kburtram: Fixed in b66db2d8f. Binder resolution now matches complete qualified overlay parts before catalog lookup, while preserving unqualified and temp-name behavior. Qualified script-created table and module references are covered. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlLanguage/core/lexer.ts
kburtram: Fixed in b66db2d8f. The misleading nonexistent batchSplitter.ts reference was replaced with the documented execution-splitter contract, and a shared table-driven GO corpus now pins plain GO, counts, comments, whitespace, and trailing-junk behavior. (Source: GPT-5.6 Sol)

extensions/mssql/test/unit/sqlLanguageProviders.test.ts
kburtram: Fixed in b66db2d8f. Provider tests now build a real CatalogBuilder snapshot and exercise SnapshotPinnedView mapping through the actual language engine, including object, column, key/readiness, definition, case-rule, and large filtered-prefix behavior. (Source: GPT-5.6 Sol)

extensions/mssql/src/sqlScripting/scriptWriter.ts
```diff
@@ -0,0 +1,115 @@
...
+ * anchors. The header MUST end with a newline so characters are unaffected.
+ */
+export function withHeader(
+    header: string,
+    body: { text: string; anchors: readonly ScriptAnchor[] },
+): { text: string; anchors: readonly ScriptAnchor[] } {
+    if (header.length === 0) {
+        return body;
+    }
+    let headerLines = 0;
```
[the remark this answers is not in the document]
kburtram: No code change. This is an explicit MUST precondition, and the complete call-site audit shows both production callers construct headers with terminating CRLF: fidelityHeader always appends CRLF when non-empty, and createTableEmitter appends CRLF before any fidelity header. The focused writer test also passes a terminated mixed-ending header. Adding runtime normalization or a throw for a hypothetical future contract violation would add behavior to a hot utility without fixing a current defect; callers should continue to honor the documented contract. (Source: GPT-5.6 Sol)

kburtram:
Review-fix validation sign-off for 66db2d8f:

- mssql build/typecheck/bundles: passed
- focused native language, observability, and service tests: 666 passed, 0 failed
- strict/native SQL scripting tests: 88 passed, 0 failed
- full mssql extension-host unit lane: 4,742 passed, 0 failed, 10 expected skips
- SQL Database Projects unit lane: 205 passed, 0 failed, 6 expected skips
- mssql ESLint: passed
- full repository Prettier check: passed
- staged pre-commit checks and git diff --check: passed
- online VSIX packaging: passed (1,542 files, 104.9 MB)
- review disposition: 26 original threads fixed or explicitly scoped, one immediate follow-up precondition nit resolved with call-site evidence
- post-push quiet-period audit: rechecked after six minutes; 0 unresolved review threads

The worktree is clean and the pushed PR head matches the validated commit. Hosted checks triggered by the push are tracked separately by GitHub. (Source: GPT-5.6 Sol)

kburtram (commented):
## Round 2: verification of `b66db2d8f`

I re-verified every one of the 27 threads against the fix diff (+714/−73 across 25 files), re-ran the gates in a clean clone, and re-ran the same compiled-engine probes that produced the round-1 evidence. **One residual defect** (inline comment — the alias-column-list fix misses the space-before-paren form); everything else is fixed correctly, several with better solutions than I suggested.

### Gates at `b66db2d8f` (clean clone)
Emit ✔, eslint 0 errors + prettier ✔ on the 25 changed files, all 15 language/scripting suites + observability → **714 passing, 0 failing** (682 in round 1; +32 regression tests: EOF-opener coverage, NBSP, GO contract corpus, same-length cache collision, real-CatalogBuilder provider suite, 1,205-procedure prefix regression, CS dedupe, `#temp` INSERT/UPDATE, UPDATE TOP, `db..`, `IS NOT DISTINCT FROM`, WITHIN GROUP, AT TIME ZONE, alias-equals, CURRENT OF, GOTO, script-view suppression, star-expansion no-dup, comment-sanitize canaries, writer CR accounting, async router timeout). CI green at head.

### False-positive matrix, re-run on the compiled fix head (same harness as round 1)
```
SELECT Total = SUM(C01) …                     clean ✓ (was 207)
UPDATE … WHERE CURRENT OF cur                 clean ✓ (was 207)
STRING_AGG(…) WITHIN GROUP (ORDER BY …)       clean ✓ (was 102)
… WHERE C02 AT TIME ZONE 'UTC' >= 1           clean ✓ (was 102 ×2)
FROM (VALUES (1,2),(3,4)) v(a, b)             clean ✓ (was 208)
FROM (SELECT …) d (c1, c2)                    STILL 208 + 207   ← inline comment (space before paren)
… IS NOT DISTINCT FROM C02                    clean ✓ (was 208)
myLabel: … GOTO myLabel                       clean ✓ (was 2812 error)
CREATE VIEW dbo.MyV … GO … FROM dbo.MyV       clean ✓ (was 208)
FROM dbo.Orders␠WHERE … (NBSP)                clean ✓ (was 208)
UPDATE TOP (100) … SET …                      clean ✓ (was 208 'TOP')
SELECT * FROM Db..Orders                      clean ✓ (was 208 'Db')
controls: real 207 / 208 / 102 all still fire ✓
```
Extra edge probes all clean: legacy hint `o (NOLOCK)`, `UPDATE TOP (10) PERCENT`, `WITHIN GROUP … OVER (…)`.

### Suppression matrix (was the Tier A bypass)
Trailing `--` comment, `--` at EOL, unterminated block after the statement, string opener at exact EOF, mid-string, mid-block-comment: **all 0 items now** ✓; empty doc/statement start still complete ✓. One deliberate-looking behavior change to be aware of: the new lexical pre-check suppresses at the caret *immediately after a closed string's final quote* (`SELECT 'abc'│` → 0 items; one character later → normal), which `classifyContext`'s closed-string boundary rule used to treat as outside. Harmless in practice (nobody completes at that exact caret) — just noting it contradicts the older comment in `context.ts`; not asking for a change.

### Perf spot-check (10k-object catalog, unchanged within noise)
FROM-target completion p50 4.3 ms / p95 15 ms; member access 0.1 ms; 2k-statement diagnostics 76 ms cold / cached ~0; the grow-the-search-window fix terminates correctly (verified against the 1,205-proc fixture) and adds no measurable cost at 10k.

### Fixes verified beyond the tests
Same-key `analyze`/`diagnosticsMemo` now compare exact text (my probe collision can't recur); the lexer EOF flush preserves total coverage without double-emitting (resume branches reset `constructStart` before the tail flush); `db..tbl` reaches the honest cross-database path via raw parts; binder qualified-overlay lookup folds per the catalog rule and takes the *last* definition; `caseSensitivityKnown` fails closed and suppresses-and-counts binder claims; expand-after-dot now emits unqualified names (verified: `C01, C02, …`, no `o.o.`); DML placeholders/module refusals sanitize `*/` (canary test); ScriptWriter counts CR/CRLF/LF correctly in both `append` and `withHeader`; the two TVF shapes now match the live server exactly; the GO corpus + documented-contract rewording resolves the dangling `batchSplitter.ts` reference; router disposition (cooperative timeout documented + async-timeout circuit test) and the deferred empty-prefix cache disposition are both reasonable.

### Sign-off
Fix the one spaced alias-list residual (a two-line change + the spaced form in the two tests) and this PR is ready from my side: 714/0 locally, CI green, the library boundary intact (no commands/settings/activation paths, not in the extension bundle), and every round-1 finding closed with regression coverage. Full re-review of that fix not required — I'll re-run the matrix when it lands and post the confirmation here.

_(source: Fable)_

extensions/mssql/src/sqlLanguage/core/sketch/index.ts
```diff
@@ -0,0 +1,1300 @@
...
+                kind,
+                alias: aliasRead.alias,
+                span: { start: spanStart, end: spanEnd },
+                innerScopeId,
+            });
+            // A derived/VALUES source can declare a column alias list after
+            // its correlation name: `(VALUES (...)) v(a, b)`. Skip the whole
+            // balanced list so its comma cannot start a phantom source.
+            j =
+                aliasRead.alias !== undefined && b.punct(aliasRead.next) === "("
```
kburtram: **[A-residual] The alias column-list skip checks the raw token index — any trivia before the `(` defeats it, so `v (a, b)` still creates a phantom source.**

`readAlias` returns `next` as a plain `index + 1` (the token immediately after the alias, i.e. Whitespace when the paren is space- or newline-separated), and the new guard tests `b.punct(aliasRead.next) === "("` without skipping trivia. Verified on the compiled fix head:

```
SELECT v.a FROM (VALUES (1, 2)) v(a, b);                    → clean ✓
SELECT v.a FROM (VALUES (1, 2)) v (a, b);                   → mssql(208)
SELECT d.a FROM (SELECT 1, 2 FROM dbo.Orders) AS d (a, b);  → mssql(208)
SELECT d.a FROM (SELECT 1, 2 FROM dbo.Orders) AS d\n(a, b); → mssql(208)
```

Space before the alias list is a completely ordinary formatting choice, so this is the same false-positive class the fix targeted — it just needs the significant-token walk: `const open = b.next(aliasRead.next, endExclusive); j = aliasRead.alias !== undefined && b.punct(open) === "(" ? b.skipBalanced(open, endExclusive) : aliasRead.next;`. Balance-skipping a significant paren after an alias is still safe (the only legal forms there are the column alias list and legacy hints; a new source requires a comma first). Please add the spaced form to both the sketch test and the honesty corpus — both new tests use the unspaced form, which is why this survived.

_(source: Fable)_

kburtram: Fixed in 2bc5811e3 — and went one step further than the comment asked: the alias list is now located with the significant-token walk AND its names are captured on the source, with the binder applying them as the positional column renames they are. So besides killing the phantom-source 208, `v (a, b)` / `AS d (c1, c2)` / newline-before-paren now expose the *renamed* columns: VALUES sources gain real column names (`v.b` resolves, completion after `v.` offers a, b), renamed projections resolve by their new names, and a renamed-away original correctly stops resolving (matching server behavior). Validated: eslint 0 errors, no typecheck errors in any sqlLanguage/sqlScripting file, all 15 suites + observability → 715 passing / 0 failing, the full round-1 false-positive matrix now 12/12 clean with all controls still firing, and legacy `o (NOLOCK)` hints unaffected. With this, everything from both review rounds is closed — sign-off from my side.

_(source: Fable)_

extensions/mssql/src/sqlScripting/dmlTemplateEmitter.ts
```diff
@@ -0,0 +1,177 @@
...
+    const skipped = columns.length - target.length;
+    if (skipped > 0) {
+        notes.push(`${skipped} identity/computed column(s) omitted from the column list`);
+    }
+    const writer = new ScriptWriter();
+    writer.anchored({ kind: "header" }, "INSERT INTO");
+    writer.append(` ${quoteIdentifier(info.schema)}.`);
+    writer.anchored({ kind: "objectName" }, quoteIdentifier(info.name));
+    writer.append(" (\r\n");
+    target.forEach((column, index) => {
```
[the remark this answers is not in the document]
kburtram: Valid — and it reproduced exactly as described: an all-identity/computed table emitted `INSERT INTO dbo.IdOnly (
)
VALUES (
);` and the line-94 twin left UPDATE with an empty SET list. Fixed in 2c20cc170 (+ test-assertion tidy in 2a54ec071): INSERT with no writable columns now emits `DEFAULT VALUES` with a fidelity note (your suggestion — it inserts exactly such a row), and UPDATE leaves an explicit `SET /* add column assignments */` placeholder plus note, matching the established missing-primary-key WHERE pattern. New regression test covers both shapes; normal tables verified unchanged. Validated: prettier/eslint clean, all language/scripting suites → 716 passing / 0 failing.

_(source: Fable)_

aasimkhan30 (approved):

kburtram:
For a more detailed design review of this PR — architecture, invariants, diagrams, and the full validation/review evidence — see the design and validation report in the notes repo:

**[Native T-SQL Language Service: Design, Implementation, and Validation Review (PDF)](https://github.com/kburtram/notes/blob/main/vscode-ext-refactor/design_reports/language_service_design_and_validation_report.pdf)** ([LaTeX source](https://github.com/kburtram/notes/blob/main/vscode-ext-refactor/design_reports/language_service_design_and_validation_report.tex))

The report is pinned to final head `2a54ec071` and covers: the claim-hierarchy and pinned-metadata design, the analysis pipeline and scheduling/routing containment model, strict scripting provenance (including the degenerate-DML `DEFAULT VALUES`/placeholder guards), the full differential false-positive matrix across review heads with the repair-loop methodology, the 709-owned/716-targeted test matrix, hosted CI and packaging evidence, all 29 dispositioned review threads, and the private-preview integration/graduation gates for the first consumer PR. 29 pages; report snapshot as of the `ee60edf` notes commit.

_(source: Fable)_

PR #22863, opened 2026-09-01
Add private-preview AI inline SQL completions

kburtram:
## Summary

This PR adds the first product consumer of the native SQL language/metadata foundations merged in #22836 and #22860: private-preview AI inline completions for classic SQL editor documents.

The branch is based directly on current `main` after PR05. It is not stacked on the open Debug Console work and does not include Query Studio, OE v2, or other unreleased `dev/refactor` features.

## What is included

- Inline-completion request orchestration, result shaping, cancellation, latency accounting, diagnostics integration, and editor registration.
- Structural schema context over the accepted metadata service and PR05 system-object catalog.
- A classic VS Code/STS token source for connection metadata and Azure access tokens.
- Model display, selection profiles, provider/vendor filtering, session overrides, and token approximation.
- Optional GitHub Copilot, Anthropic, OpenAI, and xAI language-model adapters.
- VS Code SecretStorage/environment API-key resolution and localized set/clear commands for direct SDK providers.
- Bounded internal completion trace serialization/loading/storage for later diagnostics work.
- Focused provider, schema-context, model-selection, SDK translation/key-resolution, trace, token-source, activation, manifest, and preview-boundary tests.

## Private-preview and default-off behavior

AI completions are effective only when both settings are enabled before activation:

```text
mssql.enableExperimentalFeatures
AND mssql.copilot.inlineCompletions.enabled
```

Both default to `false`, and their descriptions state that reload is required. The extension publishes an activation-snapshot context key and only composes the inline-completion service/provider and direct SDK providers when the full gate was effective at activation. Each request checks the live gate again and fails closed after disablement.

The six API-key commands have identical full-gate `enablement` and Command Palette `when` clauses. The three static language-model provider contributions also use that complete `when` clause, so neither commands nor provider names appear in the default UI.

All direct SDK providers default to disabled independently. Trace capture defaults off, prompt redaction defaults on, and closed-console recording defaults off.

## Privacy and network boundaries

- Schema context contains structural metadata, not credentials, tokens, query results, or stored query history.
- API keys are read from VS Code SecretStorage or explicitly configured environment-variable names; they are not placed in settings or diagnostics.
- Direct provider objects and their network-capable SDK clients are not constructed when the private-preview gate is off.
- Internal trace capture is bounded and opt-in; redaction is the default.
- No Debug Console, replay, trace-opening, sync, or other diagnostic command is contributed by this PR.

## Deliberate exclusions

- Query Studio editor/UI integration.
- PR03 Debug Console and session-diagnostics UX.
- AI-specific capture/replay pages or trace-opening commands.
- Object Explorer v2 and query-execution refactors.
- Any default-on behavior or migration of established completion UX.

Those boundaries keep this PR independently reviewable while leaving later UX iteration and diagnostic tooling on their own tracks.

## Validation

- Full repository build: passed for Extension Toolkit, MSSQL extension/typecheck/bundle/webviews/notebook, SQL Projects, Data Workspace, and localization.
- Full repository lint: passed.
- Localization generation and modified JSON parsing: passed.
- Complete no-coverage VS Code-hosted MSSQL suite: **4,884 tests, 0 failures**.
- SQL Projects hosted suite: **211 tests, 0 failures**.
- Focused inline-provider selection: **50/50 passed**.
- Focused sound/observability/model/SDK regression selection: **40/40 passed**.
- Final provider-visibility manifest selection: **5/5 passed**.
- Online VSIX packaging with released STS `6.0.20260827.1`: passed; `mssql-1.46.0.vsix`, 1,480 files, 102.85 MB.
- Commit hooks, staged lint, and diff checks: passed.

The coverage-enabled runner completes the test assertions but then fails while translating an upstream `https://raw.githubusercontent.com/...` source-map URL into a Windows output path. The identical complete test lane passes with coverage disabled; this is the existing Windows coverage-writer issue rather than a product-test failure.

## Follow-up assessment

Before this becomes default-on, the remaining work is primarily scenario/evaluation and UX policy rather than a prerequisite for reviewing the engine boundary here: validate profiles and prompt budgets on representative databases, decide which optional direct-provider settings belong in long-term UX, and integrate feature-specific diagnostics only after the PR03 Debug Console UX stabilizes. Query Studio and OE v2 remain later independent extractions.

(Source: GPT-5.6 Sol)


extensions/mssql/test/unit/utils.ts
```diff
@@ -32,6 +32,40 @@ export function stubTelemetry(sandbox?: sinon.SinonSandbox): {
...
+ * Creates the small TextDocument surface needed by text-processing unit tests.
+ */
+export function createTestDocument(text: string, uri: string): vscode.TextDocument {
+    const lines = text.split(/\r?\n/);
+
+    function offsetAt(position: vscode.Position): number {
+        let offset = 0;
+        for (let line = 0; line < position.line; line++) {
+            offset += lines[line].length + 1;
+        }
```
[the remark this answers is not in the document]
kburtram: Valid. `split(/
?
/)` erases the terminator width while `offsetAt` adds `+1` per line, so CRLF fixtures produce offsets short by one per preceding line and `getText(range)` slices drift. Test-helper only, so low-priority — but the fix is small: compute line-start offsets from the original text (accumulate `text.indexOf("
", …) + 1`, which counts the `
` naturally) instead of assuming a 1-char terminator. Fine to batch with the review-round fixes.

_(source: Fable)_

extensions/mssql/src/copilot/inlineCompletionDebug/tracePersistence.ts
```diff
@@ -0,0 +1,153 @@
...
+            {
+                redactPrompts: getTraceRedactPromptsSetting(),
+                maxFileSizeMB: getTraceMaxFileSizeMBSetting(),
+            },
+        );
+        const filePath = path.join(folder, createTraceFileName(trace._savedAt));
+        const serialized = JSON.stringify(trace, undefined, 2);
+        fs.writeFileSync(filePath, serialized, "utf8");
+        traceLogger.info(`Saved inline completion trace to ${filePath}`);
+        return { filePath };
```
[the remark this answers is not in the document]
kburtram: Valid nit. The function is already async and awaits `fs.promises.mkdir`; the write should be `await fs.promises.writeFile(...)`. VS Code awaits the promise returned by `deactivate()` (within its shutdown timeout), so the async form is safe on the deactivate path and stops blocking the host thread. One-liner; batch.

_(source: Fable)_

extensions/mssql/src/copilot/inlineCompletionDebug/tracePersistence.ts
```diff
@@ -0,0 +1,153 @@
...
+export function getConfiguredTraceFolder(context: vscode.ExtensionContext): string {
+    const configured = vscode.workspace
+        .getConfiguration()
+        .get<string>(Constants.configCopilotInlineCompletionsTraceFolder, "")
+        .trim();
+    if (configured.length === 0) {
+        return vscode.Uri.joinPath(context.globalStorageUri, DEFAULT_TRACE_FOLDER_NAME).fsPath;
+    }
+
+    return expandHome(configured);
```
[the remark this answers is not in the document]
kburtram: Valid. `expandHome(configured)` can return a relative path, which then resolves against the extension-host CWD — surprising and contradicting the "Absolute or home-relative folder" description. Suggest `path.isAbsolute(expanded) ? expanded : <globalStorage default>` with a one-time warning. Related: my review flags the broader scope question on this setting (window-scoped, so workspace-settable) — worth handling both in the same pass.

_(source: Fable)_

kburtram (commented):
# Review: private-preview AI inline SQL completions (PR06)

**Bottom line:** the architecture is right and most of the hard things are done well — the activation-snapshot + live-recheck gate split is real, API keys and Azure tokens have no leak path I or three focused review passes could find, telemetry is a bucketed whitelist, prompt budgeting/cancellation/sanitization in the inline provider are thorough, and the schema-context payload is genuinely structural (my canary probe planted secrets-like default/computed definitions, description prose, and an email in a real catalog — none reach the payload). What needs fixing before preview users touch it is a small cluster where the *privacy story as documented* diverges from the code: the trace "redaction" that doesn't redact the document text, a settings path that accepts literal API keys, the PR05 remote-LM consent contract this PR bypasses, and a lease-leak race — plus a set of B-level lifecycle/correctness items. Tiers below; every A/B claim was verified in code and the two headline ones reproduced empirically before posting.

### What I verified (beyond reading the diff)
- **Gates** at `c39256d60` in a clean clone: fresh `npm install` (no lockfile drift), emit ✔, webviews typecheck ✔, eslint **0 errors** + prettier ✔ on all 109 changed files; all 15 PR06 suites → **135 passing, 0 failing** (50 provider + 34 SDK + 51 others). Local `tsgo` errors are confined to four main-side files byte-identical to `origin/main` — environmental; CI is green at the head.
- **Privacy canary probe** (compiled payload builder, real `CatalogBuilder` snapshot): `sk-live-…` default definition, computed definition, table/column MS_Description, and an email — all absent from `buildRawSchemaContextPayload` output; the only identifying strings are `server`/`database`. Deterministic (byte-identical repeat), 7.5 ms p50 / 129 KB at 10,201 objects.
- **Redaction canary probe** (compiled trace serializer, `redactPrompts: true`): prompt/messages/schema-context fields correctly redacted; `locals` document text, SQL diagnostics, and `error.message`/`stack` all persist — the Tier A finding, proven empirically.
- **Dependency sanity**: `@anthropic-ai/sdk@0.91.1` + `openai@6.49.0` via the internal 1ES feed, official packages, two benign transitives; xAI rides the OpenAI-compatible surface; no typosquats.
- **Method**: three fan-out review passes (provider+gating; schema context+leases; SDK+secrets+traces) with every A/B claim spot-checked in code or reproduced against compiled code before posting; the two independent passes converged on the same top finding.

### Tier A — fix before preview exposure
1. **`trace.redactPrompts` doesn't redact `locals`/errors** — "redacted" traces contain the user's SQL (`traceSerializer.ts:27`; canary output in the comment; two review passes found it independently).
2. **Lease-leak race in the schema-context resolver** — check-await-set orphanes a refcounted lease; dedicated metadata session + poll timers leak until shutdown; same bug class as the PR #22836 store race (`completionSchemaContextService.ts:135`).
3. **Remote-LM consent divergence** — the PR05 `remoteLm`/`allowObjectNames` privacy contract is bypassed; server/database header + object names go to third-party APIs under a default-true setting whose description doesn't say so (`sqlInlineCompletionProvider.ts:2543`). Decide it explicitly.
4. **Literal API keys via the env setting fallback** — the setting value becomes the key itself when no env var matches; invites plaintext keys into settings.json/Settings Sync, contradicting the PR body (`apiKeyResolution.ts:167`).

### Tier B
SDK providers never re-check gates on the response path + `isEnabled()` defaults open (`sdkLanguageModelProviderBase.ts:105`); dead `modelFamily` setting (`sqlInlineCompletionProvider.ts:812`); undeclared `schemaContext` setting — the PR04 lesson again (`completionSchemaContextCore.ts:462`); window-scoped trace/debug settings let a workspace enable capture and pick the write path (`package.json:2039`); `unwrapQuotedResponse` corrupts quoted-literal completions (`:2338`); Anthropic first-message promotion + trailing-assistant prefill 400 for external callers (`messageTranslation.ts:20`); model-catalog data errors — sonnet-4-6 output 64K→128K, grok-4.20-0309 input 2M vs documented 1M (`sdkModelCatalog.ts:114`); cold-catalog 120 s wait without cancellation (`sqlInlineCompletionProvider.ts:525`); CS-catalog case-folded dedup (`completionSchemaContextCore.ts:986`); trace loader unbounded/unvalidated (`traceLoader.ts:367`); `when` on `languageModelChatProviders` is an undocumented property — runtime check needed (`package.json:5755`).

### Tier C — notes (inline: guard umbrella-blindness, LF-into-CRLF insertions)
Body-only: schema-budget compares JSON size not prompt size (safe direction; under-fills) and the shrink floor can exit over-budget on one very wide table; normalized cache is FIFO not LRU; debug fetch-cache key omits inventory caps; dead+divergent bucket helpers in the core vs provider; `readOptionalIntegerOverride` swallows explicit 0; "unlimited" profile relevance scoring is ~10⁷–10⁸ ops/keystroke (opt-in — worth a cap comment); `splitQualifiedName` splits at the first dot; Anthropic `provideTokenCount` is a live network call (bounded, 5 s); prepare/client caches hold full-key copies in memory (never exported); missing-key toast can be triggered by any extension's model enumeration (once, dismissible); redacted traces still carry `documentUri` (local paths can embed usernames — worth a docs note); per-attempt skipped-telemetry volume while typing comments; engine editions 6/9/11/12 get display names but only `scope:"all"` system objects (add the why-comment).

### Verified OK (highlights, each traced with file:line evidence during review)
Activation snapshot + per-request live gate for the inline provider, with mid-session fail-closure tests; API-key flow closed (SecretStorage/env → SDK constructor only; `password:true` input, prefix-validated, secrets-change invalidation; no logger/telemetry/trace path; inline errors sent with `includeErrorMessage:false`); Azure token source exemplary (single-flight, expiry/identity checks, canary-tested serialization); telemetry is an enumerated bucketed whitelist; prompt windows hard-capped with estimate→exact-count→trim→recount budgeting; cancellation wired through debounce/request/stream with time bounds on every await; SDK streams abort on cancel; errors contained to `[]`; disposal complete incl. LRU-bounded caches; manifest: all six key commands triple-gated in both `enablement` and palette `when`, new booleans default false, reload semantics declared and enforced; system-object catalog re-exports the PR05 single source (no drift); no per-keystroke lease acquire (warm path is a map hit); readiness degrades honestly ("columns unknown", never fabricated); sanitization pipeline (fences/sentinels/echo/stop-truncation/suffix-overlap) is thorough and well-tested; trace fixture is fully synthetic.

### Copilot threads (replied on each)
All three valid, low-priority, batchable: CRLF offsets in the `createTestDocument` helper; `writeFileSync` in the already-async trace save; relative `trace.folder` paths escaping to CWD (this last one is also part of the B-scope comment on trace-setting scopes).

### Test-gap summary (the highest-leverage ones)
Redaction canary with sentinels in `locals`/`error` (would have caught A1); concurrent-`resolve()` dedup + eviction-disposes-lease for the resolver (A2); serialized-payload field-surface pin for the remote-LM prompt (A3); key-canary through an SDK error path; SDK response-path fail-closure mid-session; configured-`modelFamily` selection; quoted-literal completion; CS-catalog schema-context fixture; oversized/garbage trace files; a `Constants.config*`-vs-manifest cross-check test.

### CI
Green at head (all completed lanes; the coverage-runner source-map issue is the known main-side Windows defect, reproduced identically without this PR).

### Merge readiness
Not yet, but close: the four Tier A items are small, contained fixes (a redaction keylist + canary; a pending-promise map; a consent decision + header/doc change; removing one fallback branch), and I'd want the SDK response-path gate check and the two model-catalog numbers in the same pass since they're cost/traffic-facing. Everything else can batch or follow up. Happy to fix directly in `work_verify` as before if you want.

_(source: Fable)_

extensions/mssql/src/copilot/inlineCompletionDebug/traceSerializer.ts
```diff
@@ -0,0 +1,128 @@
...
+    customPromptLastSavedAt?: number;
+}
+
+export interface SerializeSessionTraceOptions {
+    redactPrompts?: boolean;
+    maxFileSizeMB?: number;
+}
+
+const REDACTED = "[REDACTED]";
+const REDACTED_KEYS = new Set([
```
kburtram, line 42: **[A] `trace.redactPrompts: true` does not redact the document text in `locals`, nor `error.message`/`error.stack` — "redacted" trace files still contain the user's SQL.**

`REDACTED_KEYS` covers `userPrompt`/`systemPrompt`/`customSystemPrompt`/`rawResponse`/`sanitizedResponse`/`finalCompletionText`, plus the special-cased `promptMessages[].content` and `schemaContextFormatted` — but every debug event's `locals` bag stores the raw prompt inputs (`linePrefix`, `recentPrefix`, `statementPrefix`, `lineSuffix`, `suffix`, `sqlDiagnostics` — `sqlInlineCompletionProvider.ts:338-386`, up to ~7 KB of document text per event × 500 events), and `error.message`/`error.stack` survive too. Since the prompt is *built from those same fields*, redacting `promptMessages` while keeping `locals` redacts nothing in practice. Verified empirically against the compiled serializer with `redactPrompts: true`:

```
PROMPT-CANARY / MSG-CANARY / SCHEMA-CANARY   absent (redaction works)
locals.linePrefix  "…LOCALS-CANARY-TABLE"    *** PRESENT ***
locals.suffix      "LOCALS-SUFFIX-CANARY"    *** PRESENT ***
locals.sqlDiagnostics "…DIAG-CANARY"         *** PRESENT ***
error.message      "…ERROR-CANARY"           *** PRESENT ***
error.stack        "at STACK-CANARY:5"       *** PRESENT ***
```

The setting's own description promises "Redact prompt, response, completion, and schema-context text in saved completion trace files", and the deactivate path writes these files whenever `trace.captureEnabled` is on. Fix: redact (or key-allowlist) `locals` and strip `error.message`/`error.stack` under redaction; add a serializer canary test with a sentinel in `locals` — the current test's `locals` contains only `document.languageId`, which is why this survived.

_(source: Fable)_

extensions/mssql/src/copilot/completionSchemaContextService.ts
```diff
@@ -0,0 +1,343 @@
...
+        // fingerprints from the same facts via prepareConnection.
+        const fingerprint = profileFingerprint({
+            ...facts,
+            user: profilePrincipal(facts),
+            authKind,
+        });
+
+        let acquisition = this.acquisitions.get(fingerprint);
+        if (!acquisition) {
+            acquisition = await this.acquire(fingerprint, facts, authKind);
```
kburtram: **[A] Concurrent `resolve()` for the same connection races check-then-await-then-set — the losing lease is orphaned and its dedicated metadata session + poll timers leak until shutdown.**

`this.acquisitions.get(fingerprint)` misses, then `acquire()` awaits `prepareConnection` + `store().acquireDatabase(...)` before `this.acquisitions.set(...)`. Two overlapping completion requests for the same connection (realistic: the cold path awaits the disk-cache load inside `acquireDatabase`, and VS Code keeps cancelled `provideInlineCompletionItems` calls running since the token is only consulted *after* `getSchemaContext`) both miss and both call `acquireDatabase`. The store dedups entry *creation*, but each call increments `refCount` and returns a distinct lease — the second `set` overwrites the first `ClassicAcquisition`, the orphaned lease is never disposed, `refCount` never reaches 0, and the store's idle-TTL/LRU reclamation can never fire for that entry. This is byte-for-byte the same bug class as the `acquireDatabase` race fixed in PR #22836 review round 1.

Fix: a pending-promise map per fingerprint (mirror the store's `databaseCreations`), or re-check the map after the await and `lease.dispose()` the loser. Related (worth deciding, not blocking): acquisitions are released only by LRU overflow past 8 or service dispose — there is no hook on connection close/change, so up to 8 dedicated metadata sessions keep digest-polling servers the user has disconnected from for the rest of the VS Code session. A concurrent-resolve test plus an eviction-disposes-lease test would pin both (the only service test stubs the resolver entirely).

_(source: Fable)_

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
```diff
@@ -0,0 +1,3242 @@
...
+): string {
+    if (!schemaContext) {
+        return "-- unavailable";
+    }
+
+    const metadata = schemaContext.selectionMetadata;
+    const inventoryChunkSize = metadata?.inventoryChunkSize ?? 8;
+    const lines: string[] = [];
+    lines.push(
+        `-- connection: ${schemaContext.server ?? "unknown server"} / ${
```
kburtram: **[A] The remote-LM prompt bypasses the PR05 privacy contract: schema/object names plus the `-- connection: <server> / <database>` header go to third-party APIs with no `allowObjectNames`-equivalent consent.**

PR #22860's shared builder is explicitly gated for exactly this destination: `buildSchemaContext` takes `privacy: { destination: "local" | "remoteLm"; allowObjectNames }` and **refuses** (`degraded: "privacyPolicy"`, "no silent pseudonymization in v1 — degrade explicitly") when `remoteLm` lacks consent. This PR doesn't use that builder — the reimplemented payload path has no destination/consent parameter anywhere — and the rendered context (real schema/table/column/routine names *plus* this server/database header) is sent to `copilot`, `anthropic-api`, `openai-api`, and `xai-api`, the last three being direct HTTP clients to external APIs.

I verified the payload is otherwise clean (canary probe: default/computed SQL definitions, MS_Description prose, and an email canary all absent; user/account/tenant feed only the sha256 fingerprint) — so the PR body's "structural metadata" claim holds. But server/database names are connection *identity*, not structure, `mssql.copilot.inlineCompletions.useSchemaContext` defaults **true** with a description that never says identifiers leave the machine, and `canSendRequest` is meaningful consent for Copilot models but not for the extension's own SDK-registered providers.

Ask: make the divergence deliberate and visible — either route through (or mirror) the PR05 gate with an explicit consent bit, or at minimum drop/neutralize the server/database header for non-Copilot vendors and say "sent to the selected model provider" in the `useSchemaContext` description. A test pinning the exact serialized field surface of the payload would keep future fields from silently shipping to remote LMs.

_(source: Fable)_

extensions/mssql/src/copilot/sdkLanguageModels/apiKeyResolution.ts
```diff
@@ -0,0 +1,190 @@
...
+        if (!isNonEmpty(raw)) {
+            return undefined;
+        }
+
+        const configuredValue = raw.trim();
+        const envVarName = configuredValue.startsWith("$")
+            ? configuredValue.slice(1)
+            : configuredValue;
+        const resolvedFromNamedEnv = process.env[envVarName];
+        return isNonEmpty(resolvedFromNamedEnv) ? resolvedFromNamedEnv.trim() : configuredValue;
```
kburtram: **[A] The env-setting fallback treats the configured value as a literal API key — inviting plaintext keys into settings.json and Settings Sync, contradicting the PR's own security story.**

`getConfiguredEnvValue` tries the setting value as an environment-variable name and, when no such variable exists, returns the raw setting string as the key itself. Two consequences: (a) users can (and the nls description even documents it: "may contain a key or name an existing environment variable") paste real keys into `settings.json` — plaintext on disk, roamed by Settings Sync, sidestepping the SecretStorage design this PR otherwise gets right, and directly contradicting the PR body's "they are not placed in settings"; (b) a typo'd env-var name is silently transmitted to the provider as a credential. The `application` scope prevents workspace injection, which is why this isn't worse.

Fix: drop the literal fallback (env-var indirection only — require the `$VAR` form or treat the value strictly as a name), update the description, and toast on an unresolvable name instead of sending it. Keys stay in SecretStorage or real env vars, matching the stated contract.

_(source: Fable)_

extensions/mssql/src/nonproduction/sdkLanguageModels/sdkLanguageModelProviderBase.ts
```diff
@@ -0,0 +1,369 @@
...
+        this._prepareCache = undefined;
+        this.invalidateClient();
+        this._onDidChange.fire();
+    }
+
+    public async prepareLanguageModelChat(
+        _options: unknown,
+        token: vscode.CancellationToken,
+    ): Promise<LanguageModelChatInformation[]> {
+        if (token.isCancellationRequested || !this.isEnabled()) {
```
kburtram, line 106: **[B] The SDK providers never re-check the gates on the response path, and `isEnabled()` defaults open — a disabled provider (or disabled preview) keeps serving network requests until reload.**

The inline-completion provider re-checks the live gate per request as documented — but these direct SDK providers are registered as public `vscode.lm` chat providers, usable by any LM consumer once registered, and: `prepareLanguageModelChat` checks only the per-provider `enabled` (never `isInlineCompletionFeatureEnabled()`); `provideLanguageModelChatResponse` checks nothing at all; Anthropic's `provideTokenCount` makes a live network call carrying document text with no check. After mid-session disablement of the umbrella, the feature flag, or the per-provider setting, an already-acquired model handle keeps making API calls until reload — the config listener only logs. And `isEnabled()` falls back to `true` when the setting is undeclared (line 241), the fail-open direction, while `index.ts` correctly defaults `false`.

Fix: check the full gate + per-provider `enabled` at the top of `provideLanguageModelChatResponse`, `provideTokenCount`, and `prepare` (throw `LanguageModelError.NoPermissions`), and flip the fallback default to `false`. A mid-session fail-closure test for the response path would pin it — the existing suites only cover the inline provider's gate.

_(source: Fable)_

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
```diff
@@ -0,0 +1,3242 @@
...
+                    false,
+                ) ?? false
+        );
+    }
+
+    private async getLanguageModel(
+        modelSelectorOverride?: string,
+        modelPreference?: InlineCompletionModelPreference,
+    ): Promise<vscode.LanguageModelChat | undefined> {
+        const effectiveSelector =
```
kburtram, line 832: **[B] `mssql.copilot.inlineCompletions.modelFamily` is a dead setting — no settings-only configuration can ever make it take effect.**

`effectiveSelector = modelSelectorOverride ?? (modelPreference ? undefined : this.getConfiguredModelSelector())`, and `modelPreference` is *always* defined from settings: the configured profile id falls back to `"balanced"`, invalid values are coerced to a preset, and every preset supplies `modelPreference` for both categories. `getModelSelectorForCompletionCategory` never consults the configured `modelFamily` (only debug-store overrides and `continuationModelFamily`, which does work). So a user setting `modelFamily: "gpt-5.5"` sees no effect; the file-header comment and the nls description both describe behavior the code doesn't have, and the default value coincidentally matching the top preset preference (`claude-sonnet-4-6`) masks it. The test setup pins `configuredModelFamily = ""`, so nothing exercises it.

Fix: fold the configured `modelFamily` into the selector resolution ahead of the preset preference (or remove the setting and keep it debug-only, updating the docs). Either way add the test: set the setting to a secondary available model and assert selection.

_(source: Fable)_

extensions/mssql/src/copilot/completionSchemaContextCore.ts
```diff
@@ -0,0 +1,2774 @@
...
+    },
+};
+
+export function getSqlInlineCompletionSchemaContextRuntimeSettings(
+    modelMaxInputTokens?: number,
+    debugSchemaContextOverrides?: InlineCompletionDebugSchemaContextOverrides | null,
+): SqlInlineCompletionSchemaContextRuntimeSettings {
+    const workspaceSettings = asRecord(
+        vscode.workspace
+            .getConfiguration()
```
kburtram: **[B] `mssql.copilot.inlineCompletions.schemaContext` is read here (and change-watched by the service) but not declared in `package.json` — the PR04 lesson repeated.**

The constant exists, this reads it as the schema-budget override object, and the service flushes its cache when it changes — but it's absent from `contributes.configuration` (I cross-checked every `mssql.copilot.*` constant against the manifest; this is the only undeclared *setting*). Undeclared means no discoverability, no schema validation, and it silently shapes what's sent to remote models. Declare it (object schema with the override fields) or remove the workspace read and keep it debug-store-only. Worth adding the generic guard while here: a unit test asserting every `Constants.config*` read is declared in the manifest — that would have caught this and the PR04 instances.

_(source: Fable)_

extensions/mssql/package.json
```diff
@@ -1891,6 +1953,303 @@
...
+                    "description": "%mssql.copilot.inlineCompletions.enabledCategories.description%",
+                    "scope": "window"
+                },
+                "mssql.copilot.inlineCompletions.debug.recordWhenClosed": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "%mssql.copilot.inlineCompletions.debug.recordWhenClosed.description%",
+                    "scope": "window"
+                },
+                "mssql.copilot.inlineCompletions.trace.captureEnabled": {
```
kburtram, line 2046: **[B] The trace/debug settings are `window`-scoped — a workspace can enable capture, disable redaction, and choose the write destination.**

`inlineCompletions.enabled`, `debug.recordWhenClosed`, `trace.captureEnabled`, `trace.redactPrompts`, and `trace.folder` are all workspace-settable. For a user whose application-scoped umbrella is already on, opening a repo can therefore turn the feature on, enable always-capture, turn redaction off, and point `trace.folder` at an arbitrary path (`expandHome` + `mkdir -p` + write on deactivate) — producing a file of the user's SQL and schema names at a workspace-chosen location (compounded by the `locals` redaction gap). The SDK `baseUrl`/`env` settings are correctly `application`-scoped, so there's no network redirection — this is a local-write/defense-in-depth issue. Recommend `machine` scope (or `machine-overridable` + `restricted`) for `trace.folder`, `trace.captureEnabled`, `trace.redactPrompts`, and `debug.recordWhenClosed`.

_(source: Fable)_

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
```diff
@@ -0,0 +1,3242 @@
...
+}
+
+function stripModelPreamble(text: string): string {
+    return text.replace(
+        /^\s*(?:completion|insert text|ghost text|sql|query|answer|result|output)\s*:\s*/i,
+        "",
+    );
+}
+
+function unwrapQuotedResponse(text: string): string {
```
kburtram, line 2380: **[B] `unwrapQuotedResponse` corrupts completions that are legitimately quoted SQL.**

Any trimmed response that starts and ends with `'`/`"`/curly quotes is unwrapped unconditionally, before the sentinel/meta checks. A continuation completion that *is* a string literal — cursor after `WHERE status = ` and the model returns `'active'`, or `'2024-01-01'`, or a quoted identifier `"Order Details"` — is stripped to `active` / `2024-01-01` / `Order Details` and inserted as broken SQL. Suggest unwrapping only when the interior looks like a quoted whole answer (contains SQL keywords/whitespace structure) or skipping the unwrap in continuation mode when the interior parses as a plain literal; add the `WHERE status = ` → `'active'` case to the sanitization suite (no test covers a quoted-literal completion today).

_(source: Fable)_

extensions/mssql/src/copilot/sdkLanguageModels/messageTranslation.ts
```diff
@@ -0,0 +1,99 @@
...
+    system: string;
+    messages: MessageParam[];
+}
+
+export function translateForAnthropic(
+    messages: vscode.LanguageModelChatMessage[],
+): AnthropicMessageTranslation {
+    let system = "";
+    let chatStart = 0;
+    if (messages.length > 0 && messages[0].role === vscode.LanguageModelChatMessageRole.User) {
```
kburtram: **[B] Two translation-shape hazards for external callers: the first user message is unconditionally promoted to system prompt, and a trailing assistant turn produces an Anthropic prefill request that current models reject.**

1. `translateForAnthropic`/`translateForOpenAI` treat `messages[0]` (role User) as the system prompt. These providers are registered as public `vscode.lm` chat providers, so any extension can call them; a single-user-message request becomes system-only (OpenAI) or `system` + a synthetic `"Please respond."` user turn (Anthropic) — semantically wrong with no warning. The mssql pipeline always sends ≥2 user messages, so it works internally. Either promote only when ≥2 messages exist or document the convention on the provider.
2. History ending in an assistant turn (`[system, user, assistant]` — the module's own test produces exactly this) yields an Anthropic messages array ending with `assistant`, i.e. an assistant prefill — which the Claude 4.6+ generation (every model in this catalog) rejects with a 400. Mirror the existing leading-turn guard: append a closing user turn or drop the trailing assistant message.

Neither is reachable from the inline-completion path today, which keeps this at B — but the public registration makes both real for the first chat consumer.

_(source: Fable)_

extensions/mssql/src/nonproduction/sdkLanguageModels/sdkModelCatalog.ts
```diff
@@ -0,0 +1,119 @@
...
+        case "xai":
+            return Constants.configCopilotSdkProvidersXAiAdditionalModels;
+    }
+}
+
+function getDefaultSdkModels(kind: SdkProviderKind): SdkModelCatalogEntry[] {
+    switch (kind) {
+        case "anthropic":
+            return defaultAnthropicSdkModels;
+        case "openai":
```
kburtram, line 114: **[B] Two model-catalog data errors worth fixing before users hit token-budget 400s.**

Checked the catalog against current provider documentation: the IDs are all real (no hallucinated models), but (1) `claude-sonnet-4-6` lists `maxOutputTokens: 64000` while the actual cap is 128K — this clamps every caller through `getMaxTokens`, halving the available output for the default model family; (2) the two `grok-4.20-0309-*` entries list `maxInputTokens: 2,000,000` while xAI's own model pages list 1,000,000 for the 0309 variants — an overstated input budget is worse than an understated one, because the schema-context trimming trusts it and oversized prompts then 400 at the API. Also stylistic: `claude-haiku-4-5-20251001` is a dated snapshot while its siblings use undated aliases — Anthropic guidance prefers the undated alias for forward-compat. A once-per-release re-verify note on this file would help; none of these numbers are covered by tests.

_(source: Fable)_

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
```diff
@@ -0,0 +1,3242 @@
...
+                    intentMode,
+                );
+                this.sendInlineCompletionTelemetry("noPermission", telemetrySnapshot);
+                recordDebugEvent("noPermission");
+                return [];
+            }
+
+            if (useSchemaContext) {
+                recordPendingStage("buildingSchemaContext");
+                schemaContext = await withTimeout(
```
kburtram: **[B] Cold-catalog requests can hold an automatic completion for up to 120 s, and the schema-context build ignores the cancellation token entirely.**

`getSchemaContext` is awaited under a 6-minute `withTimeout` without the token; inside, the resolver's `aiContext` policy joins first hydration with `timeoutMs: 120_000`. So on a cold catalog every automatic completion attempt waits (the work is shared — single hydration lane — so no server amplification, but the request and its resources are held long after the user moved on, and the module header's "a catalog still hydrating yields no schema context for that request" doesn't match: it waits rather than yields). Suggest: for `Automatic` triggers, a short wait (1–2 s) with immediate `fallbackWithoutMetadata`, keeping the long join for `Invoke`; and pass the cancellation token down as the freshness policy's `signal`. A cancelled-during-schema-context test would pin it (only an already-cancelled token during debounce is tested today).

_(source: Fable)_

extensions/mssql/src/copilot/completionSchemaContextCore.ts
```diff
@@ -0,0 +1,2774 @@
...
+        if (normalizedObjects.length >= maxObjectCount) {
+            break;
+        }
+
+        const qualifiedName = toQualifiedName(object.schema, object.name);
+        if (!qualifiedName) {
+            continue;
+        }
+
+        const objectKey = qualifiedName.toLowerCase();
```
kburtram: **[B] Case-insensitive dedup silently drops case-distinct objects/columns on case-sensitive catalogs.**

`normalizeSchemaObjects`/`normalizeObjectColumns` (and the selection maps at 1445/1742) key by `toLowerCase()`, so on a CS database `dbo.Orders` and `dbo.ORDERS` collapse to one — the second silently vanishes from the prompt. The payload builder upstream deliberately ranks with `ordinalCompare` to support such catalogs, and the catalog-backed source knows `caseSensitive` — but `RawSchemaContextPayload` doesn't carry it and the core ignores it. Same fix shape as the PR05 completion-dedupe repair: fold only when the catalog is case-insensitive (thread `caseSensitive` through the payload). No CS-catalog fixture exists in the schema-context tests.

_(source: Fable)_

extensions/mssql/src/copilot/inlineCompletionDebug/traceLoader.ts
```diff
@@ -0,0 +1,190 @@
...
+            fileSizeBytes: stat.size,
+            included: options.included,
+            loaded: false,
+            imported: options.imported,
+            loadError: getErrorMessage(error),
+        };
+    }
+}
+
+export async function loadTraceFile(filePath: string): Promise<InlineCompletionDebugExportData> {
```
kburtram: **[B] The trace loader reads untrusted JSON with no size cap and casts `events` with zero per-event validation — harden before the debug panel lands.**

`loadTraceFile` does `readFile` + `JSON.parse` unbounded (the folder is user-configurable and `maxFileSizeMB` only applies at *write* time — a multi-hundred-MB file stalls/OOMs the host), and `normalizeTraceFile` validates top-level fields but casts `value.events as InlineCompletionDebugEvent[]` (line 118), so garbage event objects flow straight into `importSession` and, later, the webview. The sharper risks are handled (per-file try/catch turns JSON garbage into `loadError`; no prototype-pollution path since nothing merges by assignment), and these exports currently have no production caller — but that's exactly why the hardening should land before the consumer PR does: a size check before read, per-event shape validation, and an oversized/garbage-file test (loader tests cover only the happy path + filename filtering). Also line 100's `value.version === 1 ? 1 : 1` always yields 1 — a dead conditional that should be a real version check.

_(source: Fable)_

extensions/mssql/package.json
```diff
@@ -5392,6 +5751,32 @@
                     "dark": "editor.foreground"
                 }
             }
+        ],
+        "languageModelChatProviders": [
```
kburtram, line 6039: **[B] `when` on `contributes.languageModelChatProviders` is not a documented property of that contribution point — the "hide providers outside private preview" commit needs a runtime check.**

The documented schema for `languageModelChatProviders` is `vendor`/`displayName`/`managementCommand` (`when` is documented for `languageModelTools`). If VS Code ignores the unknown key, the three vendors ("Anthropic API", "OpenAI API", "xAI API") appear in the Manage Language Models picker for **every** user regardless of the preview gate. Impact is limited to UI visibility — selecting a vendor yields no models because code-side registration is gated at activation and `prepare` returns `[]` without the enabled flag/key — but the head commit's stated purpose rests on this property working, and `privatePreviewManifest.test.ts` only asserts the string exists in the JSON, which can't catch it. Please verify against the target VS Code build; if unsupported, either accept picker visibility and document it, or gate visibility another way.

_(source: Fable)_

extensions/mssql/src/copilot/copilotEnableSettingsGuard.ts
```diff
@@ -0,0 +1,196 @@
...
+        return vscode.workspace.getConfiguration(copilotSection, scope);
+    }
+
+    private isInlineCompletionsEnabled(): boolean {
+        return (
+            vscode.workspace
+                .getConfiguration()
+                .get<boolean>(Constants.configCopilotInlineCompletionsEnabled, false) ?? false
+        );
+    }
```
kburtram, line 184: **[C] The guard is umbrella-blind mid-session — it can assert "MSSQL inline SQL completions are active" when they aren't.**

`isInlineCompletionsEnabled()` reads only the sub-flag. After the user turns off `mssql.enableExperimentalFeatures` (feature inert pending reload), a `github.copilot.enable` change or a sub-flag off→on toggle still shows the "active" prompt and can write `github.copilot.enable` on a false premise. One-line fix: include `isInlineCompletionFeatureEnabled()` in the guard's check.

_(source: Fable)_

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
```diff
@@ -0,0 +1,3242 @@
...
+    "full",
+    "cross",
+    "outer",
+    "where",
+    "group",
+    "order",
+    "having",
+]);
+
+export function fixLeadingWhitespace(
```
kburtram, line 1659: **[C] `fixLeadingWhitespace` joins with bare `\n` — accepting a multi-line completion in a CRLF document yields mixed line endings.**

The sanitization pipeline works in LF throughout and this helper prepends/joins with `\n` before insertion, so multi-line completions accepted into CRLF files mix EOLs (visible in diffs; some formatters/linters flag it). Cheap fix: detect `document.eol` at insertion time and join with it.

_(source: Fable)_

extensions/mssql/package.json
```diff
@@ -1891,6 +1953,303 @@
...
+                    "description": "%mssql.copilot.inlineCompletions.modelFamily.description%",
+                    "scope": "window"
+                },
+                "mssql.copilot.inlineCompletions.continuationModelFamily": {
+                    "type": "string",
+                    "default": "",
+                    "description": "%mssql.copilot.inlineCompletions.continuationModelFamily.description%",
+                    "scope": "window"
+                },
+                "mssql.copilot.inlineCompletions.modelVendors": {
```
karlb-dev, line 2002: **Workspace-scoped `modelVendors` can send SQL + schema context to an arbitrary `vscode.lm` vendor.**

This setting is `window` scope and `selectConfiguredLanguageModels` calls `selectChatModels({ vendor })` for whatever strings are in the array — no intersection with the known set (`copilot` / `anthropic-api` / `openai-api` / `xai-api`). Once the application-level umbrella is on, a repo `.vscode/settings.json` can add a vendor from a workspace-recommended extension and this provider will ship statement prefix/suffix, schema names, and nearby diagnostics there. The trace knobs being window-scoped is already called out; this is the live network path. Pin `modelVendors` (and likely `useSchemaContext` / `includeSqlDiagnostics`) to `application` or `machine`, and drop unknown vendors rather than forwarding to them.

(Source: Cody · Grok Bot)

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
```diff
@@ -0,0 +1,3242 @@
...
+            }
+
+            const telemetrySnapshot = sendResultTelemetry("success");
+            telemetrySnapshot.documentUri = document.uri.toString();
+            telemetrySnapshot.positionLine = position.line;
+            telemetrySnapshot.completionLineCount = getCompletionLineCount(finalCompletionText);
+            const storedEvent = recordDebugEvent("success");
+
+            return [
+                new vscode.InlineCompletionItem(
```
karlb-dev, line 732: **Ghost text is built against a `Position` captured before the schema/model awaits — we never re-check `document.version`.**

After the 350ms debounce this request can sit in `getSchemaContext` / `sendRequest` for a long time (120s hydration join + 120s LM timeout). We consult the cancellation token after schema build and after `collectText`, but a still-alive request returns `InlineCompletionItem` at the original `position`. Overlapping `provideInlineCompletionItems` already continues through schema hydration when VS Code has cancelled; if that result is still accepted, the insert point is stale. Capture `document.version` and the cursor offset at the start of the try, and return `[]` if either changed before constructing the item.

(Source: Cody · Grok Bot)

karlb-dev (commented):
Two items not covered by the earlier Copilot/Fable pass: workspace-controlled `modelVendors` can redirect live SQL/schema prompts to an untrusted `vscode.lm` vendor, and ghost text is returned at a `Position` captured before the long schema/model awaits with no `document.version` check.

(Source: Cody · Grok Bot)

extensions/mssql/test/unit/utils.ts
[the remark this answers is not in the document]
kburtram: Fixed in f0961863c. createTestDocument now computes exact line-start offsets from the document's actual CRLF or LF delimiters and exposes matching eol and offsetAt behavior. CRLF completion regression coverage was added. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/inlineCompletionDebug/tracePersistence.ts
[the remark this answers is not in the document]
kburtram: Fixed in f0961863c. A configured trace folder must now resolve to an absolute path; relative values emit a warning and fall back to extension global storage. The fallback behavior has unit coverage. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/inlineCompletionDebug/traceSerializer.ts
kburtram: Fixed in f0961863c. Redaction now recursively sanitizes every string value in locals, redacts error message and stack while retaining the error type, and covers document URI/name plus prompt, schema, and response fields. Canary-string tests verify that sensitive values do not survive serialization. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
kburtram: Fixed in f0961863c. The configured general model-family selector is now included after category-specific overrides and before profile preference, so it is no longer shadowed by configured profiles. Its manifest default is empty and the precedence has unit coverage. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/sdkLanguageModels/apiKeyResolution.ts
kburtram: Fixed in f0961863c. The configuration value is now interpreted only as the name of an environment variable; a missing variable or apparent literal secret is ignored. SecretStorage and named environment-variable resolution remain supported, with regression tests. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
kburtram: Fixed in f0961863c. Schema context and diagnostics now require explicit opt-in, default off, and use application scope with clearer disclosure. The server/database header was removed from prompts, so schema identifiers are sent only when the user explicitly enables schema context. (Source: GPT-5.6 Sol)

extensions/mssql/src/nonproduction/sdkLanguageModels/sdkLanguageModelProviderBase.ts
kburtram: Fixed in f0961863c. Model discovery, response generation, and token counting now recheck both the umbrella private-preview gate and the provider-specific gate. Provider gates also default off, and the checks run again after asynchronous key resolution. Regression coverage was added. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/completionSchemaContextService.ts
kburtram: Fixed in f0961863c. Concurrent requests for the same fingerprint now share a single pending acquisition promise and lease. Late-acquired leases are released if the consumer was disposed while acquisition was pending. Concurrency and late-disposal regressions are covered. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/completionSchemaContextCore.ts
kburtram: Fixed in f0961863c. The undeclared schemaContext setting constant, configuration read, and change listener were removed. Schema-context tuning remains available only through the existing debug override path. (Source: GPT-5.6 Sol)

extensions/mssql/package.json
kburtram: Fixed in f0961863c. Trace capture, recording, folder, redaction, and size-limit settings are now machine scoped. Prompt-related controls that are not host-specific remain application scoped. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
kburtram: Fixed in f0961863c. Automatic metadata acquisition is capped at 1.5 seconds while explicit Invoke keeps the 120-second budget. VS Code cancellation is bridged to an AbortSignal and passed into ensureFresh. Injectable seams and tests cover timeout and cancellation behavior. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/sdkLanguageModels/messageTranslation.ts
kburtram: Fixed in f0961863c. A lone user message remains a user message, and Anthropic translation can no longer end with an assistant prefill. The first user message is promoted only when another message follows. Regression tests cover both cases. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
kburtram: Fixed in f0961863c. Quote unwrapping is now limited to a whole quoted response that is SQL-shaped or contains a newline. Ordinary SQL literals such as 'active' and quoted identifiers such as Order remain intact, with regression tests. (Source: GPT-5.6 Sol)

extensions/mssql/src/nonproduction/sdkLanguageModels/sdkModelCatalog.ts
kburtram: Fixed in f0961863c. The catalog now uses the current Sonnet output limit, Grok context size, and Haiku alias. The corrected metadata is covered by catalog tests. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/completionSchemaContextCore.ts
kburtram: Fixed in f0961863c. The snapshot's case-sensitivity flag is now threaded through schema selection and payload construction. Identity maps, deduplication, foreign-key matching, and column trimming use collation-aware keys, with a regression test for case-distinct schemas, objects, and columns. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/inlineCompletionDebug/traceLoader.ts
kburtram: Fixed in f0961863c. Trace loading now checks a 64 MiB cap before reading, requires the supported trace version exactly, and validates every event plus nested prompt, tag, and error structure before returning data. Malformed and oversized traces have regression coverage. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
kburtram: Fixed in f0961863c. The provider captures the document version and insertion offset before asynchronous work, then verifies both immediately before constructing the completion item. A changed document now causes the stale result to be discarded, with regression coverage. (Source: GPT-5.6 Sol)

extensions/mssql/package.json
kburtram: Not changing this contribution. The repository's exact tested VS Code target is 1.136.0-insider at commit 5c9173276314a5688bc10275e5a25c8d0fe9f4f7, whose languageModelChatProviderType contribution schema includes when. Current official VS Code contribution-point documentation also defines languageModelChatProviders[].when for Manage Models visibility: https://code.visualstudio.com/api/references/contribution-points. The runtime guard remains defense in depth, but removing the manifest condition would expose disabled providers in this target. (Source: GPT-5.6 Sol)

extensions/mssql/package.json
kburtram: Fixed in f0961863c. Model-vendor configuration is now application scoped, constrained by a manifest enum and uniqueItems, and checked against a runtime allowlist. Unknown vendor identifiers are ignored even if injected outside the settings UI, with regression coverage. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/copilotEnableSettingsGuard.ts
kburtram: Fixed in f0961863c. The Copilot settings guard's enabled predicate now includes the umbrella private-preview gate, so it cannot prompt when the overall preview is disabled. Regression coverage was updated accordingly. (Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/sqlInlineCompletionProvider.ts
kburtram: Fixed in f0961863c. Completion output is normalized to the active document's EOL sequence before the edit is constructed. CRLF regression coverage and the exact-EOL test document helper verify both text and offsets. (Source: GPT-5.6 Sol)

kburtram:
### Validation sign-off — bd43b0e02

Review fixes are in f0961863c, and the branch has been merged with the current main baseline in bd43b0e02.

Validated locally:

- Full repository lint passed.
- Full workspace build passed across mssql, extension/webview/notebook bundles, extension-toolkit, SQL Database Projects, and Data Workspace.
- Focused AI-completions and debug-trace hosted suites passed: 121 passed, 0 failed, 0 skipped.
- The complete VS Code-hosted extension suite passed with coverage disabled (exit code 0).
- Perftest dependency install, build, and typecheck passed for all five workspaces.
- All runnable non-live perftests passed: 347 passed, with 1 intentional skip.
- Full online packaging passed for all extension packages. The mssql VSIX packaged successfully at 118.79 MiB with STS 6.0.20260827.1.
- The final Git worktree is clean, and commits f0961863c and bd43b0e02 are pushed.

Environmental limitations observed:

- The coverage-enabled extension command completed its tests but the coverage writer then failed on Windows while trying to interpret an external dependency source-map URL from ts-async as a local https:\ path. The same complete hosted suite passed with coverage disabled; this is a post-test coverage-path tooling issue.
- The perftest live central-store scenario could not connect because no SQL Server was listening at localhost:14333. The CLI suite passed 251/251 when that external-infrastructure test file was excluded; the other perftest suites also passed.

(Source: GPT-5.6 Sol)

extensions/mssql/src/copilot/languageModelSelection.ts
```diff
@@ -0,0 +1,73 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from "vscode";
+export {
+    defaultInlineCompletionModelVendors,
+    getConfiguredInlineCompletionModelVendors,
+} from "./languageModels/shared/vendorAllowList";
+import { getConfiguredInlineCompletionModelVendors } from "./languageModels/shared/vendorAllowList";
+import { formatModelSelector, parseModelSelector } from "./languageModels/shared/modelDisplay";
```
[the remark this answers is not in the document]
kburtram: No change needed. A re-export-from clause does not introduce a local binding in TypeScript, so the separate import supplies the local callable binding without creating a duplicate identifier. The full extension typecheck and full workspace build both pass on bd43b0e02, confirming there is no compile error. Rewriting this would be stylistic churn only and is below the triage bar for a post-validation follow-up. (Source: GPT-5.6 Sol)

kburtram:
## Round 2: fix verification + build-channel mechanism

### Fix verification (`f0961863c` + merge `bd43b0e02`)

Re-verified every one of the 23 threads against the fix diff and re-ran the full local battery in a clean clone: fresh install (no drift), emit ✔, eslint 0 errors, prettier ✔, all PR06 suites → **150 passing / 0 failing** (135 → +15 new regression tests). Highlights checked in code line-by-line:

- **Trace redaction** now recursively redacts every string in `locals`, redacts `error.message` (keeping the type, dropping the stack), and covers `documentUri`/`documentFileName` — with canary tests. The `-- connection: server/db` header is gone from prompts, and `useSchemaContext`/`includeSqlDiagnostics` are opt-in, application-scoped, with honest "send … to the selected language model provider" descriptions.
- **Lease race** fixed with the pending-promise map, including disposal of a lease that lands after the resolver is disposed — both covered by tests.
- **Literal API keys** can no longer come from settings: the env setting resolves strictly as a variable name; literals are ignored (test pins it).
- **SDK gates** re-checked in `prepare`, response, and token counting (including after async key resolution), defaults fail closed.
- `modelFamily` wired into selector precedence; catalog numbers corrected (Sonnet 128K out, Grok 1M in, Haiku alias); `unwrapQuotedResponse` preserves SQL literals; translation guards for single-message and trailing-assistant shapes; abortable 1.5 s automatic schema-context budget with the token bridged to `ensureFresh`; stale document version/offset discard; CS-aware identity keys end-to-end; trace loader capped at 64 MiB with strict per-event validation; trace settings machine-scoped with a manifest test; vendor allow-list enforced at runtime. All good — no residuals from my side.

### Build-channel mechanism (this round's design change)

Per the shipping constraint that the direct Anthropic/OpenAI/xAI SDK providers are for internal testing only, this adds a generic compile-time channel mechanism — the TypeScript equivalent of `#ifdef`, except excluded code is genuinely absent from the bundle and VSIX:

- **Channels**: `development` (default for source builds) / `internal` / `insiders` / `stable`, selected via `MSSQL_BUILD_CHANNEL` or `--channel=`.
- **Code organization**: gated areas live under `src/nonproduction/<area>/` with one import seam (`index.ts`) and an inert twin (`index.stub.ts`). `src/nonproduction/channels.json` declares each area's channels.
- **Bundle exclusion**: an esbuild plugin substitutes the stub for excluded areas, rejects deep imports into them, audits the metafile so excluded sources cannot leak into `dist`, and records the channel in `dist/build-channel.json`.
- **VSIX manifest**: packaging refuses a channel mismatch with the built bundle, strips the excluded areas' settings/commands/palette entries/language-model-provider contributions (declared in `src/nonproduction/manifest-contributions.json`), and restores `package.json` afterwards — the same mutate-restore pattern the offline packaging already uses.
- **First area**: `sdkLanguageModels` (the three direct-API providers + key resolution + translation + catalog) moved to `src/nonproduction/sdkLanguageModels/`, channels `development`/`internal`/`insiders`. The SDK deps moved to `devDependencies`. Stable ships Copilot models only; the shared model-selection/display code and all runtime preview gates are unchanged.
- **Tests**: `tsc`/`out/` still compile everything, so unit tests run identically in every channel; `test/unit/nonproductionChannels.test.ts` pins seam/stub export parity, channel validity, and manifest-map accuracy.

Landed as `c5617dd31`. Validation: extension typecheck fully clean (after rebuilding the stale local extension-toolkit — the earlier telemetry-type noise was exactly that), 142 focused tests passing including the new channel-invariants suite, the `development` bundle includes the real SDK sources, the `stable` bundle contains only the stub (metafile audit) with zero SDK identifiers/endpoints/dependency references in `dist`, and the stable manifest strip removes 15 settings + 6 commands + 6 palette entries + 3 language-model providers and restores `package.json` byte-identically. Redaction canary also re-run on the fix head: all ten sentinels (locals incl. nested, diagnostics, error message/stack, documentUri) redacted.

_(source: Fable)_

extensions/mssql/src/nonproduction/sdkLanguageModels/sdkLanguageModelProviderBase.ts
```diff
@@ -0,0 +1,383 @@
...
+            additionalProps: {
+                vendor: this.vendor,
+                family: model.family,
+                latencyBucket: getLatencyBucket(Date.now() - startedAt),
+                result,
+            },
+            additionalMeasurements: {
+                inputTokens: usage?.inputTokens ?? 0,
+                outputTokens: usage?.outputTokens ?? 0,
+            },
```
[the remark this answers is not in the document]
kburtram: Valid observation, but leaving as-is at this stage of the review. The conflation only affects the two aggregate measurements on the SDK-provider invocation event (result/latency/vendor dimensions are unaffected), the SDK providers are a non-production build-channel area as of c5617dd31 (development/internal/insiders only — these events can never come from stable users), and cancellation/no-usage cases are identifiable by the paired result property ("cancelled"/"error") when slicing the metric. Omitting the keys when usage is absent is a reasonable polish for the follow-up diagnostics PR where this telemetry gets consumed; noting it there rather than re-opening the review cycle for it now.

_(source: Fable)_

kburtram:
## Round 3: Copilot "Needs a closer look" findings — fixed in 640d3f23f

Copilot's review of 650078adc created no inline threads; its three "previously missed" findings sit in the collapsed review body. All three are legitimate and all three are fixed in 640d3f23f:

1. **Trace truncation measured compact JSON, but the file is written pretty-printed** (`tracePersistence.ts:64`) — confirmed. `truncateTraceToMaxSize` sized `JSON.stringify(trace)` while `saveInlineCompletionTraceNow` wrote `JSON.stringify(trace, undefined, 2)`, so a trace that just fit the budget could exceed it on disk. Fix: `serializeTraceFile()` in `traceSerializer.ts` is now the single on-disk representation; the truncation loop measures that exact text and persistence writes it. New test `measures the size limit against the pretty-printed file text` fails on the old code (no `_truncated`) and passes now.
2. **`trace.maxFileSizeMB` uncapped while the loader hard-limits at 64 MB** (`tracePersistence.ts:124`) — confirmed. Fix: one shared constant (`MAX_TRACE_FILE_SIZE_MB = 64` / `MAX_TRACE_FILE_BYTES`) in the serializer. The loader's `MAX_TRACE_LOAD_BYTES` derives from it, the serializer clamps its byte budget to it, `getTraceMaxFileSizeMBSetting()` clamps larger configured values (with a warning in the trace log), and the manifest now declares `"maximum": 64` with an honest description (xlf regenerated). Tests cover the setting clamp/fallback, constant parity between save and load, and the manifest min/max/default.
3. **Cancelled `prepareLanguageModelChat` showed the missing-API-key prompt** (`sdkLanguageModelProviderBase.ts:114`) — confirmed. `!apiKey || token.isCancellationRequested` routed a cancellation that landed during key resolution into `showMissingApiKeyMessage()`. Fix: cancellation returns `[]` before the key check. Tests: `stays silent when cancelled while resolving the API key` (cancels the token from inside `secrets.get`; fails on the old code) plus `prompts once for a missing API key`, which pins that the prompt still fires, exactly once, in the non-cancelled case.

Also folded in: the OpenAI provider suite used a sinon-chai assertion without registering the plugin, so it only passed when another suite happened to load first. Replaced with a plain sinon check so the file runs standalone.

### CI

The `build-and-test` failure on 650078adc was infrastructure: `vscode-test` hit `AggregateError [ETIMEDOUT]` downloading VS Code Insiders before a single unit test ran (unit step exit code 1, then both test-reporter steps failed on empty reports). A re-run of the same head passed (attempt 2, fully green). The 640d3f23f push triggers a fresh run.

### Validated locally (clean clone at 640d3f23f)

- `tsgo` extension typecheck clean; eslint 0 errors; prettier clean on every touched file; `npm run localization` diff limited to the one description string.
- Hosted unit run of the full PR06 suite set: **260 passed / 0 failed** (all three SDK provider suites, trace loader/serializer/persistence, private-preview manifest, nonproduction channels, inline completion provider, schema context, language completion).
- The two new behavioral tests verified to fail with the fixes reverted (2/2) and pass with them applied.
- Bundles: `development` includes the new serializer symbols; the `stable` bundle still contains zero SDK provider code, so the channel stub seam is unaffected.

_(source: Fable)_

PR #22865, opened 2026-09-01
Add private-preview Object Explorer v2 core

kburtram:
## Summary

This change extracts the core Object Explorer v2 implementation from `dev/refactor` as an independent, direct-to-`main` private-preview PR. It builds on the merged SQL Data Plane, metadata-cache, and native language-service foundations; it does not depend on the open AI-completions or Debug Console PRs.

The scope is intentionally a working, testable browsing core rather than every feature envisioned for the final Object Explorer experience. Follow-up PRs can add Query Studio actions and container lifecycle support without making this review carry those dependencies.

## What is included

- A separate `Object Explorer v2 (Private Preview)` tree backed by saved connection profiles and connection groups.
- SQL Data Plane session ownership with explicit connecting, connected, disconnecting, failed, lost, and cancellation states.
- MetadataStore-backed server/database hierarchy and lazy expansion for databases, schemas, tables, views, columns, keys, programmability, security, storage, Service Broker, and other supported catalog sections.
- Bounded asynchronous expansion so a slow or waking server cannot indefinitely block the tree, plus honest loading/error/stale/no-access states.
- Schema grouping, system-database visibility, per-folder filtering, database-object search, copy-name, and copy-qualified-name operations.
- Connection-group create/edit/delete/move and drag/drop integration using the existing shared connection configuration.
- Explicit handoff to established classic Object Explorer handlers for already-supported administration experiences such as backup/restore, profiler, schema compare, table editing, import, DACPAC/BACPAC, notebooks, and related actions.
- Diagnostics/performance classification and four updated OE v2 perftest scenarios using the actual private-preview settings.
- Localized view, command, status, folder, prompt, and error text.
- Eight focused OE v2 test suites plus preview-service and manifest regression coverage.

## Architecture

```text
Saved profiles / connection groups
               |
               v
       OE v2 tree controller
          |             |
          v             v
  session registry   pure node/layout layer
          |
          v
     SQL Data Plane
          |
          v
 metadata coordinator --> shared MetadataStore/cache
```

The v2 session registry owns SQL Data Plane sessions; it does not create a parallel classic `ConnectionManager` session. Metadata coordinators lease and refresh the shared catalog, while the hierarchy/path/readiness layers remain data-only and independently testable. A classic connection is created only when the user explicitly chooses an action whose implementation still belongs to classic Object Explorer.

## Private-preview boundary

The feature is off by default. The view and every one of its 37 contributed commands require the complete gate:

```text
mssql.enableExperimentalFeatures
&& mssql.sqlDataPlane.enabled
&& mssql.objectExplorer.v2.enabled
```

All three settings must be enabled before extension activation; the setting description explicitly requires a window reload for first-time composition. The activation snapshot, command `enablement`, view/menu/keybinding `when` clauses, and runtime handler checks use the same dependency path. With any prerequisite off, no OE v2 view, command, menu, button, or keybinding appears, and classic Object Explorer remains unchanged and available.

## Deliberate follow-ups / exclusions

- **Query Studio integration:** New Query, SELECT TOP, script-as-create/drop/execute, and other Query Studio routes are not contributed here. They will arrive with the Query Studio PRs, after that consumer exists on `main`.
- **Container lifecycle:** saved container profiles keep their display identity and icons, but Docker discovery/preflight and start/stop/delete lifecycle commands are excluded. They can be reviewed as a side or follow-up PR.
- **Entra token composition:** this branch does not duplicate the VS Code/STS token adapter currently owned by the independent AI-completions PR. SQL and Integrated Authentication paths are available in this core; token-requiring profiles fail honestly until the shared adapter is available and composed in a follow-up.
- **Debug Console / AI completions:** no code or UI from either open track is included.

The omitted implementation remains preserved on `dev/refactor` for later, narrower ports.

## Validation

- Full repository build: passed (Extension Toolkit, MSSQL, SQL Database Projects, and Data Workspace).
- Full repository lint and formatting checks: passed.
- MSSQL TypeScript emit/typecheck and extension bundle: passed.
- Localization extraction and generated bundle/XLIFF refresh: passed (3,238 strings).
- Focused hosted validation: 101 passed across the eight OE v2 suites plus private-preview service/manifest coverage.
- Full coverage-free MSSQL hosted suite: all OE v2 and affected preview tests passed. The only failures were five existing Windows `QueryCompletionSoundService` path/process assertions; both the implementation and tests are byte-identical to `origin/main`, and the failures reproduce in isolation.
- Perftest install/build/typecheck: passed. All runnable non-live suites passed 347 tests with one intentional skip. The optional central-store integration was unavailable because no SQL Server was listening on `localhost:14333`; excluding that infrastructure-only file, the CLI lane passed 251/251.
- Online production packaging with released STS `6.0.20260827.1`: passed, producing an ignored local 1,480-file VSIX at 102.78 MiB.
- Manifest audit: 37/37 commands use the full private-preview gate; the view uses the same gate; deferred Query Studio and Docker lifecycle commands are absent.
- Initial CodeQL correctly rejected `Math.random()` in the legacy-handoff owner-URI suffix. Follow-up commit `1f6486c82` uses `crypto.randomUUID()`; typecheck, focused ESLint, extension emit/bundle, and the hosted handoff suite (5/5) passed after the repair.

## Review guidance

Material findings in session lifecycle, cancellation, metadata freshness/readiness, hierarchy correctness, classic handoff targeting, private-preview visibility, localization, or disposal are in scope. Requests to add Query Studio delivery, container lifecycle, AI-completion UI, or Debug Console integration should be handled in their owning follow-up PRs.


extensions/mssql/src/objectExplorer/v2/commands/sqlIdentifierFormatter.ts
```diff
@@ -0,0 +1,19 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * SQL identifier formatting for OE v2 native commands (oe_view_design
+ * §11.4): bracket-quote identifiers (']' doubled), qualified names, and
+ * qualified names. Pure module; identifiers always pass through bracketQuote.
+ */
```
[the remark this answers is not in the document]
kburtram: Fixed in `9e7111e8e`: removed the duplicated phrase and clarified that the formatter bracket-quotes identifiers and composes qualified names. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/commands/oeV2GroupCommands.ts
```diff
@@ -0,0 +1,335 @@
...
+            }
+            // Creating from a group node's context menu nests under it.
+            const parentId = groupIdOf(node);
+            const dialog = instantiationService.createInstance(
+                ConnectionGroupWebviewController,
+                undefined,
+            );
+            dialog.revealToForeground();
+            emitGroupMutation("create", "ok");
+            void parentId; // dialog assigns root parent; nesting via move/DnD
```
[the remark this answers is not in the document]
kburtram: Fixed in `9e7111e8e`: removed the unused node/parent calculation and corrected the comment to state that the classic dialog creates at the root; nesting remains available through move or drag-and-drop. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Readiness.ts
```diff
@@ -0,0 +1,140 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Readiness → child-node policy (oe_view_design §9.4/§13): ONLY `readyEmpty`
+ * produces a no-items child. Loading, failed, stale, partial, permission
+ * denied, unsupported, and data-plane-unavailable states each synthesize
+ * their own explicit status/error child — failure is never emptiness.
+ */
```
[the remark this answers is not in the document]
kburtram: Fixed in `9e7111e8e`: the header now documents both no-items cases—explicit `readyEmpty` and `ready` with zero children—while keeping failure and other readiness states distinct from emptiness. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/activation.ts
```diff
@@ -0,0 +1,537 @@
...
+            async (node?: OeV2Node) => {
+                if (!deps.isEnabled()) {
+                    return;
+                }
+                const connectionId = connectionIdOf(node);
+                if (!connectionId || !controller) {
+                    return;
+                }
+                const connected = await controller.connectProfile(connectionId);
+                if (!connected) {
```
kburtram: **B — Cancel Connection produces a spurious "could not connect" error.** `connectProfile` resolves `false` whenever the registry snapshot is not `connected`, and after `cancelConnect()` the pending open settles through the registry's superseded branch with `state: "disconnected"` and no `failureReason`. So the flow *Connect → Cancel Connection* shows `ObjectExplorerV2.couldNotConnect()` seconds later, when the cancelled open finally settles under its backend deadline.

Reproduced against the compiled registry (`out/`): `connect()` → `cancelConnect()` → resolve the opener → the `connect()` promise resolves with `state: "disconnected"`, `failureReason: undefined`, and the late session is closed (good), but the caller sees plain failure.

Fix: only toast when the registry reports a real failure, e.g. `const state = registry?.stateOf(connectionId); if (!connected && state === "failed") …` — or have `connectProfile` return the final state (`"connected" | "cancelled" | "failed"`) so both this handler and the auto-connect path can distinguish the user's own cancel from an error. A test in `objectExplorerV2AsyncExpansion.test.ts` that cancels mid-open and asserts no error surfaces would pin it.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/sessions/oeV2SessionRegistry.ts
```diff
@@ -0,0 +1,318 @@
...
+            return snapshotOf(existing);
+        }
+        if (existing && existing.state === "connecting") {
+            // JOIN the in-flight attempt (live-run finding: the deployment
+            // wizard's explicit connect races the single-new-profile
+            // auto-connect). A "still connecting" snapshot here read as
+            // failure to callers awaiting the outcome.
+            return existing.pending ?? snapshotOf(existing);
+        }
+        const entry: Entry = {
```
kburtram, line 125: **B — Reconnecting a `lost` connection leaks the previous session and its state subscription.** `connect()` only short-circuits on `connected`/`connecting`; for an existing `lost` (or `failed`) entry it builds a fresh `Entry` and overwrites the map slot without disposing `existing.stateSubscription` or closing `existing.session`. The lost `ISqlSession` therefore stays registered in `SqlDataPlaneService`'s session map (it shows up in data-plane status and is never released until extension dispose), and its old listener can still fire against the detached entry.

Reproduced against the compiled registry: connect → emit `lost` → `connect()` again → new session live, `first.close()` call count = 0.

`recordPreparationFailure()` already does the right cleanup (`previous?.stateSubscription?.dispose(); void previous?.session?.close()`); factor that into a small `retire(existing)` helper and call it here before creating the new entry. `lostConnection` advertises `canConnect`, so this is the normal recovery path, not an edge case.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/activation.ts
```diff
@@ -0,0 +1,537 @@
...
+    const removeSavedProfile = async (connectionId: string): Promise<void> => {
+        const config = (deps.groupConfig ?? (() => undefined))();
+        if (!config) {
+            return;
+        }
+        const profile = (await config.getConnections()).find(
+            (candidate) => stableProfileId(candidate as never) === connectionId,
+        );
+        if (profile) {
+            await config.removeConnection(profile as never);
```
kburtram: **B — Remove Connection bypasses `ConnectionStore.removeProfile`, leaving the saved password and the recent-connections entry behind.** Classic `ObjectExplorerService.removeNode` calls `connectionStore.removeProfile(profile, false)`, which removes the profile from settings **and** from the recently-used list **and** deletes the credential-store entry when `savePassword === true`. This handler goes straight to `ConnectionConfig.removeConnection`, so a v2 "Remove Connection" orphans the stored password in the OS credential store / secret storage and keeps the profile in the MRU list. The command reuses classic's confirmation wording ("delete" semantics), so users will reasonably expect classic's cleanup.

`deps.profiles` is already the `ConnectionStore` instance (see `mainController.ts` wiring), so the fix is to widen the seam (`removeProfile(profile, keepCredentialStore?)`) and call it here instead of `groupConfig().removeConnection`. Worth a test in `objectExplorerV2Commands.test.ts` asserting the credential path is exercised.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/activation.ts
```diff
@@ -0,0 +1,537 @@
...
+                if (!outcome.ok && outcome.error) {
+                    void vscode.window.showErrorMessage(outcome.error);
+                }
+            }),
+        ),
+        vscode.commands.registerCommand("mssql.objectExplorerV2.showStatus", async () => {
+            if (!deps.isEnabled()) {
+                return;
+            }
+            const channel = vscode.window.createOutputChannel(ObjectExplorerV2.outputChannelName);
```
kburtram: **B — `showStatus` creates a new OutputChannel on every invocation and never disposes it.** Each run of *Object Explorer v2: Show Status* adds another "MSSQL Object Explorer v2" entry to the Output panel's dropdown for the life of the window. Create the channel once (lazily) inside `activateObjectExplorerV2`, push it onto `context.subscriptions`, and `clear()` + `appendLine` on each invocation — the same pattern `models/logger.ts` uses.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/metadata/oeV2MetadataCoordinator.ts
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    // -- server catalog -------------------------------------------------------
+
+    async ensureServer(): Promise<ServerCatalogLease> {
+        if (this.serverLease) {
+            return this.serverLease;
+        }
+        if (!this.serverPending) {
+            this.serverPending = this.store.acquireServer(this.prepared).then((lease) => {
```
kburtram: **B — A rejected lease acquisition is cached forever, so "Refresh to retry" cannot recover.** `serverPending` is assigned the `acquireServer(...).then(...)` chain and is only ever cleared in `dispose()`; `ensureDatabase` likewise deletes `databasePending` only inside the success `.then`. If `acquireServer`/`acquireDatabase` rejects once (the store's `assertAcquisitionAllowed`/`assertLive` guards, or a construction failure), every later `ensureServer()` / `ensureDatabase(db)` / `refreshServer()` / `refreshDatabase(db)` returns the same rejected promise for the lifetime of the coordinator — i.e. until the user disconnects and reconnects — while the tree keeps rendering `ObjectExplorerV2.metadataFailed` ("Refresh to retry").

Fix: clear the pending slot on rejection as well (`.catch(err => { this.serverPending = undefined; throw err; })`, and `.finally(() => this.databasePending.delete(database))` for the database map), so a retry re-attempts acquisition. A unit test with a store whose first `acquireDatabase` rejects and second resolves would pin it.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/activation.ts
```diff
@@ -0,0 +1,537 @@
...
+        // Ctrl+C keybinding parity: fall back to the focused tree selection.
+        () => activeView?.selection[0],
+        deps.isEnabled,
+    );
+
+    // B27: while any connection is opening/closing, tick the tree so the
+    // slow-connect elapsed description ("connecting… (12s)") stays live.
+    const connectingTicker = setInterval(() => {
+        if (registry?.anyConnecting()) {
+            controller?.refresh();
```
kburtram: **C — The 2 s connecting ticker invalidates and re-renders the whole tree.** `controller.refresh()` drops the cached profile tree and fires a root-level change, so while any connection is opening (a sleeping serverless can take 30–60 s) the view re-reads all saved profiles/groups from settings and re-renders every node every two seconds, just to advance the "connecting… (Ns)" description. A targeted `fireChange(connectionNode)` (or a `controller.touchConnection(id)` that fires the change for that node's id only) keeps the elapsed counter live without the full refresh.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/activation.ts
```diff
@@ -0,0 +1,537 @@
...
+                        throw new Error("OE v2 data-plane connect failed");
+                    }
+                    const server = (await controller.children()).find(
+                        (node) => node.kind === "connectedServer",
+                    );
+                    if (!server) {
+                        throw new Error("no connected server node after connect");
+                    }
+                    const security = (await controller.children(server)).find(
+                        (node) => node.label === "Security",
```
kburtram: **C — PERF_MODE probes match on localized labels.** Folder labels go through `ObjectExplorerV2.folderLabel(...)` (l10n), so `node.label === "Security"` / `"Logins"` only holds under an English display language; the `objectExplorerV2SecurityExpand` scenario would fail on a localized runner with "no Security folder". Match on the path instead (`node.path.kind === "serverFolder" && node.path.folder === "security"`, and the `logins` folder id), which is what the perftest registry keys on anyway.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/sessions/oeV2ProfileAdapter.ts
```diff
@@ -0,0 +1,109 @@
...
+export async function readProfileTree(source: ConnectionProfileSource): Promise<OeV2ProfileTree> {
+    const [groups, connections] = await Promise.all([
+        source.readAllConnectionGroups().catch(() => [] as OeV2StoredGroup[]),
+        source.readAllConnections(false).catch(() => [] as OeV2StoredProfile[]),
+    ]);
+    const groupRecords: OeV2GroupRecord[] = groups
+        .filter((group) => group.id)
+        .map((group) => ({
+            groupId: group.id!,
+            name: group.name ?? "Group",
```
kburtram: **C — Hard-coded fallback group label.** `"Group"` is user-facing (it becomes the tree label for a stored group without a name). Route it through `locConstants.ts` (e.g. `ObjectExplorerV2.unnamedGroup`) so the fallback localizes with the rest of the view.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/tree/oeV2TreeController.ts
```diff
@@ -0,0 +1,747 @@
...
+        connectionId: string,
+        database: string,
+        term: string,
+        limit = 50,
+    ): Promise<{ schema: string; name: string; kind: string }[]> {
+        const runtime = this.runtimes.get(connectionId);
+        if (!runtime) {
+            return [];
+        }
+        await runtime.coordinator.ensureDatabase(database).catch(() => undefined);
```
kburtram: **C — Search awaits lease acquisition without the expansion bound.** Every expand path goes through `boundedWait(..., EXPAND_WAIT_MS)` so a sleeping/serverless connection cannot hang the tree, but `searchObjects` awaits `ensureDatabase(database)` unbounded, so the Search quick pick can sit in its loading state indefinitely on the same kind of connection. Wrapping it in the same `boundedWait` (and surfacing "metadata still loading — try again" when it returns `undefined`) keeps the two paths consistent.

(source: Fable)

extensions/mssql/test/unit/objectExplorerV2CommandRegistry.test.ts
```diff
@@ -0,0 +1,301 @@
...
+        expect(target.database).to.equal("AppDb");
+        expect(target.schema).to.equal("dbo");
+        expect(target.objectName).to.equal("Orders");
+    });
+});
+
+suite("Object Explorer v2 legacy redirect (B25)", () => {
+    let execute: sinon.SinonStub;
+
+    setup(() => {
```
kburtram, line 220: **B — The B25 legacy-redirect suite only passes when an earlier suite happened to initialize `IconUtils`.** `redirectToClassic` builds a real `TreeNodeInfo`, whose constructor eagerly calls `ObjectExplorerUtils.iconPath` → `IconUtils.getIcon` → `vscode.Uri.joinPath(this._extensionUri, …)`. Nothing in this file initializes `IconUtils`, so run standalone (`vscode-test --run out/test/unit/objectExplorerV2CommandRegistry.test.js`, fresh `out/`) four tests fail deterministically: the three `outcome.ok === true` assertions (lines 235/247/265) and the "classic exploded" assertion (line 284) all see the guarded error `The legacy feature could not run with an Object Explorer v2 node (…)` instead of a successful redirect. On CI's full run an alphabetically earlier suite calls the helper first, which is why the job is green.

`objectExplorerV2Handoff.test.ts` already does the right thing (`initializeIconUtils(); // TreeNodeInfo resolves icons in its ctor` in its `setup`). Add the same `initializeIconUtils()` call from `./utils` to this `setup`, and the suite becomes order-independent. (This is also what produced the "4 unrelated test-host races" noted in the PR's validation write-up — it is not a race, and not tied to the Insiders build.)

(source: Fable)

extensions/mssql/src/objectExplorer/v2/tree/oeV2ConnectionLabel.ts
```diff
@@ -0,0 +1,196 @@
...
+ * the classic function. Karl's addition over v1: when two sibling nodes tie
+ * on the full label, each tooltip gains the properties that differ.
+ */
+
+/** Classic auth-type wire values (constants.ts). */
+const SQL_AUTH = "SqlLogin";
+const AZURE_MFA = "AzureMFA";
+const INTEGRATED = "Integrated";
+
+const DEFAULT_DATABASE_LABEL = "<default>";
```
kburtram: **A — Connection labels/tooltips are hard-coded English and ship in the tree.** This is the one Copilot item from the last two rounds that is still open at HEAD (the file is byte-identical to the `dev/refactor` original). The strings are user-facing: `connectionDisplayLabel` becomes the tree label via `oeV2ProfileAdapter.ts`, and `connectionTooltipLines`/`disambiguationLines` feed the tooltip in `oeV2NodeFactory.ts`. Remaining literals: `"<default>"` (23); the tooltip keys `"Server"`, `"Database"`, `"Authentication Type"`, `"User"`, `"Port"`, `"SQL Container Name"`, `"SQL Container Version"`, `"Application Intent"`, `"Connection Timeout"`, `"Command Timeout"`, `"Always Encrypted"`, `"Replication"` (90–101); `"Azure MFA"` / `"Windows Authentication"` (122); `"Enabled"` / `"Disabled"` (125/128); `"(not set)"` (192).

Identical-meaning constants already exist in `locConstants.ts` and are what classic uses (`connectionNode.ts`, `connectionInfo.ts`): `defaultDatabaseLabel`, `server`, `database`, `authenticationType`, `user`, `port`, `sqlContainerName`, `sqlContainerVersion`, `applicationIntent`, `connectionTimeout`, `commandTimeout`, `alwaysEncrypted`, `replication`, `azureMFA`, `windowsAuthentication`, `enabled`, `disabled`. Only `"(not set)"` needs a new `ObjectExplorerV2` entry. In a localized VS Code the v2 label therefore diverges from the classic label for the same profile; the parity tests only pass because they run in English.

The header's "PURE module (tree/ may not import vscode/l10n)" rationale no longer holds — `oeV2Browse.ts`, `oeV2Readiness.ts`, and `oeV2NodeFactory.ts` in the same folder already import `locConstants`. Fix: import the constants above, add `ObjectExplorerV2.notSet`, and update the stale purity comments here (line 12) and in `oeV2Node.ts`.

(source: Fable)

extensions/mssql/test/unit/objectExplorerV2Browse.test.ts
```diff
@@ -0,0 +1,804 @@
...
+
+    teardown(() => {
+        // THE TRIPWIRE: every browse test runs entirely without STS v1.
+        sinon.assert.notCalled(sendRequest);
+        sinon.assert.notCalled(connect);
+        sendRequest.restore();
+        connect.restore();
+    });
+
+    test("expanding a disconnected saved profile opens it and renders server children", async () => {
```
kburtram, line 395: **B — Five browse tests were dropped in the extraction that do not belong to an excluded feature.** `git diff dev/refactor..HEAD` on this file removes eight tests. Three are legitimately out of scope (two DOCK-3 container tests, the AzureMFA token-source test). The other five cover core behaviour that is still in this PR and now has no coverage anywhere under `test/unit/objectExplorerV2*`:

- "a second connect during an in-flight open JOINS it (wizard vs auto-connect race)"
- "cancel while connecting returns to disconnected; the late session self-closes"
- "connect → server node; Databases folder renders catalog states honestly"
- "server catalog failure renders an error child, never an empty list"
- "database → structural folders → objects, functions merge kinds, schemas list"

Only one assertion in the last of these (`tables[0].capabilities.canSelectTop`) touched a deferred Query Studio capability. Together they were the only tests of: registry join/cancel semantics, the System Databases folder, inaccessible-database rendering (`collapsible=false`, "no access"), `showSystemDatabases=false`, server-catalog failure → error child, the function kind-merge, the Schemas listing, and the group-by-schema level. Please restore them, deleting just the `canSelectTop` line — the cancel/join ones in particular would have caught the registry issues noted on `oeV2SessionRegistry.ts`.

(source: Fable)

kburtram (commented):
## Object Explorer v2 core — review of `9e7111e8e`

Read the whole change against the `dev/refactor` originals, the private-preview gating contract from PR04, and the classic handlers the handoff targets; ran the full local battery in a clean clone; reproduced the behavioural findings under node against the compiled modules. Inline comments carry the detail — this is the outline of what makes the branch mergeable.

### A — should fix before merge (2)

- **Localization, connection labels/tooltips** (`tree/oeV2ConnectionLabel.ts`): `"<default>"`, the tooltip field names, `"Azure MFA"`/`"Windows Authentication"`, `"Enabled"`/`"Disabled"`, `"(not set)"` are hard-coded English in shipped tree text; identical-meaning constants already exist in `locConstants.ts`. This and the next item are the two Copilot "previously missed" findings that are still open at HEAD.
- **Localization, redirect error strings** (`legacy/oeV2LegacyRedirect.ts`): every `RedirectOutcome.error` is shown via `showErrorMessage` and none are localized.

### B — recommended (13)

Lifecycle / correctness (each reproduced under node unless noted):
- Cancel Connection → the Connect command later toasts "could not connect" (`activation.ts`, `connectProfile` cannot tell cancel from failure).
- Reconnecting a `lost` connection never closes the old session or its subscription (`oeV2SessionRegistry.ts`).
- Remove Connection bypasses `ConnectionStore.removeProfile`, so the saved password and MRU entry survive (`activation.ts`).
- A rejected lease acquisition is cached for the coordinator's lifetime, defeating "Refresh to retry" (`oeV2MetadataCoordinator.ts`; reachability of the reject is narrow today).
- `showStatus` creates a new OutputChannel per invocation (`activation.ts`).
- "Compare Schemas..." reaches the classic handler with one positional arg, which it ignores → empty dialog (`oeV2LegacyRedirect.ts`).
- The handoff passes the cached profile object by reference and classic connect writes the password onto it (`oeV2LegacyRedirect.ts`, `oeV2LegacyNodeAdapter.ts`).
- The handoff's 10-minute idle TTL disconnects the URI under long-lived Restore/Import/DACPAC dialogs (`oeV2ClassicHandoffService.ts`).
- "Delete group and contents" leaves the subtree's sessions/handoffs open (`oeV2GroupCommands.ts`).
- `catalogGate` renders `absent`/`loading` sections as "No items" (`oeV2Browse.ts`; latent — not proven reachable).
- Group-by-schema sorts schemas case-sensitively while every other list is case-folded (`oeV2Browse.ts`).

Tests:
- Five non-excluded browse tests were dropped in extraction (join-in-flight, cancel, catalog states, server-catalog failure, structural folders) — restore them minus the one `canSelectTop` line (`objectExplorerV2Browse.test.ts`).
- The B25 legacy-redirect suite depends on another suite having called `initializeIconUtils()`; standalone it fails 4/12 deterministically — this is the "four test-host races" from the validation note, and it is neither a race nor Insiders-specific (`objectExplorerV2CommandRegistry.test.ts`).

### C — nits and questions (14)

Ticker does a full-tree `refresh()` every 2 s while connecting; PERF probes match localized labels; `"Group"` fallback label; unbounded `ensureDatabase` in search; `sortLast` comparator; `decodePath` validation; `\s` test literal; dead `synthesizeChildren` + unreachable return-value branch; stale notice after sub-folders; External Tables ignoring `groupBySchema` (question); unreachable `copyName`/`legacyActions` + `policiesForNode` ignoring `objectKind` + dead `label`/`title` fields; search description shows the raw kind token; Copy Connection String missing on `lostConnection`; real timers in the TTL test.

### Verified locally (clean clone at `9e7111e8e`)

- `tsgo` extension typecheck, eslint (v2 + tests + touched shared files), prettier, extension emit, esbuild bundle: all clean.
- Hosted OE v2 selection (8 suites + preview service + manifest): 93 passed / 4 failed — the 4 are the isolation gap above and pass inside the full run.
- Full MSSQL unit suite: **4,847 passed / 6 failed / 10 skipped**. All 6 failures are environmental: the 5 Windows-host `queryCompletionSoundService` assertions that are byte-identical to main, and the ES localization test that expects a generated `bundle.l10n.es.json` my clone never built.
- Perftest: install/build/typecheck pass; 31 + 49 + 16 + 251 tests pass; the only failure is the live central-store test that needs a SQL Server on `localhost:14333` (same as every PR in this series).
- Pure modules exercised under node from `out/`: session registry (cancel/lost), hierarchy registry invariants, path codec, browse rendering, identifier formatting.

### Checked and found sound

- **Gating**: every one of the 37 `mssql.objectExplorerV2.*` commands, the view, all `view/title` / `view/item/context` menus, the Command Palette entries (32 hidden, 5 gated), and the Ctrl+C keybinding (`focusedView == mssql.objectExplorerV2`) carry the complete umbrella + SQL Data Plane + OE v2 gate; every handler re-checks the live `isEnabled()`; disabling at runtime disposes the view, sessions, coordinators and handoffs; first-time enable correctly requires reload.
- **Manifest ↔ code conformance**: all 33 registered ids are contributed and vice versa; every `oe2:cmd=`/`oe2:can*`/`oe2:kind=` flag used in a `when` clause is produced by `commandFlagsFor`/`contextValueFor`; deferred Query Studio and Docker commands are absent.
- **No-v1 boundary**: browse never touches `ConnectionManager`/STS v1 (tripwire test); a classic connection is created only by explicit handoff, with a `crypto.randomUUID()` owner URI, reuse per v2 connection, and close on disconnect/remove/unregister/deactivate.
- **Handoff targeting** for backup/restore/profiler/table designer/table explorer/schema designer/DAB/rename/drop/import/copy-connection-string/Copilot/DACPAC/notebook: node types, metadata, URNs and scoped profiles match what the classic handlers read; DB-scoped connections are re-kinded correctly.
- **Extraction drops** vs `dev/refactor`: only Query Studio, Docker lifecycle/preflight, the Entra token source, and literal→l10n replacements; the five dropped tests above are the one exception.
- **Privacy**: diag/perf emissions carry classes and a truncated fingerprint only; no SQL, server names, credentials or error text in telemetry; failure text is surfaced only in the user's own tree/toasts.
- Connection-group create/edit/delete/move/drag-drop semantics match classic (cycle guard, same-group no-op, foreign MIME ignored, `groupConfig()` undefined handled everywhere); identifier bracket-quoting doubles `]` correctly.

### Merge-readiness checklist

1. Localize `oeV2ConnectionLabel.ts` and `oeV2LegacyRedirect.ts` (A ×2) and refresh the l10n bundle/xlf.
2. Fix the cancel toast, lost-session cleanup, Remove Connection credential cleanup, and rejected-lease caching; restore the five browse tests so the registry semantics are pinned.
3. Schema-compare arity, profile copy on handoff, handoff TTL policy, group-delete teardown.
4. `initializeIconUtils()` in the B25 suite so the command-registry file runs standalone; then the local OE v2 selection should be 97/97.
5. C-tier items at the author's discretion; the External Tables question just needs an answer.

Follow-ups explicitly out of scope, as the PR states: Query Studio actions, container lifecycle, Entra composition, Debug Console/AI-completion surfaces.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Browse.ts
```diff
@@ -0,0 +1,1080 @@
...
+    section: string,
+): OeV2Node[] | undefined {
+    if (!status || status.readiness === "absent" || status.readiness === "loading") {
+        return [loadingNode(scope, connectionId)];
+    }
+    if (status.readiness === "failed" || !snapshot) {
+        return [errorNode(scope, ObjectExplorerV2.metadataFailed, connectionId)];
+    }
+    const sectionState = (snapshot.readiness as Record<string, string>)[section];
+    if (sectionState === "failed") {
```
kburtram, line 416: **B — `catalogGate` treats an `absent`/`loading` section as ready, so a not-yet-hydrated section renders "No items".** `SectionState` is `"absent" | "loading" | "ready" | "failed" | "stale" | "lite"` (`catalogModel.ts`), but only `failed` is gated here; every other value falls through to "render real children", and an empty section list then becomes the `noItems` child — the exact "silent-empty lie" the readiness layer is designed to prevent. Reproduced with the compiled module: `objectFolderChildren(columns)` over a snapshot whose `readiness.columns` is `absent`, `loading`, `lite`, or `stale` returns `['noItems: No items']` instead of `Loading…`.

Caveat: I could not prove this is reachable today — the live hydrate path only emits `ready`/`failed` for columns/keys/foreignKeys/parameters, and I did not trace whether the persisted-cache rehydrate policy intersection can hand back `absent`/`loading` sections. It is a one-line defensive fix either way: return `[loadingNode(scope, connectionId)]` for `"absent"`/`"loading"` and keep `ready`/`lite`/`stale` rendering.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Browse.ts
```diff
@@ -0,0 +1,1080 @@
...
+        const withObjects = new Set(
+            snapshot
+                .listObjects(undefined, objectKinds)
+                .filter((o) => visibleWithFacets(facets.get(o.objectId)))
+                .map((o) => o.schema),
+        );
+        if (withObjects.size === 0) {
+            return [noItemsNode(scope, connectionId)];
+        }
+        return [...withObjects].sort().map((name) => {
```
kburtram: **B — Group-by-schema orders schemas case-sensitively, unlike every other list in the view.** Default `Array.prototype.sort()` is UTF-16 code-unit order, while `listObjects`/`listSchemas` use the catalog's case-folded `ordinalCompare`. Reproduced: schemas `dbo`, `Sales` render as `Sales`, `dbo` under group-by-schema, but `dbo`, `Sales` in the Schemas folder and in the flat Tables list — so toggling the setting visibly reorders schemas. Use a case-insensitive comparator here (a local `localeCompare`-free fold keeps the module free of service imports).

(source: Fable)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Hierarchy.ts
```diff
@@ -0,0 +1,727 @@
...
+            }
+            if (def.isSystemFolder && facts.isSystemDatabase !== true) {
+                return false;
+            }
+            if (def.presence === "nonEmpty" && options.hasItems && !options.hasItems(def)) {
+                return false;
+            }
+            return true;
+        })
+        .sort((a, z) => (a.sortLast === z.sortLast ? a.order - z.order : a.sortLast ? 1 : -1));
```
kburtram: **C — `resolveFolders` comparator is inconsistent when `sortLast: false` and `undefined` mix.** `a.sortLast === z.sortLast` is false for `false` vs `undefined`, so the comparator returns `-1` for both `cmp(a, b)` and `cmp(b, a)` (reproduced: `{a: order 1, sortLast: false}, {b: order 0}, {c: order 2, sortLast: false}` sorts to `c, b, a`). Today's registry only uses `true`/absent, so it is latent; `!!a.sortLast === !!z.sortLast` makes it total.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Path.ts
```diff
@@ -0,0 +1,321 @@
...
+                    folder: dec(parts[4]),
+                };
+            case "object":
+                return {
+                    kind,
+                    connectionId: dec(parts[1]),
+                    database: dec(parts[2]),
+                    schema: dec(parts[3]),
+                    name: dec(parts[4]),
+                    objectKind: parts[5] as OeV2ObjectKind,
```
kburtram: **C — `decodePath` accepts truncated ids and unvalidated enum casts.** `decodePath("oe2:connection")` yields `{ connectionId: "undefined" }` (`decodeURIComponent(undefined)`), and `"oe2:object/c1/db/dbo/T/bogusKind"` decodes with `objectKind: "bogusKind"`, which makes `objectChildren` throw `Cannot read properties of undefined (reading 'map')`; the `folder` cast on the next branch has the same shape. There are no production callers yet (tests only), so this is hygiene for when ids start coming back from VS Code state: return `undefined` when a required segment is missing or the `objectKind`/`folder` value is not in the closed set.

(source: Fable)

extensions/mssql/test/unit/objectExplorerV2Hierarchy.test.ts
```diff
@@ -0,0 +1,689 @@
...
+        const nodes = serverAuxFolderChildren("c1", "security/nonsense", {}, undefined, undefined);
+        expect(nodes[0].kind).to.equal("error");
+    });
+
+    test("path codec: serverObjectItem round-trips", () => {
+        const path = {
+            kind: "serverObjectItem" as const,
+            connectionId: "c1",
+            folder: "security/logins",
+            name: "CONTOSO\svc account",
```
kburtram: **C — This literal does not contain the backslash it means to test.** `"CONTOSO\svc account"` is `"CONTOSOsvc account"` (`\s` is not an escape sequence), so the hostile-identifier round-trip never exercises `\`. Use `"CONTOSO\\svc account"`.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Readiness.ts
```diff
@@ -0,0 +1,140 @@
...
+
+export type ChildSynthesis =
+    | { kind: "children" } // render the real children
+    | { kind: "noItems" }
+    | { kind: "loading" }
+    | { kind: "status"; message: string; retryable: boolean }
+    | { kind: "error"; message: string; retryable: boolean };
+
+/** Decide what a container renders for a given readiness state. */
+export function synthesizeChildren(readiness: OeV2Readiness, childCount: number): ChildSynthesis {
```
kburtram: **C — Dead code.** `synthesizeChildren` has no callers in `src/` (only the shell test exercises it); the rendering paths use `loadingNode`/`errorNode`/`noItemsNode` directly. Related: in `oeV2Browse.ts` the parameter list filter `parameter.ordinal > 0 || parameter.name !== ""` (line 1049) drops every ordinal-0 unnamed row, which is exactly the return-value row, so the `(return)` / `ObjectExplorerV2.returnValue` branch a few lines below is unreachable. Either delete both, or drop the filter if the return value is meant to render under Parameters.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Browse.ts
```diff
@@ -0,0 +1,1080 @@
...
+                  ),
+            freshness,
+        );
+    } else if (def.section === "aux") {
+        // Pure parent folder: registry children ARE the content.
+        content = before.length + after.length === 0 ? [noItemsNode(scope, connectionId)] : [];
+    } else {
+        content = databaseAuxLeafContent(scope, connectionId, database, def, facts, aux);
+    }
+    return [...before, ...content, ...after];
```
kburtram: **C — The stale-metadata notice renders after the leading sub-folders.** `withStaleNotice` is applied to `content` only, so a stale Tables folder renders as `System Tables`, then the "Metadata not validated…" status child, then the objects, then `External Tables`. Applying the notice to the combined `[...before, ...content, ...after]` puts it first, where the other status children sit.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Browse.ts
```diff
@@ -0,0 +1,1080 @@
...
+                database,
+                def,
+                snapshot!,
+                groupBySchema,
+                schema,
+                aux,
+            ),
+            freshness,
+        );
+    } else if (def.facetFlag === "isExternal" && def.objectKinds) {
```
kburtram, line 484: **C — Question: External Tables ignores `groupBySchema`.** With the setting on, Tables/Views gain a schema level but this branch still lists `dbo.Ext (External)` flat. If that is intentional v1 parity, a one-line comment here would save the next reader the question; otherwise route it through `databaseFolderContent` with the facet predicate. Same question, lower stakes, for the raw `state_desc` note at line 338 (`offline`, `read-only` are server strings lower-cased rather than the localized wording classic uses).

(source: Fable)

extensions/mssql/src/objectExplorer/v2/legacy/oeV2LegacyRedirect.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+export async function redirectToClassic(
+    feature: string,
+    node: OeV2Node,
+    deps: { facts: RedirectFactsSource; handoff: OeV2ClassicHandoffService },
+): Promise<RedirectOutcome> {
+    const policy = LEGACY_COMMAND_POLICIES.find((entry) => entry.feature === feature);
+    if (!policy || !node.connectionId) {
+        return { ok: false, error: "This action is not available here." };
```
kburtram: **A — Hard-coded user-facing error strings in the redirect library.** `RedirectOutcome.error` is shown verbatim by `activation.ts` (`void vscode.window.showErrorMessage(outcome.error)` in both the `legacyActions` picker and every `legacyRedirect` command), so these are shipped UI text: this line, `"This action is not available on this node."` (66), `"This action needs a database — use a database node."` (71), `"Connect this profile in Object Explorer v2 first."` (75), the interpolated `"node kind not adaptable"` (108), and the template at 117–119 (`The legacy feature could not run with an Object Explorer v2 node (…). Use Classic Object Explorer for this command.`). This is the second still-open Copilot item; `locConstants.ts` has no equivalent constants yet.

Fix: add `ObjectExplorerV2.legacyActionUnavailable`, `legacyActionNotOnThisNode`, `legacyActionNeedsDatabase`, `legacyActionConnectFirst`, and `legacyFeatureFailed(reason)` and return those (keep the raw error text for the diag event only). `objectExplorerV2CommandRegistry.test.ts` line 258 currently pins the English (`expect(outcome.error).to.include("database")`) and will need to assert on the constant instead.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/legacy/oeV2LegacyRedirect.ts
```diff
@@ -0,0 +1,122 @@
...
+    });
+    try {
+        if (policy.level === "h1") {
+            await vscode.commands.executeCommand(policy.classicCommand, ownerUri);
+        } else {
+            const adapted = toLegacyTreeNode(effectiveNode(policy, node), ownerUri, profile);
+            if (!adapted) {
+                throw new Error("node kind not adaptable");
+            }
+            await vscode.commands.executeCommand(policy.classicCommand, adapted);
```
kburtram: **B — "Compare Schemas..." from a v2 node opens Schema Compare with no source.** The classic `mssql.schemaCompare` handler (`mainController.ts` ~2075) only reads the node when it receives at least two positional arguments (`if (args.length >= 2) { sourceNode = args[0]; targetNode = args[1]; … }`); with a single argument it calls `onSchemaCompare(undefined, undefined, false)` and the dialog opens empty. Classic tree menus always pass `(node, selection[])`, so v1 clears the gate; this `executeCommand(policy.classicCommand, adapted)` passes one argument and silently drops the node the user right-clicked, while the v2 menu label promises a pre-filled source.

Fix: give policies an optional `extraArgs` (or special-case `schemaCompare`) so the call becomes `executeCommand(cmd, adapted, undefined)`, and add a redirect test asserting the classic call receives the adapted node as argument 0 for this feature.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/legacy/oeV2LegacyRedirect.ts
```diff
@@ -0,0 +1,122 @@
...
+    // Defense in depth behind the menu gating: a database-scoped feature on
+    // a top-level connection needs the connection to BE a database (K4).
+    if (policy.databaseScoped && node.kind === "connectedServer" && node.database === undefined) {
+        return { ok: false, error: "This action needs a database — use a database node." };
+    }
+    const facts = await deps.facts.handoffFacts(node.connectionId);
+    if (!facts) {
+        return { ok: false, error: "Connect this profile in Object Explorer v2 first." };
+    }
+    const profile = facts.stored as unknown as IConnectionProfile;
```
kburtram: **B — The handoff passes the v2 profile-tree's cached object by reference, and classic connect writes the password onto it.** `handoffFacts` returns `stored: profile.stored` — the same object `readProfileTree` cached in the controller's `tree`. It goes straight to `ensureOwnerUri` → `ConnectionManager.connect(ownerUri, profile)` → `prepareConnectionInfo` → `handlePasswordBasedCredentials`, which assigns in place (`connectionManager.ts` ~1396: `connectionCreds.password = password;`, plus `normalizeServerName` / `azureAccountToken`), and that same object then becomes the ConnectionManager's credentials record. `oeV2LegacyNodeAdapter.ts` line 49 also passes it uncopied when no database override is set, and classic handlers (schema designer, table designer) call `prepareConnectionInfo`/`updateConnectionProfile` on it.

Net effect: the resolved password sits on the v2 profile cache until the next `refresh()`, which contradicts the adapter's "NO credentials ride these records" contract. It is in-memory only (tooltips exclude it, nothing persists it), so this is hygiene rather than a leak, but it is cheap to close: `const profile = { ...facts.stored } as IConnectionProfile;` here, and always spread in `toLegacyTreeNode` (`{ ...profile, ...(database !== undefined ? { database } : {}) }`). A test that the seam's `connect` argument is not `===` the stored record would pin it.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/legacy/oeV2ClassicHandoffService.ts
```diff
@@ -0,0 +1,132 @@
...
+            void this.close(connectionId);
+        }
+    }
+
+    private touch(entry: HandoffEntry, connectionId: string): void {
+        if (entry.idleTimer) {
+            clearTimeout(entry.idleTimer);
+        }
+        const ttl = this.options.idleTtlMs ?? 600_000;
+        entry.idleTimer = setTimeout(() => void this.close(connectionId), ttl);
```
kburtram: **B — The 10-minute idle TTL closes the handoff connection underneath long-lived classic dialogs.** `touch()` only runs from `ensureOwnerUri`, so the timer measures time since the *handoff*, not since last use. Several h2 handlers keep the owner URI for the dialog's lifetime: Restore stores it in state and uses it for `getRestoreConfigInfo` and the restore itself, Flat File Import passes it on every request, and the DACPAC / drop / rename flows resolve the same URI via `getUriForConnection`. A user who spends more than ten minutes in Restore or Import and then clicks Run hits `close()`'s disconnect and gets a "not connected" failure from STS. (Backup is safe — it opens its own `${db}_${ownerUri}` connection.)

The explicit lifecycle already covers every real end-of-life (`close` on v2 disconnect/remove, `dispose` on view unregister and deactivate), so the simplest fix is to drop the idle TTL for features that hand the URI to a dialog, or have `close()` skip while the URI is still connected/in use. At minimum, document the TTL where those dialogs consume the URI.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/commands/oeV2GroupCommands.ts
```diff
@@ -0,0 +1,333 @@
...
+                    { modal: true },
+                    LocalizedConstants.Common.delete,
+                );
+                mode = choice === LocalizedConstants.Common.delete ? "delete" : undefined;
+            }
+            if (!mode) {
+                emitGroupMutation("delete", "canceled");
+                return;
+            }
+            await config.removeGroup(groupId, mode);
```
kburtram, line 206: **B — "Delete group and contents" removes the saved profiles but leaves their v2 sessions and handoff connections open.** `ConnectionConfig.removeGroup(id, "delete")` drops every profile in the subtree from settings. The single-profile path in `activation.ts` deliberately tears down first (`handoff.close` → `controller.disconnectProfile` → remove), but this path does none of that, so data-plane sessions, metadata coordinators, and any classic handoff connection for profiles in the deleted subtree stay alive with no tree node left to disconnect them (until the idle TTL or view dispose). Classic has the same gap, so this is parity-neutral, but v2 already has the hooks: compute the subtree's profile ids (the same walk `wouldCreateCycle` does) and run the two callbacks for each before `removeGroup`, injecting them through `OeV2GroupCommandDeps`.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/commands/oeV2LegacyCommandPolicy.ts
```diff
@@ -0,0 +1,177 @@
...
+    {
+        feature: "createNotebook",
+        label: "New SQL Notebook (legacy)",
+        classicCommand: "mssql.notebooks.createNotebook",
+        level: "h0",
+        nodeKinds: ["connectedServer", "database"],
+    },
+];
+
+export function policiesForNode(kind: OeV2NodeKind, nodeDatabase?: string): LegacyCommandPolicy[] {
```
kburtram: **C — `copyName` and `legacyActions` are contributed but unreachable, and `policiesForNode` ignores `objectKind`.** Both commands have `commandPalette.when: false` and no `view/item/context` entry, although `oe2:canCopyName` is emitted for most nodes — dead surface (same in `dev/refactor`, so not an extraction drop). If `legacyActions` is ever wired, this function offers `editTable`/`tableExplorer` for *every* `object` kind (a view or procedure would launch Table Designer / Table Explorer, and the adapter hard-codes `nodeType: "Table"`); the registry's `tableObject` predicate is right, the policy filter is not. Either add the menu entries (`viewItem =~ /\boe2:canCopyName\b/`, and `legacyActions` on nodes with policies) or remove the two commands and the picker handler, and add an `objectKind` parameter here mirroring `tableObject`. Related dead fields: `LegacyCommandPolicy.label` (line 23) and `OeV2CommandDef.title` (`oeV2CommandRegistry.ts` line 47) are never read — activation uses `ObjectExplorerV2.legacyActionLabel(...)` and package.json uses `%nls%` keys — so they are duplicate English strings that will drift.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/commands/oeV2NativeCommands.ts
```diff
@@ -0,0 +1,152 @@
...
+                if (matches.length === 0) {
+                    void vscode.window.showInformationMessage(
+                        ObjectExplorerV2.noObjectsMatch(term),
+                    );
+                    return;
+                }
+                const picked = await vscode.window.showQuickPick(
+                    matches.map((match) => ({
+                        label: `${match.schema}.${match.name}`,
+                        description: match.kind,
```
kburtram: **C — The search quick pick shows the internal kind token.** `match.kind` is the metadata-store enum value (`table`, `scalarFunction`, …) rendered untranslated as the item description; map it through a loc table the way `folderLabel` does for folders. Also `if (!term)` (line 121) lets a whitespace-only term through to `searchObjects` (harmless — it just reports no matches for `' '`), so `term.trim()` first.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/commands/oeV2CommandRegistry.ts
```diff
@@ -0,0 +1,369 @@
...
+        id: "mssql.objectExplorerV2.copyConnectionString",
+        title: "Copy Connection String",
+        feature: "copyConnectionString",
+        route: "legacyRedirect",
+        placements: [
+            {
+                flag: "copyConnectionString",
+                menuGroup: "2_MSSQL_connection@4",
+                appliesTo: (facts) =>
+                    facts.kind === "connectedServer" || facts.kind === "disconnectedConnection",
```
kburtram: **C — Copy Connection String is not offered on lost connections.** v1 shows it on `Server | disconnectedServer`; a v2 `lostConnection` node is still a saved profile with an h0 route (no handoff connection needed), but `commandFlagsFor({ kind: "lostConnection" })` is empty here while edit/move/remove menus do include `lostConnection`. Add `|| facts.kind === "lostConnection"` here and in the policy's `nodeKinds`.

(source: Fable)

extensions/mssql/test/unit/objectExplorerV2Handoff.test.ts
```diff
@@ -0,0 +1,216 @@
...
+        expect(await service.ensureOwnerUri("p1", "sfp_x", PROFILE, "profiler")).to.equal(
+            undefined,
+        );
+        expect(service.hasHandoff("p1")).to.equal(false);
+
+        const { connections, calls } = seam();
+        const ttlService = new OeV2ClassicHandoffService(connections, { idleTtlMs: 5 });
+        const uri = await ttlService.ensureOwnerUri("p1", "sfp_x", PROFILE, "profiler");
+        expect(uri).to.not.equal(undefined);
+        await sleep(30);
```
kburtram: **C — TTL test uses real timers.** `idleTtlMs: 5` plus a 30 ms real `sleep` will flake on a loaded CI runner; `sinon.useFakeTimers()` and `clock.tick(6)` make it deterministic. While in here, the gaps that would have caught the redirect findings above: schema-compare argument arity, the seam's `connect` argument not being `===` the stored profile, `policiesForNode` for non-table objects, and `handleDrop` with malformed JSON / a foreign MIME type.

(source: Fable)

extensions/mssql/src/objectExplorer/v2/activation.ts
kburtram: Fixed in f2ed818f7. The connect handler now shows an error only when the registry ends in the failed state; cancellation returns to disconnected silently, and the restored cancellation test pins the late-session cleanup behavior. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/sessions/oeV2SessionRegistry.ts
kburtram: Fixed in f2ed818f7. Reconnect now retires the previous state subscription and closes the previous session before replacing a lost, failed, or disconnected registry entry. A regression test verifies exactly one close and one subscription disposal. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/activation.ts
kburtram: Fixed in f2ed818f7. Remove Connection now resolves the saved profile and calls ConnectionStore.removeProfile(profile, false), preserving the full settings, MRU, and credential-cleanup contract. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/activation.ts
kburtram: Fixed in f2ed818f7. The status command now lazily creates and reuses one OutputChannel, clears it before each display, and disposes it during OE v2 teardown. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/metadata/oeV2MetadataCoordinator.ts
kburtram: Fixed in f2ed818f7. Server and database pending acquisitions are now cleared in finally blocks on both success and rejection, so Refresh can retry. The new coordinator test verifies a first rejection followed by a successful retry for both lease types. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/activation.ts
kburtram: Fixed in f2ed818f7. The two-second ticker now calls refreshTransientConnections, which emits targeted changes only for connecting or disconnecting rows without invalidating the cached profile tree. A focused event test covers this behavior. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/activation.ts
kburtram: Fixed in f2ed818f7. The perf probes now locate Security and Logins from their structured serverFolder path values instead of comparing localized display labels. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/sessions/oeV2ProfileAdapter.ts
kburtram: Fixed in f2ed818f7. The unnamed-group fallback now uses ObjectExplorerV2.unnamedGroup, and the generated localization bundle and XLIFF were refreshed. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2TreeController.ts
kburtram: Fixed in f2ed818f7. Search now wraps ensureDatabase in the same bounded-wait policy used by expansion. If metadata is still unavailable after the bound, the command presents a localized retry-later message instead of hanging. (Source: GPT-5.6 Sol)

extensions/mssql/test/unit/objectExplorerV2CommandRegistry.test.ts
kburtram: Fixed in f2ed818f7. The B25 suite setup now calls initializeIconUtils before constructing TreeNodeInfo. The command-registry file passes standalone with 15 passed and 0 failed. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2ConnectionLabel.ts
kburtram: Fixed in f2ed818f7. The default database marker, tooltip field names, authentication names, boolean values, and not-set marker now use shared localization constants. The generated bundle and XLIFF were refreshed. (Source: GPT-5.6 Sol)

extensions/mssql/test/unit/objectExplorerV2Browse.test.ts
kburtram: Fixed in f2ed818f7. Restored the five non-excluded browse tests: join-in-flight, cancellation and late cleanup, catalog states, server-catalog failure, and structural-folder traversal. The intentionally deferred canSelectTop assertion remains excluded. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Browse.ts
kburtram: Fixed in f2ed818f7. catalogGate now renders both absent and loading section readiness as a loading node, never as No items. A regression test covers both states. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Browse.ts
kburtram: Fixed in f2ed818f7. Schema grouping now uses a deterministic case-folded ordinal comparator consistent with catalog ordering. The regression test verifies dbo sorts before Sales. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Hierarchy.ts
kburtram: Fixed in f2ed818f7. The comparator normalizes sortLast with boolean coercion so false and undefined are equivalent, then applies order normally. A regression test covers the mixed false and omitted case. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Path.ts
kburtram: Fixed in f2ed818f7. decodePath now enforces the exact segment count for every path variant and validates objectKind and objectFolder tokens against their supported sets. Corrupt-path tests cover missing segments and invalid enums. (Source: GPT-5.6 Sol)

extensions/mssql/test/unit/objectExplorerV2Hierarchy.test.ts
kburtram: Fixed in f2ed818f7. The test literal now contains the intended escaped backslash in the Windows-style login name, so the round-trip assertion exercises the actual value. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Readiness.ts
kburtram: Fixed in f2ed818f7. Removed the unused synthesizeChildren API and its isolated dead-code test. Return-value parameters are no longer filtered out, and the browse fixture now verifies that a function return value renders. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Browse.ts
kburtram: Fixed in f2ed818f7. The stale notice now wraps the combined structural, content, and trailing-child list, so it is always the first child. A regression test pins the ordering ahead of System Tables. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2Browse.ts
kburtram: Fixed in f2ed818f7. External Tables now honors groupBySchema at both the schema-list and schema-child levels. Known database states also use localized labels instead of lower-cased wire tokens; the generated localization files and grouping regression were added. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/legacy/oeV2LegacyRedirect.ts
kburtram: Fixed in f2ed818f7. Every user-visible legacy redirect outcome now comes from ObjectExplorerV2 localization constants or the localized legacyFeatureFailed formatter, with generated bundle and XLIFF updates. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/legacy/oeV2LegacyRedirect.ts
kburtram: Fixed in f2ed818f7. The schema-compare policy now supplies the required second positional argument as an unset target, while preserving the selected v2 node as source. The redirect test verifies the complete command argument list. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/legacy/oeV2LegacyRedirect.ts
kburtram: Fixed in f2ed818f7. Cached profiles are cloned at the redirect boundary, again before the classic connect seam, and in the legacy node adapter, preventing classic password normalization from mutating profile-tree state. Identity-isolation tests cover the seams. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/legacy/oeV2ClassicHandoffService.ts
kburtram: Fixed in f2ed818f7. The arbitrary idle TTL and timer were removed. Handoff connections now remain valid for long-running dialogs and close through explicit disconnect, profile or group removal, unregister, or deactivation ownership paths. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/commands/oeV2GroupCommands.ts
kburtram: Fixed in f2ed818f7. Delete group and contents now enumerates profiles across the full descendant-group subtree and closes each handoff and v2 session before removing shared storage. A regression test covers direct and nested connections. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/commands/oeV2LegacyCommandPolicy.ts
kburtram: Fixed in f2ed818f7. Copy Name now has a view-item menu contribution gated by oe2:canCopyName. The redundant unreachable legacyActions picker and its contribution were removed, dead label and title fields were removed, and policy plus redirect validation now restrict table-only commands to table objects. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/commands/oeV2NativeCommands.ts
kburtram: Fixed in f2ed818f7. Search input is trimmed before use, and result descriptions now map object-kind tokens through localized ObjectExplorerV2 labels. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/commands/oeV2CommandRegistry.ts
kburtram: Fixed in f2ed818f7. lostConnection is now included in the Copy Connection String registry placement and legacy policy, and the adapter maps it to disconnectedServer for the classic handler. Registry and redirect regressions cover the route. (Source: GPT-5.6 Sol)

extensions/mssql/test/unit/objectExplorerV2Handoff.test.ts
kburtram: Fixed in f2ed818f7. The real-timer test disappeared with the unsafe TTL. Added deterministic coverage for schema-compare arity, profile clone identity, non-table policy filtering, and malformed JSON plus foreign MIME drag/drop payloads. (Source: GPT-5.6 Sol)

kburtram:
Addressed the two suppressed Copilot localization findings in f2ed818f7: Object Explorer v2 connection labels and tooltips now use the shared localized constants, and every user-visible legacy redirect error now uses ObjectExplorerV2 localization. The generated bundle and XLIFF were refreshed. The focused OE v2 selection passes 114/114, and the full MSSQL suite improved to 4,865 passed with only the five pre-existing Windows QueryCompletionSoundService path failures remaining. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/commands/oeV2NativeCommands.ts
```diff
@@ -0,0 +1,159 @@
...
+            "mssql.objectExplorerV2.copyName",
+            async (node?: OeV2Node) => {
+                if (!isEnabled()) {
+                    return;
+                }
+                const name = node?.objectName ?? node?.schema ?? node?.database ?? node?.label;
+                if (name) {
+                    await vscode.env.clipboard.writeText(name);
+                    emitCommand("copyName", "native", node);
+                }
```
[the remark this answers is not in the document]
kburtram: Fixed in 2b2bf300d. Copy Name now uses objectName only for actual object and databaseObject nodes; every other selected node copies its own label, so columns, parameters, keys, and foreign keys cannot inherit a parent identity. The regression covers object, database object, column, parameter, and key cases. (Source: GPT-5.6 Sol)

extensions/mssql/src/objectExplorer/v2/tree/oeV2NodeFactory.ts
```diff
@@ -0,0 +1,193 @@
...
+        ...(description ? { description } : {}),
+        tooltip: lines.join("\n"),
+        collapsible: true,
+        connectionId: profile.profileId,
+        // DB-scoped connections carry their database (K4 backup targeting).
+        ...(profile.database ? { database: profile.database } : {}),
+        ...(containerName ? { containerName } : {}),
+        readiness: NOT_APPLICABLE,
+        capabilities: capabilitiesFor(kind),
+        icon:
```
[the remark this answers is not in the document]
kburtram: Fixed in 2b2bf300d. A disconnecting row still uses the transient visual node kind, but now receives no Cancel Connection capability while teardown is in flight. The regression verifies both the capability and serialized context omit canCancelConnect. (Source: GPT-5.6 Sol)

PR #22746, opened 2026-08-14
Schema Compare: add ability to select from all connections

Benjin:
## Summary
Shows all saved connection profiles in Schema Compare, including profiles that are not currently connected. Database lists are cached per connection, configured databases remain visible while loading, and connection changes provide immediate loading and error feedback without retaining the previous database list.

The separate button for opening Connection Dialog has been removed from source and target selection.

Fixes #22731

## Screenshots
**Before:** (only showing active connections)

image

**After:** (all saved connections are available)
image

image

PR #22877, opened 2026-09-02
Fix SCMP connection profile matching

aasimkhan30:
## Description

Fixes opening `.scmp` files when equivalent saved connections represent the server port differently. Ignores the embedded connection string during profile lookup and normalizes embedded and separate ports.

Regression from #22746.

## Code Changes Checklist

## Reviewers: Please focus on

Connection matching for `.scmp` database endpoints.

### DIFF
--- a/extensions/mssql/src/models/utils.ts
+++ b/extensions/mssql/src/models/utils.ts
@@ -334,15 +334,28 @@ export class ConnectionMatcher {
         current: IConnectionProfile,
         expected: IConnectionProfile,
     ): boolean {
-        function normalizeServerName(server: string): string {
-            if (server === ".") {
-                return "localhost";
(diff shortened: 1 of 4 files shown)


extensions/mssql/src/models/utils.ts
Benjin, line 439: Tbh, I think we can just strip out this function (and others like it) so that everything just uses ConnectionMatcher directly.  What do you think?

aasimkhan30: Agreed. I tracked consolidating these helpers under ConnectionMatcher in #22889 so we can keep this PR focused on the SCMP fix.

lewis-sanchez (approved):

PR #22896, opened 2026-09-04
Preserve precise Fluent grid scroll position

aasimkhan30:
## Summary

- preserve the pixel offset within the top visible Fluent result-grid row
- restore the exact vertical scroll position instead of snapping to the row boundary
- retain row-only fallback behavior for previously saved state

## Validation

- `npm run build:extension:typecheck`
- `npm run build:webviews`
- `npm run lint -- --target mssql`
- `npm test -- --target mssql --grep "Fluent Result Grid"` (84 passed)
- `npm test -- --target mssql` (full suite passed)

## Dependency

This is stacked on #22893 because it extends the Fluent result-grid state introduced there.


aasimkhan30:
Closing this stacked PR; replacing it with a standalone PR targeting main.

PR #22902, opened 2026-09-05
Fix Object Explorer node stuck on "Loading..." after applying a filter

### LINKED ISSUE #22874: [Bug]: Table Filtering Seems to No longer Work
### Description

Expected MSSQL Extension to filter unwanted tables from the schema browser
However when I filter using the "Not Contains" it used to filter as expected, but now it just says "loading" forever and freezes the entire extension so have to kill VS Code and start over.

Image

### Steps to Reproduce

1.  Connect to SQL Server
2. Expand Tables
3. Select "Filter" Icon aka the funnel
4. Under the "Name" property choose the operator "Not Contains" and type in the value what tables you want to exclude
5. Press Apply 
6. Nothing happens it just stays stuck on "Loading"
what is interesting is if you chose "Contains" it works fine

### Affected Area

- [ ] Query editor
- [ ] Query results panel
- [x] Object Explorer
- [ ] Preview/Edit data
- [ ] Table Designer
- [ ] Schema Designer
- [ ] Schema Compare
- [ ] SQL Database projects
- [ ] Query Plan Visualizer

_No response_

### Environment Information

VS Code Version :1.136
MSSQL Extension Version : 1.45.1
Operating System : Windows 11
Connection Type : SQL Server

### Confirmation

aasimkhan30:
Fixes #22874

## Problem

Applying an Object Explorer filter (for example `Name` / `Not Contains`) leaves the node spinning on "Loading..." forever, and the extension becomes unresponsive until VS Code is killed.

## Root cause

Applying a filter marks the node for refresh and fires the tree data change event. VS Code answers that event by calling `getChildren` again, and the loading node the service returns fires the event once more.

Because `shouldRefresh` stayed set for the whole duration of the in-flight expand, those internal callbacks were indistinguishable from a fresh user refresh:

1. `getNodeChildren` reads `element.shouldRefresh` (still `true`) and sees an in-flight fetch, so it queues another refresh and returns `setLoadingUiForNode`.
2. `setLoadingUiForNode` fires the refresh callback, so VS Code calls `getChildren` again — back to step 1.
3. When the expand finally completes, the queued marker starts a **whole new expand**, and the cycle restarts.

Whenever the expand outlasts VS Code's tree refresh debounce, this never settles: the spinner never resolves and the tree re-renders continuously.

Applying a filter reliably lands in that window because a marked-for-refresh node sends a `RefreshRequest`, which re-populates the node on the server via SMO rather than expanding cached state — comfortably slower than the debounce on a real server. A fast cached expand beats the debounce and settles, which is why this looked like "only filtering is broken".

## Fix

Consume the refresh request up front in `getNodeChildren` and carry it forward as an explicit argument through `getOrCreateNodeChildrenWithSession` → `expandExistingNode` / `createSessionAndExpandNode` → `expandNode`. An internal loading-UI callback now reads as an ordinary re-entrant `getChildren` and returns the cached loading node without queuing anything, while a genuine second user refresh still queues correctly.

`expandNode` keeps defaulting its new `shouldRefresh` parameter to the node's flag, so its existing external callers are unchanged.

## Testing

- New regression test drives the real feedback loop — refresh callback → `getChildren` → loading node → refresh callback — with an expand slow enough to still be in flight when the tree calls back. It fails on `main` (the loop keeps issuing requests) and passes here with exactly one `RefreshRequest`.
- Adjusted one exact-args assertion in `objectExplorerProvider.test.ts` for the new parameter.
- Full MSSQL unit suite: 4857 passed, 0 failed, 10 skipped.
- Lint and typecheck clean

### DIFF
--- a/extensions/mssql/src/objectExplorer/objectExplorerService.ts
+++ b/extensions/mssql/src/objectExplorer/objectExplorerService.ts
@@ -376,6 +376,9 @@ export class ObjectExplorerService {
                          * the database finishes resuming.
                          */
                         shouldRefresh = true;
+                        expandActivity.update({
+                            additionalProps: { isRefresh: shouldRefresh.toString() },
+                        });
                         continue;
                     }
                 }
@@ -621,14 +624,8 @@ export class ObjectExplorerService {
             `getNodeChildren: ${getNodeDescriptor(element)}, hadCache=${hadCache}, hasInFlight=${hasInFlight}`,
         );
 
-        /**
-         * Consume the refresh request up front and carry it forward as a local. Showing the loading
-         * node fires the tree-data-change event, which makes VS Code call back into this method
-         * while the load is still running; if the flag were still set, that internal callback would
-         * be mistaken for a second user-initiated refresh, queue another refresh, and show the
-         * loading node again — a cycle that never settles whenever the expand outlasts VS Code's
-         * tree refresh debounce.
-         */
+        // Consume the refresh before showing the loading node. That update makes VS Code request
+        // the children again, and the cleared flag prevents the callback from queuing a refresh.
         element.shouldRefresh = false;
 
         if (wasRefresh) {
--- a/extensions/mssql/src/objectExplorer/objectExplorerService.ts
+++ b/extensions/mssql/src/objectExplorer/objectExplorerService.ts
@@ -200,14 +200,19 @@ export class ObjectExplorerService {
     }
 
     /**
-     * Expands a node in the Object Explorer tree. If the node has the shouldRefresh flag set, it will be refreshed.
+     * Expands a node in the Object Explorer tree.
      * @param node The node to expand
      * @param sessionId The session ID to use for the expansion
+     * @param shouldRefresh Whether to re-populate the node on the server instead of expanding its
+     *   cached state. Defaults to the node's `shouldRefresh` flag for callers that still signal the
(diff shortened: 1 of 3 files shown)


extensions/mssql/src/objectExplorer/objectExplorerService.ts
```diff
@@ -616,6 +621,16 @@ export class ObjectExplorerService {
             `getNodeChildren: ${getNodeDescriptor(element)}, hadCache=${hadCache}, hasInFlight=${hasInFlight}`,
         );
 
+        /**
+         * Consume the refresh request up front and carry it forward as a local. Showing the loading
```
Benjin: Probs can be made more concise.

extensions/mssql/test/unit/objectExplorerService.test.ts
```diff
@@ -833,6 +836,89 @@ suite("OE Service Tests", () => {
             ).to.be.false;
         });
 
+        test("a filtered refresh settles even when the expand outlasts the tree refresh debounce", async () => {
+            /**
+             * Applying an Object Explorer filter marks the node for refresh and fires the tree
```
Benjin: Lol, agent-speak is so unintelligble.  Can you clean up this comment?

Benjin (dismissed):

Benjin (approved):

PR #22903, opened 2026-09-06
Update slickgrid-react to 10.10.0 and drop the workarounds it fixes

aasimkhan30:
## Description

Updates `slickgrid-react` 10.8.2 → 10.10.0 and deletes the FluentResultGrid overrides that only existed to work around bugs now fixed upstream.

Seven issues filed against slickgrid-universal on 2026-08-25 are all fixed in 10.10.0: [#2749](https://github.com/ghiscoding/slickgrid-universal/issues/2749), [#2750](https://github.com/ghiscoding/slickgrid-universal/issues/2750), [#2751](https://github.com/ghiscoding/slickgrid-universal/issues/2751), [#2752](https://github.com/ghiscoding/slickgrid-universal/issues/2752), [#2753](https://github.com/ghiscoding/slickgrid-universal/issues/2753), [#2754](https://github.com/ghiscoding/slickgrid-universal/issues/2754), [#2755](https://github.com/ghiscoding/slickgrid-universal/issues/2755). Each had a local workaround. Two commits: version bump, then removals.

Deleted `fluentResultGridCellRangeSelector.ts` outright and trimmed `fluentResultGridSelectionModel.ts` from 177 to ~73 lines; Shift-click, Ctrl/Cmd toggle, multi-range drag and active-cell sync now come from the base model via the new `enableMultiSelection` option. **384 lines removed.**

Two things stay on purpose. The row-number gutter branch, because the base model only consults `rowSelectColumnIds` when `selectionType !== "cell"` and this grid is cell-only. And `enableExcelCopyBuffer: false`, because the grid is virtualized and copy must go through SQL Tools Service to format the full selection rather than the loaded row window.

**Needs manual verification:** copy layout for two *adjacent* Ctrl-selected blocks. The deleted code merged adjacent ranges; upstream's toggle may not, which is what #2752 was about.

Thanks to @ghiscoding for making the fixes! 

## Code Changes Checklist


ghiscoding:
@aasimkhan30 side note, I found another bug after releasing v10.10.0 which was that using the new multi-selection with row selections caused incorrect row ranges... however I think you're not affected since I believe you're only using cell multi-selection (not row selection which has the bug). The fix will be included in the upcoming v10.11.0 (not yet released), the new version will also include a new grid option to allow "automatic height for multiline column headers" (not sure if that can be useful for you or not but that's the new feature). I'll probably release at the end of the week, just waiting to see if you catch any other bugs. Cheers

Benjin (approved):

PR #22911, opened 2026-09-09
Generate SELECT/UPDATE/DELETE/INSERT: resolve real table name via FROM-clause parsing

FredericoMSAlmeida:
## Description

Generate SELECT/UPDATE/DELETE/INSERT from the results grid always fell back to a placeholder table name "UnknownTable", because SQL Tools Service never populates IDbColumn.baseTableName/baseSchemaName (see the debug findings doc this branch implements: docs/superpowers/plans/2026-09-08-grid-query-generation-debug-findings.md).

This adds a conservative, client-side fallback: it parses the query's own FROM clause to recover the real table name for the simple single-table case, and only uses it when the real column metadata is missing. Existing metadata always wins when present. Anything ambiguous (JOIN, UNION, subquery, CTE, comma-join, multi-statement, linked-server names, escaped identifiers, APPLY/PIVOT) bails to the old "UnknownTable" behavior rather than risking a wrong guess.

Implementation plan: docs/superpowers/plans/2026-09-09-grid-query-table-name-fallback.md

Pieces:
- parseSingleTableFromClause - standalone FROM-clause parser (extensions/mssql/src/queryResult/fromClauseTableParser.ts)
- QueryRunner.getBatchQueryText - resolves the executed query snapshot for a batch, not the live (possibly since-edited) editor text
- ResolveTableNameRequest RPC connecting the webview to the host-side resolver
- sqlScriptGenerator.ts fallback threading through buildQualifiedTableName/generateSelect/generateUpdate/generateDelete/generateInsertForRows
- contextMenu.plugin.ts wiring to call the RPC before building SQL

Also fixes a stray debug console.log in handleContextMenu and updates a pre-existing test whose assertions assumed only one RPC call per Generate action.

Outstanding: manual F5 Extension Development Host verification (running the real extension, Generate SELECT on a real single-table query, then a JOIN query) has not been done in this sandboxed environment and should happen before/at merge.

## Code Changes Checklist

- [x] New or updated unit tests added
- [x] All existing tests pass (npm run test) - the only test failures on main today are pre-existing, unrelated to this branch (a macOS-only sound-service test and one flaky timeout test, in files this branch never touches)
- [x] Code follows contributing guidelines
- [x] Telemetry/logging updated if relevant - n/a, no new telemetry needed
- [x] No regressions or UX breakage - pending manual F5 verification noted above


aasimkhan30:
Closing this PR for now. This is an interesting idea, but I think the implementation should be in sqltoolsservice.
Please create a feature request with a few examples of the options and behavior you’d like to see, and I can pick it up from there.

PR #22936, opened 2026-09-15
connection: add Kerberos configuration guidance

Benjin:
## Summary

- Update the existing macOS and Linux Kerberos setup guide
- show a Kerberos setup tooltip on the Windows Authentication option only when the extension runs on macOS or Linux
- show messages when a kerberos error is encounters while trying to connect (Object Explorer, Connection Dialog)

image


aasimkhan30 (dismissed):

aasimkhan30:
@Benjin, do we surface this information in the extension like a link to this readme?

Benjin:
@aasimkhan30

> Do we surface this information in the extension like a link to this readme?

I don't think so, but that's a good point.  I can add a tooltip when the user's on mac or linux and has integrated auth selected.

aasimkhan30 (dismissed):

aasimkhan30 (approved):

